PR #12, opened 2025-01-17
Initial text search RFC

allenss-amazon:


rfc/text-field.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Motivation (Required)
+
+Text searching is useful in many applications.
+
+## Terminology
+
+In the context of this RFC.
+
+| Term | Meaning |
```
joehu21: Can we add the definition of "stop words"?

allenss-amazon: Stop words are used in the search lexicon to indicate words that are ignored for text matching purposes. For our purposes they are effectively "not present". In other words, the text that's actually indexed and queried against is the text that had the stop words removed.

zvi-code:
@allenss-amazon, If we are thinking of it as a design, is this also the right place to set the goals for: memory-usage? performance?
I think it's worth mentioning here the goals in terms of multi-threaded operation, update costs and query costs (big O notation).

rfc/text-field.md
```diff
@@ -0,0 +1,170 @@
...
+
+Each ```TEXT``` field has the a set of configurables some control the process to convert a string into a vector of terms, others control the contents of the generated index.
+
+| Metadata | Type | Meaning | Default Value |
+| --- | --- | --- | --- |
+| Punctuation | vector<character> | Characters that separate words during the tokenization process. |  
+| Stop Words | vector<String> | List of stop words to be removed during decomposition |
+| Stemming Language | Enumeration | Snowball stemming library lanugage control |
+| Suffix Tree | Boolean | Controls the presence/absence of a suffix tree. |
+| Word Offsets | Boolean | Indicates if the postings for a word contain word offsets or just a count. |
```
QuChen88: Would be nice if you can include a definition of `posting` in the RFC, and more details on how words get indexed

rfc/text-field.md
```diff
@@ -0,0 +1,170 @@
...
+### Tokenization process
+
+A tokenization process is applied to strings of text to produce a vector of terms.
+Tokensization is applied in two contexts. 
+First as part of the ingestion process for text fields of keys.
+Second to process query string words and phrases.
+
+The tokenization process has four steps.
+
+1. The text is tokenized, removing punctuation and redundant whitespace, generating a vector of words.
```
KarthikSubbarao, line 49: Just listing a few questions. How would this handle the following?

1. Contractions (e.g. you're, I'm, They'll, etc) Will these be split into characters? URLs can be considered another form of a contraction
2. Emojis - Since we are not removing based on unicode, will emoji's survive tokenization?

rfc/text-field.md
```diff
@@ -0,0 +1,170 @@
...
+| prefix tree | A mapping from input string to an output object. Insert/delete operations are O(length(input)). Additional operations include the ability to iterate over the entries that share a common prefix in O(length(prefix)) time. |
+
+## Design considerations
+
+The text searching facility provides machinery that decomposes text fields into terms and field-indexes them.
+The query facility of the ```FT.SEARCH``` and ```FT.AGGREGATE``` commands is enhanced to select keys based on combinations of words, wildcard, fuzzy and phrase matching.
+
+### Tokenization process
+
+A tokenization process is applied to strings of text to produce a vector of terms.
```
KarthikSubbarao, line 42: Are we describing the tokenization algorithm that we will write ourselves or is the algorithm from a library we plan to? 

Note: For better accuracy, different languages have their own tokenizers that handle linguistic nuances. But it might be fine to start with a general purpose approach

rfc/text-field.md
```diff
@@ -0,0 +1,170 @@
...
+Step 1 of the tokenization process is controlled by a specified list of punctuation characters. 
+Sequences of one or more punctuations characters delimit word boundaries.
+Note that individual punctuation characters can be escaped using the normal backlash prefix syntax, causing the next character to be treated as non-punctuation.
+
+### Text Field-Index Structure
+
+The text field-index is a mapping from a term to a list of locations.
+Depending on the configuration, the list of locations may be just a list of keys or a list of key/term-offset pairs.
+
+The mapping structure is built from one or two prefix trees. When present, the second prefix tree is built from each inserted term sequenced in reverse order, effectively providing a suffix tree.
```
KarthikSubbarao, line 79: In case of deletion of text documents, when would these structures be updated?

rfc/text-field.md
```diff
@@ -0,0 +1,170 @@
...
+1. The text is tokenized, removing punctuation and redundant whitespace, generating a vector of words.
+2. Latin upper-case characters are converted to lower-case.
+3. Stop words are removed from the vector.
+4. Words with more than a fixed number of characters are replaced with their stemmed equivalent term according to the selected language.
+
+Step 1 of the tokenization process is controlled by a specified list of punctuation characters. 
+Sequences of one or more punctuations characters delimit word boundaries.
+Note that individual punctuation characters can be escaped using the normal backlash prefix syntax, causing the next character to be treated as non-punctuation.
+
+### Text Field-Index Structure
```
KarthikSubbarao, line 74: This is an interesting mapping structure which made more sense once I read through this with actual text examples in mind. If there are any images / drawings, that might help visually illustrate how it would look

rfc/text-field.md
```diff
@@ -0,0 +1,353 @@
...
+| stemming    | A process of mapping similar words into a common base word. For example, the words _drives_, _drove_ and _driven_ would be replaced with the  _drive_.                                                                      |
+| term        | The output of stemming a word.                                                                                                                                                                                              |
+| prefix tree | A mapping from input string to an output object. Insert/delete operations are O(length(input)). Additional operations include the ability to iterate over the entries that share a common prefix in O(length(prefix)) time. |
+
+## Design considerations
+
+The text searching facility provides machinery that decomposes text fields into terms and field-indexes them.
+The query facility of the ```FT.SEARCH``` and ```FT.AGGREGATE``` commands is enhanced to select keys based on combinations of words, wildcard, fuzzy and phrase matching.
```
KarthikSubbarao: Do we plan to support the ability to search on `PHONETIC` fields?

allenss-amazon: No.

rfc/text-field.md
```diff
@@ -0,0 +1,353 @@
...
+The wildcard can be at any position within the term, providing prefix, suffix and infix style matching.
+Note, in this proposal, the second prefix tree is required to perform infix and suffix matching.
+
+#### Phrase matching
+
+Exact phrase matching is specified by enclosing a sequence of terms in double quotes ```"```. In order to perform phrase matching, the field-index must be configured to contain term-offsets.
+
+#### Phrase matching
+
+Exact phrase matching is specified by enclosing a sequence of terms in double quotes ```"```. In order to perform phrase matching, the field-index must be configured to contain term-offsets.
```
KarthikSubbarao: This seems like a duplicate

rfc/text-field.md
```diff
@@ -0,0 +1,353 @@
...
+#### Term matching
+
+There are three types of term matching: exact, wildcard and fuzzy. Exact term matching is self-descriptive, i.e., only keys containing exactly the specified text are matched.
+
+Wildcard matching provides a subset of reg-ex style matching of terms.
+Initially, only a single wildcard specifier ```*``` is allowed which matches any number of characters in a term.
+The wildcard can be at any position within the term, providing prefix, suffix and infix style matching.
+Note, in this proposal, the second prefix tree is required to perform infix and suffix matching.
+
+#### Phrase matching
```
KarthikSubbarao: It might be assumed already, but the different TEXT queries types includes the following. We will support all 3 types, right? 
    1. Single token: `FT.SEARCH books-idx '@title:dogs'`
    2. Phrase: `FT.SEARCH books-idx '@title:"hello world"'`
    3. Grouped: `FT.SEARCH books-idx '(@title:dogs | @title:cats)'`

rfc/text-field.md
```diff
@@ -0,0 +1,353 @@
...
+│ │  └────────────┘  │          │ key3: [8]  │        │  └────────────┘  │     │
+│ │                  │          └────────────┘        │                  │     │
+│ │                  │                                │                  │     │
+│ └──────────────────┘                                └──────────────────┘     │
+│                                                                              │
+└──────────────────────────────────────────────────────────────────────────────┘
+```
+
+
+### Search query operators
```
KarthikSubbarao, line 113: Do we plan to support QMA - Query Modifying Attributes?
Note: There can be multiple QMA per query string.


Some of them include:

$weight (Boosting)
```
FT.SEARCH books-idx "(@title:dogs | @title:cats) | (@title:dogs) => { $weight: 5.0; }"
```
 $inorder
```
FT.SEARCH memes '@phrase:(amazing hello world)=>{$inorder: true;}' NOCONTENT
```
$slop
```
FT.SEARCH quotes-idx '@text:(quick brown)=>{ $slop: 0 }'
```
$phonetic
```
FT.SEARCH people-idx '(@name:"jon") => { $phonetic: true }'
```

See [Ref](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/query_syntax/#query-attributes).

allenss-amazon: Yes. $inorder and $slop support exact phrase. $weight is meaningful for BM-25 (or other scoring algo) and this wasn't planned for v1 but will almost certainly happen in future releases. $phonetic is unclear.

rfc/text-field.md
allenss-amazon: Single wildcard and Fuzzy are word matching. Exact phrase is a combination of word matching. What you describe as "grouping" is binary combination of search operators. Binary combination of search operators is already assumed.

rfc/text-field.md
```diff
@@ -0,0 +1,353 @@
...
+│  Input Text     │     │ Tokenization  │     │ Case         │     │ Stop Word     │
+│  "The quick     │ ──> │ ["The",       │ ──> │ ["the",      │ ──> │ ["quick",     │
+│   brown fox."   │     │  "quick",     │     │  "quick",    │     │  "brown",     │
+└─────────────────┘     │  "brown",     │     │  "brown",    │     │  "fox"]       │
+                        │  "fox"]       │     │  "fox"]      │     └───────┬───────┘
+                        └───────────────┘     └──────────────┘             │
+                                                                           V
+                                                                      ┌───────────────┐
+                                                                      │ Stemming      │
+                                                                      │ ["quick",     │
```
yairgott: It would be more helpful if the example used a word that can actually be stemmed, perhaps  "The quick brownish fox." ?

rfc/text-field.md
yairgott, line 1: It would be helpful to include a few supported `FT.SEARCH` query expressions, each with a brief explanation.

rfc/text-field.md
```diff
@@ -0,0 +1,353 @@
...
+                                                                      │  "brown",     │
+                                                                      │  "fox"]       │
+                                                                      └───────────────┘
+```
+
+Step 1 of the tokenization process is controlled by a specified list of punctuation characters. 
+Sequences of one or more punctuations characters delimit word boundaries.
+Note that individual punctuation characters can be escaped using the normal backslash prefix syntax, causing the next character to be treated as non-punctuation.
+
+### Text Field-Index Structure
```
KarthikSubbarao, line 74: Could we also talk about how the following operations work with the new index structure?

1. RDB Save and RDB Load
2. Slot Migration
3. FlushAll / Index Deletion
4. Defragmentation

rfc/text-field.md
allenss-amazon: This is a research item.

rfc/text-field.md
allenss-amazon: See the design documents for the deletion algorithm.

rfc/text-field.md
allenss-amazon: added

PR #29, opened 2025-02-01
Redesign FT.SEARCH command line parsing in anticipation of merging with future FT.AGGREGATE code.

allenss-amazon:
Previously, the command line was parsed up to the query string. The parse of the query string had the side effect of identifying all $param substitution variables. This list was then used to generate a custom key/value parser which was then combined with the other processing clauses to finish parsing of the command line.

Now, the parsing of the query string is performed after the parsing of the entire command line. The major change here is that instead of creating a custom key/value parser that only recognized the $param substitution variables a generic hash-map of the substitution variables is generated in the first pass. During the subsequent parsing of the query string, the various places where a $param substitution is allowed is done by looking up the value in the generated hash map.

At the end of the parsing, (Verify), the hash map is walked to ensure that every $param was used at least once.

The new code will also reject an attempt to define the same $param twice. The old code would silently accept this and use the last value.


.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
     "clangd.path": "/usr/bin/clangd",
     "C_Cpp.codeAnalysis.runAutomatically": true,
     "editor.formatOnSave": true,
-    "C_Cpp.intelliSenseEngine": "disabled",
+    "C_Cpp.intelliSenseEngine": "default",
```
yairgott: would this enable intelliSenseEngine checks? what is the benefit of setting it to "default" and would it conflict with clang-tidy?

.vscode/settings.json
```diff
@@ -19,4 +19,83 @@
     "bazel.buildifierFixOnFormat": false,
     "clangd.inactiveRegions.useBackgroundHighlight": true,
     "clangd.enableCodeCompletion": true,
+    "files.associations": {
```
yairgott: why this is needed?

src/commands/ft_search_parser.cc
```diff
@@ -70,15 +70,25 @@ constexpr absl::string_view kAsParam{"AS"};
...
-  absl::string_view query;
-  absl::string_view k;
-  absl::string_view ef;
-  std::string score_as;
-};
+absl::StatusOr<absl::string_view> SubstituteParam(
+    query::VectorSearchParameters &parameters, absl::string_view source) {
+  if (source.empty() || source[0] != '$') {
+    return source;
+  } else {
```
yairgott, line 77: nit- here and in other places, no need to use `else` when the `if` section alway returns. It would avoid the need to indent

src/commands/ft_search_parser.cc
```diff
@@ -218,96 +242,38 @@ ConstuctLimitParser() {
...
+              "Parameter count must be an even number.");
+        }
+        while (count > 0) {
+          VMSDK_ASSIGN_OR_RETURN(auto key_str, itr.Get());
+          itr.Next();
+          VMSDK_ASSIGN_OR_RETURN(auto value_str, itr.Get());
+          itr.Next();
+          absl::string_view key = vmsdk::ToStringView(key_str);
+          absl::string_view value = vmsdk::ToStringView(value_str);
+          if (parameters.parse_vars.params.find(key) !=
```
yairgott: Better to call `parameters.parse_vars.params.insert(key, std::make_pair(0, value))` and if the key already exists, the success would be false.

src/commands/ft_search_parser.cc
```diff
@@ -70,15 +70,25 @@ constexpr absl::string_view kAsParam{"AS"};
...
+      itr->second.first++;
+      return itr->second.second;
+    }
+  }
+}
 
-absl::StatusOr<ParamVariables> ParseVectorFilter(std::string_view filter) {
+absl::Status ParseKnnInner(query::VectorSearchParameters &parameters,
+                           std::string_view filter) {
   absl::InlinedVector<absl::string_view, 8> params =
```
yairgott, line 92: Maybe for another PR, it would be nice to consolidate the ParseKnnInner parsing with the other parsers to use vmsdk::ArgsIterator. I think that we can do that by templetazing vmsdk::ArgsIterator with the type of the argv (the default type would be RedisModuleString but one can use absl::string_view).  This would simplify the code here by a lot.

yairgott (changes requested):
Looks overall good to me. Thanks for driving this!

I left a couple of comments inline.

.vscode/settings.json
allenss-amazon: oops. not sure what happened there. I didn't check that.

.vscode/settings.json
allenss-amazon: same here. no intention to change the tooling.

.vscode/settings.json
```diff
@@ -19,4 +19,8 @@
...
```
yairgott: Is this required?

src/commands/ft_search_parser.cc
yairgott: As per our offline discussion, we can wait with that until real need for such shows up.

.vscode/settings.json
allenss-amazon: No, and I diffed my PR against main and it showed no changes in this file. :( :( :(

yairgott (approved):

PR #33, opened 2025-02-04
CMake support

eifrah-aws:
With this PR, `valkey-search` can be built with `CMake`.

Build instructions where added to the `BUILD_cmake.md`.

Point to take into consideration while reviewing this PR:

## OpenSSL 
OpenSSL used, is the system SSL (this is similar to how `valkey` works)

## Submodules

Submodules are being pulled and built during the `cmake` stage. The first time you run `cmake` it might took a while (due to `git clone + build`), consecutive calls for `cmake` are **much** faster. 

The rationale behind this, is that submodules are rarely changed so there is no need to rebuild them after each `make` call (even if CMake will just report `Built target`)

## How to build & run

To build:

```bash
mkdir build-release
cd $_
cmake .. -DCMAKE_BUILD_TYPE=Release
make -j$(nproc)
```

To run:

```bash
valkey-server --loadmodule lib/valkeysearch.so
```


CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+cmake_minimum_required(VERSION 3.24)
+project(VSS)
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
```
yairgott, line 7: CMake should match the compilation flags defined in [.bazelrc](https://github.com/valkey-io/valkey-search/blob/main/.bazelrc), ensuring the necessary optimizations, such as SIMD, are enabled.

Additionally, I propose supporting the following build configurations:

1. Default (prod) build: A production-ready build with maximum optimization. It’s not intended for rapid iteration, so longer build times are acceptable. Beyond the flags specified in `.bazelrc`, we should also enable -O2 and link-time optimization (LTO).
2. Test build: Used for ongoing development, optimized for fast build times. One caveat is that some tests run very slowly with `-O0`. Therefore, we use `-O1` for testing in `.bazelrc` to strike a balance between performance and debugging visibility.
3. Test (prod) build: Used for presubmit verification, typically executed as a GitHub Action. This build should mirror the production configuration to catch potential issues before deployment.
4. Debug build: Designed for troubleshooting and local e2e testing. It should prioritize fast build times while preserving symbols and debug information for better diagnostics.
5. ASAN/TSAN build: similarly to debug and also enable [A|T]SAN.
6. ASAN test build: similarly to test build and also enable ASAN.

I'm fine with 5 and 6 as a separate PR.

BUILD_cmake.md
```diff
@@ -0,0 +1,65 @@
...
+cmake ..
+make -j$(nproc)
+```
+
+## Integration with IDE
+
+During the `CMake` stage of the build, `CMake` generates a JSON file named `compile_commands.json` and places it under the
+build folder. This file is used by many IDEs and text editors for providing code completion (via `clangd`).
+
+A small caveat is that these tools will look for `compile_commands.json` under the workspace root folder.
```
yairgott: Please update the following with the correct path:
1. [vscode clang compile commands path](https://github.com/valkey-io/valkey-search/blob/main/.vscode/settings.json#L2) 
2. [check_changes.sh](https://github.com/valkey-io/valkey-search/blob/main/ci/check_changes.sh#L38)
3. [entrypoint.sh](https://github.com/valkey-io/valkey-search/blob/main/ci/entrypoint.sh#L9) to create compile_commands.json while running vscode under the devcontainer.
4. [refresh_comp_db](https://github.com/valkey-io/valkey-search/blob/main/ci/refresh_comp_db.sh#L16) to rebuild compile_commands.json on every file changed while running vscode under the devcontainer.

yairgott (changes requested):
Thanks for driving this! I left a couple of comments inside, PTAL.

CMakeLists.txt
yairgott: Some of the compilation flags specified in [`.bazelrc`](https://github.com/valkey-io/valkey-search/blob/main/.bazelrc#L39-L52) are intended to suppress third party, protobuf and grpc compilation warnings, so it probably makes sense to use them just for the subcomponents build.

BUILD_cmake.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Install basic tools
+
+```bash
+sudo apt update
+sudo apt install -y clangd          \
+                    build-essential \
+                    g++             \
+                    cmake           \
+                    libgtest-dev    \
```
yairgott, line 15: In addition, the deb package protobuf-compiler is required

BUILD_cmake.md
```diff
@@ -0,0 +1,65 @@
...
+                    libssl-dev
+```
+
+
+## Build the module
+
+```bash
+git clone git@github.com:valkey-io/valkey-search.git
```
yairgott: It would be very useful if we would provide a build helper script, .e.g.:
./build.sh debug
./build.sh release
./build.sh clean

yairgott (changes requested):
I noticed that the tests are missing

BUILD_cmake.md
eifrah-aws: `protobuf-compiler` is not needed. During the cmake stage, we check out and build `grpc` (which is placed under `.deps/`). Once checked out and built, we use the `protoc` compiler that comes with the `grpc`:

```cmake
set(MODULE_BASE_DIR "${CMAKE_BINARY_DIR}/.deps")
set(MODULES_BIN_DIR "${MODULE_BASE_DIR}/install/bin")
set(MODULES_LIB_DIR "${CMAKE_BINARY_DIR}/.deps/install/lib")

set(grpc_cpp_plugin_location ${MODULES_BIN_DIR}/grpc_cpp_plugin)
set(Protobuf_PROTOC_EXECUTABLE ${MODULES_BIN_DIR}/protoc)
```

BUILD_cmake.md
eifrah-aws: I think that it will be easier to create a symbolic lfrom `${CMAKE_BINARY_DIR}/compile_commands.json` to `${CMAKE_SOURCE_DIR}/compile_commands.json` after `cmake` is done (and we could also remove the "caveat" paragraph I added above)

CMakeLists.txt
eifrah-aws: - I have added the missing flags to the `cmake` build
- LTO added to Release configuration
- `Debug` builds now uses `-O0` and `-DNEBUG`
- Fixed typo
- At the end of cmake stage, we now create a symbolic link for compile_commands.json

Clean build (without grpc which is built *once* during the cmake stage of the build) is taking: `1m39s`. Changing several files and then building, takes `~10seconds` (debug build) - which sounds reasonable to me

BUILD_cmake.md
yairgott: This yield the below error but the following works: `git clone https://github.com/valkey-io/valkey-search.git` 

yairg_google_com@test-cmake:~$ git clone git@github.com:valkey-io/valkey-search.git
Cloning into 'valkey-search'...
The authenticity of host 'github.com (140.82.114.3)' can't be established.
ECDSA key fingerprint is SHA256:p2QAMXNIC1TJYWeIOttrVc98/R1BUFWu3/LiyKgUfQM.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added 'github.com,140.82.114.3' (ECDSA) to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

BUILD_cmake.md
```diff
@@ -0,0 +1,65 @@
...
+cd valkey-search
+git submodule update --remote --init --recursive --depth 1
```
yairgott: The documentation should direct to use the build.sh script. Also, it would be very conviniet if build.sh would issue the 'git submodule update ...'

CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+cmake_minimum_required(VERSION 3.24)
```
yairgott: Can we relax this minimum version requirement? this cmake version is supported just with Ubuntu 24 which is quite new and I believe that requiring such adds friction. Ubuntu 20 comes with cmake 3.16.3 and Ubuntu 22 comes with cmake 3.22.1. Both failed for me while trying to change the min required version.

BUILD_cmake.md
yairgott: hmm, I'm still seeing the following error which is resolved just after installing protobuf-compiler:
[ 36%] Running grpc protocol buffer compiler on /home/yairg_google_com/valkey-search/src/coordinator/coordinator.proto
/bin/sh: 1: protobuf::protoc: not found
make[2]: *** [src/coordinator/CMakeFiles/coordinator_grpc_proto.dir/build.make:74: src/coordinator/coordinator.grpc.pb.h] Error 127
make[1]: *** [CMakeFiles/Makefile2:1902: src/coordinator/CMakeFiles/coordinator_grpc_proto.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

build.sh
```diff
@@ -0,0 +1,87 @@
...
+echo "Root directory: ${ROOT_DIR}"
+
+function print_usage() {
+cat<<EOF
+Usage: build.sh [options...]
+
+    --help | -h         Print this help message and exit
+    --cmake             Run cmake stage
+    --verbose | -v      Run verbose build
+    --release           Build for release
```
yairgott: I suggest that the following changes:
1. Run the command `git submodule update --remote --init --recursive --depth 1` as part of this script. Its execution time is negligible and it is idempotent. 
2. Auto detect if `--configure` is needed to be run. Eliminate the `--configure` param.
3. If no argument provided, build with release mode.
4. Replace the argument `--release` with `--debug` such that unless specify build in release mode. 

After those changes:
- Any `build.sh` execution update the submodules and run `configure` is needed
- For release build: `./build.sh`
- For debug build: `build.sh --debug`

Does this make sense?

cmake/Modules/valkey_search.cmake
```diff
@@ -0,0 +1,75 @@
...
+  if(VALKEY_SEARCH_RELEASE_BUILD)
+    set_property(TARGET ${name} PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)
+    message(STATUS "Adding shared library ${name} - LTO enabled")
+  else()
+    message(STATUS "Adding shared library ${name}")
+  endif()
+  valkey_search_target_update_compile_flags(${name})
+endfunction()
+
+# Setup global compile flags to match BAZEL build flags
```
yairgott: Lets remove all mentioning of bazel at it would die after submitting this PR

CMakeLists.txt
yairgott: I still see a lot of warnings emitted while compiling third-parties. A couple of examples:
1. /home/yairg_google_com/valkey-search3/build-debug/.deps/grpc/src/core/lib/promise/detail/promise_factory.h:174:5: warning: ‘always_inline’ function might not be inlinable [-Wattributes]
  174 |     PromiseFactoryImpl(F&& f) {
  2. /home/yairg_google_com/valkey-search/build-debug/.deps/grpc/src/core/tsi/ssl_transport_security.cc:627:34: warning: ‘int ENGINE_ctrl_cmd_string(ENGINE*, const char*, const char*, int)’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  627 |           !ENGINE_ctrl_cmd_string(engine, "DIR_LOAD", "2", 0) ||
      |            ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
3. /home/yairg_google_com/valkey-search/build-debug/.deps/grpc/src/core/lib/transport/call_state.h:187:3: warning: control reaches end of non-void function [-Wreturn-type]
  187 |   }

cmake/Modules/valkey_search.cmake
```diff
@@ -0,0 +1,75 @@
...
+  target_compile_options(${TARGET} PRIVATE -gdwarf-5)
+  target_compile_options(${TARGET} PRIVATE -gz=zlib)
+  target_compile_options(${TARGET} PRIVATE -ffast-math)
+  target_compile_options(${TARGET} PRIVATE -funroll-loops)
+  target_compile_options(${TARGET} PRIVATE -ftree-vectorize)
+  target_compile_options(${TARGET} PRIVATE -fopenmp)
+  target_compile_options(${TARGET} PRIVATE -mavx2)
+  target_compile_options(${TARGET} PRIVATE -mfma)
+  target_compile_options(${TARGET} PRIVATE -ffp-contract=off)
+  target_compile_options(${TARGET} PRIVATE -flax-vector-conversions)
```
yairgott, line 100: Release build should use `-O3`, aligning with Valkey optimized build.

build.sh
```diff
@@ -0,0 +1,87 @@
...
+CMAKE_TARGET=""
+
+echo "Root directory: ${ROOT_DIR}"
+
+function print_usage() {
+cat<<EOF
+Usage: build.sh [options...]
+
+    --help | -h         Print this help message and exit
+    --cmake             Run cmake stage
```
yairgott: IIUC this should be `--configure` rather than `--cmake`

CMakeLists.txt
yairgott: I also see: /usr/bin/ld: missing --end-group; added as last command line option

build.sh
yairgott, line 1: When running build.sh --cmake twice in a row without any changes, the second run still takes a long time and appears to be recompiling.

cmake/Modules/valkey_search.cmake
```diff
@@ -0,0 +1,75 @@
...
+  if(VALKEY_SEARCH_DEBUG_BUILD)
+    target_compile_options(${TARGET} PRIVATE -O0)
```
yairgott, line 105: We should use -O1 for running the unit tests; otherwise, they take too long to complete. I noticed some misalignment with .bazelrc. For example, in the testing configuration: [.bazelrc#L35-L36](https://github.com/valkey-io/valkey-search/blob/main/.bazelrc#L35-L36). I suggest revisiting this, not just for the unit tests.

yairgott: Also, you're missing the compilation flag, define in .bazelrc, `BAZEL_BUILD` which leads to test failures.  It makes sense to rename this to CMAKE_BUILD. The good news is that it looks like this flag now is needed just in testing/search_test.cc and you can remove all other usage of such  from the other unit tests. Please verify that all tests are passing.

build.sh
```diff
@@ -0,0 +1,87 @@
...
+    # run CMake for debug build & build
+    build.sh --configure
+
+    # run CMake for release build & build
+    build.sh --release --configure
+EOF
+}
+
+function configure() {
+    mkdir -p ${ROOT_DIR}/build-${BUILD_CONFIG}
```
yairgott: I suggest making the build directory hidden, e.g., .build-debug. Otherwise, it needs to be explicitly excluded when running commands like grep. In general, hiding build directories is more common and reduces confusion.

testing/CMakeLists.txt
yairgott, line 1: I noticed that all the binaries, including the unit tests, resides flat under the `bin` dir. It would be much more cleaner if we preserve the code directory structure. For example, the unittest fanout_test should reside under bin/testing/query/fanout_test. The shared object libvalkeysearch should reside under bin/src/libvalkeysearch.so

In this context, I noticed that under the dir bin there is a binary with a name `example` which is unclear why it's there.

yairgott, line 1: Missing vmsdk/testing

build.sh
yairgott, line 1: How do you suggest that to run the tests? Bazel support bazel test $TARGET also bazel test testing/... which provides a nice clean view of the test runs. It's important to support something similar as developers would frequently be running the unit tests.

build.sh
yairgott: Building with --release fails with the following error:
/usr/bin/ld: missing --end-group; added as last command line option
lto1: fatal error: multiple prevailing defs for ‘_ZTIN13valkey_search7indexes18EntriesFetcherBaseE’
compilation terminated.
lto-wrapper: fatal error: /usr/bin/c++ returned 1 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
make[2]: *** [testing/CMakeFiles/ft_create_parser_test.dir/build.make:117: bin/ft_create_parser_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:3457: testing/CMakeFiles/ft_create_parser_test.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 69%] Linking CXX static library ../lib/libvalkey_search.a
[ 69%] Built target valkey_search
make: *** [Makefile:91: all] Error 2

yairgott (changes requested):

BUILD_cmake.md
eifrah-aws: indeed, I used the "ssh" version instead of the "https" version

BUILD_cmake.md
eifrah-aws: 1. There is no need for the `submodule` call - removed
2. The docs now refers to the `build.sh` script

build.sh
eifrah-aws: 1. I removed this line completely as it is no longer needed
2. The script already detects if configure was run, however, if a new source file was added, you need to run `cmake` again (this is how cmake works), so we need the `--configure` option
3. Making "release" the default is OK I guess. (the reason for the current defaults: I mostly code these days in Rust, where cargo defaults to create release builds, one needs to manually specify the "--release" switch)
4. Same as 3: will do it

eifrah-aws: Will look into this

build.sh
eifrah-aws: CMake does not offer such capability. We could leverage the build script for this. For starter, we could:
- Place all tests artifacts in a separate folder (e.g. `bin/tests/`)
- `build.sh --run-test=[all|test-name]` if `all` is provided, we simply globe and run everything under the test folder (failing at the first error). if `test-name` is provided, we run that specific binary

WDYT?

testing/CMakeLists.txt
eifrah-aws: This was done by me on purpose. It's how I am used to work...

By using this approach, new comers, can locate the artifacts in a standard locations without needing to know which cmake target created it and search for the binary on that folder. Your way requires people to be familiar with the directory structure of the project

The "example" is probably built by one of the external cmake files, I will try to locate which one created it and exclude it from the build.

BUILD_cmake.md
eifrah-aws: Fixed

CMakeLists.txt
eifrah-aws: I will look into relaxing it to 3.16 (I am pretty sure there was a reason I used cmake 3.24). Ubuntu 20 is pretty outdated, usually, it is common to support 2 LTS versions (e.g. Ubuntu 22 & 24)

build.sh
eifrah-aws: Done

cmake/Modules/valkey_search.cmake
eifrah-aws: This is done automatically by CMake for build type Release

build.sh
eifrah-aws: Done

build.sh
eifrah-aws: This seems to be because the grpc / pb files are being re-generated unconditionally.

cmake/Modules/valkey_search.cmake
eifrah-aws: Done

cmake/Modules/valkey_search.cmake
eifrah-aws: I added `--run-test` to the `build.sh` script. It can be used like this:

```bash
# Run a single test by name
build.sh --run-test=search_test

# Run all tests
build.sh --run-test
```

After aligning all flags, all tests are passing

eifrah-aws: I am going to leave the change from BAZEL_BUILD -> CMAKE_BUILD into a separate PR

build.sh
eifrah-aws: Fixed

testing/CMakeLists.txt
eifrah-aws: I re-organized the directory structure a bit (assuming release build configuration):

- All test executables are placed under `.build-release/tests`
- Any external modules executable is placed under their directory (inside the build folder), so for example the binary `example` is kept under `.build-release/highwayhash` folder

Marking this issue as resolved, let me know if we want to further change it

eifrah-aws: Done

CMakeLists.txt
eifrah-aws: CMake minimum requirement reduced to 3.16

build.sh
yairgott: SGTM

cmake/Modules/valkey_search.cmake
yairgott: SG!

BUILD_cmake.md
```diff
@@ -0,0 +1,50 @@
...
+
+```bash
+git clone https://github.com/valkey-io/valkey-search.git
+cd valkey-search
+```
+
+Next, build the module for the `release` configuration by typing this into your terminal:
+
+```bash
+./build.sh --run-tests=all
```
yairgott: it seems that --run-tests=all is redundant, right?

src/coordinator/CMakeLists.txt
yairgott, line 1: I see the following error while running `./build.sh`:
[ 33%] Linking CXX executable sip_hash_test
[ 33%] Built target sip_hash_test
[ 34%] Linking CXX executable vector_test
[ 34%] Linking CXX executable example
[ 34%] Built target vector_test
[ 34%] Built target example
[ 35%] Linking CXX static library libtime_sliced_mrmw_mutex.a
[ 35%] Built target time_sliced_mrmw_mutex
[ 36%] Linking CXX executable highwayhash_test
[ 36%] Built target highwayhash_test
[ 37%] Linking CXX static library libmemory_allocation_overrides.a
[ 37%] Built target memory_allocation_overrides
[ 38%] Linking CXX static library libstring_interning.a
[ 38%] Built target string_interning
[ 39%] Linking CXX static library liblog.a
[ 39%] Built target log
[ 39%] Linking CXX static library libattribute_data_type.a
[ 39%] Built target attribute_data_type
[ 40%] Linking CXX executable ../../tests/status_macros_test
/usr/bin/ld: missing --end-group; added as last command line option
[ 40%] Built target status_macros_test
[ 40%] Linking CXX static library libcoordinator_grpc_proto.a
[ 40%] Built target coordinator_grpc_proto
make: *** [Makefile:91: all] Error 2

BUILD_cmake.md
eifrah-aws: Do you mean that the "=all" is redundant?

src/coordinator/CMakeLists.txt
eifrah-aws: I guess it's related to the previous error? (protoc)

BUILD_cmake.md
eifrah-aws: You shouldn't install `protobuf-compiler` as I explained earlier. The build is using the `protoc` compiled from `grpc`. Using another version can cause a mismatch in generated code, please do not install `protobuf-compiler`. If there is an issue, we should address it. I performed the steps on a clean instance - it worked. It also worked for @zvi-code (which uses arm ubuntu) so I wonder what is the difference in the setup

eifrah-aws: 1. You don't need to install `protobuf-compiler`
2. All references to `protobuf::protoc` were removed from the CMake files, did you run `./build.sh --configure` ?

Please do another `git pull` (I added a line that removes the `CMakeCache.txt` when running `./build.sh --configure`). Also, it would be great to see the **full** `./build.sh --configure` output

BUILD_cmake.md
yairgott: There is misalignment between the text and the bash command: The text says "Next, build the module for the `release` configuration by typing this into your terminal:" and the bash command runs the tests.

BUILD_cmake.md
eifrah-aws: This is now fixed. GCC / Clang versions are now set 12 / 16 (this is enforced by CMake)

BUILD_cmake.md
eifrah-aws: I updated the text:

```
Next, build & test the module for the `release` configuration by typing this into your terminal
```

src/coordinator/CMakeLists.txt
eifrah-aws: This is now fixed

BUILD_cmake.md
```diff
@@ -0,0 +1,75 @@
...
+sudo apt update
+sudo apt install -y clangd          \
+                    build-essential \
+                    g++             \
+                    cmake           \
+                    libgtest-dev    \
+                    libssl-dev
+```
+
+### `Ubuntu 22.04`
```
yairgott: I think that it makes better sense if we capture this as a note which is not particular to Ubuntu 22.04 as it's relevant also to Debian distribution under specific version and also lower Ubuntu versions. We can just extend the note from line #9 and move it here to say something along the lines:
**NOTE**: Building requires gcc/g++ version 12+ or clang/clang++ version 16+. 

For Debian/Ubuntu, in case a lower version is installed, you may upgrade to gcc/g++ 12 with:

```bash
sudo apt update
sudo apt install -y gcc-12 g++-12
sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-12 90
sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-12 90
```

eifrah-aws: Updated

yairgott (approved):
Well done!

PR #35, opened 2025-02-04
RFC: use versioning for edge repair in HNSW

vitarb:
Adding an RFC for improving hnsw library to allow better handling of deleted nodes in the graph.

### DIFF
--- a/rfc/replace-deleted.md
+++ b/rfc/replace-deleted.md
@@ -1,0 +1,238 @@
+# RFC: Versioning-Based Lazy (and Optional Synchronous) Repair in HNSW
+
+**Author(s):** Vitaly Arbuzov 
+**Created:** 02/04/2025 
+**Status:** Draft / Proposed
+
+---
+
+## 1. Overview and Motivation
+
+### 1.1 Background
+
+Hierarchical Navigable Small World (HNSW) graphs are widely used for Approximate Nearest Neighbor (ANN) search due to their efficient graph-based structure. However, in production systems that support dynamic updates (e.g., new data points arriving or existing ones being removed), deleted nodes are sometimes reused (via `allow_replace_deleted_` in hnswlib). This reuse introduces two major problems:
+
+- **Stale Inbound Links**: When a node is deleted and subsequently reused, any existing inbound links from other nodes’ adjacency lists point to a different “state.” These stale references mislead search queries.
+- **Recall Degradation**: Searches that traverse stale edges can produce suboptimal or incorrect nearest-neighbor results—an issue with potentially severe impact on high-recall vector search applications.
+
+### 1.2 Motivation for a Versioning-Based Repair Mechanism
+
+To mitigate the impact of stale links without fully rebuilding the index, this RFC proposes a lightweight versioning system that:
+
+- **Tracks Node Reuse**: Each node has a version counter incremented every time the node is reused (after being deleted).
+- **Detects Stale Edges**: Adjacency lists store both an identifier and the version of the target node at link creation time. During search, edges whose stored version doesn’t match the node’s current version are treated as invalid (stale).
+- **Supports Lazy Repair**: An asynchronous background process progressively removes or updates invalid edges, preventing them from repeatedly impacting query results.
+- **Enables Optional Synchronous Repair**: For environments where immediate consistency is essential (and some write latency is acceptable), stale edges can be repaired on the spot—trading off query-time locking for guaranteed consistency.
+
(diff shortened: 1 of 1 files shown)


rfc/replace-deleted.md
```diff
@@ -0,0 +1,238 @@
...
+- **Stale Inbound Links**: When a node is deleted and subsequently reused, any existing inbound links from other nodes’ adjacency lists point to a different “state.” These stale references mislead search queries.
+- **Recall Degradation**: Searches that traverse stale edges can produce suboptimal or incorrect nearest-neighbor results—an issue with potentially severe impact on high-recall vector search applications.
+
+### 1.2 Motivation for a Versioning-Based Repair Mechanism
+
+To mitigate the impact of stale links without fully rebuilding the index, this RFC proposes a lightweight versioning system that:
+
+- **Tracks Node Reuse**: Each node has a version counter incremented every time the node is reused (after being deleted).
+- **Detects Stale Edges**: Adjacency lists store both an identifier and the version of the target node at link creation time. During search, edges whose stored version doesn’t match the node’s current version are treated as invalid (stale).
+- **Supports Lazy Repair**: An asynchronous background process progressively removes or updates invalid edges, preventing them from repeatedly impacting query results.
```
yairgott, line 24: Can you clarify what you aim to achieve with node versioning?

Just a reminder: ValkeySearch uses a time-slice mechanism with mutual exclusion between reads and writes to ensure consistency. This guarantees that a mutation is either applied to all indexes or to none, ensuring atomicity. This consistency model applies to all indexes, not just vector indexes.

rfc/replace-deleted.md
```diff
@@ -0,0 +1,238 @@
...
+
+---
+
+## 1. Overview and Motivation
+
+### 1.1 Background
+
+Hierarchical Navigable Small World (HNSW) graphs are widely used for Approximate Nearest Neighbor (ANN) search due to their efficient graph-based structure. However, in production systems that support dynamic updates (e.g., new data points arriving or existing ones being removed), deleted nodes are sometimes reused (via `allow_replace_deleted_` in hnswlib). This reuse introduces two major problems:
+
+- **Stale Inbound Links**: When a node is deleted and subsequently reused, any existing inbound links from other nodes’ adjacency lists point to a different “state.” These stale references mislead search queries.
```
yairgott, line 15: Today reuse is not enabled.

rfc/replace-deleted.md
```diff
@@ -0,0 +1,238 @@
...
+- **Recall Degradation**: Searches that traverse stale edges can produce suboptimal or incorrect nearest-neighbor results—an issue with potentially severe impact on high-recall vector search applications.
```
yairgott, line 16: I'm uncertain about this given the current implementation, which doesn't reuse deleted nodes. On the other hand, since HNSWlib handles deletion by marking nodes as removed without reclaiming them, a large number of deleted nodes could lead to increased query runtimes.

rfc/replace-deleted.md
```diff
@@ -0,0 +1,238 @@
...
+*Note:* Although this additional memory overhead is significant, it is often acceptable in large-scale systems that already allocate substantial memory for vector data. The actual overhead will vary based on the exact layer distribution and on whether a 16-bit or 32-bit version counter is used.
+It's also possible to use fewer bits in the spin counter reducing total memory requirements, as in case of accidental match between node and edge version (after a full spin) connection will simply stay a little longer and be repaired after the next update and encounter.
+
+### 4.3 Repair Queue and Batch Processing
+
+- **Repair Queue**:  
+  Each encounter with a stale edge enqueues a repair task (e.g., `(owner_node_id, edge_index)` or `(neighbor_id, stored_version)`).
+
+- **Batch Repairs**:  
+  A background process periodically locks node adjacency lists, removes or fixes stale entries, and moves on.
```
yairgott, line 150: I think you should provide more details on how the node removal algorithm would work. Specifically:
1. Handling Removal of an Entry Point at a Higher Level: How would you accommodate the scenario where the removed node is an entry point at a higher layer?
2. Finding Incoming Links: How would you identify all nodes that have links pointing to the removed node, given that HNSW does not maintain reverse links?
3. Fixing Stale Entries: Could you clarify what you mean by "stale" entries? Additionally, how would you handle and clean up these entries to maintain graph integrity and avoid performance degradation?

yairgott (changes requested):

allenss-amazon:
closing, no activity in more than 8 months.

PR #47, opened 2025-02-16
Automatic configuration of threadpool size based on physical CPU core count

dpbnasika:
Automatic configuration of threadpool size based on physical CPU core count.

### DIFF
--- a/src/commands/ft_create_parser.cc
+++ b/src/commands/ft_create_parser.cc
@@ -53,6 +53,7 @@
 #include "vmsdk/src/command_parser.h"
 #include "vmsdk/src/status/status_macros.h"
 #include "vmsdk/src/type_conversions.h"
+#include "vmsdk/src/utils.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
 namespace valkey_search {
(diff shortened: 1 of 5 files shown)


dpbnasika:
@allenss-amazon , @yairgott some one can review this PR and let me know. i ran the tests in my local, they are passing.

allenss-amazon:
Is matching the # of threads to the # of cores actually the best configuration? @yairgott can you comment on that?

yairgott:
> @yairgott Can this be merged as it is for now for 'default' threads?, or does it need any refactoring?. IMO, creating new issues to remove some of the magic numbers in the code would be ideal, "few at a time" (as it could improve a bit of readability of the code as well).!!.

I suggest that the refactoring be done in another CL. I left a comment inline about the CPU cores vs vCPUs.

src/valkey_search.cc
```diff
@@ -85,8 +86,8 @@ constexpr absl::string_view kUseCoordinator{"--use-coordinator"};
 constexpr absl::string_view kLogLevel{"--log-level"};
 
 struct Parameters {
-  int reader_threads{10};
-  int writer_threads{30};
+  int reader_threads{static_cast<int>(std::thread::hardware_concurrency())};
```
yairgott: This captures the amount of vCPUs not the amount of physical CPU cores. Based on perf testings, we observed that the max performance is reached with matching the amount of threads to the amount of physical CPU cores. 

It seems that the standard doesn't have such API. The only cross platform C++ library which I found to support such is boost but I don't think that it would make sense to add boost dependency just for this. I suggest that we would use `std::thread::hardware_concurrency()` for any OS beside linux and and have some dedicated code to Linux which captures the amount of physical cores. For instance, such can be fetched from the procfs.

 wdyt?

dpbnasika: @yairgott  I agree with you, i will change that and add a piece of code specific for Linux. I will use either proc/cpuinfo and parse the file (or) use linux sys file system cpu info to get that,  I will try to add that and let you know.!

dpbnasika: @yairgott should we add that functionality as a tightly coupled one inside the valkey_search.cc source or shall i create a separate header & src file and add it somewhere in the src/utils folder and configure those build files?. what do you think and suggest?. (do you see this function implementation or anything related could be extended in the future or just remains only here in this particular case ?). lmk, what do you think?..

yairgott: I suggest that we put this function under vmsdk. Fwiw, vmsdk stands for valkey module sdk which we eventually aim to extract it out into a separate repo. Similar concept as Valkeymodule-rs but for C++. 
I suggest putting this either in utils.h/cpp or you can create a new file concurrency.h/cpp. In such case, you can move into to it some additional functionality from utils.cpp such as the one related to the main thread execution verification.

wdyt?

PR #49, opened 2025-02-25
Add ACL support to Valkey Search.

jeon1226:


src/acl.cc
yairgott, line 1: I suggest that we add the acl functionality to vmsdk. Soon we would need such in the json module.

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+                                 bool nocase, bool *skip_longer_matches) {
+  while (pattern_len && string_len) {
+    switch (pattern[0]) {
+      case '*':
+        while (pattern_len && pattern[1] == '*') {
+          pattern++;
+          pattern_len--;
+        }
+        if (pattern_len == 1) {
+          return 1; /* match */
```
yairgott: nit: here and in other places, return true/false

testing/acl_test.cc
yairgott, line 1: I've noticed that valkey has fuzz test: https://github.com/valkey-io/valkey/blob/unstable/src/util.c#L199

Can we add it as a unittest?

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+  std::vector<acl::ValkeyAclGetUserReplyView> acl_views;
+  GetAclViewFromCallReplyImpl(acl_views, reply);
+  return acl_views;
+}
+
+absl::Status AclManager::AclPrefixCheck(
+    RedisModuleCtx *ctx,
+    const std::unordered_set<absl::string_view> &module_allowed_cmds,
+    const std::vector<std::string> &module_prefixes) {
+  RedisModuleString *username = RedisModule_GetCurrentUserName(ctx);
```
yairgott: Please use vmsdk::UniquePtrRedisString to manage the lifecycle of username, instead of calling RedisModule_FreeString: https://github.com/valkey-io/valkey-search/blob/main/vmsdk/src/managed_pointers.h#L52C26-L52C46

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+  RedisModuleCallReply *reply =
```
yairgott: also vmsdk::UniquePtrRedisCallReply for managing the reply object.

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+  };
+  VMSDK_ASSIGN_OR_RETURN(auto acl_views, GetAclViewFromCallReply(reply));
+
+  auto acl_keys = GetKeysWithAllowedCommands(module_allowed_cmds, acl_views);
+
+  // If no index prefixes are given from the module, it should have permission
+  // to access ALL keys.
+  if (module_prefixes.empty() && IsPrefixAllowed("", acl_keys)) {
+    return absl::OkStatus();
+  } else {
```
yairgott: nit: else is not required and just adds indentation to the code.

testing/acl_test.cc
yairgott, line 1: I wonder if it would be useful to port also the valkey unittests to make sure that our implementation is aligned with the engine.

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+    const data_model::IndexSchema &index_schema_proto) {
+  std::vector<std::string> module_prefixes;
+  module_prefixes.reserve(index_schema_proto.subscribed_key_prefixes_size());
+  for (const auto &prefix : index_schema_proto.subscribed_key_prefixes()) {
+    module_prefixes.emplace_back(prefix);
+  }
+  return AclPrefixCheck(ctx, module_allowed_cmds, module_prefixes);
+}
+
+}  // namespace valkey_search
```
yairgott: I think it's saying that an empty line is missing.

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+    return absl::FailedPreconditionError(
+        "Cannot get an ACL from the server, got an invalid response");
+  }
+
+  std::vector<acl::ValkeyAclGetUserReplyView> acl_views;
+  GetAclViewFromCallReplyImpl(acl_views, reply);
+  return acl_views;
+}
+
+absl::Status AclManager::AclPrefixCheck(
```
yairgott: I wonder how expensive this function is. Maybe there is a way to cache the results and be notified on acl changes to flush the cache?

yairgott (changes requested):

src/acl.cc
yairgott: I think we can hold this off for now until we see concrete use-case which require this.

src/acl.cc
yairgott: This is probably more applicable while we'll support non-vector queries. For now, I suggest that we just run sanity performance regression test to make sure there are no surprises. This perf test shouldn't take it to the extreme, decent amount of ACLs with an index using a dozen or so prefixes.

src/acl.h
```diff
@@ -0,0 +1,73 @@
...
+
+  absl::Status AclPrefixCheck(
+      RedisModuleCtx *ctx,
+      const std::unordered_set<absl::string_view> &module_allowed_cmds,
+      const data_model::IndexSchema &index_schema_proto);
+  /*
+  Internally, we use RedisModule_Call API to get the ACL ruls from the server,
+  i.e. "ACL GETUSER alice"
+
+  The reply of RedisModule_Call API is complicated,
```
yairgott: It looks like the main reason for `AclManager` is to avoid mocking the RedisModule_* APIs. In general, testing dictating the design is suboptimal. In addition, by overriding `AclManager::GetAclViewFromCallReply` we're missing coverage on key code areas, which will force more integration tests later.

I think that a better approach is to:
1. mock RedisModule_* . Note that we already have solid framework for doing so. Most of those APIs [already](https://github.com/valkey-io/valkey-search/blob/21b93d5d50eb2a499a1229301893dc9f1178cfe1/vmsdk/src/testing_infra/module.h#L233) have mocking support and adding mocks for the remaining APIs should be relatively simple.
2.  To simplify mock usage, we can represent the expected reply as a json string.
3. During the mock API calls, we can use the parsing context to generate JSON path queries and validate the calls against the JSON.

Happy to further discuss this.

src/commands/ft_create.cc
```diff
@@ -40,6 +41,8 @@ absl::Status FTCreateCmd(RedisModuleCtx *ctx, RedisModuleString **argv,
                          int argc) {
   VMSDK_ASSIGN_OR_RETURN(auto index_schema_proto,
                          ParseFTCreateArgs(ctx, argv + 1, argc - 1));
+  VMSDK_RETURN_IF_ERROR(AclManager::Instance().AclPrefixCheck(
+      ctx, kCommandCategories.at(kCreate), index_schema_proto));
```
yairgott: Here and in all other AclPrefixCheck calls: I wonder if using `ctx` might lead RedisModule_Call to fail due to ACL enforced by the engine. The alternative is to use [valkey_search::ctx_](https://github.com/valkey-io/valkey-search/blob/21b93d5d50eb2a499a1229301893dc9f1178cfe1/src/valkey_search.h#L134)

testing/acl_test.cc
jeon1226: The purpose of the fuzztest in valkey is to see if there's a crash. The fuzztest we want to add here for ValkeySearch ACL has the same purpose?

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+}
+
+namespace {
+
+/* This pattern matching code mostly came from valkey/src/util.c,
+ with some modifications for a pattern-to-pattern match, not a pattern-to-string
+ match. It returns true when (1) the pattern matches the string, AND (2) the
+ pattern ends with wildcards.
+ */
+bool StringEndsWithWildCardMatch(const char *pattern, int pattern_len,
```
allenss-amazon, line 22: This code look quite brittle. Tracking the pointers and lengths should be combined into convenience functions.  I'd recommend using string_view here with it's associated manipulation functions.

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+                                 const char *string, int string_len,
+                                 bool nocase, bool *skip_longer_matches) {
+  while (pattern_len && string_len) {
+    switch (pattern[0]) {
+      case '*':
+        while (pattern_len && pattern[1] == '*') {
```
allenss-amazon: This looks like a potential memory violation. How do we know if [1] is valid?

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+        pattern++;
+        pattern_len--;
+        negate = pattern[0] == '^';
+        if (negate) {
+          pattern++;
+          pattern_len--;
+        }
+        match = 0;
+        while (1) {
+          if (pattern[0] == '\\' && pattern_len >= 2) {
```
allenss-amazon: Please reorder to avoid accessing [0] before the length check.

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+    } else if (module_allowed_cmds.contains(
+                   absl::string_view(acl_cmd.data() + 1, acl_cmd.size() - 1))) {
+      if (cmd.starts_with('+')) {
+        allowed = true;
+      } else if (cmd.starts_with('-')) {
+        allowed = false;
+      } else {
+        // This should not happen. All ACL command expression starts with +, -,
+        // or all/no command alias.
+        abort();
```
allenss-amazon: RedisModule_Assert

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+permit read and write operations. Vector Search only read the keys with given
+prefixes and never writes, so in here we assume that we need %R permission is
+required, not %W.
+*/
+bool IsPrefixAllowed(const absl::string_view &module_prefix,
+                     const std::vector<absl::string_view> &acl_keys) {
+  bool result = false;
+  for (const auto &acl_key : acl_keys) {
+    if (acl_key == "allkeys") {
+      result = true;
```
allenss-amazon, line 259: return true?

src/acl.cc
```diff
@@ -0,0 +1,358 @@
...
+    } else if (acl_key == "resetkeys") {
+      result = false;
+    } else if (acl_key.starts_with("%W")) {
+      continue;
+    } else {
+      int offset = acl_key.find('~') + 1;  // either one of ~, %R~, and %RW~
+      result = result || StringEndsWithWildCardMatch(
```
allenss-amazon: Once you get a true, why bother continuing in the loop?

src/acl.cc
jeon1226: Not sure how to get notified on acl changes - is there a built in topic for acl change? If not we need to change the server code.

The ValkeySearch ACL RFC states that no security information is cached. https://github.com/valkey-io/valkey-rfc/pull/18/files#diff-ecbdf00816a3a420fd7225875f5d296283e9fada434cef7baf279184f34b9679R43 So I didn't consider to store any acl information.

allenss-amazon: I believe you could hook out the ACL update functions.

src/acl.cc
allenss-amazon: Personally, in situations like this, I like what was done. The extra indentation isn't excessive.

src/acl.cc
yairgott: Actually, the recommended way is to use absl CHECK

testing/acl_test.cc
yairgott: Yes. If such exists, I don't see a reason not to just wrap it in a unittest.

src/acl.cc
yairgott: hmm, this may be related to code styling but note that throughout the code base this guideline is followed.

src/acl.cc
jeon1226: @allenss-amazon Do you mean the hook (a server event) should be added to Valkey? I cannot find the current Module API for an ACL update event.
https://valkey.io/topics/modules-api-ref/#section-server-hooks-implementation

src/acl.cc
jeon1226: `pattern_len` is not zero, and in this case we believe `pattern_len` is maintained properly as a length of `pattern`

src/acl.cc
jeon1226: The key patterns of Valkey Acl rule is cumulative and path dependent.

"allkeys nokeys allkeys" is all key allowed, however,
"allkeys nokeys allkeys nokeys" is no key allowed

so we need to see all items in `acl_keys` through the end.

src/acl.cc
jeon1226: The same logic as above. Even if we got all permissions at this point, if the last `acl_key` is `nokeys` or `-@all` no keys are allowed.

src/acl.cc
jeon1226: My biggest worry is, a hand crafted pattern matching code is doing something weird, while the original code is verified to work well. In this case I want to use the library code as much as possible. If that is not an option, I'll copy and paste the library code and modify it as small as possible.

The code here is from valkey/src/util.c. I agree that the pointer and length approach is scary, but deviating from the original code too much could be more dangerous, considering this is supposed to be a temp solution (after this this should be changed into some valkey module api with the same functionality). The safety issue (crash) can be handled by fuzz testing.

src/acl.cc
jeon1226: Oh, it is not a null terminated string, so you're right.

src/acl.cc
yairgott: Not sure if there is a better way but one approach is to override the ACL command and proxy it between the engine and the caller. Before returning the response back to the caller, fetch and cache the ACLs from the engine using ValkeyModule_Call.

src/commands/commands.h
```diff
@@ -30,18 +30,45 @@
 #ifndef VALKEYSEARCH_SRC_COMMANDS_COMMANDS_H_
 #define VALKEYSEARCH_SRC_COMMANDS_COMMANDS_H_
 
+#include <unordered_set>
```
yairgott: Better to use absl::flat_hash_set

yairgott (changes requested):

src/acl.cc
```diff
@@ -0,0 +1,357 @@
...
+  for (const auto acl_cmd : std::views::split(acl_cmds, ' ')) {
+    auto cmd = absl::string_view(acl_cmd.data(), acl_cmd.size());
+    if (absl::EqualsIgnoreCase(cmd, "+@all") ||
+        absl::EqualsIgnoreCase(cmd, "allcommands")) {
+      allowed = true;
+    } else if (absl::EqualsIgnoreCase(cmd, "-@all") ||
+               absl::EqualsIgnoreCase(cmd, "nocommands")) {
+      allowed = false;
+    } else if (module_allowed_cmds.contains(
+                   absl::string_view(acl_cmd.data() + 1, acl_cmd.size() - 1))) {
```
yairgott, line 199: accessing acl_cmd.data() + 1 without size check is not safe.

src/acl.cc
```diff
@@ -0,0 +1,357 @@
...
+against the key patterns. We only need the key patterns that the corresponding
+commands matches our module commands.
+*/
+std::vector<absl::string_view> GetKeysWithAllowedCommands(
+    const std::unordered_set<absl::string_view> &module_allowed_cmds,
+    const std::vector<acl::ValkeyAclGetUserReplyView> &acl_views) {
+  std::vector<absl::string_view> keys;
+  for (const auto &acl_view : acl_views) {
+    if (IsAllowedCommands(module_allowed_cmds, acl_view.cmds)) {
+      for (const auto acl_key : std::views::split(acl_view.keys, ' ')) {
```
yairgott: use `absl::StrSplit(acl_view.keys, ' ', absl::SkipEmpty());` to skip empty.

src/acl.cc
```diff
@@ -0,0 +1,357 @@
...
+}  // namespace acl
+
+namespace {
+
+bool IsAllowedCommands(
+    const std::unordered_set<absl::string_view> &module_allowed_cmds,
+    absl::string_view acl_cmds) {
+  bool allowed = false;
+
+  for (const auto acl_cmd : std::views::split(acl_cmds, ' ')) {
```
yairgott: use `absl::StrSplit(acl_cmds, ' ', absl::SkipEmpty());` to skip empty.

src/acl.cc
```diff
@@ -0,0 +1,357 @@
...
+  bool result = false;
+  for (const auto &acl_key : acl_keys) {
+    if (acl_key == "allkeys") {
+      result = true;
+    } else if (acl_key == "resetkeys") {
+      result = false;
+    } else if (acl_key.starts_with("%W")) {
+      continue;
+    } else {
+      int offset = acl_key.find('~') + 1;  // either one of ~, %R~, and %RW~
```
yairgott: should you check for std::string_view::npos?

yairgott (changes requested):

testing/acl_test.cc
jeon1226: I checked valkey unit tests related to acl (v1 and v2) and utils, and I didn't find any sophisticated tests for blobs.

src/acl.cc
```diff
@@ -0,0 +1,357 @@
...
+  }
+  return absl::PermissionDeniedError(
+      "The user doesn't have a permission to execute a command");
+}
+
+absl::Status AclPrefixCheck(
+    RedisModuleCtx *ctx,
+    const std::unordered_set<absl::string_view> &module_allowed_cmds,
+    const data_model::IndexSchema &index_schema_proto) {
+  std::vector<std::string> module_prefixes;
```
yairgott, line 357: nit: here and everywhere else, IMO a clearer name is `schema_prefixes`

yairgott (approved):

PR #64, opened 2025-03-06
Fix parsing double/float from string

yairgott:
absl::SimpleAtod has a special handling for "nan" where it returns true but copy "nan" to the out value. Addressing this by checking for the string "nan".

This PR addresses the issue described in https://github.com/valkey-io/valkey-search/issues/53

### DIFF
--- a/src/commands/filter_parser.cc
+++ b/src/commands/filter_parser.cc
@@ -118,7 +118,8 @@ absl::StatusOr<double> FilterParser::ParseNumber() {
   while (!IsEnd() && (std::isdigit(Peek()) || Peek() == '.')) {
     number_str += expression_[pos_++];
   }
-  if (absl::SimpleAtod(number_str, &value)) {
+  if (absl::AsciiStrToLower(number_str) != "nan" &&
+      absl::SimpleAtod(number_str, &value)) {
     return value * multiplier;
   }
   return absl::InvalidArgumentError(
--- a/src/indexes/numeric.cc
+++ b/src/indexes/numeric.cc
@@ -53,7 +53,7 @@ namespace valkey_search::indexes {
 namespace {
 std::optional<double> ParseNumber(absl::string_view data) {
   double value;
-  if (!absl::SimpleAtod(data, &value)) {
+  if (absl::AsciiStrToLower(data) == "nan" || !absl::SimpleAtod(data, &value)) {
     return std::nullopt;
   }
   return value;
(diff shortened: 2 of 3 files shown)


src/commands/filter_parser.cc
```diff
@@ -118,7 +118,8 @@ absl::StatusOr<double> FilterParser::ParseNumber() {
   while (!IsEnd() && (std::isdigit(Peek()) || Peek() == '.')) {
     number_str += expression_[pos_++];
   }
-  if (absl::SimpleAtod(number_str, &value)) {
+  if (absl::AsciiStrToLower(number_str) != "nan" &&
```
allenss-amazon, line 121: This would cause a nan to be treated the same as a missing or invalid field, right? Imo, that's wrong. Nan is a valid double precision number. That's how Redisearch would treat it.

Also -nan is valid.

allenss-amazon: I think you could just extend the matching for +/-Inf to include nan.

src/indexes/numeric.cc
```diff
@@ -53,7 +53,7 @@ namespace valkey_search::indexes {
 namespace {
 std::optional<double> ParseNumber(absl::string_view data) {
   double value;
-  if (!absl::SimpleAtod(data, &value)) {
+  if (absl::AsciiStrToLower(data) == "nan" || !absl::SimpleAtod(data, &value)) {
```
allenss-amazon, line 56: same here. Also, this seems different than the handling above for +/- infinity. IMO, these should get combined into a single parse-float function.

vmsdk/src/type_conversions.h
```diff
@@ -101,7 +101,7 @@ static inline absl::StatusOr<T> ToNumeric(absl::string_view str) {
 template <>
 inline absl::StatusOr<float> To(absl::string_view str) {
   float value;
-  if (!absl::SimpleAtof(str, &value)) {
+  if (absl::AsciiStrToLower(str) == "nan" || !absl::SimpleAtof(str, &value)) {
```
allenss-amazon, line 104: same here.

allenss-amazon (commented):
I think "nan" and "-nan" should be accepted and converted to true NaN values for processing later.

src/commands/filter_parser.cc
yairgott: As discussed offline, RediSearch treats any non numeric value as Nan.  The only change which is required, which is covered by this PR, is to address absl's special handling of Nan.

allenss-amazon:
After conversation this is ok.

allenss-amazon (approved):
after discussion

PR #68, opened 2025-03-13
Implement new RDB format

### LINKED ISSUE #27: Implement the new RDB format

murphyjacob4:
Closes #27

See #28 for the detailed design.

Some additional comments:

1. This commit wraps HSNW and bruteforce index serialization headers in a proto for easy serialization and future modification. This should have minimal impact on the actual serialized size, but improves our extensibility greatly and makes the index save much easier (proto handles portability issues like endianness)
2. Use nested iterators to iterate through the RDBSections, SupplementalContent sections, and chunked supplemental content. I implemented these with a simple iterator interface (Next(), HasNext()) instead of using C++ idiomatic iterators (begin(), end() and ++/== for iteration) since it is a one way iterator, and for chunked data we need to preload one ahead to know if we have next content or not.
3. Aggregate all RDB loading and saving logic into new rdb_serialization.cc/h files. The module now has a single module type and a single entrypoint for saving/loading aux metadata. Data types that would like to save an RDB section are able to register with this centralized point through a callback system.
4. Added unit tests in this commit, but integration tests will follow
5. Finally, made some edits to proto loading in Cmake to support proto loading another proto as a dependency. Adding @eifrah-aws as a reviewer to help check my work


src/rdb_serialization.h
```diff
@@ -0,0 +1,363 @@
...
+
+  virtual absl::Status SaveDouble(double val) {
+    RedisModule_SaveDouble(rdb_, val);
+    if (RedisModule_IsIOError(rdb_)) {
+      return absl::InternalError("RedisModule_SaveDouble failed");
+    }
+    return absl::OkStatus();
+  }
+
+  virtual absl::Status SaveStringBuffer(const char *str, size_t len) {
```
yairgott: nit: receive absl:string_view

src/rdb_serialization.h
```diff
@@ -0,0 +1,363 @@
...
+                         semantic_version & 0xFF);
+}
+
+/* SafeRDB wraps a RedisModuleIO object and performs IO error checking,
+ * returning absl::StatusOr to force error handling on the caller side. */
+class SafeRDB {
+ public:
+  explicit SafeRDB(RedisModuleIO *rdb) : rdb_(rdb) {}
+
+  virtual absl::Status LoadSizeT(size_t &val) {
```
yairgott: Consider:
1. Having the type as part of the method name doesn't has any value. Instead we can just overload the Load function name. This also applies to Save*.
2. Load should not receive any parameter but return `absl::Status<T>`

src/rdb_serialization.h
```diff
@@ -0,0 +1,363 @@
...
+    if (!closed_ && !Close().ok()) {
+      VMSDK_LOG(WARNING, nullptr)
+          << "Failed to write final chunk on closing output stream";
+    }
+  }
+  absl::Status SaveChunk(const char *data, size_t len) override;
+  absl::Status Close();
+
+ private:
+  SafeRDB *rdb_;
```
yairgott, line 369: consider using rdb_ == nullptr as indicator of close and eliminate the additional variable closed_.

src/indexes/index_base.h
```diff
@@ -77,7 +77,7 @@ class IndexBase {
   virtual int RespondWithInfo(RedisModuleCtx* ctx) const = 0;
   virtual bool IsTracked(const InternedStringPtr& key) const = 0;
   IndexerType GetIndexerType() const { return indexer_type_; }
-  virtual absl::Status SaveIndex(RDBOutputStream& rdb_stream) const = 0;
+  virtual absl::Status SaveIndex(RDBChunkOutputStream chunked_out) const = 0;
```
yairgott, line 80: Here and other places, prefer passing by reference.

src/indexes/vector_base.cc
```diff
@@ -447,9 +454,15 @@ void VectorBase::ExternalizeVector(RedisModuleCtx *ctx,
 
 absl::Status VectorBase::LoadTrackedKeys(
     RedisModuleCtx *ctx, const AttributeDataType *attribute_data_type,
-    const data_model::TrackedKeys &tracked_keys) {
+    SupplementalContentChunkIter iter) {
```
yairgott: Prefer passing the iter by reference.

testing/index_schema_test.cc
```diff
@@ -1058,74 +1011,24 @@ TEST_F(IndexSchemaRDBTest, SaveAndLoad) ABSL_NO_THREAD_SAFETY_ANALYSIS {
...
-  std::vector<absl::string_view> key_prefixes = {"prefix1", "prefix2"};
-  std::string index_schema_name_str("index_schema_name");
-  int dimensions = 100;
-  int initial_cap = 12;
-  int m = 16;
-  int ef_construction = 100;
-  int ef_runtime = 5;
-  int block_size = 250;
-
-  FakeRDBIOStream rdb_stream(dump_rdb, dump_rdb_len);
```
yairgott, line 1118: I believe that you can delete now the files testing/proto_package_rename_compatibility.h/cc.

cmake/Modules/protobuf_generate.cmake
```diff
@@ -200,13 +200,13 @@ endfunction()
 # "PROTO_FILE_SRC_DIR" contains the folder relative path to the proto file from
 # top level source directory. "PROTO_BASE_FILE_NAME" contains the file name (no
 # extension). "OUT_LIBNAME" is the user provided target name to create
-function(create_proto_library PROTO_FILE_SRC_DIR PROTO_BASE_FILE_NAME
+function(create_proto_library PROTO_PATH
```
eifrah-aws, line 203: Please update the comment above this function to reflect the changes

third_party/hnswlib/bruteforce.h
eifrah-aws, line 1: Since this file (and others below) are from the "third_party" folder of the code base, I would only change / edit these file without changing the coding style (e.g. 2 spaces vs 4 spaces) - this reduces the diff overhead and makes it easier to review

eifrah-aws (dismissed):
I mainly focused on CMake changes - LGTM

src/indexes/index_base.h
murphyjacob4: RDBChunkOutputStream is a move-only type. I am using pass by value with std::move - this way the RDBChunkOutputStream ownership is transferred and then closed on destruction. Closing on destruction becomes more difficult if it is passed by reference - since the caller will need to explicitly close the output stream. Let me know if this ownership model is confusing or you prefer an alternative

src/indexes/vector_base.cc
murphyjacob4: Since SupplementalContentChunkIter is a move-only type - it should be cheap to pass by value. Just like you would pass std::unique_ptr by value. It signifies the ownership of the iter is being transferred.

But I can make this explicit by passing rvalue reference

src/rdb_serialization.h
murphyjacob4: Is there a big difference between LoadSizeT and templatizing and doing Load<size_t>? Load<T> won't work since we can't do templatized virtual methods, and for testability it helps to subclass this.

src/rdb_serialization.h
murphyjacob4: I actually prefer using a seperate variable - I think it is more self-documenting than overloading the rdb_==nullptr to mean a special state.

third_party/hnswlib/bruteforce.h
murphyjacob4: Yeah - I added this in the PR description. I think we missed formatting these in the initial commit to the repo. We should probably do reformatting in a different PR though.

I will go ahead with reverting these

cmake/Modules/protobuf_generate.cmake
murphyjacob4: Done

src/rdb_serialization.h
murphyjacob4: For 2 - makes sense and done

src/rdb_serialization.h
murphyjacob4: Done

testing/index_schema_test.cc
murphyjacob4: Good point. Done

src/rdb_serialization.cc
```diff
@@ -0,0 +1,335 @@
...
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+
+namespace valkey_search {
+
+absl::flat_hash_map<data_model::RDBSectionType, RDBSectionCallbacks>
+    kRegisteredRDBSectionCallbacks = {};
+
+absl::StatusOr<std::unique_ptr<data_model::SupplementalContentChunk>>
+SupplementalContentChunkIter::Next() {
+  if (curr_chunk_.ok()) {
```
yairgott, line 58: This implementation is a little strange as curr_chunk_ is a head of next by 1. Why not just read the next chunk?

src/rdb_serialization.h
yairgott: For better readability, you can create a private Close function which returns rdb_== nullptr

src/rdb_serialization.h
yairgott: Yeah, probably for Load* is harder to overload but Save* seems to me more straights forward.

src/indexes/index_base.h
yairgott: Can SaveIndex be changed to receive a rvalue reference? e.g. SaveIndex(RDBChunkOutputStream && chunked_out)? 

I'm not sure that I follow the motivation for passing by value with move. Why can't the callee issue close on the chunked_out when done?

src/indexes/vector_base.cc
yairgott: Yeah,  rvalue reference would be much clearer.

src/indexes/index_base.h
murphyjacob4: It is to use RAII - so we want the scope of the variable to be tied to the end of the usage of the RDBChunkOutputStream. By passing by value, it forces the value to be move constructed in the function, which will be destructed at the end of its scope.

Consider (simplified):

```
save_handler(SafeRDB *rdb) {
  RDBChunkOutputStream chunked_out1(rdb);
  save_stuff(std::move(chunked_out1));
  RDBChunkOutputStream chunked_out2(rdb);
  save_stuff(std::move(chunked_out2));
}

save_stuff(RDBChunkOutputStream &&chunked_out) {
   // Output to the stream
}
```

If it is an rvalue reference, no move construction happens, so the destruction of both chunked_out1 and chunked_out2 happens at the end of save_handler. If it is pass by value, like:

```
save_stuff(RDBChunkOutputStream chunked_out) {
```

Then move construction happens, and chunked_out1 is moved to the save_stuff function and destructed at the end of the first call to save_stuff, and chunked_out2 at the end of the second call. By using RAII and move-only values, it guarantees that the object will be destructed only once, but we need to scope the destruction to the proper location or we might end up missing the EOF marker.

src/rdb_serialization.h
murphyjacob4: I think this is just a matter of preference

src/rdb_serialization.cc
murphyjacob4: We need to read one ahead since otherwise we can't implement HasNext(). Ideally we could have a "peek" style RDB module API to check what the next value is without consuming it - but in lieu of that we need to store the next value ahead of time to be able to know if we have a next value or if we have hit EOF

src/rdb_serialization.h
murphyjacob4: If we overload the `Save` function, I think we still need templatization. I.e:

```
SafeRDB rdb;
int64_t value = 34;
rdb.Save(value);
```

Will give:

```
more than one instance of overloaded function "valkey_search::SafeRDB::Save" matches the argument list:C/C++(308)
```

Having the different functions for different types matches the wrapped Valkey module API, so it doesn't seem too messy for me

murphyjacob4:
Hi folks, we need to get this in for the RC1 release. Please take another pass and approve if there is no further feedback. Thanks!

src/indexes/index_base.h
yairgott: Make sense, thanks for the elaborated explanation!

src/rdb_serialization.cc
yairgott: Yeah, logically it would make more sense if HasNext would buffer the processed data. Not sure if this would complicate the implementation.

src/rdb_serialization.h
yairgott: I agree but in general, I find that maintaining a single variable reduces the risk of the 2 variables becoming not aligned.

src/rdb_serialization.cc
murphyjacob4: If we are anyways going to buffer it - is there a difference between doing it on `HasNext` and on construction/`Next`? If there is no difference - I would prefer we keep the existing implementation to reduce unnecessary work

murphyjacob4: HasNext having side effects feels unclean to me - having construction side effects feels slightly less evil

src/rdb_serialization.h
yairgott: hmm, not sure what am I missing but I evaluated using dummy code:
```
yairg@yairg1:/tmp$ cat a.cc
#include <stdio.h>

class Test {
 public:
  bool Print(int a) {
    printf("hello %d\n", a);
    return true;
  }
  bool Print(size_t a) {
    printf("hello %lu\n", a);
    return true;
  }
};

int main() {
  Test t;
  size_t a = 1;
  int b = 2;
  if (t.Print(a)) {
    t.Print(b);
  }
  return 0;
}
yairg@yairg1:/tmp$ g++ a.cc
yairg@yairg1:/tmp$ ./a.out 
hello 1
hello 2
```

yairgott: I'm speculating that if you were using a type which is explicitly supported by SafeRDB::Save, .e.g. size_t or double, you wouldn't get this warnings.

src/rdb_serialization.h
yairgott: Anyway, this is not a must.

yairgott (approved):

src/rdb_serialization.h
murphyjacob4: Hmm, correct, looks like we have some spots where int64_t was passed, and it won't implicitly convert this to size_t.

Regardless - I think inferring what module API to call based on the passed in type is less clean than making this explicit. With the explict approach - it forces the caller to consider the type and it wouldn't be inferred based on the variable type. I can imagine someone accidentally changing the value passed in to `Save` from "int" to "size_t" between versions, which would implicitly change the save format and cause a RDB compat issue. Ideally we could use templates to do this - but as previously mentioned there are challenges there.

To demonstrate my point, consider

v1
```
unsigned int do_some_calculation() {
    // do some work
}

void my_func() {
    auto result = do_some_calculation();
    SafeRDB rdb;
    rdb.Save(result);
}
```

v2
```
size_t do_some_calculation() {
    // do some work
}

void my_func() {
    auto result = do_some_calculation();
    SafeRDB rdb;
    rdb.Save(result);
}
```

This would actually create a big compatibility issue - but will be hard to catch if someone is just editing do_some_calculation (which could even be in a different source file).

yairgott: Along the lines of your example, calling SaveSizeT with int64_t wouldn't lead to any compilation issues while it would if you overload Save. I don't think that there is a bullet proof solution. I would just note that serialization mechanisms are commonly based on output streams which overload the input operator.

PR #73, opened 2025-03-21
fixing integration tests to work with cmake

yairgott:


.devcontainer/Dockerfile
```diff
@@ -25,9 +23,16 @@ RUN apt-get install -y \
...
+    python3-pip \
+    python3.12-venv \
+    locales-all \
+    locales \
     sudo
 
-RUN rm -rf /var/lib/apt/lists/*
+RUN apt install lsb-release curl gpg
+RUN curl -fsSL https://packages.redis.io/gpg | sudo gpg --dearmor -o /usr/share/keyrings/redis-archive-keyring.gpg
+RUN echo "deb [signed-by=/usr/share/keyrings/redis-archive-keyring.gpg] https://packages.redis.io/deb $(lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/redis.list
```
eifrah-aws, line 34: hmm... as a member of the valkey org, kinda a hard for me to see a link for redis ;)
If valkey is what we need, why not downloading the source and build it from sources? its easier and pretty fast

.devcontainer/Dockerfile
```diff
@@ -25,9 +23,16 @@ RUN apt-get install -y \
...
+RUN apt update && apt-get install -y memtier-benchmark
```
eifrah-aws, line 35: Any particular reason we are using memtier and not valkey-benchmark?

build.sh
```diff
@@ -23,14 +24,15 @@ function print_usage() {
...
-    --test-errors-stdout When a test fails, dump the captured tests output to stdout.
+    --help | -h               Print this help message and exit.
+    --configure               Run cmake stage (aka configure stage).
+    --verbose | -v            Run verbose build.
+    --debug                   Build for debug version.
+    --clean                   Clean the current build configuration (debug or release).
+    --run-tests               Run all tests. Optionally, pass a test name to run: "--run-tests=<test-name>".
+    --no-build                By default, build.sh always triggers a build. This option disables this behavior.
+    --test-errors-stdout      When a test fails, dump the captured tests output to stdout.
+    --run-integration-tests   Run integration tests.
```
eifrah-aws, line 35: Question: do we want to add the ability to run a single integ test?

cmake/Modules/valkey_search.cmake
```diff
@@ -112,9 +112,11 @@ function(valkey_search_target_update_compile_flags TARGET)
   endif()
   target_compile_options(${TARGET} PRIVATE -Wno-sign-compare)
   target_compile_options(${TARGET} PRIVATE -Wno-uninitialized)
+  target_compile_options(${TARGET} PRIVATE -g)
 endfunction()
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-missing-requires")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
```
eifrah-aws: Q: this one sets it globally, do we want to set debugging symbols regardless of the build type?

testing/integration/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+        RESULT_VARIABLE VENV_RESULT
+        OUTPUT_QUIET ERROR_QUIET
+    )
+    if(NOT VENV_RESULT EQUAL 0)
+        message(FATAL_ERROR "Failed to create virtual environment.")
+    endif()
+endif()
+
+# Set paths to virtual environment executables
+if(WIN32)
```
eifrah-aws: Do we support windows?

testing/integration/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+    execute_process(
+        COMMAND ${PYTHON_VENV_EXECUTABLE} -m pip install -r ${CMAKE_SOURCE_DIR}/requirements.txt
+        RESULT_VARIABLE PIP_RESULT
+        OUTPUT_QUIET ERROR_QUIET
+    )
+    if(NOT PIP_RESULT EQUAL 0)
+        message(FATAL_ERROR "Failed to install dependencies from requirements.txt")
+    endif()
+else()
+    message(WARNING "requirements.txt not found! Python dependencies may be missing.")
```
eifrah-aws: This should be fatal, as it means that files were deleted from the repository. So I would change the logic here to:

```cmake
if (EXISTS "${CMAKE_SOURCE_DIR}/requirements.txt")
  message(FATAL_ERROR "requirements.txt not found! Python dependencies may be missing.")
endif ()

# other logic here...
```

testing/integration/run.sh
```diff
@@ -0,0 +1,184 @@
...
+    printf "Checking if valkey-server build is required..."
+    BUILD_SERVER=$(is_server_build_required)
+    printf "${GREEN}${BUILD_SERVER}${RESET}\n"
+     if [[ "${BUILD_SERVER}" == "yes" ]]; then
+        printf "${BOLD_PINK}Building valkey-server...${RESET}\n"
+        
+        rm -rf ${VALKEY_SERVER_DIR}
+        git clone --branch ${VALKEY_VERSION} --single-branch https://github.com/valkey-io/valkey.git ${VALKEY_SERVER_DIR}
+        cd ${VALKEY_SERVER_DIR}
+        make
```
eifrah-aws: 1. To reduce build time, I would use here: `make -j$(nproc)`. 
2. Another option, since valkey also supports CMake, we can add it as submodule (similar to gRPC) and build it. The build process of valkey using cmake is fast.

.devcontainer/Dockerfile
murphyjacob4: The integration code was originally written to work with Memtier. The tool works well and is open source. We can definitely move to valkey-benchmark later but it requires rewriting the test cases since the parameters are not transferrable.

.devcontainer/Dockerfile
murphyjacob4: I think memtier-benchmark is located inside the Redis Debian repository

murphyjacob4 (approved):
LGTM for code changes - will leave build system review to @eifrah-aws

.devcontainer/Dockerfile
yairgott: we need memtier for running the integration tests.

.devcontainer/Dockerfile
yairgott: Computability issues. The used memtier command not supported by valkey-benchmark.

cmake/Modules/valkey_search.cmake
yairgott: IMO, the symbol table should be included in all builds. That being said, I wanted to reach out to you about it. Even after using `-g`, the coredump doesn't show a valid backtrace.

eifrah-aws: we can take this offline

eifrah-aws (approved):

cmake/Modules/valkey_search.cmake
yairgott: What do you mean by 'offline'?

testing/integration/CMakeLists.txt
yairgott: Removed

testing/integration/CMakeLists.txt
yairgott: Done

testing/integration/run.sh
yairgott: I prefer option #1. I would expect that running integration tests would be done mainly by developers.

build.sh
yairgott: hmm, I was thinking that running from build.sh wouldn't provide you this control for now. Such is only available if you directly call `testing/integration/run.sh`. wdyt?

cmake/Modules/valkey_search.cmake
yairgott: The stacktrace of a coredump backtrace doesn't show the symbols, even with `-g`. For now, removing `-g` and we'll followup on this in a different PR.

vmsdk/src/utils.cc
```diff
@@ -93,6 +93,25 @@ std::string WrongArity(absl::string_view cmd) {
   return absl::StrCat("ERR wrong number of arguments for ", cmd, " command");
 }
 
+bool IsFakeClient(RedisModuleCtx *ctx) {
```
zvi-code: naming suggestion: the primary client in the replica is a "real" client. while fake clients are local and are not associated to network connection. I would consider replacing with a check if it's not a "user\application" client.

yairgott: Renamed to IsRealUserClient

PR #74, opened 2025-03-26
Fixing integration test

yairgott:
Fixes:
1. Running the docker in github workspaces as a privileged user.
2. Fixing valkey_search_test to turn-off memory tracking after `OnLoad` is called.

PR #76, opened 2025-03-26
fixing github integration tests

yairgott:
Fixes:

1. Running the docker in github workspaces as a privileged user.
2. Fixing valkey_search_test to turn-off memory tracking after OnLoad is called.
3. Fixing github workflows grouping


eifrah-aws (approved):
A remake of previously approved PR (#74)

PR #85, opened 2025-04-01
allow building with external libs

eifrah-aws:
- Built and packaged third party dependencies in a `deb` file (Ubuntu Noble only) for reducing CI time
- Added new `ci/build_ubuntu.sh` for building without submodules
- `build.sh` can now build the module without fetching the submodules by passing the command line argument `--use-system-modules`. Do note, that if passed, the script relies on the user to install them in the system (`grpc`, `googletest` & `libhighwayhash`). A hint on where the third party modules were installed can be provided by CMake environment variable `CMAKE_PREFIX_PATH`
- A script for building a `deb` was added under `submodules` directory (currently it only supports Ubuntu variants - but can be extended in the future)

In any case, the "old" way of building is still supported (i.e. the script will fetch and build the third party modules)

### DIFF
--- a/ci/build_ubuntu.sh
+++ b/ci/build_ubuntu.sh
@@ -1,0 +1,97 @@
+#!/bin/bash -e
+
+CI_DIR=$(readlink -f $(dirname $0))
+ROOT_DIR=$(readlink -f ${CI_DIR}/..)
+BUILD_SH_ARGS=""
+
+# Constants
+RESET='\e[0m'
+GREEN='\e[32;1m'
+RED='\e[31;1m'
+
+function LOG_INFO() {
+    printf "${GREEN}INFO ${RESET} $1\n"
+}
+
+function LOG_ERROR() {
+    printf "${RED}ERROR${RESET} $1\n"
+}
+
+## Parse command line arguments
+while [ $# -gt 0 ]
+do
+    arg=$1
+    case $arg in
+    --run-integration-tests)
+        shift || true
+        BUILD_SH_ARGS="${BUILD_SH_ARGS} --run-integration-tests"
+        LOG_INFO "BUILD_SH_ARGS=${BUILD_SH_ARGS}"
+        ;;
+    --run-tests)
+        shift || true
+        BUILD_SH_ARGS="${BUILD_SH_ARGS} --run-tests"
+        LOG_INFO "BUILD_SH_ARGS=${BUILD_SH_ARGS}"
+        ;;
+    *)
+        LOG_ERROR "Invalid command line argument: $arg"
+        exit 1
+        ;;
+    esac
+done
+
+function get_deb_suffix() {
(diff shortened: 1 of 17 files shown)


submodules/package-submodules.sh
```diff
@@ -0,0 +1,109 @@
...
+    cp ${SRC_DIR}/highwayhash/.build-release/libhighwayhash.a ${INSTALL_DIR}/lib
+    cp -fr ${SRC_DIR}/highwayhash/highwayhash ${INSTALL_DIR}/include
+fi
+
+# Pack everything into a deb file
+
+function get_arch_spec() {
+    local uname_m=$(uname -m)
+    if [[ "${uname_m}" == "x86_64" ]]; then
+        echo "amd64"
```
yairgott, line 64: should it be "x86_64" ?

ci/build_ubuntu.sh
```diff
@@ -0,0 +1,97 @@
...
+        exit 1
+        ;;
+    esac
+done
+
+function get_deb_suffix() {
+    local ARCH=""
+    local uname_m=$(uname -m)
+    if [[ "${uname_m}" == "x86_64" ]]; then
+        ARCH="amd64"
```
yairgott, line 46: should it be "x86_64" ?

yairgott (approved):

submodules/package-submodules.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash -e
+
```
yairgott, line 2: can you add a short description what this script does?

ci/build_ubuntu.sh
eifrah-aws: No, there is a specific list of strings that are allowed in deb control file.
You can see them by running:

```bash
dpkg-architecture -L
```

submodules/package-submodules.sh
eifrah-aws: No, valid values for "x86_64" are:

```bash
uclibc-linux-amd64
musl-linux-amd64
amd64
kfreebsd-amd64
kopensolaris-amd64
hurd-amd64
darwin-amd64
dragonflybsd-amd64
freebsd-amd64
netbsd-amd64
openbsd-amd64
solaris-amd64
```

Run: `dpkg-architecture -L`

submodules/package-submodules.sh
eifrah-aws: Will do on the next PR

debs/valkey-search-deps-ubuntu-noble-amd64.deb
zuiderkwast, line 1: Including two 19MB binary files in the source code repo doesn't seem like a good idea to me.

They could probably be uploaded to some github artifact place instead and used in CI, or use some github actions cache.

PR #86, opened 2025-04-05
L2Sqr NEON, unrolled loop, + prefetching

xbasel:
**DRAFT - no ready**

L2Sqr SIMD version for ARM Neon + unrolled loop  + prefetching

graviton 3:
```Benchmark Results (1M elements):
  Scalar time:     0.771 sec
  NEON time:       0.125 sec
  Speedup:          6.19x
```

Please note that there's relative error in scalar vs simd  most likely due to floating point  rounding and summation order, fp arithmetic is not really associative
this was not compiled with `-ffast-math` (if enabled, the improvement won't be as good, but there will still be improvement)

Please note that there's already SIMD impl in https://github.com/valkey-io/valkey-search/blob/main/third_party/simsimd/include/simsimd/spatial.h , but I believe this impl will outperform it as has unrolled loop and it prefetches the memory 

=====
side note, please note that the scalar impl is already PARTIALLY simded, as I can see this in the generate bytecode:
```
ldr     q16, [x0, x2]
ldr     q5, [x1, x2]
fsub    v5.4s, v16.4s, v5.4s
fmul    v5.4s, v5.4s, v5.4s
```

however, the summation, is not simded (I see fadd, which is scalar)

### DIFF
--- a/third_party/hnswlib/space_l2.h
+++ b/third_party/hnswlib/space_l2.h
@@ -29,6 +29,54 @@ L2Sqr(const void *pVect1v, const void *pVect2v, const void *qty_ptr) {
     return (res);
 }
 
+#if defined(__aarch64__)
+#include <arm_neon.h>
+
+
+#define LOAD_AND_ACCUM(VEC_IDX, OFFSET)                                  \
+    float32x4_t a##VEC_IDX = vld1q_f32(pVect1 + i + OFFSET);             \
+    float32x4_t b##VEC_IDX = vld1q_f32(pVect2 + i + OFFSET);             \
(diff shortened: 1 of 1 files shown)


yairgott:
Can you provide benchmark numbers which show the change benefits?

For benchmarking, we have been using a forked version of [VectorDBBench](https://github.com/murphyjacob4/VectorDBBench). As a client, please use `memorydb` and for `--case-type`, please use Performance768D10M or Performance768D1M.

allenss-amazon:
closing, no activity in 6 mo.

PR #87, opened 2025-04-07
RFC for geospatial indexing

adamk-janea:


rfc/geospatial.md
```diff
@@ -0,0 +1,116 @@
...
+
+The `GEO` field supports the following query operations:
+
+- **Radius Search**: Finds keys within a specified radius from a given point.
+- **Bounding Box Search**: Finds keys within a specified rectangular area.
+- **Nearest Neighbor Search**: Finds the nearest keys to a given point.
+
+These operations allow efficient retrieval of geospatial data based on proximity and spatial relationships.
+
+## Query Language Extensions
```
allenss-amazon: Redisearch already has a syntax for supporting GEO searches. IMO, we should reuse this same syntax for the same operations. Where you're offering additional searching options we should utilize that same style.

Unfortunately, the current query syntax parser isn't easily extended to do this, but a restructuring of that parser will be required for other projects. I think you could safely assume that this restructuring will happen during the development of the other parts of the code.

benedetto73: We changed the syntax of GEO searches in the RFC to match Redisearch.

rfc/geospatial.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Example with GEOSHAPE:
+
+```
+FT.SEARCH idx:pizza "@store_zone:[WITHIN $zone] PARAMS 2 zone "POLYGON((-16 25, 35 39, 38 67, -24 60, -23 36))"
+```
+
+## Implementation Details
+
+- **Boost.Geometry Integration**: Utilizes the R-tree implementation from Boost.Geometry for efficient spatial indexing.
```
yairgott, line 115: Are there any other libraries you have considered/evaluated?

huseyinacacak-janea: Yes, we also considered CGAL (Computational Geometry Algorithms Library) during our evaluation. While CGAL offers a rich set of computational geometry tools, we ultimately chose to proceed with Boost.Geometry for several important reasons.
We began our research by exploring the fundamental differences between quad-trees and R-trees, particularly in the context of geospatial indexing. Through this investigation, we concluded that R-trees are generally better suited for handling geospatial data, especially for use cases involving range queries, where their bounding-box-based approach provides better spatial locality and efficiency.
Once we determined that R-trees were the more appropriate choice, we evaluated several implementations and identified Boost.Geometry as the most compelling option. This decision was supported by [this study](https://www.osti.gov/servlets/purl/2001969) which highlights the robustness and performance of Boost’s R-tree implementation.
Additionally, we performed our own benchmarking study comparing Boost's R* and quadratic split strategies. According to Boost's documentation, the quadratic algorithm offers faster insertions, while the R* is optimized for search efficiency. In our testing, we observed that the quadratic split was considerably faster for insertions, and that both algorithms performed similarly in search operations.

rfc/geospatial.md
```diff
@@ -0,0 +1,130 @@
...
+- **Data Storage**: Geospatial data is stored in fields designated as `GEO` type or `GEOSHAPE` type, with the indexing mechanism parsing these fields to extract coordinates.
+- **Performance Considerations**: Ensures efficient insertions, deletions, and queries, even with large datasets.
+
+## Backward Compatibility
+
+The introduction of the `GEO` and `GEOSHAPE` field types and associated commands is backward compatible. Existing functionality remains unchanged, and the new features are additive.
+
+## Open Questions
+
+- **Coordinate Reference Systems**: Should multiple coordinate reference systems be supported, or should the system standardize on WGS 84?
```
yairgott, line 126: IMO, starting with WGS 84 is a good approach while keeping in mind that additional reference systems could be supported in the future.

rfc/geospatial.md
```diff
@@ -0,0 +1,130 @@
...
+- **3D Geospatial Data**: Should three-dimensional data (latitude, longitude, altitude) be supported in the future?
```
yairgott, line 127: It would be good to keep this in mind as a potential future enhancement during implementation.

rfc/geospatial.md
```diff
@@ -0,0 +1,130 @@
...
+- **Advanced Spatial Queries**: Should additional spatial predicates (e.g., intersects, contains) be supported beyond radius and bounding box searches?
```
yairgott, line 128: Same as the previous comment.

rfc/geospatial.md
```diff
@@ -0,0 +1,130 @@
...
+In the context of this RFC:
+
+- **Key**: A Valkey key, which could be the name of the key or the contents of a HASH or JSON key.
+- **Field**: A component of an index. Each field has a type and a path.
+- **Index**: A collection of fields and field-indexes. The object created by the `FT.CREATE` command.
+- **Field-Index**: A data structure associated with a field that accelerates the operation of search operators for this field type.
+- **Geospatial Data**: Data that represents the geographic location and shape of objects.
+- **R-tree**: A tree data structure used for spatial access methods, i.e., for indexing multi-dimensional information such as geographical coordinates.
+
+## Design
```
yairgott, line 29: Are there any known incompatibilities or caveats compared to RediSearch?

rfc/geospatial.md
yairgott: SG and Thanks for the explanation!

rfc/geospatial.md
zori-janea: Aside from `OPTIMIZED_FOR`, we expect the commands to have the same syntax and almost identical behavior.
For the second point, some implementation details might cause differences in certain edge cases.
We'll address and resolve these issues during testing.

yairgott (approved):

PR #95, opened 2025-04-22
adding support for blocking clients during keyspace notification

yairgott:
On keyspace event notification the function `ShouldBlockClient` is called which checks that the ctx holds a read user client, via `vmsdk::IsRealUserClient`.


testing/integration/run.sh
```diff
@@ -4,7 +4,7 @@ ROOT_DIR=$(readlink -f $(dirname $0))
 BUILD_CONFIG=release
 TEST=all
 CLEAN="no"
-VALKEY_VERSION="8.0.0"
+#VALKEY_VERSION="8.0.0"
```
murphyjacob4: I wonder if always building against unstable makes sense. We probably want to know when something breaks immediately.

We should also have past version compat testing (for now, it would just be 8.1.1 and unstable, but we would add all supported versions to the list). But that seems like a different initiative we can fund

yairgott: I think that the github presubmit should work against a stable valkey version.

I agree that we should run the stability on the unstable but lets not tight it to pre-submit but run it periodically.

yairgott: I'll open an issue for this.

src/commands/ft_search.cc
```diff
@@ -242,7 +243,7 @@ absl::Status FTSearchCmd(RedisModuleCtx *ctx, RedisModuleString **argv,
     }
 
     vmsdk::BlockedClient blocked_client(ctx, async::Reply, async::Timeout,
-                                        async::Free, 0);
+                                        async::Free, parameters->timeout_ms);
```
murphyjacob4, line 246: Should we file an issue for integration testing of timeout functionality?

vmsdk/src/blocked_client.cc
```diff
@@ -0,0 +1,134 @@
...
+#include "vmsdk/src/blocked_client.h"
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/log/check.h"
+#include "absl/synchronization/mutex.h"
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+
+namespace vmsdk {
+absl::Mutex blocked_clients_mutex;
+absl::flat_hash_map<unsigned long long, BlockedClientEntry> gBlockedClients;
```
murphyjacob4: Do we want to enforce a style guide for the project? It doesn't have to be Google C++ style, but I feel like we should align naming. I think other global naming we just did something akin to `blocked_clients`

Regarding Google C++ style - they forbid globals with non-trivial destructors. Consider wrapping in an absl::NoDestructor to ensure destruction order doesn't do anything weird

murphyjacob4 (approved):

src/commands/ft_search.cc
allenss-amazon: yes.

vmsdk/src/blocked_client.cc
```diff
@@ -0,0 +1,134 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
allenss-amazon, line 29: We've been requested to move to SPDX marking here see #5

vmsdk/src/blocked_client.cc
allenss-amazon: Yes on the naming style. Is there an assert in the shutdown code to ensure this is empty?

vmsdk/src/blocked_client.cc
```diff
@@ -0,0 +1,134 @@
...
+// Used for testing
+absl::flat_hash_map<unsigned long long, BlockedClientEntry> &
+TrackedBlockedClients() {
+  return gBlockedClients;
+}
+
+BlockedClient::BlockedClient(RedisModuleCtx *ctx, bool handle_duplication) {
+  if (handle_duplication) {
+    unsigned long long tracked_client_id = RedisModule_GetClientId(ctx);
+    if (tracked_client_id != 0) {
```
allenss-amazon, line 60: What triggers duplication? What's the scenario where this happens?

vmsdk/src/blocked_client.h
```diff
@@ -0,0 +1,81 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
```
allenss-amazon, line 28: SPDX please.

vmsdk/src/blocked_client.h
```diff
@@ -0,0 +1,81 @@
...
+  BlockedClient(const BlockedClient &other) = delete;
+  BlockedClient &operator=(const BlockedClient &other) = delete;
+
+  operator RedisModuleBlockedClient *() const { return blocked_client_; }
+  void SetReplyPrivateData(void *private_data);
+  void UnblockClient();
+  void MeasureTimeStart();
+  void MeasureTimeEnd();
+  ~BlockedClient() { UnblockClient(); }
+
```
allenss-amazon, line 66: These are the the kinds of supportive APIs that need additional documentation. For example, is OK to call UnblockClient twice? I see that the destructor always calls it, so yes?? As for the measurement calls, what is the pattern? What if I called them multiple times?

vmsdk/testing/blocked_client_test.cc
```diff
@@ -0,0 +1,147 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
```
allenss-amazon, line 28: spdx

vmsdk/src/blocked_client.cc
yairgott: SG but I suggest that we'll switch to SPDX as part of this task for all files. For now, I suggest to  continue with requiring following the old format to maintain uniformity.

vmsdk/src/blocked_client.cc
yairgott: Coding style is already enforced. See https://github.com/valkey-io/valkey-search/blob/main/.github/workflows/clang_tidy_format.yml .

Good point about the absl::NoDestructor.

vmsdk/src/blocked_client.cc
yairgott: Duplication can only be supported when the callbacks are empty. In practice, this is only happening on keyspace notifications. When the callbacks are not empty, the blocking state callbacks are being overriden. 
I'll add a comment explaining this.

vmsdk/src/blocked_client.h
yairgott: It's safe to call UnblockClient more than once. UnblockClient nullptr blocked_client_ and would be early return if blocked_client_ is nullptr.

src/commands/ft_search.cc
yairgott: Will do

yairgott: https://github.com/valkey-io/valkey-search/issues/103

PR #97, opened 2025-04-22
Make backfill batch/block size configurable

Shivshankar-Reddy:
Ran benchmark test with GIST dataset from ann-benchmark on valkey-search module resulted performance fluctuations and observed that lot of backfill/resize during the test. For large datasets the vector backfill process is triggered so frequent as it is hard coded to 10240. So making the constant to be configurable would help to adjust based on the number of documents be loaded and the frequency of backfill can be controlled based on the expected dataset size. this helps for reducing the indexing time and also perfomance fluctuation. 

Added '--resize-block-size' option in module load optins and this will be reflcted on kIndexSchemaBackfillBatchSize and kHNSWBlockSize for consistency and default value is 10240 as before. The values get modified for above default 10240 and does not reflect for lower sizes.

### DIFF
--- a/src/indexes/vector_base.h
+++ b/src/indexes/vector_base.h
@@ -65,7 +65,7 @@
 
 namespace valkey_search::indexes {
 
-inline constexpr uint32_t kHNSWBlockSize = 1024 * 10;
+uint32_t kHNSWBlockSize{1024 * 10};
 
 std::vector<char> NormalizeEmbedding(absl::string_view record, size_t type_size,
                                      float* magnitude = nullptr);
--- a/src/indexes/vector_base.h
+++ b/src/indexes/vector_base.h
@@ -65,8 +65,6 @@
 
 namespace valkey_search::indexes {
 
-uint32_t g_hnsw_block_size{1024 * 10};
-
 std::vector<char> NormalizeEmbedding(absl::string_view record, size_t type_size,
                                      float* magnitude = nullptr);
 
--- a/src/indexes/vector_base.h
+++ b/src/indexes/vector_base.h
@@ -65,7 +65,7 @@
 
 namespace valkey_search::indexes {
 
-uint32_t kHNSWBlockSize{1024 * 10};
+uint32_t g_hnsw_block_size{1024 * 10};
 
 std::vector<char> NormalizeEmbedding(absl::string_view record, size_t type_size,
                                      float* magnitude = nullptr);
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -57,6 +57,7 @@
 #include "src/metrics.h"
 #include "src/rdb_section.pb.h"
 #include "src/rdb_serialization.h"
+#include "src/valkey_search.h"
 #include "src/vector_externalizer.h"
 #include "vmsdk/src/log.h"
 #include "vmsdk/src/managed_pointers.h"
@@ -639,8 +640,10 @@ void SchemaManager::OnServerCronCallback(RedisModuleCtx *ctx,
                                          [[maybe_unused]] RedisModuleEvent eid,
                                          [[maybe_unused]] uint64_t subevent,
                                          [[maybe_unused]] void *data) {
+  auto index_schema_backfill_batch_size =
+      ValkeySearch::Instance().GetDefaultBlockSize();
   SchemaManager::Instance().PerformBackfill(ctx,
-                                            g_index_schema_backfill_batch_size);
+                                            index_schema_backfill_batch_size);
 }
 
 }  // namespace valkey_search
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -116,6 +116,8 @@ SchemaManager::SchemaManager(
   }
 }
 
+constexpr uint32_t kIndexSchemaBackfillBatchSize{10240};
+
 absl::Status GenerateIndexNotFoundError(absl::string_view name) {
   return absl::NotFoundError(
       absl::StrFormat("Index with name '%s' not found", name));
@@ -640,10 +642,7 @@ void SchemaManager::OnServerCronCallback(RedisModuleCtx *ctx,
                                          [[maybe_unused]] RedisModuleEvent eid,
                                          [[maybe_unused]] uint64_t subevent,
                                          [[maybe_unused]] void *data) {
-  auto index_schema_backfill_batch_size =
-      ValkeySearch::Instance().GetDefaultBlockSize();
-  SchemaManager::Instance().PerformBackfill(ctx,
-                                            index_schema_backfill_batch_size);
+  SchemaManager::Instance().PerformBackfill(ctx, kIndexSchemaBackfillBatchSize);
 }
 
 }  // namespace valkey_search
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -115,8 +115,6 @@ SchemaManager::SchemaManager(
   }
 }
 
-constexpr uint32_t kIndexSchemaBackfillBatchSize{10240};
-
 absl::Status GenerateIndexNotFoundError(absl::string_view name) {
   return absl::NotFoundError(
       absl::StrFormat("Index with name '%s' not found", name));
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -639,7 +639,8 @@ void SchemaManager::OnServerCronCallback(RedisModuleCtx *ctx,
                                          [[maybe_unused]] RedisModuleEvent eid,
                                          [[maybe_unused]] uint64_t subevent,
                                          [[maybe_unused]] void *data) {
-  SchemaManager::Instance().PerformBackfill(ctx, kIndexSchemaBackfillBatchSize);
+  SchemaManager::Instance().PerformBackfill(ctx,
+                                            g_index_schema_backfill_batch_size);
 }
 
 }  // namespace valkey_search
--- a/src/valkey_search.cc
+++ b/src/valkey_search.cc
@@ -75,7 +75,7 @@ namespace valkey_search {
 
 static absl::NoDestructor<std::unique_ptr<ValkeySearch>> valkey_search_instance;
 constexpr size_t kMaxWorkerThreadPoolSuspensionSec{60};
-std::atomic<uint32_t> ValkeySearch::hnsw_block_size_{10240};
+static std::atomic<uint32_t> hnsw_block_size{10240};
 const absl::string_view kHNSWBlockSizeConfig{"vs-hnsw-block-size"};
 
 namespace options {
@@ -141,6 +141,10 @@ ValkeySearch &ValkeySearch::Instance() { return **valkey_search_instance; };
 
 void ValkeySearch::InitInstance(std::unique_ptr<ValkeySearch> instance) {
   *valkey_search_instance = std::move(instance);
+}
+
+uint32_t ValkeySearch::GetHNSWBlockSize() const {
+  return hnsw_block_size.load();
 }
 
 static std::string ConvertToMB(double bytes_value) {
@@ -543,7 +547,7 @@ void ValkeySearch::ResumeWriterThreadPool(RedisModuleCtx *ctx,
 long long ValkeySearch::BlockSizeGetConfig(
     [[maybe_unused]] const char *config_name,
     [[maybe_unused]] void *priv_data) {
-  return hnsw_block_size_.load();
+  return hnsw_block_size.load();
 }
 
 int ValkeySearch::BlockSizeSetConfig([[maybe_unused]] const char *config_name,
@@ -557,7 +561,7 @@ int ValkeySearch::BlockSizeSetConfig([[maybe_unused]] const char *config_name,
     }
     return REDISMODULE_ERR;
   }
-  hnsw_block_size_ = static_cast<uint32_t>(value);
+  hnsw_block_size = static_cast<uint32_t>(value);
   return REDISMODULE_OK;
 }
 
--- a/src/valkey_search.cc
+++ b/src/valkey_search.cc
@@ -76,6 +76,8 @@ namespace valkey_search {
 static absl::NoDestructor<std::unique_ptr<ValkeySearch>> valkey_search_instance;
 constexpr size_t kMaxWorkerThreadPoolSuspensionSec{60};
 
+static uint32_t current_block_size{10240};
+
 namespace options {
 
 // Maintaining the parameter `--threads` for backward compatibility. Safe to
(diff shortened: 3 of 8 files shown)


src/valkey_search.cc
```diff
@@ -85,13 +85,15 @@ constexpr absl::string_view kReaderThreadsParam{"--reader-threads"};
 constexpr absl::string_view kWriterThreadsParam{"--writer-threads"};
 constexpr absl::string_view kUseCoordinator{"--use-coordinator"};
 constexpr absl::string_view kLogLevel{"--log-level"};
+constexpr absl::string_view kResizeBlockSize{"--resize-block-size"};
```
eifrah-aws: nit: I would rename this into `--block-size` (no need for the "resize")

src/valkey_search.cc
```diff
@@ -107,6 +109,8 @@ absl::StatusOr<Parameters> Load(RedisModuleString **argv, int argc) {
                         GENERATE_FLAG_PARSER(Parameters, use_coordinator));
   parser.AddParamParser(kLogLevel,
                         GENERATE_VALUE_PARSER(Parameters, log_level));
+  parser.AddParamParser(kResizeBlockSize,
```
eifrah-aws: This seems like a valid mitigation option, i.e. I would prefer to be able to modify it during runtime using valkey's configuration options (e.g. `CONFIG SET vs-block-size 20480`. The current approach forces a valkey restart

src/valkey_search.cc
Shivshankar-Reddy: Updated the option to `block-size`. Thanks!

src/valkey_search.cc
Shivshankar-Reddy: Hi @eifrah-aws, Thank you for the review!
Yes, you are correct it makes more flexible to configure the option and I think its possible to do. However, we need to add `vs-block-size` config in the valkey core configs and then it will be available to update the options with `CONFIG SET`.  The main problem is that this is not a generic config that we can propose in the core, and I am afraid they accept it. Currently, we dont have any such module specific config if I am not wrong.

- There is another way we can do it by introducing this VM_UpdateRuntimeArgs api which is implemented in https://github.com/valkey-io/valkey/pull/1041, But valkey-search makes use of the redis module apis so this api is not available at this point. Once the module apis are updated with latest valkey module apis I think we can enhance the option to update in runtime. WDYT?

Thanks,

src/indexes/vector_base.h
```diff
@@ -65,7 +65,7 @@
 
 namespace valkey_search::indexes {
 
-inline constexpr uint32_t kHNSWBlockSize = 1024 * 10;
+uint32_t kHNSWBlockSize{1024 * 10};
```
yairgott: Variables starting with the letter `k` indicate constant (based on Google coding style which we follow).

yairgott: why not keeping this `kHNSWBlockSize` as constexpr? maybe just rename it to kDefaultHNSWBlockSize?

src/valkey_search.cc
eifrah-aws: I am pretty sure its doable today:

https://valkey.io/topics/modules-api-ref/#section-module-configurations-api

Shivshankar-Reddy: Thank you @eifrah-aws for sharing the details, I will look into it.

src/indexes/vector_base.h
Shivshankar-Reddy: if kHNSWBlockSize is constexpr compiler did not allow to change as part of options. Sure I will change the name.

src/valkey_search.cc
```diff
@@ -470,6 +474,11 @@ absl::Status ValkeySearch::LoadOptions(RedisModuleCtx *ctx,
       "write-worker-", options.writer_threads);
   writer_thread_pool_->StartWorkers();
 
+  if (options.block_size > 10240) {
```
yairgott: why the `if` is required?

src/valkey_search.cc
```diff
@@ -470,6 +474,11 @@ absl::Status ValkeySearch::LoadOptions(RedisModuleCtx *ctx,
...
+    kIndexSchemaBackfillBatchSize = options.block_size;
```
yairgott: Can we keep just a single source of truth? For instance, `VectorHNSW<T>::block_size_` could be `std::optional<uint32_t>` and in  `VectorHNSW<T>::ResizeIfFull` , we can use the following:
`auto block_size = block_size_.value_or(ValkeySearch::Instance().GetDefaultBlockSize())`

src/indexes/vector_base.h
yairgott: Right but you assign the value to options.block_size

src/valkey_search.cc
yairgott: On second thought, `VectorHNSW<T>::block_size_` can be completely be removed. It mainly exist today to support testing. The test can use the ValkeySearch API to load the options and the block size can be exposed via `ValkeySearch::Instance().GetDefaultBlockSize()`. wdyt?

src/indexes/vector_base.h
```diff
@@ -65,7 +65,7 @@
 
 namespace valkey_search::indexes {
 
-inline constexpr uint32_t kHNSWBlockSize = 1024 * 10;
+uint32_t g_hnsw_block_size{1024 * 10};
```
yairgott: Google coding standard does not specify special prefixing global variables, these need to follow the regular naming, see: https://google.github.io/styleguide/cppguide.html#Variable_Names

yairgott (changes requested):

src/valkey_search.cc
Shivshankar-Reddy: Its not need, just enforced if the module options loaded with lower value. removed it in latest commit.

src/valkey_search.cc
Shivshankar-Reddy: The latest commit implents as suggested. Thanks!

src/indexes/vector_base.h
Shivshankar-Reddy: I removed redundant global variable in latest commit and refering the single value.

src/valkey_search.cc
Shivshankar-Reddy: This reference was helpful implement the config set, Thanks again and please have a look at latest commit.

src/indexes/vector_base.h
Shivshankar-Reddy: Thank you for the quick review! The commit was not final, that was the reason I moved the title to `[WIP]`. Anyways the latest commit covers the previous comments and also the naming, could you please have a look when you have some time.

Shivshankar-Reddy:
Below is the logs from valkey-cli manual testing results that the config we can set and get durinng run time. Also verified the logs of resizefill reflects the values.

```
127.0.0.1:6379> CONFIG set search.vs-block-size 10240
OK
127.0.0.1:6379>
127.0.0.1:6379> CONFIG get search.vs-block-size
1) "search.vs-block-size"
2) "10240"
127.0.0.1:6379>
127.0.0.1:6379> CONFIG set search.vs-block-size 20480
OK
127.0.0.1:6379> CONFIG get search.vs-block-size
1) "search.vs-block-size"
2) "20480"
127.0.0.1:6379>

```

src/valkey_search.cc
```diff
@@ -521,13 +539,56 @@ void ValkeySearch::ResumeWriterThreadPool(RedisModuleCtx *ctx,
...
+    [[maybe_unused]] const char *config_name,
+    [[maybe_unused]] void *priv_data) {
+  return block_size_.load();
+}
+
+int ValkeySearch::BlockSizeSetConfig([[maybe_unused]] const char *config_name,
+                                     long long value,
+                                     [[maybe_unused]] void *priv_data,
+                                     RedisModuleString **err) {
+  if (value > UINT32_MAX) {
```
yairgott: `value` should also be > 0

src/valkey_search.h
```diff
@@ -136,6 +144,8 @@ class ValkeySearch {
 
   std::unique_ptr<coordinator::Server> coordinator_;
   std::unique_ptr<coordinator::ClientPool> client_pool_;
+
+  static std::atomic<uint32_t> block_size_;
```
yairgott: This seems to be redundant

src/schema_manager.cc
```diff
@@ -115,8 +116,6 @@ SchemaManager::SchemaManager(
   }
 }
 
-constexpr uint32_t kIndexSchemaBackfillBatchSize{10240};
```
yairgott: This batch size is different: it controls the backfill batch size and should be in-depended of the hnsw block size.

src/valkey_search.cc
```diff
@@ -85,13 +86,15 @@ constexpr absl::string_view kReaderThreadsParam{"--reader-threads"};
 constexpr absl::string_view kWriterThreadsParam{"--writer-threads"};
 constexpr absl::string_view kUseCoordinator{"--use-coordinator"};
 constexpr absl::string_view kLogLevel{"--log-level"};
+constexpr absl::string_view kBlockSize{"--vs-block-size"};
```
yairgott: Suggest to be more explicit about which block size, .e.g.: `--vs-hnsw-block-size`

This also true for internal variables or functions, e.g.:
kHNSWBlockSize
hnsw_block_size_
GetHNSWBlockSize
etc

src/valkey_search.cc
```diff
@@ -521,13 +539,56 @@ void ValkeySearch::ResumeWriterThreadPool(RedisModuleCtx *ctx,
...
                                   int argc) {
   ctx_ = RedisModule_GetDetachedThreadSafeContext(ctx);
 
   // Register a single module type for Aux load/save callbacks.
   VMSDK_RETURN_IF_ERROR(RegisterModuleType(ctx));
 
+  // Register vs-block-size configuration
+  if (RedisModule_RegisterNumericConfig(
+          ctx,
+          "vs-block-size",             // Name
```
yairgott: See my earlier comment about "vs-block-size"

src/valkey_search.cc
```diff
@@ -470,6 +475,19 @@ absl::Status ValkeySearch::LoadOptions(RedisModuleCtx *ctx,
       "write-worker-", options.writer_threads);
   writer_thread_pool_->StartWorkers();
 
+  if (options.block_size) {
+    RedisModuleString *err = nullptr;
+    if (BlockSizeSetConfig("vs-block-size", options.block_size, nullptr,
```
yairgott: Here and other places, avoid directly using the string `"vs-block-size"`. Ideally, it would be defined once and sometime it would be used with "--" and sometime not.

testing/vector_test.cc
```diff
@@ -325,8 +325,15 @@ TEST_F(VectorIndexTest, ResizeHNSW) ABSL_NO_THREAD_SAFETY_ANALYSIS {
                                    kM, kEFConstruction, kEFRuntime),
         "attribute_identifier_1",
         data_model::AttributeDataType::ATTRIBUTE_DATA_TYPE_HASH);
-    const uint32_t block_size = 1024;
-    index.value()->SetBlockSize(block_size);
+    RedisModuleString* err = nullptr;
+    EXPECT_EQ(
+        ValkeySearch::BlockSizeSetConfig("vs-block-size", 1024, nullptr, &err),
```
yairgott: See my earlier comment about "vs-block-size"

yairgott (changes requested):

src/valkey_search.cc
Shivshankar-Reddy: config set restricts that but you are right!

src/valkey_search.h
Shivshankar-Reddy: this is declaration of class member and in .cc just initialized.

src/schema_manager.cc
Shivshankar-Reddy: I thought, to be consitent with backfill size of hnsw, sure I removed it.

Shivshankar-Reddy:
Hi @yairgott , As suggested I have covered changes to address the comments, please have a look and provide your thoughts. thanks!

src/valkey_search.cc
```diff
@@ -542,21 +543,21 @@ void ValkeySearch::ResumeWriterThreadPool(RedisModuleCtx *ctx,
...
-  return block_size_.load();
+  return hnsw_block_size_.load();
 }
 
 int ValkeySearch::BlockSizeSetConfig([[maybe_unused]] const char *config_name,
                                      long long value,
                                      [[maybe_unused]] void *priv_data,
                                      RedisModuleString **err) {
-  if (value > UINT32_MAX) {
+  if (value < 0 || value > UINT32_MAX) {
```
yairgott: value of 0 can not work as it would be invalid to create a block with size of 0.

src/valkey_search.h
yairgott: But you can declare this variable inside `src/valkey_search.cc`, similarly to `valkey_search_instance`. This variable is not accessed directly but only through `GetHNSWBlockSize` from outside `src/valkey_search.cc`.

src/valkey_search.cc
eifrah-aws: LGTM

src/valkey_search.h
eifrah-aws: Also, placing static variables in header files is not recommended. Basically, each `.cc` file that includes this header will get its own copy of the variable

src/valkey_search.cc
eifrah-aws: Nit comment: (if we are about to change the variable name, again): when a module registers a configuration variable within Valkey, the module name is prepended to the name, so it becomes: `search.block-size` - this makes the `vs-` redundant. So changing it to `hnsw-block-size` should do the trick as it will become: `search.hnsw-block-size`

src/valkey_search.h
```diff
@@ -136,6 +144,8 @@ class ValkeySearch {
 
   std::unique_ptr<coordinator::Server> coordinator_;
   std::unique_ptr<coordinator::ClientPool> client_pool_;
+
+  static std::atomic<uint32_t> hnsw_block_size_;
```
yairgott: Following on the earlier comment, this should be removed and `hnsw_block_size` should be declared in valkey_search.cc as a global variable.

yairgott (changes requested):

src/valkey_search.cc
```diff
@@ -137,6 +143,10 @@ void ValkeySearch::InitInstance(std::unique_ptr<ValkeySearch> instance) {
   *valkey_search_instance = std::move(instance);
 }
 
+uint32_t ValkeySearch::GetHNSWBlockSize() const {
+  return hnsw_block_size.load();
```
yairgott: Lets load with `std::memory_order_relaxed` as enforcing sequential consistency is not required in this case.

yairgott (changes requested):

yairgott (approved):

PR #101, opened 2025-04-24
Fixing asan issues

yairgott:
Addressed the following issues which were found by ASAN:
1. hnswlib: call to `free` after allocation by `new char[]`.
2. `VectorFlat<T>::ModifyRecord` may lead to a dangling vector pointer if the vector was not changed. This is because `VectorBase::UpdateMetadata` called `VectorFlat<T>::TrackVector ` which override the old vector `shared_ptr`. Later a call to `VectorFlat<T>::ModifyRecordImpl` is issued which reads the freed old vector for the purpose of checking if the new vector is the same as the previous. Addressed this by consolidating all the verifications, including the check if the new vector value differs, inside `VectorBase::UpdateMetadata`.
3. Changed the mocked sender id generation for testing to be aligned with the engine sender id length, `REDISMODULE_NODE_ID_LEN`
4. Define `absl::Mutex mutex` in vmsdk/testing/thread_pool_test.cc in the right scope.

Additionally, added presubmit workflow to run the unittests in asan mode.


.github/workflows/unittests.yml
```diff
@@ -32,3 +32,4 @@ jobs:
 
     - name: Run Tests
       run: docker run --rm  -v "$(pwd):/workspace" --user "ubuntu:ubuntu" presubmit-image sudo ci/build_ubuntu.sh --run-tests
+
```
murphyjacob4, line 35: Any reason to remove the newline at the end of the file?

src/index_schema.cc
```diff
@@ -917,13 +917,11 @@ void IndexSchema::OnLoadingEnded(RedisModuleCtx *ctx) {
...
                            << " stale entries of " << key_size
-                           << " total keys for "
-                           << "{Index: " << name_
+                           << " total keys for " << "{Index: " << name_
                            << ", Attribute: " << attribute.first << "}";
   }
   VMSDK_LOG(NOTICE, ctx) << "Deleting " << deletion_attributes.size()
-                         << " stale entries for "
-                         << "{Index: " << name_ << "}";
+                         << " stale entries for " << "{Index: " << name_ << "}";
```
murphyjacob4:
```suggestion
                         << " stale entries for {Index: " << name_ << "}";
```

src/index_schema.cc
```diff
@@ -917,13 +917,11 @@ void IndexSchema::OnLoadingEnded(RedisModuleCtx *ctx) {
     });
     VMSDK_LOG(NOTICE, ctx) << "Deleting " << stale_entries
                            << " stale entries of " << key_size
-                           << " total keys for "
-                           << "{Index: " << name_
+                           << " total keys for " << "{Index: " << name_
```
murphyjacob4:
```suggestion
                           << " total keys for {Index: " << name_
```

src/indexes/vector_base.cc
```diff
@@ -353,30 +359,38 @@ absl::StatusOr<uint64_t> VectorBase::TrackKey(const InternedStringPtr &key,
...
   }
+  TrackVector(id, vector);
   key_by_internal_id_.insert({id, key});
   return id;
 }
 
-absl::Status VectorBase::UpdateMetadata(const InternedStringPtr &key,
-                                        float magnitude,
-                                        const InternedStringPtr &vector) {
+absl::StatusOr<bool> VectorBase::UpdateMetadata(
```
murphyjacob4, line 373: For a complex return type like this, can we add a comment explaining what it means? Guessing it is error == failure, true == updated, false == not updated? If so, would be good to document under which situation is it not updated

third_party/hnswlib/hnswalg.h
murphyjacob4, line 1: Is this just formatting changes? There is a huge diff and it is hard to compare left and right side

third_party/hnswlib/hnswlib.h
murphyjacob4, line 1: Same here - is it just formatting changes?

.github/workflows/unittests.yml
murphyjacob4: Oh sorry, the diff displayed this weird so I thought you were removing a newline. Looks like it is added

murphyjacob4 (approved):

Issue #55, opened 2025-03-02
Incompatible output sorting in FT.SEARCH command


allenss-amazon:
In RedisSearch, the FT.SEARCH command returns keys sorted by the key name.

In ValkeySearch the FT.SEARCH command returns keys sorted by the vector distance (highest distance to lowest). 

One could argue that highest to lowest distance sorting is also backwards from what anybody would want when combined with the LIMIT clause (who would want to have 5 vectors that aren't close but aren't far??)

Here is some "proof", I"m happy to walk you through this 1:1. It'll make more sense that way: But this is a side-by-side run of some sample data and commands through both code bases.

```
test/aggregate.py::TestAggregateCompatibility::test_search_reverse[hash] Generated schema: n1 numeric n2 numeric n3 numeric t1 tag t2 tag t3 tag v1 vector HNSW 6 DIM 3 TYPE FLOAT32 DISTANCE_METRIC L2
Generated schema: $.n1 AS n1 numeric $.n2 AS n2 numeric $.n3 AS n3 numeric $.t1 AS t1 tag $.t2 AS t2 tag $.t3 AS t3 tag v1 vector HNSW 6 DIM 3 TYPE FLOAT32 DISTANCE_METRIC L2
load_data with index reverse vector numbers   hash
Start Loading Data, data set has 15 records
Doing:  ['FT.CREATE hash_idx1 ON HASH PREFIX 1 hash: SCHEMA n1 numeric n2 numeric n3 numeric t1 tag t2 tag t3 tag v1 vector HNSW 6 DIM 3 TYPE FLOAT32 DISTANCE_METRIC L2']
HSET hash:0 n1 -5 n2 5 n3 -5 t1 one.one0 t2 two.two0 t3 all_the_same_value v1 b'\x00\x00pA\x00\x00pA\x00\x00pA'
HSET hash:1 n1 -4 n2 4 n3 9 t1 one.one2 t2 two.two-2 t3 all_the_same_value v1 b'\x00\x00`A\x00\x00`A\x00\x00`A'
HSET hash:2 n1 -3 n2 3 n3 8 t1 one.one4 t2 two.two-4 t3 all_the_same_value v1 b'\x00\x00PA\x00\x00PA\x00\x00PA'
HSET hash:3 n1 -2 n2 2 n3 7 t1 one.one6 t2 two.two-6 t3 all_the_same_value v1 b'\x00\x00@A\x00\x00@A\x00\x00@A'
[...]

PR #104, opened 2025-04-29
FT.SEARCH - Add logic to decide sorting order (descending or ascending distance)

KarthikSubbarao:
FT.SEARCH - Add logic to decide sorting order (descending or ascending distance).

https://github.com/valkey-io/valkey-search/issues/55


yairgott (approved):

KarthikSubbarao:
IMO, it would be interesting to allow two different sorting orders based on the distance metric.

If the distance metric is L2 (Euclidian), using ascending order (nearest / lowest first) is better.

If the distance metric is cosine, using descending order (highest) is better because it represents the closest in similarity.

yairgott (approved):

KarthikSubbarao:
In case of L2, for distance, a lower score is better.

In case of cosine / IP, for similarity, a higher score is better. However, we use 1 - cosine and return cosine dissimilarity. So, using a lower score is better.

So, it should be fine to always use asc sort order.

src/indexes/vector_base.cc
```diff
@@ -129,10 +129,10 @@ T CopyAndNormalizeEmbedding(T *dst, T *src, size_t size) {
     magnitude += src[i] * src[i];
   }
   magnitude = std::sqrt(magnitude);
-  if (magnitude == 0.0f) {
-    return magnitude;
+  T norm = 1.0f;
+  if (magnitude != 0.0f) {
```
allenss-amazon: T norm = (magnitude == 0.0) ? 1.0.f : (1.0f / magnitude);

KarthikSubbarao: Thanks - Updated

src/indexes/vector_base.cc
```diff
@@ -271,7 +268,8 @@ absl::StatusOr<std::deque<Neighbor>> VectorBase::CreateReply(
       knn_res.pop();
       continue;
     }
-    ret.emplace_back(Neighbor{vector_key.value(), ele.first});
+    // Sorting in asc order.
+    ret.emplace_front(Neighbor{vector_key.value(), ele.first});
```
murphyjacob4, line 272: @KarthikSubbarao do we need to make a similar change for the coordinator?

PR #105, opened 2025-05-01
chore: avoid reserved gid, group name when setup for local dev env

proost:
when i run 
```
id -g
id -gn
```

commands In My Mac OS, I got

```
20
staff
```

`ubuntu:stable` has already "staff" group. so when build docker got error
```
groupmod: group 'staff' already exists
```

If conflict with reserved group, then use default group in the script.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -55,6 +55,13 @@ ENV USER_GID=${USER_GID}
 ARG USER_GNAME=ubuntu
 ENV USER_GNAME=${USER_GNAME}
 
+# Check if the target group name exists and has a different GID. If so, delete it.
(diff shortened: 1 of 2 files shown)


yairgott:
The idea is that you'll be able to use the exact user id and group id inside the container so you can dir and apply changes to the cloned dir from outside. 

I think that the change should be in the [docker file](https://github.com/valkey-io/valkey-search/blob/main/.devcontainer/Dockerfile#L58),  to delete the group if its id & name doesn't match the env variables USER_GID, USER_GNAME.

wdyt?

proost:
@yairgott
I understand. Change it: [37713ee](https://github.com/valkey-io/valkey-search/pull/105/commits/37713eedb11d62e8d77c56261a71f24526e4648b)

yairgott (approved):

PR #107, opened 2025-05-04
Method `GetPhysicalCPUCoresCount` is broken for `Linux / aarch64`

eifrah-aws:
On aarch64, `cpuinfo` output is different than the one generated by `amd64` machines. This fix uses `lscpu` to get the number of physical cores by parsing `lscpu` output and multiplying the number of sockets by the number of cores-per-socket. In case of an error, it will return the default from `std::thread::hardware_concurrency`.

Tests were added to `concurrency_test` binary


allenss-amazon (approved):
ship it.

PR #109, opened 2025-05-05
Exposing search subscription info just if built with the macro DEBUG_INFO

yairgott:


allenss-amazon (commented):
I'm good with a compile-time switch. I dislike "VERBOSE". IMO, this is more developer oriented, so something that's more internally oriented like "INTERNAL" or "DEBUG_xxx", etc.

src/valkey_search.cc
```diff
@@ -247,6 +247,7 @@ void ValkeySearch::Info(RedisModuleInfoCtx *ctx, bool for_crash_report) const {
       RedisModule_InfoAddFieldLongLong(ctx, skipped_count_str.c_str(),
                                        stat.skipped_cnt);
     };
+#ifdef VERBOSE_INFO
```
murphyjacob4: I like Allen's suggestion for: DEBUG_INFO

```suggestion
#ifdef DEBUG_INFO
```

murphyjacob4 (approved):

PR #118, opened 2025-05-12
Valkey-Search doc related files

eifrah-aws:
This PR is complemantry PR for this one:
https://github.com/valkey-io/valkey-doc/pull/275

### DIFF
--- a/src/commands/ft._list.json
+++ b/src/commands/ft._list.json
@@ -1,0 +1,15 @@
+{
+  "FT._LIST": {
+    "acl_categories": [
+      "FAST",
+      "READ",
+      "SEARCH"
+    ],
+    "arguments": [],
+    "arity": 1,
+    "complexity": "O(1)",
+    "group": "search",
+    "module_since": "1.0.0",
+    "summary": "Lists the currently defined indexes"
+  }
+}
--- a/src/commands/ft.create.json
+++ b/src/commands/ft.create.json
@@ -1,0 +1,14 @@
+{
+  "FT.CREATE": {
+    "acl_categories": [
+      "SLOW",
+      "WRITE",
+      "SEARCH"
+    ],
+    "arguments": [],
+    "complexity": "Construction time O(N log N), where N is the number of indexed items",
+    "group": "search",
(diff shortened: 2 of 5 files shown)


src/commands/ft._list.json
```diff
@@ -0,0 +1,15 @@
+{
```
yairgott, line 1: I wonder if it would make better sense to place all the docs under some directory, say docs, .e.g. //docs/commands

eifrah-aws: I agree. I basically copied the logic from other modules where the root makefile (which is located under `valkey-doc`) is collecting JSON files from `src/commands` folder

eifrah-aws: I suggest to postpone this as this requires a major change to the current Makefile (in valkey-doc)

src/commands/ft._list.json
```diff
@@ -0,0 +1,15 @@
...
+  "FT._LIST": {
+    "acl_categories": [
+      "FAST",
+      "READ",
+      "SEARCH"
+    ],
+    "arguments": [],
+    "arity": 1,
+    "complexity": "O(1)",
+    "group": "search",
```
yairgott, line 11: This is the ACL group mapping, right?

The assigned groups per command is listed [here](https://github.com/valkey-io/valkey-search/blob/main/src/commands/commands.h#L63).

yairgott (changes requested):

src/commands/ft.create.json
```diff
@@ -0,0 +1,14 @@
+{
+  "FT.CREATE": {
+    "acl_categories": [
+      "SLOW",
```
yairgott: Here and the other commands, the ACL categories should aligned to the categories mapped in the [code](https://github.com/valkey-io/valkey-search/blob/main/src/commands/commands.h#L65).

yairgott (changes requested):

src/commands/ft.create.json
eifrah-aws: Done

yairgott (approved):

PR #125, opened 2025-05-15
Command registration permissions alignment with ACL prefix permissions

yairgott:
While the ACL enhancement introduced in valkey-search allows command filtering based on prefixes, the engine enforces ACL checks on incoming commands independently. Currently, there's a misalignment between the permissions used during command registration and the permissions evaluated by valkey-search’s prefix-based ACL. This mismatch can result in the engine blocking legitimate commands. For example, if a user is allowed to execute only slow commands, an FT.SEARCH invocation may be blocked by the engine because it wasn't registered with the slow ACL category.

I manually verified this PR and integration tests should be added in the future.

In addition, this PR sets `admin` permission to FT._LIST, [similarly](https://redis.io/docs/latest/commands/ft._list/) to RediSearch. This is needed to be captured under the Valkey documentation.

### DIFF
--- a/src/commands/commands.h
+++ b/src/commands/commands.h
@@ -30,7 +30,6 @@
 #ifndef VALKEYSEARCH_SRC_COMMANDS_COMMANDS_H_
 #define VALKEYSEARCH_SRC_COMMANDS_COMMANDS_H_
 
-#include "absl/container/flat_hash_map.h"
 #include "absl/container/flat_hash_set.h"
 #include "absl/status/status.h"
 #include "absl/strings/string_view.h"
@@ -51,6 +50,7 @@ constexpr absl::string_view kReadCategory{"@read"};
 constexpr absl::string_view kWriteCategory{"@write"};
 constexpr absl::string_view kFastCategory{"@fast"};
 constexpr absl::string_view kSlowCategory{"@slow"};
+constexpr absl::string_view kAdminCategory{"@admin"};
 
 constexpr absl::string_view kCreateCommand{"FT.CREATE"};
 constexpr absl::string_view kDropIndexCommand{"FT.DROPINDEX"};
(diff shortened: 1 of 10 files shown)


src/commands/commands.h
```diff
@@ -51,24 +50,32 @@ constexpr absl::string_view kReadCategory{"@read"};
...
-};
+const absl::flat_hash_set<absl::string_view> kCreateCmdPermissions{
+    kSearchCategory, kWriteCategory, kFastCategory};
+const absl::flat_hash_set<absl::string_view> kDropIndexCmdPermissions{
+    kSearchCategory, kWriteCategory, kFastCategory};
+const absl::flat_hash_set<absl::string_view> kSearchCmdPermissions{
+    kSearchCategory, kReadCategory, kSlowCategory};
+const absl::flat_hash_set<absl::string_view> kInfoCmdPermissions{
+    kSearchCategory, kReadCategory, kFastCategory};
+const absl::flat_hash_set<absl::string_view> kLISTCmdPermissions{
```
eifrah-aws: nit: change `kLISTCmdPermissions` -> `kListCmdPermissions`

eifrah-aws (changes requested):
You should also fix the `src/commands/ft._list.json` file to include "ADMIN"

src/module_loader.cc
```diff
@@ -44,39 +45,63 @@
...
+  }
+  return absl::StrJoin(permissions, " ");
+}
+
 vmsdk::module::Options options = {
     .name = "search",
+    .acl_categories =
+        {
+            "search",
+        },
```
murphyjacob4: nit: Could do ACLPermissionFormatter(kSearchCategory)

src/module_loader.cc
```diff
@@ -44,39 +45,63 @@
...
     .version = MODULE_VERSION,
     .info = valkey_search::ModuleInfo,
     .commands =
         {
             {
                 .cmd_name = valkey_search::kCreateCommand,
-                .permissions = "write deny-oom",
+                .permissions = ACLPermissionFormatter(
+                    valkey_search::kCreateCmdPermissions),
+                .deny_oom = true,
```
murphyjacob4: So is deny_oom pulled out of the permissions since it isn't a command category?

To me, it might make sense to have two fields, one for "categories" and one for "permissions"? This way deny_oom isn't special cased, e.g. if there are other non-category permissions we want to add later.

murphyjacob4 (approved):

murphyjacob4 (approved):

jeon1226:
Thank you for merging the two source of truth of cmd category information.

Not sure how the command category information is handled in the engine, but if we assume that the category information in the engine (`ACL CAT @category`) it will be a problem in the future as well, since we have two source of truth, one from the module and one from the engine, and we need to find a way to enforce ACL using the category information from the engine not the module. When I implemented this I considered to call `ACL CAT @category` every time the command is executed, but it would add quite overhead so I didn't do that. The alternative considered was to use a cached ACL information (cmd category as well) and add a hook to this ACL commands to update the cache whenever ACL in the engine is updated, but Valkey didn't support the hook for ACL change.

Issue #39, opened 2025-02-07
Dynamically Configurable ThreadPool sizes


allenss-amazon:
Making the threadpool sizes dynamically configurable would be a huge boon development testing and characterization.

dpbnasika:
hey @allenss-amazon , i would like to work on this.

allenss-amazon:
Go for it! Having this as a feature will really help developers. There's all sorts of testing situations where being able to dynamically change the # of threads is really helpful.

yairgott:
One of the drawback of dynamically adjusting the thread pool size is that it involve more complication and potentially more contention. 

I would like to better understand the use-case when it's expected to be useful.

allenss-amazon:
Performance measurement and tuning. I've often found that when investing a problem that's live an persisting that by manipulating the # of thread and their mutual priority that either the situation resolved OR a better understanding of the problem is obtained.

dpbnasika:
@allenss-amazon i have gone through the source, is this where it is hardcoded?. and do we need this to dynamically configure?.

`struct Parameters {
  int reader_threads{10};
  int writer_threads{30};
  std::optional<int> threads;
  bool use_coordinator{false};
  std::optional<std::string> log_level;
};`

but i would like to have a bit intro to the dev environment, perhaps, in the call i would also like to join on wednesday, as i am very new to this project and would like to understand the workflow and get a basic understanding of the project. would that be fine?.

allenss-amazon:
I believe that those are where the default values are set. I believe that those can be overridden at module load time.

You're welcome to join the wednesday call.

dpbnasika:
@yairgott we could close this too i think!

PR #126, opened 2025-05-16
Dynamic thread pool

eifrah-aws:
Extended `ThreadPool` class with a "Resize" method
which allows the caller to increase or reduce the
thread count at runtime. In order to reduce the thread count,
the ThreadPool class associates a metadata per running thread `Thread`
the thread, within its "main" loop, checks if it was signaled to exit.

If it does, than the thread leaves its main loop while moving itself to
a "pending join" queue. The idea is that at given interval (using existing
timer events) the engine will periodically call `ThreadPool::JoinTerminatedWorkers`
to join threads that were exit. 

The default `ThreadPool::JoinWorkers` now internally also calls `ThreadPool::JoinTerminatedWorkers`

Fixed flaky ThreadPool.priority test

Added new `Vector<T>` class which is a thread-safe wrapper around `std::vector` and provides convenience methods such as `PopIf(predicate)`, `PopBackMulti(..)` and `ClearWithCallback(cb)`

Related issue: https://github.com/valkey-io/valkey-search/issues/39

### DIFF
--- a/vmsdk/src/thread_pool.cc
+++ b/vmsdk/src/thread_pool.cc
@@ -33,6 +33,7 @@
 #include <cstddef>
 #include <memory>
 #include <numeric>
+#include <ranges>
 #include <string>
 #include <utility>
 
@@ -44,9 +45,24 @@
 #include "absl/synchronization/mutex.h"
 namespace {
 
+class ThreadRunContext {
+ public:
+  ThreadRunContext(vmsdk::ThreadPool *pool,
+                   std::shared_ptr<vmsdk::ThreadPool::Thread> data)
+      : pool_{pool}, thread_{data} {}
+
+  vmsdk::ThreadPool *GetPool() { return pool_; }
+  std::shared_ptr<vmsdk::ThreadPool::Thread> GetThread() { return thread_; }
+
+ private:
+  vmsdk::ThreadPool *pool_ = nullptr;
+  std::shared_ptr<vmsdk::ThreadPool::Thread> thread_ = nullptr;
+};
+
 void *RunWorkerThread(void *arg) {
-  vmsdk::ThreadPool *pool = static_cast<vmsdk::ThreadPool *>(arg);
-  pool->WorkerThread();
+  ThreadRunContext *ctx = static_cast<ThreadRunContext *>(arg);
+  ctx->GetPool()->WorkerThread(ctx->GetThread());
+  delete ctx;  // shallow delete
   return nullptr;
 }
 }  // namespace
@@ -54,19 +70,14 @@ void *RunWorkerThread(void *arg) {
 namespace vmsdk {
 
 ThreadPool::ThreadPool(const std::string &name_prefix, size_t num_threads)
-    : threads_(num_threads),
+    : initial_thread_count_(num_threads),
       priority_tasks_(static_cast<int>(ThreadPool::Priority::kMax) + 1),
       name_prefix_(name_prefix) {}
 
 void ThreadPool::StartWorkers() {
   CHECK(!started_);
   started_ = true;
-  for (size_t i = 0; i < threads_.size(); ++i) {
-    pthread_create(&threads_[i], nullptr, RunWorkerThread, this);
-#ifndef __APPLE__
-    pthread_setname_np(threads_[i], (name_prefix_ + std::to_string(i)).c_str());
-#endif
-  }
+  IncrThreadCountBy(initial_thread_count_);
 }
 
 bool ThreadPool::Schedule(absl::AnyInvocable<void()> task, Priority priority) {
@@ -106,13 +117,17 @@ void ThreadPool::JoinWorkers() {
     }
     suspend_workers_ = false;
   }
-  if (!started_) {
-    return;
-  }
-  for (auto thread : threads_) {
-    pthread_join(thread, nullptr);
-  }
+
+  threads_.ClearWithCallback(
+      [](auto thread) { pthread_join(thread->thread_id, nullptr); });
   started_ = false;
+
+  JoinTerminatedWorkers();
+}
+
+void ThreadPool::JoinTerminatedWorkers() {
+  pending_join_threads_.ClearWithCallback(
+      [](auto thread) { pthread_join(thread->thread_id, nullptr); });
 }
 
 absl::Status ThreadPool::SuspendWorkers() {
@@ -132,7 +147,7 @@ absl::Status ThreadPool::SuspendWorkers() {
     }
     suspend_workers_ = true;
     blocking_refcount_ =
-        std::make_unique<absl::BlockingCounter>(threads_.size());
+        std::make_unique<absl::BlockingCounter>(threads_.Size());
     condition_.SignalAll();
   }
   blocking_refcount_->Wait();
@@ -154,7 +169,7 @@ absl::Status ThreadPool::ResumeWorkers() {
     }
     suspend_workers_ = false;
     blocking_refcount_ =
-        std::make_unique<absl::BlockingCounter>(threads_.size());
+        std::make_unique<absl::BlockingCounter>(threads_.Size());
   }
 
   blocking_refcount_->Wait();
@@ -176,7 +191,7 @@ void ThreadPool::AwaitSuspensionCleared()
   }
 }
 
-void ThreadPool::WorkerThread() {
+void ThreadPool::WorkerThread(std::shared_ptr<Thread> thread) {
   while (true) {
     absl::AnyInvocable<void()> task;
     {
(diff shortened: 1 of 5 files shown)


vmsdk/src/thread_pool.cc
```diff
@@ -44,29 +45,39 @@
 #include "absl/synchronization/mutex.h"
 namespace {
 
+class ThreadRunContext {
+ public:
+  ThreadRunContext(vmsdk::ThreadPool *pool,
+                   std::shared_ptr<vmsdk::ThreadPool::Thread> data)
```
allenss-amazon: data -> thread

vmsdk/src/vector.h
```diff
@@ -0,0 +1,86 @@
...
+#include <optional>
+#include <vector>
+
+#include "absl/algorithm/container.h"
+#include "absl/synchronization/mutex.h"
+
+/// A thread safe vector protected by a mutex
+namespace vmsdk {
+template <typename T>
+class Vector {
```
allenss-amazon: Nit: I'd call this something more descriptive, like ThreadSafeVector.

allenss-amazon (commented):
A couple of naming nits, otherwise it looks good to me.

allenss-amazon (approved):

vmsdk/src/thread_pool.cc
```diff
@@ -44,29 +45,39 @@
...
+  ThreadRunContext(vmsdk::ThreadPool *pool,
+                   std::shared_ptr<vmsdk::ThreadPool::Thread> thread)
+      : pool_{pool}, thread_{thread} {}
+
+  vmsdk::ThreadPool *GetPool() { return pool_; }
+  std::shared_ptr<vmsdk::ThreadPool::Thread> GetThread() { return thread_; }
+
+ private:
+  vmsdk::ThreadPool *pool_ = nullptr;
+  std::shared_ptr<vmsdk::ThreadPool::Thread> thread_ = nullptr;
```
yairgott, line 59: Assigning nullptr here is redundant.

vmsdk/src/thread_pool.h
```diff
@@ -66,14 +76,48 @@ class ThreadPool {
...
+  /// Resize the pool size to `count` threads. If `wait_for_resize` is `true`,
+  /// this method waits for resize operation to complete, otherwise, the resize
+  /// operation is done asynchronously.
+  void Resize(size_t count, bool wait_for_resize = false);
+
+  /// A struct representing a worker thread
+  struct Thread {
+    bool IsShutdown() const { return shutdown_flag.load(); }
+    void Shutdown(absl::AnyInvocable<void()> callback = nullptr) {
+      if (callback != nullptr) {
```
yairgott, line 94: This if statement seems redundant

vmsdk/src/thread_pool.h
```diff
@@ -66,14 +76,48 @@ class ThreadPool {
...
+      if (shutdown_callback.has_value()) {
+        (*shutdown_callback)();
+      }
+    }
+
+    pthread_t thread_id = 0;
+    std::atomic_bool shutdown_flag = false;
+    /// If not null, the thread will call this callback when it exits via the
+    /// shutdown_flag
+    std::optional<absl::AnyInvocable<void()>> shutdown_callback = std::nullopt;
```
yairgott, line 111: Assigning std::nullopt is redundant

vmsdk/src/thread_safe_vector.h
```diff
@@ -0,0 +1,86 @@
...
+    }
+    vec_.clear();
+  }
+
+  /// This is the same as calling `ClearWithCallback()` with `nullptr` as the
+  /// callback
+  void Clear() { ClearWithCallback(nullptr); }
+
+ private:
+  std::vector<T> vec_;
```
yairgott, line 83: use ABSL_GUARDED_BY(mutex_)

vmsdk/src/thread_safe_vector.h
```diff
@@ -0,0 +1,86 @@
...
+  absl::Mutex mutex_;
```
yairgott, line 84: should be declared as mutable

vmsdk/src/thread_safe_vector.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
```
yairgott, line 1: why this class is needed? It seems to me that simply using std::vector with a mutex in ThreadPool would be better approach.

yairgott, line 1: Missing copyright header

vmsdk/src/thread_safe_vector.h
```diff
@@ -0,0 +1,86 @@
...
+  ThreadSafeVector(const ThreadSafeVector&) = delete;
+  ThreadSafeVector& operator=(const ThreadSafeVector&) = delete;
+  ThreadSafeVector() = default;
+
+  size_t Size() const { return vec_.size(); }
+  bool IsEmpty() const { return Size() == 0; }
+
+  /// Pop the first item that matches the predicate (starting from the top of
+  /// the list)
+  std::optional<T> PopIf(std::function<bool(T)> predicate) {
```
yairgott, line 27: Here and all the other functions where mutex_ is acquired, use ABSL_LOCKS_EXCLUDED(mutex_)

vmsdk/src/thread_pool.cc
```diff
@@ -176,15 +191,25 @@ void ThreadPool::AwaitSuspensionCleared()
...
       AwaitSuspensionCleared();
       auto condition = absl::Condition(this, &ThreadPool::QueueReady);
       while (!condition.Eval()) {
-        condition_.Wait(&queue_mutex_);
+        condition_.WaitWithTimeout(&queue_mutex_, absl::Seconds(1));
+        if (thread->IsShutdown()) {
+          thread->InvokeShutdownCallback();
+          // remove this thread from the threads list and place it in the
+          // pending join list
+          threads_.PopIf([thread](std::shared_ptr<Thread> t) {
```
yairgott, line 207: `thread` should be captured by reference.

vmsdk/src/thread_pool.cc
```diff
@@ -44,29 +45,39 @@
 #include "absl/synchronization/mutex.h"
 namespace {
 
+class ThreadRunContext {
+ public:
```
yairgott, line 49: copy constructor and assignment operator should be deleted.

vmsdk/src/thread_pool.h
```diff
@@ -66,14 +76,48 @@ class ThreadPool {
...
       ABSL_LOCKS_EXCLUDED(queue_mutex_);
-  void WorkerThread() ABSL_LOCKS_EXCLUDED(queue_mutex_);
+
+  /// Resize the pool size to `count` threads. If `wait_for_resize` is `true`,
+  /// this method waits for resize operation to complete, otherwise, the resize
+  /// operation is done asynchronously.
+  void Resize(size_t count, bool wait_for_resize = false);
+
+  /// A struct representing a worker thread
+  struct Thread {
```
yairgott, line 91: copy constructor and assignment operator should be deleted.

vmsdk/src/thread_safe_vector.h
```diff
@@ -0,0 +1,86 @@
...
+      for (auto& item : vec_) {
+        callback(item);
+      }
+    }
+    vec_.clear();
+  }
+
+  /// This is the same as calling `ClearWithCallback()` with `nullptr` as the
+  /// callback
+  void Clear() { ClearWithCallback(nullptr); }
```
yairgott, line 80: Is this needed?

vmsdk/src/thread_safe_vector.h
```diff
@@ -0,0 +1,86 @@
...
+namespace vmsdk {
+template <typename T>
+class ThreadSafeVector {
+ public:
+  // This type is neither copyable nor assignable.
+  ThreadSafeVector(const ThreadSafeVector&) = delete;
+  ThreadSafeVector& operator=(const ThreadSafeVector&) = delete;
+  ThreadSafeVector() = default;
+
+  size_t Size() const { return vec_.size(); }
```
yairgott, line 22: mutex should be acquired.

vmsdk/src/thread_safe_vector.h
yairgott, line 27: should use absl::AnyInvocable<...>

vmsdk/src/thread_safe_vector.h
eifrah-aws: A leftover from previous iteration. Will remove it

vmsdk/src/thread_safe_vector.h
eifrah-aws: IIRC, all `const` methods of any container in the STL can be accessed concurrently without locking.

vmsdk/src/thread_safe_vector.h
eifrah-aws: I tried it. It does not work with `std::find_if` (nor its absl equivalent `absl::c_find_if`)

vmsdk/src/thread_safe_vector.h
yairgott: hmm, AFAIK any const method should acquire at least a read mutex.

vmsdk/src/thread_safe_vector.h
yairgott: You might need to wrap those.

PR #127, opened 2025-05-18
Allow changing the reader/writer threads count

eifrah-aws:
With this PR, user can control the threads counts
using

```
config set search.reader-threads <value>
config get search.reader-threads

config set search.writer-threads <value>
config get search.writer-threads
```

In addition, added new helper class `ValkeySearchConfig` which provides a convenient methods for registering numeric configuration items

threads

### DIFF
--- a/src/valkey_search.cc
+++ b/src/valkey_search.cc
@@ -59,6 +59,7 @@
 #include "src/schema_manager.h"
 #include "src/utils/string_interning.h"
 #include "src/vector_externalizer.h"
+#include "valkey_search_config.h"
 #include "vmsdk/src/command_parser.h"
 #include "vmsdk/src/concurrency.h"
 #include "vmsdk/src/latency_sampler.h"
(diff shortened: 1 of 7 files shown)


src/valkey_search.cc
```diff
@@ -556,51 +584,60 @@ void ValkeySearch::ResumeWriterThreadPool(RedisModuleCtx *ctx,
...
+      ctx, kReaderThreadsConfig, default_threads_count, 1, kMaxThreadsCount,
+      // Callback for "config set"
+      [this](long long val) {
+        if (!reader_thread_pool_) {
+          return;
+        }
+        VMSDK_LOG(NOTICE, nullptr) << "Reader threads count set to: " << val;
+        reader_thread_pool_->Resize(val);
+      },
+      // Callback for "config get"
```
allenss-amazon, line 616: we will eventually have 10's if not 100's of config variables. I think this boilerplate code can be centralized in a class or multiple classes. 

I'd rather see configuration variables declared as global statics (and have them automatically self-register)

PR #129, opened 2025-05-19
Allow changing the reader/writer threads count at runtime

eifrah-aws:
With this PR, user can control the threads counts
using

```
config set search.reader-threads <value>
config get search.reader-threads

config set search.writer-threads <value>
config get search.writer-threads
```

In addition, refactored the configuration registration mechanism. Calling `Builder<type>("my-config",...).Build()` will create an instance of configuration entry named `my-config` that can be controlled from the Valkey's `CONFIG SET|GET search.my-config` command AND from the command line by passing `--my-config ...` 

**Example 1:**

In order to register a configuration entry with Valkey, named: `search.hnsw-block-size`, the developer needs to place similar lines of code in one of his source files:

```c++
// Include the vmsdk::config namespace
#include "vmsdk/src/module_config.h"

...

// Controls the HNSW resize increments
namespace config = vmsdk::config;

static auto hnsw_block_size =
    config::NumberBuilder("hnsw-block-size",      // name
                          kHNSWDefaultBlockSize,  // default size
                          kHNSWMinimumBlockSize,  // min size
                          UINT_MAX)               // max size
        .WithValidationCallback(ValidateHNSWBlockSize)
        .Build();

// Define some boolean flag and initialize it with false
// we can later change it using `config set search.my-bool-flag yes`
auto my_bool_flag =
    config::BooleanBuilder("my-bool-flag", false).Build();
```

With the above example, it is now also possible to pass these configuration from the
command line like this:

```bash
valkey-server "--loadmodule libsearch.so --my-bool-flag yes --hnsw-block-size 20480"
```

**Example 2:**

We can also register a configuration entry with an action to take when the configuration item is modified:

```c++
// Include the vmsdk::config namespace
#include "vmsdk/src/module_config.h"

...

namespace config = vmsdk::config;

// Controls the number of reader threads
static auto reader_threads_count =
    config::NumberBuilder(
        kReaderThreadsConfig,  // name
        kDefaultThreadsCount,  // default size
        1,                     // min size
        kMaxThreadsCount)      // max size
        .WithModifyCallback(   // set an "On-Modify" callback
            [](long long new_value) {
              UpdateThreadPoolCount(
                  ValkeySearch::Instance().GetReaderThreadPool(), new_value);
            })
        .Build();
```

Note that the callback is called if you set it programmatically like this:

```c++
VMSDK_RETURN_IF_ERROR(reader_threads_count->SetValue(100));
```

or using `valkey-cli`:

```bash
config set search.reader-threads 100
```

threads


vmsdk/src/module_config.cc
```diff
@@ -0,0 +1,150 @@
...
+  auto entry = reinterpret_cast<Number *>(priv_data);
+  CHECK(entry) << "null private data for Number configuration entry.";
+  return entry->GetValue();
+}
+
+int Number::OnSetNumericConfig(const char *config_name, long long value,
+                               void *priv_data, RedisModuleString **err) {
+  auto entry = reinterpret_cast<Number *>(priv_data);
+  CHECK(entry) << "null private data for configuration Number entry.";
+  entry->SetValue(value);
```
allenss-amazon: This code fails to allow some kind of "validation" function. IMO, an optional validation function should be provided.

allenss-amazon (commented):
1. Need to provide some form of optional validation function. Just having ranges is insufficient for the general case.
2. Should provide some mechanism for adding in the configuration flags like REDISMODULE_CONFIG_IMMUTABLE, etc. 
3. Need some kind of full end to end test, i.e., python based test

vmsdk/src/module_config.cc
eifrah-aws: Done. The validation is now done internally

src/valkey_search.cc
```diff
@@ -65,17 +64,87 @@
...
+static const int64_t kDefaultThreadsCount = vmsdk::GetPhysicalCPUCoresCount();
+
+namespace {
+/// Check that the new value for configuration item `hnsw-block-size` confirms
+/// to the allowed values.
+bool ValidateHNSWBlockSize(long long new_value, RedisModuleString **err) {
+  if (new_value < kHNSWDefaultBlockSize || new_value > UINT32_MAX) {
+    if (err) {
+      *err = RedisModule_CreateStringPrintf(
+          nullptr, "Block size must be between %u and %u",
```
allenss-amazon: Isn't this dead code now?

src/valkey_search.cc
```diff
@@ -65,17 +64,87 @@
...
+  }
+  pool->Resize(new_value);
+}
+}  // namespace
+
+// Configuration entries
+namespace config = vmsdk::config;
+
+// Controls the HNSW resize increments
+static auto hnsw_block_size = config::Builder<config::Number, long long>(
```
allenss-amazon: Isn't the first type redundant with the second type?

vmsdk/src/module_config.h
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  bool default_value_ = false;
+  std::atomic_bool current_value_ = 0;
+
+  FRIEND_TEST(Builder, ConfigBuilder);
+};
+
+template <typename T, typename ValkeyT>
+class ConfigBuilder {
```
allenss-amazon: I think ConfigBuilder should only have a single type, here: "ValkeyT", the other type "T" is a funciton of ValkeyT and can be computed accordingly.

src/valkey_search.cc
```diff
@@ -493,16 +573,12 @@ absl::Status ValkeySearch::LoadOptions(RedisModuleCtx *ctx,
...
-      RedisModule_FreeString(nullptr, err);
-      return absl::InternalError(absl::StrFormat(
-          "Failed to set vs-hnsw-block-size config from command-line: %s",
-          error_msg));
+    if (!ValidateHNSWBlockSize(options.hnsw_block_size, nullptr)) {
+      return absl::InvalidArgumentError(
+          absl::StrFormat("Could not set %s. Invalid value: %u",
+                          kHNSWBlockSizeConfig, options.hnsw_block_size));
     }
+    hnsw_block_size->SetValue(options.hnsw_block_size);
```
allenss-amazon: In my CR the command line parsing was also handled by the config variables. I thought that was included here but apparently not. I think it should be.

eifrah-aws: I was not aiming of making it part of this PR. But I can add it.

src/valkey_search.cc
```diff
@@ -65,17 +64,87 @@
...
+const absl::string_view kReaderThreadsConfig{"reader-threads"};
+const absl::string_view kWriterThreadsConfig{"writer-threads"};
+
+static const int64_t kDefaultThreadsCount = vmsdk::GetPhysicalCPUCoresCount();
+
+namespace {
+/// Check that the new value for configuration item `hnsw-block-size` confirms
+/// to the allowed values.
+bool ValidateHNSWBlockSize(long long new_value, RedisModuleString **err) {
+  if (new_value < kHNSWDefaultBlockSize || new_value > UINT32_MAX) {
```
yairgott: why `new_value` should be greater than kHNSWDefaultBlockSize?

It has to be greater than 0.

vmsdk/src/module_config.h
```diff
@@ -0,0 +1,246 @@
...
+  absl::Status Register(RedisModuleCtx *ctx) override;
+  bool GetValueImpl() const override {
+    return current_value_.load(std::memory_order_relaxed);
+  }
+
+  void SetValueImpl(bool val) override {
+    current_value_.store(val, std::memory_order_relaxed);
+  }
+
+  bool default_value_ = false;
```
yairgott: nit: here and other places, prefer using: 
`bool default_value_{false}`

vmsdk/src/module_config.h
```diff
@@ -0,0 +1,246 @@
...
+
+ private:
+  std::vector<Registerable *> entries_;
+  friend class Configbase;
+};
+
+template <typename T>
+class ConfigBase : public Registerable {
+ public:
+  using OnModifyCB = std::function<void(T)>;
```
yairgott, line 153: Prefer absl::AnyInvocable over std::function

src/valkey_search.cc
eifrah-aws: Agreed. We can remove it since in our case, there are only limited number of item types (long long, bool, string & flags). Removed and further simplified the developer experience

vmsdk/src/module_config.h
eifrah-aws: Done

src/valkey_search.cc
eifrah-aws: Good catch. This bring the question: do we want to set a minimal value which is greater than `0` ?

src/valkey_search.cc
eifrah-aws: Still using it in the part we I we set the block size from the command line argument.

vmsdk/src/module_config.h
eifrah-aws: What is the advantage of `absl::AnyInvokable` over `std::function`?

src/valkey_search.cc
eifrah-aws: For now, I set to to `0`. We can change this later if needed

vmsdk/src/module_config.h
eifrah-aws: Anyways, the code does not compile when using `AnyInvokable` - we can take this offline if you want :)

src/valkey_search.cc
eifrah-aws: Done

src/valkey_search_options.cc
```diff
@@ -0,0 +1,222 @@
...
+            if (ValidateLogLevel(value).ok()) {
+              vmsdk::InitLogging(nullptr, kLogLevelNames[value].data())
+                  .IgnoreError();
+            }
+          })
+          .WithValidationCallback(ValidateLogLevel)
+          .Build();
+}
+
+/// If user passed "--threads" it triumphs any value provided by
```
murphyjacob4: For --threads: would it make sense to remove this for the 1.0 release? I think it is only needed for backwards compatibility with the original Google deployment.

Same goes for the --use-coordinator no argument special case: should we make this conform before 1.0 while we have the chance?

eifrah-aws: About the `--threads`: I can remove it, but I would prefer doing it in a separate PR (there are multiple unit tests + integration tests that are using it). There are also some tests such as `threads_win` that check that "--threads" overcome "--reader/writer-threads"...

`--use-coordinator` is indeed a different beast, as it is being used as a boolean flag without Valkey convention. Fixing it will remove some code that I added for it

allenss-amazon: I had originally assumed that 1.0 would ship before this work reached completion. With the delays in 1.0, my desire to put this into 1.0 has climbed. I feel like we're close enough to being able to merge all of this and avoid the legacy issues. So I'd vote to delay 1.0 to complete this -- under the assumption that we're only talking about a day or few.

murphyjacob4: Yeah - definitely shouldn't be lumped with this PR, since we wouldn't want to block 1.0 on this PR that is more of a feature. But if we want to do breaking changes - now is the time

murphyjacob4: @yairgott wdyt?

murphyjacob4: Not opposed to blocking 1.0 on this - and cleaning up the configs as part of this

src/valkey_search_options.cc
```diff
@@ -0,0 +1,222 @@
...
+constexpr absl::string_view kLogLevel{"log-level"};
+constexpr absl::string_view kThreads{"threads"};
+
+static const int64_t kDefaultThreadsCount = vmsdk::GetPhysicalCPUCoresCount();
+
+namespace {
+/// Check that the new value for configuration item `hnsw-block-size` confirms
+/// to the allowed values.
+absl::Status ValidateHNSWBlockSize(long long new_value) {
+  if (new_value < kHNSWMinimumBlockSize || new_value > UINT32_MAX) {
```
yairgott, line 74: why new_value should be lower than kHNSWMinimumBlockSize ?

Though, new_value should be greater than 0.

src/valkey_search_options.cc
yairgott: Agreed. Lets cleanup the config and include this PR in 1.0.

yairgott: On second thought, I think it makes more sense to create a dedicated PR focused solely on cleaning up the config, rather than rushing a larger change and potentially risking the stability of 1.0.
Thoughts?

src/valkey_search.cc
```diff
@@ -143,7 +91,11 @@ void ValkeySearch::InitInstance(std::unique_ptr<ValkeySearch> instance) {
 }
 
 uint32_t ValkeySearch::GetHNSWBlockSize() const {
-  return hnsw_block_size.load(std::memory_order_relaxed);
+  return Options::Instance().GetHNSWBlockSize().GetValue();
+}
+
+void ValkeySearch::SetHNSWBlockSize(uint32_t block_size) {
+  Options::Instance().GetHNSWBlockSize().SetValue(block_size).IgnoreError();
```
yairgott: Here and other places, I don't see a reason why we would want to ignore errors, at minimum we should log those.

yairgott (changes requested):

src/valkey_search_options.cc
eifrah-aws: 1. `kHNSWMinimumBlockSize` is `0` (line 40 on this file)
2. Valkey will not pass the value to the module if it is smaller than the minimal value defined when we first registered the configuration. So this validation basically mimics that behavior if a developer adds a call somewhere in the code to `hnsw_block_size->SetValue(new_value)` or if it passes it from the command line `--loadmodule ... --hnsw-block-size -10`.

src/valkey_search_options.cc
eifrah-aws: `95%` of this PR is focusing on improving the configuration mechanism, so in theory it should have no effect on the stability of the module once the module is up and running. The only thing I can think of is the ability to change the number of threads during runtime (which we can prevent by adding the `IMMUTABLE` flag to the `--reader-threads` + `--writer-threads` for the `1.0` release and remove it `main`)

As a side note, the level of confidence I have atm with the module comes from the integration tests + unit testing + playing with it locally. Since everything passes with this change included, I have the same level of confidence.

I vote to include it.

src/valkey_search_options.cc
```diff
@@ -0,0 +1,222 @@
...
+namespace options {
+
+constexpr uint32_t kHNSWDefaultBlockSize{10240};
+constexpr uint32_t kHNSWMinimumBlockSize{0};
+constexpr uint32_t kMaxThreadsCount{1024};
+
+constexpr absl::string_view kHNSWBlockSizeConfig{"hnsw-block-size"};
+constexpr absl::string_view kReaderThreadsConfig{"reader-threads"};
+constexpr absl::string_view kWriterThreadsConfig{"writer-threads"};
+constexpr absl::string_view kUseCoordinator{"use-coordinator"};
```
murphyjacob4, line 46: I was talking with Yair offline - do you think we should just remove --use-coordinator in favor of just enabling/disabling it based on cluster mode?

allenss-amazon: I'm struggling to come up with ANY scenario where having --use-coordinator is something that we want to support in user deployments. Does it have any use in testing? I don't think so, I vote to remove it.

murphyjacob4: It is technically usable without --use-coordinator in cluster mode - but each node would only return the results from the local index. The only case I can think of where it might be useful is users who want to use hashtags or client-side coordination to retrieve results, instead of server side coordination

allenss-amazon: I have a proposal for Single slot indexes, which is sorta what this is similar to. But my proposal would let you intermix both cross-slot and single slot indexes on the same cluster, which makes it a superset. I'm a bit concerned that we might be painting ourselves into a hole by having this flag. I don't know how much testing of no-coordinator in a CME there has been, and I dont' think we really want to support that functionality over the long haul. So I'd vote to remove this flag.

yairgott: I'm also supportive of removing it at this point otherwise it would be harder to do so in the future. The single slot index suggested by Allen seems to address the scenario mentioned above.

vmsdk/src/module_config.cc
```diff
@@ -0,0 +1,300 @@
...
+      // All keys must accept values. Unless it is `--use-coordinator` (we
+      // keep this option for backward compatibility)
+      sv_val = "yes";
+      VMSDK_RETURN_IF_ERROR(UpdateConfigFromKeyVal(ctx, sv_key, sv_val));
+    } else {
+      // Read the value
+      if (!iter.HasNext()) {
+        return absl::NotFoundError(absl::StrFormat(
+            "Command line argument `%s` is missing a value", sv_key));
+      } else {
```
yairgott: nit: save the need to indent by skipping the `else` when the `if` always returns.

vmsdk/src/module_config.h
```diff
@@ -0,0 +1,349 @@
...
+    return *this;
+  }
+
+  auto &WithFlags(size_t flags) {
+    config_->SetFlags(flags);
+    return *this;
+  }
+
+  std::unique_ptr<ConfigBase<ValkeyT>> Build() {
+    return std::unique_ptr<ConfigBase<ValkeyT>>{config_};
```
yairgott, line 298: I see a few issues with the current approach:
1. `Build()` can be called multiple times, and on the second call it might return a `unique_ptr` to a dangling pointer.
2. It assumes that `Build()` will always be called once; otherwise, *config_ is leaked.
3. `ModuleConfigManager` assumes that its stored config items are valid and not deleted.

I wonder if we can address all of this by storing the config as a `shared_ptr` both internally and within `ModuleConfigManager`, and have `Build()` return a `shared_ptr` as well.

vmsdk/src/module_config.h
```diff
@@ -0,0 +1,349 @@
...
+    if (!res.ok()) {
+      return res;
+    }
+
+    SetValueImpl(value);
+    NotifyChanged();
+    return absl::OkStatus();
+  }
+
+  T GetValue() const { return GetValueImpl(); }
```
yairgott, line 193: I think that it would make better sense to return `const T&` which could be assigned to a T by the caller.

vmsdk/src/module_config.h
yairgott: It's the same concept as `std::move_only_function`. It ensures that the function will always be moved instead of copied. This has value when the function stores callables  copyable and it's good habit to use it in general.

vmsdk/src/module_config.h
eifrah-aws: I don't think its always possible to do so. For example, the concrete class `Number` uses this:

```c++
  long long GetValueImpl() const override {
    return current_value_.load(std::memory_order_relaxed);
  }
```

We can't really return a `const long long&` here.

vmsdk/src/module_config.h
```diff
@@ -0,0 +1,371 @@
...
+  size_t flags_{kDefault};
+};
+
+template <typename T>
+class ConfigBase : public Registerable {
+ public:
+  using OnModifyCB = std::function<void(T)>;
+  using ValidateCB = std::function<absl::Status(const T)>;
+
+  ConfigBase(std::string_view name) : Registerable(name) {
```
allenss-amazon: Shouldn't this be protected?

vmsdk/src/module_config.h
```diff
@@ -0,0 +1,371 @@
...
+    ModuleConfigManager::Instance().RegisterConfig(this);
+  }
+
+  ~ConfigBase() override = default;
```
allenss-amazon: shoud this unregister itself?

src/valkey_search_options.cc
```diff
@@ -0,0 +1,232 @@
...
+    REDISMODULE_LOGLEVEL_NOTICE,
+    REDISMODULE_LOGLEVEL_VERBOSE,
+    REDISMODULE_LOGLEVEL_DEBUG,
+};
+
+static const std::vector<int> kLogLevelValues = {
+    static_cast<int>(LogLevel::kWarning), static_cast<int>(LogLevel::kNotice),
+    static_cast<int>(LogLevel::kVerbose), static_cast<int>(LogLevel::kDebug)};
+
+void Options::DoInitialize() {
```
allenss-amazon: These are singletons. What is the point of putting them in an object that's manually created/destroyed. It's a bunch of extra verbiage with no point. Why can't these just be global variables? Much cleaner.

vmsdk/src/module_config.h
eifrah-aws: Done

src/valkey_search_options.cc
eifrah-aws: Indeed, it is cleaner, but this way it is easier to test.
You just reset the `Options::InitInstance` in the test `TearDown`/`SetUp`

allenss-amazon: Making the next 100 configurables harder to use doesn't justify making the testing easier to use now.

vmsdk/src/module_config.h
eifrah-aws: Done

src/valkey_search_options.cc
eifrah-aws: I disagree here. In order to register a new configuration, you just add a member to the `Options` class and initialize it in `Options::DoInitialize()`.

We could also make them public to avoid the extra `GetVARIABLE` (which is not always needed). 

Making them global / static requires adding some new method to be added (e.g. `Options::Cleanup()`) so we could use it in between tests. 
With global variables, you are also required to add some cleanup code into that function whenever you add new variable

eifrah-aws: BTW, this is not a requirement to add variables into `Options`, you can place it in any other `.cc` files and it will work in a similar way, you will just need to pay attention when testing the new variables to make sure to have them cleanup in-between tests

src/valkey_search_options.cc
eifrah-aws: I suggest to create a separate PR for:
- Remove `--threads`
- Remove `--use-coordinator` and rely on the cluster mode "(`cluster-enabeld`)" (at least until @allenss-amazon single slot index proposal will land)

Thoughts? @yairgott @murphyjacob4

src/valkey_search_options.h
```diff
@@ -0,0 +1,97 @@
...
+  /// **DEPRECATED** controls both the reader & writer pool size
+  /// use --reader/--writer-threads instead
+  std::shared_ptr<config::ConfigBase<long long>> threads_;
+  /// Controls the reader threads count
+  std::shared_ptr<config::ConfigBase<long long>> reader_threads_count_;
+  /// Controls the writer threads count
+  std::shared_ptr<config::ConfigBase<long long>> writer_threads_count_;
+  /// Should this instance launch coordinator?
+  std::shared_ptr<config::ConfigBase<bool>> use_coordinator_;
+  /// Control the modules log level verbosity
```
allenss-amazon: Why aren't these "config::Number" and "config::Boolean" ??

vmsdk/src/module_config.h
```diff
@@ -0,0 +1,379 @@
...
+/// block similar to this:
+///
+///
+/// ```c++
+/// namespace config = vmsdk::config;
+///
+/// // Create a Valkey configuration item named "config-name" of type `long
+/// // long`
+/// static auto my_config =
+///    config::Builder<long long>(
```
allenss-amazon: This doesn't actually work anymore, it will fault out at load time, because of the way that the ModuleConfigManager requires manual initialization BEFORE you can declare any config variables.

allenss-amazon (changes requested):
The redesign of the vmsdk configuration subsystem to require that all configs be dynamically created during module OnLoad initialization is exactly the opposite direction of where I wanted to go with this. 

The original concept was that configurables would be declared at file level and be initialized automatically. We want to encourage developers to create configurables for any variable that has a particular value, i.e., favor configurables over #define or const variables.

The key issue that I feel has been lost is the ease-of-use. You should be able to add a configurable in one line of code. With this design, it's approaching 10 lines of code. Also, it encourages bad modularization as the temptation is for developers to add another element to the Options structure, which is horribly wordy and will become yet another global dependency rathole.

vmsdk/src/module_config.h
eifrah-aws: Correct, fixed

eifrah-aws:
> The redesign of the vmsdk configuration subsystem to require that all configs be dynamically created during module OnLoad initialization is exactly the opposite direction of where I wanted to go with this.
> 
> The original concept was that configurables would be declared at file level and be initialized automatically. We want to encourage developers to create configurables for any variable that has a particular value, i.e., favor configurables over #define or const variables.
> 
> The key issue that I feel has been lost is the ease-of-use. You should be able to add a configurable in one line of code. With this design, it's approaching 10 lines of code. Also, it encourages bad modularization as the temptation is for developers to add another element to the Options structure, which is horribly wordy and will become yet another global dependency rathole.

I agree with the overall approach (ease-of-use), but I don't agree with the "static globals" approach. Anyways, and in-order to progress, I will "disgree-and-commit" here ;) so:

I implemented your suggestion and removed the manual instantiation of the `ModuleConfig` & removed `Options` (it is now a pure namespace `options`) 

This allows static variables to be registered before `OnLoad`.

allenss-amazon (approved):

src/valkey_search.cc
eifrah-aws: Fixed

src/valkey_search_options.cc
```diff
@@ -0,0 +1,231 @@
...
+static auto hnsw_block_size =
+    config::NumberBuilder(kHNSWBlockSizeConfig,   // name
+                          kHNSWDefaultBlockSize,  // default size
+                          kHNSWMinimumBlockSize,  // min size
+                          UINT_MAX)               // max size
+        .WithValidationCallback(ValidateHNSWBlockSize)
+        .Build();
+
+/// DEPRECATED: Register the "--threads" flag.
+static auto threads =
```
yairgott, line 125: Is this still needed? I don't see much value maintaining this.

eifrah-aws: Will create a PR to remove it and clear all usage of it in tests (we are referencing it in integ + UT)

Issue #130, opened 2025-05-19
Error while loading dump file created by valkey with the module (w


eifrah-aws:
**Problem**:

If the module is loaded and no VSS related commands were executed (no index, no data) and then valkey generates rdb file (using `BGSAVE` or `SAVE` commands), on the next startup `valkey` will exit with this error message:

```
18201:M 19 May 2025 21:01:11.475 # <search> [warning], tid: 139893335331712, /home/src/ElastiCacheValkeySearch/src/rdb_serialization.cc:238: Failed to load ValkeySearch aux section from RDB: RedisModule_LoadString failed; IO error while reading serialized RDBSection from RDB
18201:M 19 May 2025 21:01:11.475 # The RDB file contains module AUX data for the module type 'Vk-Search', that the responsible module is not able to load. Check for modules log above for additional clues.
18201:M 19 May 2025 21:01:11.475 # Short read or OOM loading DB. Unrecoverable error, aborting now.
18201:M 19 May 2025 21:01:11.475 # Short read or OOM loading DB. Unrecoverable error, aborting now.
18201:M 19 May 2025 21:01:11.475 # Internal error in RDB reading offset 0.
18201:M 19 May 2025 21:01:11.475 # Internal error in RDB reading offset 0, function at rdb.c:3891 -> Unexpected EOF reading RDB file
[offset 0] Checking RDB file dump.rdb
[offset 27] AUX FIELD valkey-ver = '8.1.1'
[offset 41] AUX FIELD redis-bits = '64'
[offset 53] AUX FIELD ctime = '1747688465'
[offset 68] AUX FIELD used-mem = '12374640'
[offset 80] AUX FIELD aof-base = '0'
[offset 92] MODULE AUX for: Vk-Search
[offset 110] Checksum OK
[offset 110] \o/ RDB looks OK! \o/
[info] 0 keys read
[info] 0 expires
[info] 0 already expired
18201:M 19 May 2025 21:01:11.475 * Cleaning up Amazon resources
18201:M 19 May 2025 21:01:11.477 * [Resp] Network server stop requested
```

**Steps to reproduce**:

- Start `valkey-server` with the module
- Connect using `valkey-cli` and run the `BGSAVE` command
- Terminate `valkey`
- Verify `dump.rdb` present
- Start `valkey-server` again

**culprit**:

The module writes in the RDB that it has `1` section but then it does not write any content.
This is because the `sections_count` method registered by the `MetadataManager` always returns `1`.

[...]

PR #131, opened 2025-05-19
Fixed corrupted dump.rdb file is created in some cases

eifrah-aws:
Fixes: https://github.com/valkey-io/valkey-search/issues/130


allenss-amazon (approved):

murphyjacob4:
Thanks for the fix! This will also fix RDB load on a non-ValkeySearch-module-enabled instance from a ValkeySearch-module-enabled instance with no indexes. We use auxsave2 to specifically support this, which may happen in the case of rolling upgrade (boot a replica with ValkeySearch, failover, then remove old primary). Looks good to me

PR #143, opened 2025-05-23
Use background context rather than IO context during Aux save.

### LINKED ISSUE #142: [BUG] Valkey + the module crashes on the next command when AOF is enabled
## Why this is happening?

During the module startup, the module registers callback for RDB actions, specifically this method: `valkey_search::AuxSaveCallback`.

In this method, we "Get" a context (I write here "Get" with double quotes since we are actually allocating one), however, we allocated a context which bumps the server nesting level (each time Valkey's `call()` is being called the nesting level is bumped by `1`).

The context is never freed... hence we leave the server in a limbo state (its nesting level (`server.execution_netsting == 1`). So on the next call Valkey fails one of its assertions and crashes.

Here is the last backtrace that increments `server.execution_nesting`:

```c++
Thread 1 "valkey-server" hit Hardware watchpoint 2: *(int *) 0x555555926f38

Old value = 0
New value = 1
enterExecutionUnit (update_cached_time=1, us=0) at /home/ubuntu/devl/valkey/src/server.c:1352
1352	    if (server.execution_nesting++ == 0 && update_cached_time) {
(gdb) bt
#0  enterExecutionUnit (update_cached_time=1, us=0) at /home/ubuntu/devl/valkey/src/server.c:1352
#1  0x00005555556c1fa8 in moduleCreateContext (out_ctx=0x7ffff7840e80, module=0x7ffff7840a00, ctx_flags=0) at /home/ubuntu/devl/valkey/src/module.c:984
#2  0x00005555556ce9b8 in VM_GetContextFromIO (io=0x7fffffffdc90) at /home/ubuntu/devl/valkey/src/module.c:7600
#3  0x00007ffff62aebb3 in valkey_search::AuxSaveCallback (rdb=0x7fffffffdc90, when=2) at /home/ubuntu/devl/valkey-search/src/rdb_serialization.cc:285
#4  0x0000555555633bc2 in rdbSaveSingleModuleAux (rdb=0x7fffffffde60, when=2, mt=0x7ffff785a240) at /home/ubuntu/devl/valkey/src/rdb.c:1276
#5  0x00005555556ce263 in rdbSaveModulesAux (rdb=0x7fffffffde60, when=2) at /home/ubuntu/devl/valkey/src/module.c:7389
#6  0x000055555563449a in rdbSaveRio (req=0, rdb=0x7fffffffde60, error=0x7fffffffde54, rdbflags=1, rsi=0x0) at /home/ubuntu/devl/valkey/src/rdb.c:1443
#7  0x0000555555668c03 in rewriteAppendOnlyFile (filename=0x7ffff7847f63 "appendonlydir/appendonly.aof.1.base.rdb") at /home/ubuntu/devl/valkey/src/aof.c:2328
#8  0x00005555556641bb in aofOpenIfNeededOnServerStart () at /home/ubuntu/devl/valkey/src/aof.c:718
#9  0x000055555573850a in main (argc=2, argv=0x7fffffffe1f8) at /home/ubuntu/devl/valkey/src/server.c:7171
(gdb) c
Continuing.
319760:M 23 May 2025 19:39:19.741 * Creating AOF base file appendonly.aof.1.base.rdb on server start
319760:M 23 May 2025 19:39:19.748 * Creating AOF incr file appendonly.aof.1.incr.aof on server start
319760:M 23 May 2025 19:39:19.748 * Ready to accept connections tcp
```

My Valkey conf (basically a modified copy of `tests/assests/default.conf`

```ini
# Server configuration for testing.

always-show-logo yes
notify-keyspace-events KEA
daemonize no
pidfile /var/run/valkey.pid
port 7000
timeout 0
bind 127.0.0.1
loglevel verbose
logfile ''
databases 16
latency-monitor-threshold 1
repl-diskless-sync-delay 0
loadmodule /home/ubuntu/~devl/valkey-search/.build-debug/libsearch.so

# Turn off RDB by default (to speedup tests)
# Note the infrastructure in server.tcl uses a dict, we can't provide several save directives
save ''

rdbcompression yes
dbfilename dump.rdb
dir ./

slave-serve-stale-data yes
appendonly yes
appendfsync everysec
no-appendfsync-on-rewrite no
activerehashing yes
auto-aof-rewrite-percentage 0

enable-protected-configs yes
enable-debug-command yes
enable-module-command yes

propagation-error-behavior panic

# Make sure shutdown doesn't fail if there's an initial AOFRW
shutdown-on-sigterm force

enable-debug-assert yes

hide-user-data-from-log no
```

## How to reproduce:

- `rm -fr appendonlydir/`
- `valkey-server my.conf` (see content above)
- `valkey-cli -p 7000 incr a`

Crash...

Catching the crash under GDB will not reveal much since its a symptom of the what I wrote above.

## How to fix

Release the context

Related Valkey issue: https://github.com/valkey-io/valkey/issues/2125

murphyjacob4:
Due to https://github.com/valkey-io/valkey/issues/2125 the IO context will not be freed which can trigger later assertions. We don't rely on it during saving so instead we now just use the background context for things like logging.

Fixes #142

### DIFF
--- a/src/coordinator/metadata_manager.cc
+++ b/src/coordinator/metadata_manager.cc
@@ -520,13 +520,13 @@ absl::Status MetadataManager::SaveMetadata(RedisModuleCtx *ctx, SafeRDB *rdb,
   if (!DoesGlobalMetadataContainEntry(metadata_.Get())) {
     // Auxsave2 will ensure nothing is written to the aux section if we write
     // nothing.
-    VMSDK_LOG(NOTICE, detached_ctx_.get())
+    VMSDK_LOG(NOTICE, ctx)
         << "Skipping aux metadata for MetadataManager since there is no "
            "content";
     return absl::OkStatus();
   }
 
-  VMSDK_LOG(NOTICE, detached_ctx_.get())
+  VMSDK_LOG(NOTICE, ctx)
       << "Saving aux metadata for MetadataManager to aux RDB";
   data_model::RDBSection section;
   std::string serialized_metadata;
--- a/src/rdb_serialization.cc
+++ b/src/rdb_serialization.cc
@@ -40,6 +40,7 @@
 #include "absl/strings/string_view.h"
 #include "src/metrics.h"
 #include "src/rdb_section.pb.h"
+#include "src/valkey_search.h"
 #include "vmsdk/src/log.h"
 #include "vmsdk/src/managed_pointers.h"
 #include "vmsdk/src/status/status_macros.h"
@@ -282,8 +283,8 @@ absl::Status PerformRDBSave(RedisModuleCtx *ctx, SafeRDB *rdb, int when) {
 }
 
 void AuxSaveCallback(RedisModuleIO *rdb, int when) {
-  auto ctx = RedisModule_GetContextFromIO(rdb);
   SafeRDB safe_rdb(rdb);
+  auto ctx = ValkeySearch::Instance().GetBackgroundCtx();
   auto result = PerformRDBSave(ctx, &safe_rdb, when);
   if (result.ok()) {
     Metrics::GetStats().rdb_save_success_cnt++;
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -521,13 +521,13 @@ absl::Status SchemaManager::SaveIndexes(RedisModuleCtx *ctx, SafeRDB *rdb,
   if (db_to_index_schemas_.empty()) {
     // Auxsave2 will ensure nothing is written to the aux section if we
     // write nothing.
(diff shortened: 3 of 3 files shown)


yairgott (changes requested):
I recall that we had some issues when using the detached ctx instead of the rdb context in the past. It was related to the externelizing of the vectors. Although this feature is disabled now, I think that it would be safer to consider the alternative which is to avoid creating the context if nothing to be saved.

murphyjacob4:
I think that was just on the load path? This context was not used on the save path other than for some logging AFAICT

yairgott (approved):

PR #144, opened 2025-05-26
Fix GetRedisLocalPort function to support TLS

yanamolo:
ATM, when constructing the coordinator service, the module uses the port obtained from `CONFIG GET PORT`. But when trying to get coordinator port to connect to other nodes, it uses the following code:
```c
if (RedisModule_GetClusterNodeInfo(ctx, sender_id_str.c_str(), node_ip,
                                     nullptr, &node_port,
                                     nullptr) != REDISMODULE_OK) {
```

RedisModule_GetClusterNodeInfo API calls moduleGetClusterNodeInfoForClient function. wich calculates the port as follows:
```c
int getNodeDefaultClientPort(clusterNode *n) {
    return server.tls_cluster ? n->tls_port : n->tcp_port;
}
```
When TLS is enabled, the current node will start coordinator on the calculated port X  while other nodes will calculate this node's port as Y. This happens because we have two methods of obtaining node's port - which return different results when TLS is enabled.
Changed `GetRedisLocalPort` function to call `RedisModule_GetClusterNodeInfo` as well, in order to use correct port in case tls is enabled/disabled.


src/valkey_search.cc
```diff
@@ -458,19 +458,21 @@ void ValkeySearch::OnForkChildCallback(RedisModuleCtx *ctx,
 }
 
 absl::StatusOr<int> GetRedisLocalPort(RedisModuleCtx *ctx) {
-  auto reply = vmsdk::UniquePtrRedisCallReply(
-      RedisModule_Call(ctx, "CONFIG", "cc", "GET", "port"));
-  if (reply == nullptr) {
-    return absl::InternalError("Failed to get port configuration");
+  int port = -1;
+  auto node_id = RedisModule_GetMyClusterID();
```
yairgott: `RedisModule_GetMyClusterID` returns NULL if the cluster is disabled. Today `GetRedisLocalPort` is called if `use_coordinator` is set, regardless if the cluster mode is on. 
@murphyjacob4 , is this the intended behavior? is the intension to allow running the coordinator on standalone HA to leverage the replica for search?

murphyjacob4: Right, we allow coordinator even in standalone, but I think ultimately we should align to just having "cluster" and "non-clustered" mode to avoid maintaining unused deployments ("cluster without coordinator" and "non-clustered with coordinator")

murphyjacob4: I think we had a similar conversation in https://github.com/valkey-io/valkey-search/pull/129

yairgott: SG! In that case, would retiring `use-coordinator` be a prerequisite for this PR? This change would implicitly enforce cluster mode when `use-coordinator` is enabled, and we want to avoid breaking backward compatibility, especially since the plan is to include it in 1.0.1.

@yanamolo, @murphyjacob4 , what do you think about falling back to the previous `GetRedisLocalPort `logic if `RedisModule_GetMyClusterID` returns NULL, instead of returning an error?

murphyjacob4: Alternatively, we can take the logic from the engine and duplicate it in our code:

Pseudocode
```
if CONFIG GET tls-cluster == yes
    return CONFIG GET tls-port
else
    return CONFIG GET port
```

yairgott:
@yanamolo , this PR is essential for [1.0.1](https://github.com/orgs/valkey-io/projects/24) release. I haven’t seen much recent activity on it and wanted to check if you’re planning to drive it to completion.

yanamolo:
@yairgott Yes, will be completed in the following days.

src/valkey_search.cc
```diff
@@ -413,20 +413,34 @@ void ValkeySearch::OnForkChildCallback(RedisModuleCtx *ctx,
...
-  if (!absl::SimpleAtoi(port_str, &port)) {
+  RedisModuleCallReply *config_reply =
+    RedisModule_CallReplyArrayElement(reply.get(), 1);
+  return RedisModule_CallReplyStringPtr(config_reply, nullptr);
+}
+
+
+absl::StatusOr<int> GetRedisLocalPort(RedisModuleCtx *ctx) {
+  int port = -1;
+  VMSDK_ASSIGN_OR_RETURN(auto tls_port_str, GetConfigGetReply(ctx, "tls-port"));
```
yairgott, line 434: My understanding is that we would like to get the confit `port` if `tls-port` is unavailable. However, the code returns an error in this case.

yanamolo: The config `tls-port` should always be available. If TLS is not set, it will return 0 - which in that case we fallback to `port`. Not able to get the config is unexpected behaviour, which throws an error

yairgott: Thanks for clarifying!

yairgott (approved):

src/valkey_search.cc
```diff
@@ -413,20 +413,34 @@ void ValkeySearch::OnForkChildCallback(RedisModuleCtx *ctx,
...
+      absl::StrFormat("Failed to get config: %s", config));
   }
-  RedisModuleCallReply *port_reply =
-      RedisModule_CallReplyArrayElement(reply.get(), 1);
-  const char *port_str = RedisModule_CallReplyStringPtr(port_reply, nullptr);
-  int port;
-  if (!absl::SimpleAtoi(port_str, &port)) {
+  RedisModuleCallReply *config_reply =
+    RedisModule_CallReplyArrayElement(reply.get(), 1);
+  return RedisModule_CallReplyStringPtr(config_reply, nullptr);
```
yairgott: We can't assume that the returned string is null terminated. `RedisModule_CallReplyStringPtr` should be called with a length pointer.

yairgott (changes requested):

src/valkey_search.cc
yanamolo: Fixed and uploaded.

yairgott (approved):

eifrah-aws (approved):

Issue #38, opened 2025-02-06
Fine-grained memory usage statistics


allenss-amazon:
Having per-index and per-field memory consumption numbers (visible in FT.INFO) would be a big help to both users and developers.

I'd propose a 'scoped" allocation object that would track the per-thread delta in consumed memory and update a stat. Doing it with a scoped object makes the code simple and clearn.

yairgott:
Good rampup task!

Sharing some context about how such a solution may look like:
1. The existing allocation/free override will be extended  to track memory allocation and free based on this allocation scope object.
2. The allocation scope object has to be store as a TLS variable to be multi threaded compatible.

ashu3103:
@allenss-amazon, can I take this one?

allenss-amazon:
Go for it!

ashu3103:
Hi @yairgott, @allenss-amazon,

Thanks for the opportunity, I just need to ask is there any documentation or tutorial about the architecture of this project to get familiar with the repo?

PR #153, opened 2025-05-29
feat: memory scope & memory usage for string interning store

proost:
resolve: https://github.com/valkey-io/valkey-search/issues/38

tracking memory allocation per index using per-thread stats.


src/index_schema.cc
```diff
@@ -184,6 +185,9 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::Create(
   auto res = std::shared_ptr<IndexSchema>(
       new IndexSchema(ctx, index_schema_proto, std::move(attribute_data_type),
                       mutations_thread_pool));
+  
+  MemoryTrackingScope scope(res->GetMemoryStats());
```
eifrah-aws: nit: prefer `{..}` construction style

vmsdk/src/memory_stats.h
```diff
@@ -0,0 +1,66 @@
...
+public:
+    MemoryStats() : allocated_bytes_(0) {}
+
+    MemoryStats(const MemoryStats&) = delete;
+    MemoryStats& operator=(const MemoryStats&) = delete;
+    MemoryStats(MemoryStats&&) = delete;
+    MemoryStats& operator=(MemoryStats&&) = delete;
+
+    void RecordAllocation(size_t size) {
+        allocated_bytes_.fetch_add(size, std::memory_order_relaxed);
```
yairgott: `allocated_bytes_` is currently an atomic variable, which implies support for a multi-threaded environment. However, since it uses a relaxed memory order, I’m concerned that concurrent access from multiple threads may result in `GetAllocatedBytes` returning an imprecise value.

That said, I’d prefer to avoid the overhead of stricter memory ordering.

At this point, I don’t see a strong justification for `MemoryStats` to support concurrent access. In the rare cases where concurrency is required, the caller can handle synchronization externally (e.g., with a mutex).

I propose changing `allocated_bytes_` to a non-atomic variable and clearly documenting that `MemoryStats` is not thread-safe.

vmsdk/src/memory_tracker.cc
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+#include "memory_tracker.h"
+#include "memory_stats.h"
+
+thread_local MemoryTrackingScope* MemoryTrackingScope::current_scope_tls_ = nullptr;
+
+thread_local MemoryTrackingScope::ScopeEventCallback MemoryTrackingScope::scope_event_callback_ = nullptr;
+
+MemoryTrackingScope::MemoryTrackingScope(MemoryStats* index_stats)
```
yairgott: nit: `stats` rather than `index_stats`

testing/ft_info_test.cc
```diff
@@ -88,6 +89,13 @@ TEST_P(FTInfoTest, FTInfoTests) {
         VMSDK_EXPECT_OK(SchemaManager::Instance().CreateIndexSchema(
             &fake_ctx_, index_schema_proto));
         EXPECT_EQ(SchemaManager::Instance().GetNumberOfIndexSchemas(), 1);
+        if (test_case.memory_bytes.has_value()) {
+          auto schema_or = SchemaManager::Instance().GetIndexSchema(
```
yairgott: nit: the best practice is to avoid using `_or` as part of a variable name.

yairgott (changes requested):
Thanks for driving this! A few comments inline. PTAL

src/index_schema.cc
```diff
@@ -398,6 +402,8 @@ bool IndexSchema::IsTrackedByAnyIndex(const InternedStringPtr &key) const {
 void IndexSchema::SyncProcessMutation(RedisModuleCtx *ctx,
                                       MutatedAttributes &mutated_attributes,
                                       const InternedStringPtr &key) {
+  MemoryTrackingScope scope(&memory_stats_);
```
eifrah-aws: nit: prefer `{..}` construction style

src/index_schema.cc
```diff
@@ -599,6 +605,7 @@ void IndexSchema::BackfillScanCallback(RedisModuleCtx *ctx,
 
 uint32_t IndexSchema::PerformBackfill(RedisModuleCtx *ctx,
                                       uint32_t batch_size) {
+  MemoryTrackingScope scope(&memory_stats_);
```
eifrah-aws: nit: prefer `{..}` construction style

src/index_schema.cc
```diff
@@ -743,6 +750,8 @@ void IndexSchema::RespondWithInfo(RedisModuleCtx *ctx) const {
                .c_str());
   RedisModule_ReplyWithSimpleString(ctx, "state");
   RedisModule_ReplyWithSimpleString(ctx, GetStateForInfo().data());
+  RedisModule_ReplyWithSimpleString(ctx, "index_size_mb");
+  RedisModule_ReplyWithLongLong(ctx, memory_stats_.GetAllocatedBytes() / 1024 / 1024);
```
eifrah-aws: All other elements of this array are of type "string", except for this on one

src/index_schema.cc
```diff
@@ -847,6 +856,8 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::LoadFromRDB(
       IndexSchema::Create(ctx, *index_schema_proto, mutations_thread_pool,
                           /*skip_attributes=*/true));
 
+  MemoryTrackingScope scope(index_schema->GetMemoryStats());
```
eifrah-aws: nit: prefer `{..}` construction style

src/index_schema.h
```diff
@@ -149,6 +150,9 @@ class IndexSchema : public KeyspaceEventSubscription,
   virtual void OnLoadingEnded(RedisModuleCtx *ctx);
 
   inline const Stats &GetStats() const { return stats_; }
+  inline MemoryStats *GetMemoryStats() { return &memory_stats_; }
```
eifrah-aws: Nit:if possible, use `&` instead of `*`

```c++
  inline MemoryStats& GetMemoryStats() { return memory_stats_; }
  inline const MemoryStats& GetMemoryStats() const { return memory_stats_; }
```

vmsdk/src/memory_tracker.h
```diff
@@ -0,0 +1,67 @@
...
+
+class MemoryStats;
+
+class MemoryTrackingScope {
+public:
+    // Test hook function type for scope lifecycle events
+    using ScopeEventCallback = std::function<void(const MemoryTrackingScope*)>;
+    
+    explicit MemoryTrackingScope(MemoryStats* index_stats);
+    ~MemoryTrackingScope();
```
eifrah-aws: If the destructor is not `virtual`, then make this class `final`

vmsdk/src/memory_tracker.cc
```diff
@@ -0,0 +1,67 @@
...
+ * INTERRUPTION) HOWEVER CAUSED ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "memory_tracker.h"
+#include "memory_stats.h"
+
+thread_local MemoryTrackingScope* MemoryTrackingScope::current_scope_tls_ = nullptr;
```
eifrah-aws: Out of curiosity, why not using `std::stack` instead of this logic?

vmsdk/src/memory_tracker.h
```diff
@@ -0,0 +1,67 @@
...
+
+    MemoryTrackingScope(const MemoryTrackingScope&) = delete;
```
eifrah-aws: IMO, we should add a macro `VMSDK_NO_COPYABLE(ClassName)` that expands into something like:

```c++
#define VMSDK_NON_COPYABLE(ClassName)              \
  ClassName(const ClassName&) = delete;            \
  ClassName& operator=(const ClassName&) = delete;
```

vmsdk/src/memory_tracker.h
```diff
@@ -0,0 +1,67 @@
...
+    MemoryTrackingScope& operator=(const MemoryTrackingScope&) = delete;
+    MemoryTrackingScope(MemoryTrackingScope&&) = delete;
+    MemoryTrackingScope& operator=(MemoryTrackingScope&&) = delete;
+
+    static MemoryTrackingScope* GetCurrentScope();
+    MemoryStats* GetStats() const;
+    
+    // Used for testing
+    static void SetScopeEventCallback(ScopeEventCallback callback);
```
eifrah-aws: Better add `FRIEND_TEST` instead and make these private (or remove them)

vmsdk/src/memory_tracker.h
```diff
@@ -0,0 +1,67 @@
...
+    static void ClearScopeEventCallback();
+
+private:
+    MemoryStats* target_stats_;
```
eifrah-aws: Initialize pointers with `nullptr`

vmsdk/src/memory_stats.h
proost: Ok, I understood.
[0364016](https://github.com/valkey-io/valkey-search/pull/153/commits/03640166dd20914b837550e56bede1a66a38742d)

testing/ft_info_test.cc
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

vmsdk/src/memory_tracker.cc
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

src/index_schema.cc
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

src/index_schema.cc
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

src/index_schema.cc
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

src/index_schema.cc
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

src/index_schema.h
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

vmsdk/src/memory_tracker.h
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

vmsdk/src/memory_tracker.cc
proost: It is more clear to me.

vmsdk/src/memory_tracker.h
proost: [75bcace](https://github.com/valkey-io/valkey-search/pull/153/commits/75bcace16fc5552a76a349a9e1dd8b3094153717)

vmsdk/src/memory_tracker.h
proost: We can't. because namespace is different.

vmsdk/src/memory_tracker.h
proost: [fefd118](https://github.com/valkey-io/valkey-search/pull/153/commits/fefd118d86be0a95e2c578c9fcc92fd3a95f386d)

allenss-amazon (commented):
IMO, this is more complicated than needed. IMO, the following scheme is much simpler.

1. Create a single thread-local variable called memory_delta. It's an int64_t (signed number is intentional).
2. The memory allocate and free functions increment/decrement memory_delta as appropriate.
3. The Scoped Memory object's constructor takes a single parameter which is the reference to the "Pool". The constructor also makes a private copy of memory_delta.
4. The Scoped Memory destructor reads the current value of memory_delta and subtracts it from the saved value of memory_delta and updates the "Pool" by that value.
5. BTW, I think the "Pool" object should be an atomic not something protected by a mutex.

Boom. done.

proost:
@allenss-amazon 
Yes, you are right. how about [b554481](https://github.com/valkey-io/valkey-search/pull/153/commits/b5544819a5b38e5246a657227318f573122aa5a9)?

vmsdk/src/memory_tracker.h
```diff
@@ -0,0 +1,64 @@
...
+public:
+    // Test hook function type for scope lifecycle events
+    using ScopeEventCallback = std::function<void(const MemoryTrackingScope*)>;
+    
+    explicit MemoryTrackingScope(std::atomic<int64_t>* pool);
+    ~MemoryTrackingScope();
+
+    VMSDK_NON_COPYABLE_NON_MOVABLE(MemoryTrackingScope);
+
+    static MemoryTrackingScope* GetCurrentScope();
```
allenss-amazon: The callback scope thing is only used for testing, right? It would be nice if we could figure out a way to test this code without inserting code into a performance critical path that's only used for testing.

vmsdk/testing/memory_allocation_test.cc
```diff
@@ -425,6 +427,192 @@ TEST_F(MemoryAllocationTest, VallocOverride) {
...
+  void* ptr1 = nullptr;
+  void* ptr2 = nullptr;
+
+  {
+    MemoryTrackingScope scope(&memory_pool);
+    EXPECT_CALL(*kMockRedisModule, Alloc(100)).WillOnce(testing::Return(reinterpret_cast<void*>(0x1000)));
+    EXPECT_CALL(*kMockRedisModule, MallocUsableSize(reinterpret_cast<void*>(0x1000))).WillRepeatedly(testing::Return(100));
+    ptr1 = __wrap_malloc(100);
+    EXPECT_EQ(memory_pool.load(), 0);
+    EXPECT_EQ(vmsdk::GetUsedMemoryCnt(), 100);
```
allenss-amazon: wouldn't it make sense to have testing for memory_delta here?

testing/index_schema_test.cc
```diff
@@ -860,6 +861,45 @@ TEST_F(IndexSchemaBackfillTest, PerformBackfill_SwapDB) {
...
+      .WillRepeatedly(Return(5));
+
+  auto index_schema = MockIndexSchema::Create(
+                          &parent_ctx, index_schema_name_str, key_prefixes,
+                          std::make_unique<HashAttributeDataType>(),
+                          nullptr)  // No thread pool for simplicity
+                          .value();
+
+  constructor_call_count = 0;
+
```
allenss-amazon: I don't think that showing that the appropriate constructor was called in one path is an adequate test. (Which is why I suspect that the Scoping callback isn't really needed).

I believe what these tests need to show is that after an index is created memory usage for the index is something and that upon destruction it's exactly back to zero usage.

testing/index_schema_test.cc
```diff
@@ -1143,6 +1183,43 @@ TEST_F(IndexSchemaRDBTest, LoadEndedDeletesOrphanedKeys) {
...
+      .WillRepeatedly(Return(&scan_ctx));
+  EXPECT_CALL(*kMockRedisModule, SelectDb(&scan_ctx, 0))
+      .WillRepeatedly(Return(REDISMODULE_OK));
+
+  FakeSafeRDB fake_rdb;
+
+  EXPECT_EQ(constructor_call_count, 0);
+
+  auto result = IndexSchema::LoadFromRDB(&fake_ctx, nullptr, 
+                                        std::make_unique<data_model::IndexSchema>(index_schema_proto),
```
allenss-amazon: again, just showing that the memory accounting was called isn't sufficient testing.

allenss-amazon (commented):
I think the memory accounting framework is starting to look good. But the integration into the remainder of the codebase is likely insufficient. For example, the string-interning facility allocates and frees memory independent of indexes. The current code doesn't recognize this. I suspect that the intern-ed strings get charged against the first index that uses them, but when they are destroyed I don't know who gets credit??

I don't fully understand the semantics of nesting the tracking objects as the test is nesting them against the same Pool accumulator -- which is probably a case that occurs, but it will also be the case that a nested tracking object will have a different Pool accumulator (for example, the string-interning case). I think this case needs to testing also.

I think we need to thought around how to ensure that the accounting gets properly done for all indexes. Worst case, we could check the per-index accounting at the end of the drop and detect whether it's zero or not.




The testing of this mechanism isn't adequate.

allenss-amazon (changes requested):
I see the attempt to track interned keys/strings and associate them back with their consuming indexes. I'm skeptical of the value of that and concerned about the overhead. IMO, having a separate system-wide intern-string pool as just another memory allocation domain -- separate from the indexes is fine. They can just be reported separately. It will be a bit confusing to users, but keys are likely the smaller part of the memory being consumed.

src/index_schema.h
```diff
@@ -149,6 +149,14 @@ class IndexSchema : public KeyspaceEventSubscription,
   virtual void OnLoadingEnded(RedisModuleCtx *ctx);
 
   inline const Stats &GetStats() const { return stats_; }
+  inline std::atomic<int64_t> &GetMemoryPool() { return memory_pool_; }
+  inline const std::atomic<int64_t> &GetMemoryPool() const { return memory_pool_; }
+
+  // NOTE: Net memory usage can be negative in temporary.
```
allenss-amazon: How can it get negative?

vmsdk/src/memory_tracker.cc
allenss-amazon: I don't understand why a stack is need at all. Under the assumption that this class is only every instantiated on a thread's run-time stack then they implicitly already nest and you don't need to explicitly track that.

I think there needs to be two types of scopes. They differ in how they nest. Let's call them the outer and inner scope, i.e, the outer scope is in the stack before the inner scope is created and conversely the inner scope is destructed before the outer scope terminates.

The difference between the two types is how the delta accrued in the inner scope affects the outer scope.

In one case, you want the inner scope to represent a new, disjoint, memory domain from the outer scope. I'll call this one "NewScope"
In the other case, you want to create a sub-domain of the outer scope -- I'll call it "SubScope"

For both scopes, the constructor reads the current memory delta and saves it in the object, I'll call this variable "start_memory".

```
struct MemoryScope {
   MemoryScope(MemoryPool& pool) : start_memory(current_memory_delta) {}
   size_t start_memory;
};
```
For NewScope, the destructor computes the memory delta within it's scope by subtracting start_memory but it resets current_memory_delta to be the same value as in the constructor:
```
struct NewScope : public MemoryScope {
   ~NewScope() {
      auto used = current_memory_delta - start_memory; // compute delta within my scope
      Pool += used; // Update the pool
      current_memory_delta = start_memory; // Remove my memory consumption from any outer nested scope
   }
}
```

The NewScope is what you want for string-interning. Because there's no direct relationship between the memory consumption pattern of the outer scope and the inner scope.

The Subscope destructor just doesn't fiddle with current_memory_delta, hence it's nesting has no effect on the outer Scope object (of any type).

```
struct SubScope : public MemoryScope {
   ~SubScope() {
      auto used = current_memory_delta - start_memory; // compute delta within my scope
      Pool += used; // Update the pool
   }
}
```
This subscope is totally useful in many contexts.

1. Cheaply maintaining a hierarchical memory consumption statistics. For example, A field-type (say Numeric) could have a SubScope of the Index's memory, which could in turn have a SubScope of the overall IndexMemory consumption. This nested scope avoids the run-time problem of adding up lots of indexes when all you want is the global index memory consumption.
2. Doing development work with fine-grained measurements of memory usage. You can declare a Subscope ANYWHERE in the codebase with full confidence you wont' disturb any existing measurement machinery. This is a powerful tool for space-oriented optimizations.

src/index_schema.h
proost: I figured out it can't

vmsdk/src/memory_tracker.cc
```diff
@@ -0,0 +1,68 @@
...
+    : MemoryScope(pool) {
+}
+
+IsolatedMemoryScope::~IsolatedMemoryScope() {
+    if (target_pool_ != nullptr) {
+        int64_t current_delta = vmsdk::GetMemoryDelta();
+        int64_t net_change = current_delta - baseline_memory_;
+        target_pool_->fetch_add(net_change);
+
+        vmsdk::SetMemoryDelta(baseline_memory_);
```
allenss-amazon: I think this needs to be outside of the check for nullptr, irght?

proost: https://github.com/valkey-io/valkey-search/pull/153/commits/1826095701726c67cbe5f759ba6ed50ff00c7575

Yes, you are right. fix test too.

vmsdk/testing/memory_allocation_test.cc
```diff
@@ -403,6 +404,243 @@ TEST_F(MemoryAllocationTest, VallocOverride) {
...
+  void* outer_ptr = nullptr;
+  void* inner_ptr = nullptr;
+
+  EXPECT_EQ(vmsdk::GetUsedMemoryCnt(), 0);
+  EXPECT_EQ(vmsdk::GetMemoryDelta(), 0);
+
+  {
+    IsolatedMemoryScope scope{&pool};
+    
+    EXPECT_CALL(*kMockRedisModule, Alloc(100))
```
allenss-amazon: To avoid ambiguity, let's have the usable size != requested the size.

vmsdk/testing/memory_allocation_test.cc
```diff
@@ -403,6 +404,243 @@ TEST_F(MemoryAllocationTest, VallocOverride) {
...
+    outer_ptr = __wrap_malloc(100);
+
+    EXPECT_EQ(vmsdk::GetUsedMemoryCnt(), 100);
+    EXPECT_EQ(pool.load(), 0);
+    EXPECT_EQ(vmsdk::GetMemoryDelta(), 100);
+
+    {
+      IsolatedMemoryScope scope{&pool};
+      
+      EXPECT_CALL(*kMockRedisModule, Alloc(75))
```
allenss-amazon: again usable_size != requested_size

vmsdk/src/memory_tracker.cc
```diff
@@ -0,0 +1,68 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "memory_tracker.h"
+#include "vmsdk/src/memory_allocation.h"
+
+thread_local MemoryScope* MemoryScope::current_scope_ = nullptr;
+
+MemoryScope::MemoryScope(std::atomic<int64_t>* pool)
```
allenss-amazon: When would there be a scope that didn't have a pool associated with it? I think the pool could be a reference and we could eliminate the tests for nullptr.

allenss-amazon (commented):
The MemoryScope class is excellent and on it's own ready to merge.

IMO, The usage of that in the class in the code base is not ready for merging. At a minimum, there's not enough testing to ensure correctness. I also believe that the places in the code where you've chosen to tap into will make it very difficult to ensure correctness. 

I'd recommend that we split this into two PRs. 

In the first PR, is the MemoryPool object and it's testing.
Let's include in that first PR, implementation within the StringPoolIntern subsystem. As that's easy to test. We should be able to complete that PR and get it merged soon.

Let's move the index accounting into the second PR.

I'd like to see more detailed accounting within the index itself. I believe that the accounting for index-schemas should be more focused on the individual fields. I believe we should place the machinery into the IndexBase object, i.e., wrap the various virtual function calls with MemoryScope(s). This will provide per-field accounting, which is very helpful in many circumstances. Also, it should make the implementation easier to check for correctness.

proost:
@allenss-amazon
I agree that dividing PRs. As you suggested, I changed this PR for memory scope & string interning pool usage.

I proposed dividing 3 PRs.
1. memory scope & string interning usage.
2. Memory usage per index.
3. Memory usage for individual fields per index.

I'm using ANN vector search at work. If I knew memory usage per index and how many records are in the index, I can easily guess how metadata use memory per index. In our case, although the guessing isn't 100% accurate, It is a close value to the exact figure. 

I understand that someone's circumstances may be different from ours. But some cases are enough just providing memory usage per index. So how do you think dividing 3 PRS?

src/utils/string_interning.cc
```diff
@@ -30,6 +34,10 @@ InternedString::~InternedString() {
   if (is_shared_) {
     StringInternStore::Instance().Release(this);
   }
+
```
allenss-amazon, line 40: Why wouldn't this be before the "if (is_shared_)" section?

Not sure why Github won't let me comment on the ::Release procedure (just below), but it's calling "erase" wouldn't we want to have a memory scope around that too?

src/utils/string_interning.cc
```diff
@@ -66,6 +74,9 @@ std::shared_ptr<InternedString> StringInternStore::InternImpl(
       return locked;
     }
   }
+
+  IsolatedMemoryScope scope {memory_pool_};
```
allenss-amazon: While somewhat less efficient, I think we're trying to establish a standard style that's reliable. IMO, we should put these declarations at the start of the functions or in a wrapper function, someplace where they aren't susceptible to missing a memory allocation/free operation.

src/utils/string_interning.cc
```diff
@@ -13,9 +13,13 @@
 #include "absl/strings/string_view.h"
 #include "absl/synchronization/mutex.h"
 #include "src/utils/allocator.h"
+#include "vmsdk/src/memory_tracker.h"
+#include "vmsdk/src/memory_allocation_overrides.h"
 
 namespace valkey_search {
 
+std::atomic<int64_t> StringInternStore::memory_pool_{0};
```
allenss-amazon: I'd like to see a class for the memory pool allocation, not an scalar variable. It's just as efficient, but provides superior compile-time checking and documentation.

src/valkey_search.cc
```diff
@@ -315,7 +315,11 @@ void ValkeySearch::Info(ValkeyModuleInfoCtx *ctx, bool for_crash_report) const {
   if (!for_crash_report) {
     vmsdk::info_field::DoSection(ctx, "string_interning", for_crash_report);
     ValkeyModule_InfoAddFieldLongLong(ctx, "string_interning_store_size",
-                                      StringInternStore::Instance().Size());
+                                     StringInternStore::Instance().Size());
+    char buffer[100];
+    vmsdk::DisplayAsSIBytes(StringInternStore::GetMemoryUsage(), buffer, sizeof(buffer));
+    ValkeyModule_InfoAddFieldCString(
+        ctx, "string_interning_memory_mb", buffer);
```
allenss-amazon: The new info_field machinery would handle this for you, please use it.

src/utils/string_interning.cc
proost: You are right. I already inserted with key so should consider this.
[23c8fac](https://github.com/valkey-io/valkey-search/pull/153/commits/23c8fac6e0c60502774affac22761517e855ba8f)

src/utils/string_interning.cc
proost: OK, I define class and using it. [23c8fac](https://github.com/valkey-io/valkey-search/pull/153/commits/23c8fac6e0c60502774affac22761517e855ba8f)

vmsdk/src/memory_tracker.cc
```diff
@@ -0,0 +1,64 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "memory_tracker.h"
+#include "vmsdk/src/memory_allocation.h"
+
+thread_local MemoryScope* MemoryScope::current_scope_ = nullptr;
```
allenss-amazon: this appears to be unused, maybe it's a vestige of earlier implementations? In so, then let's remove it.

vmsdk/src/memory_tracker.h
```diff
@@ -0,0 +1,91 @@
...
+
+protected:
+    MemoryPool& target_pool_;
+    int64_t baseline_memory_ = 0;
+
+private:
+    static thread_local MemoryScope* current_scope_;
+};
+
+class IsolatedMemoryScope final : public MemoryScope {
```
allenss-amazon, line 81: A comment about the use-case and the differences between Isolated.... and Nested.... would be very helpful.

vmsdk/testing/memory_allocation_test.cc
```diff
@@ -403,6 +405,173 @@ TEST_F(MemoryAllocationTest, VallocOverride) {
   EXPECT_EQ(__wrap_valloc(size), test_ptr);
   __wrap_free(test_ptr);
 }
+
+TEST_F(MemoryAllocationTest, IsolatedMemoryScopeAllocationIsolation) {
```
allenss-amazon, line 409: Based on the name, I would have expected there to be two different pools, one for the outer and inner scopes, respectively.

vmsdk/testing/memory_allocation_test.cc
```diff
@@ -403,6 +405,173 @@ TEST_F(MemoryAllocationTest, VallocOverride) {
...
+  EXPECT_EQ(vmsdk::GetUsedMemoryCnt(), 0);
+  EXPECT_EQ(vmsdk::GetMemoryDelta(), 0);
+
+  {
+    IsolatedMemoryScope scope{&pool};
+    
+    EXPECT_CALL(*kMockRedisModule, Alloc(112))
+        .WillOnce(testing::Return(reinterpret_cast<void*>(0x1000)));
+    EXPECT_CALL(*kMockRedisModule, MallocUsableSize(reinterpret_cast<void*>(0x1000)))
+        .WillRepeatedly(testing::Return(112));
```
allenss-amazon: I would prefer to see that the MallocUsableSize != Allocation size in order to ensure that're actually using the right value in MemoryDelta. Having those two be equal in size fails to test that.

vmsdk/testing/memory_allocation_test.cc
```diff
@@ -403,6 +405,173 @@ TEST_F(MemoryAllocationTest, VallocOverride) {
...
+  __wrap_free(inner_ptr);
+}
+
+TEST_F(MemoryAllocationTest, IsolatedMemoryScopeFreeIsolation) {
+  vmsdk::UseValkeyAlloc();
+
+  EXPECT_EQ(vmsdk::GetUsedMemoryCnt(), 0);
+  EXPECT_EQ(vmsdk::GetMemoryDelta(), 0);
+  
+  MemoryPool pool{0};
```
allenss-amazon: Again, having multiple pools, one for each scope would make more sense from a testing perspective, right?

allenss-amazon (commented):
I think we're very close now. I think there's some "old" code that is still hanging around from the original design that needs to be removed. Also, there's some small issues with the unit tests that I think are easily fixed. Once those are addressed, I think this can be merged.

vmsdk/src/memory_tracker.cc
proost: [18d0707](https://github.com/valkey-io/valkey-search/pull/153/commits/18d0707ca5529a62567dbf629d51799e7f141329)

vmsdk/testing/memory_allocation_test.cc
proost: [36a275c](https://github.com/valkey-io/valkey-search/pull/153/commits/36a275cbfd8fc8e57bf2bc7b89f91eaa47ad277c)

vmsdk/testing/memory_allocation_test.cc
proost: [36a275c](https://github.com/valkey-io/valkey-search/pull/153/commits/36a275cbfd8fc8e57bf2bc7b89f91eaa47ad277c)

vmsdk/testing/memory_allocation_test.cc
proost: [36a275c](https://github.com/valkey-io/valkey-search/pull/153/commits/36a275cbfd8fc8e57bf2bc7b89f91eaa47ad277c)

src/utils/string_interning.h
```diff
@@ -34,19 +36,30 @@ class StringInternStore {
...
+
   StringInternStore() = default;
   std::shared_ptr<InternedString> InternImpl(absl::string_view str,
                                              Allocator *allocator);
   void Release(InternedString *str);
   absl::flat_hash_map<absl::string_view, std::weak_ptr<InternedString>>
       str_to_interned_ ABSL_GUARDED_BY(mutex_);
   mutable absl::Mutex mutex_;
+
+  static void SetMemoryUsage(int64_t value) {
```
allenss-amazon, line 58: This is only used for testing, right? Let's add a comment indicating that.

testing/utils/string_interning_test.cc
```diff
@@ -71,12 +112,31 @@ TEST_P(StringInterningTest, WithAllocator) {
   }
 }
 
+TEST_F(StringInterningTest, MemoryTrackingIsolation) {
+  MemoryPool caller_pool{0};
+  std::shared_ptr<InternedString> interned_str;
+
+  NestedMemoryScope scope{caller_pool};
```
allenss-amazon: this line isn't needed, right? I found it confusing.

testing/utils/string_interning_test.cc
```diff
@@ -71,12 +112,31 @@ TEST_P(StringInterningTest, WithAllocator) {
   }
 }
 
+TEST_F(StringInterningTest, MemoryTrackingIsolation) {
```
allenss-amazon: I see that this test shows that the StringIntern class has a GetMemoryUsage function that yields zero and non-zero result.   I don't think it proves anything about MemoryTrackingIsolation, right? Let's rename it.

src/valkey_search.cc
```diff
@@ -753,6 +753,14 @@ static vmsdk::info_field::Integer &remove_subscription_skipped_count =
...
 
 #endif
+
+static vmsdk::info_field::Integer string_interning_memory("string_interning", "string_interning_memory",
+  vmsdk::info_field::IntegerBuilder()
+      .SIBytes()
+      .App()
+      .Computed(StringInternStore::GetMemoryUsage)
+      .CrashSafe());
+
```
allenss-amazon: Let's have two of these, one that's purely bytes and one that's human readable. here you have the human readable, which is definitely a nice-to-have.

allenss-amazon (commented):
I'm sorry it's taken me so long to review this, I had some personal issues. This is very close to being mergeable, I have a few nits, but one important issue -- IMO, the accounting shouldn't be for the requested memory size, but the actual allocated memory size. This will show true memory consumption, whereas using the requested size always has up to 30% uncertainty due to memory fragmentation.

src/valkey_search.cc
proost: [d9452c0](https://github.com/valkey-io/valkey-search/pull/153/commits/d9452c0adf2e16713955cf59bae48f2afdf7af02)

testing/utils/string_interning_test.cc
proost: [d9452c0](https://github.com/valkey-io/valkey-search/pull/153/commits/d9452c0adf2e16713955cf59bae48f2afdf7af02)

testing/utils/string_interning_test.cc
proost: [d9452c0](https://github.com/valkey-io/valkey-search/pull/153/commits/d9452c0adf2e16713955cf59bae48f2afdf7af02)

src/utils/string_interning.h
proost: [d9452c0](https://github.com/valkey-io/valkey-search/pull/153/commits/d9452c0adf2e16713955cf59bae48f2afdf7af02)

proost:
> IMO, the accounting shouldn't be for the requested memory size, but the actual allocated memory size. This will show true memory consumption, whereas using the requested size always has up to 30% uncertainty due to memory fragmentation.

[68adaf6](https://github.com/valkey-io/valkey-search/pull/153/commits/68adaf6860e2c41f3ac176d5dafb311edfb70a77) I'm not 100% sure about this commit. There is reason to use "ValkeyModule_MallocUsableSize", but I just found initial commit without reason. But following doc and your comment, let's change it.

Aksha1812:
Hey @proost. Thank you for working on this. Is this still being actively worked on? If this gets merged, it would be really helpful in tracking memory allocation while developing fulltext search. Let me know if I can help in anyway. If you are no longer working on this, can pick this up where you left off if that is fine with you.

proost:
@Aksha1812 
I'm working on this. I'm waiting for other reviews or to be merged.

vmsdk/src/memory_allocation_overrides.cc
```diff
@@ -175,7 +175,7 @@ void __wrap_free(void* ptr) noexcept {
     vmsdk::PerformAndTrackFree(ptr, __real_free, empty_usable_size);
   } else {
     vmsdk::PerformAndTrackFree(ptr, ValkeyModule_Free,
-                               ValkeyModule_MallocUsableSize);
+                               ValkeyModule_MallocSize);
```
allenss-amazon: Let's not change this. I believe this was changed at my request and I did some more research and believe that for our purposes MallocSize == MallocUsableSize (since we use an allocator that provides zmalloc_size)

proost: [5fc7588](https://github.com/valkey-io/valkey-search/pull/153/commits/5fc7588a269e6a83afcce136f37764889a3cb7a5)

allenss-amazon (approved):

proost:
@allenss-amazon 
I merged reverted PR which you push to main branch. But I can't run integration test build with ASan.

Issue #156, opened 2025-05-30
Potential memory leak in coordinator's `SearchIndexPartition` request handler


ChristophKoerner:
We run the following experiment on 2 shard 1 replica CME cluster (r6g.large)

1. Create an Hash index
2. Ingest Hash data (~1M keys)
3. Query the indexes on a single primary `0001-001` using below search query

```
FT.SEARCH <index> "*=>[KNN 3 @VEC $BLOB EF_RUNTIME 100" PARAMS 2 BLOB <rand vec> DIALECT 2
```

We observe, that all the memory on all nodes (including replicas) except `0001-001` increases during querying. The memory is not reclaimed after querying stops.

![Image](https://github.com/user-attachments/assets/7c601075-bdde-47af-9c22-112a51934cbc)

![Image](https://github.com/user-attachments/assets/45f34dc0-638e-4e72-84a0-dd9fb33b8310)

In a followup experiment, we verify that memory does not increase when the `RETURN 0` statement (same as `NOCONTENT`) is present in `FT.SEARCH` queries.
```
FT.SEARCH <index> "*=>[KNN 3 @VEC $BLOB EF_RUNTIME 100" RETURN 0 PARAMS 2 BLOB <rand vec> DIALECT 2
```

This indicates a potential memory leak in the search coordinators SearchIndexPartition request handler.
https://github.com/valkey-io/valkey-search/blob/e06cf7a931fd845e3a87e7320ae880bb82c5e577/src/coordinator/server.cc#L158-L174

### `free` OS memory recorded after the experiment

Primary `0001-001`
```
memory_total: 15.5 GB
memory_free: 9.4 GB
memory_used: 6.2 GB
swap_total: 15.5 GB
swap_free: 15.5 GB
swap_used: 0.0 B
total: 31.1 GB
total_free: 24.9 GB
total_used: 6.2 GB
memory_percentage: 39.7%
swap_percentage: 0.0%
```

Primary `0002-001`

```
memory_total: 15.5 GB
memory_free: 2.7 GB
memory_used: 12.9 GB
swap_total: 15.5 GB
swap_free: 15.5 GB
swap_used: 0.0 B
total: 31.1 GB
total_free: 18.2 GB
total_used: 12.9 GB
memory_percentage: 82.9%
swap_percentage: 0.0%
```

murphyjacob4:
Can we run INFO SEARCH? And just an INFO EVERYTHING on the node

[...]

PR #159, opened 2025-05-30
Use a managed pointer for thread safe context in coordinator

murphyjacob4:
As discussed in #156


murphyjacob4:
> Hmm, how can we protect against this kind of failure going forward?

I think running integration tests under ASAN should catch it

yairgott (approved):
Good catch!

PR #163, opened 2025-06-03
Search trigger for Valkey Extension Automation

Nikhil-Manglore:
This will be used to trigger the workflows in the Valkey-Extensions repository. A secret will also need to added to this repository called EXTENSION_PAT.


yairgott:
SG and thanks for the elaborated response!

yairgott (approved):

PR #165, opened 2025-06-03
Make release build as default for manual cmake.

Shivshankar-Reddy:
This PR does not modify any functional code. By default, the build.sh script generates a release build. However, when running CMake manually and building with cmake --build, it defaults to a debug build. Aligning the default CMake behavior with the script ensures consistency, so that manually generated builds also default to the release configuration.

Since debug builds use the -O0 optimization flag, they can significantly impact performance. Defaulting to a release build for manual builds would help ensure performance is more representative during testing.


yairgott:
hmm, we've opt-in to build by default a release build and I believe that this is also aligned with Valkey.

yairgott:
Closing this pull request but open for a discussion.

Issue #124, opened 2025-05-14
Indexes permanently stop ingesting data


ChristophKoerner:
Hi all,

I observed an interesting behavior upon reaching `maxmemory` where existing indexes stop ingesting new data, but newly created indexes continue to operate normally. While such behavior could be expected during `maxmemory` state, it persists for existing indexes even after space is freed again. I have a suspicion that existing indexes stop subscribing to keyspace notifications and don't resubscribe once memory is freed - but that's just a hunch based on my observation. Note that there is also no such indication on `FT.INFO` output that allows a client to verify this behavior.

Verify expected behavior:
* Create index `a`
* Create index `b`
* Insert element `i`
* => Both indexes `a` and `b` contain 1 element

```
> FT.CREATE a ON HASH PREFIX 1 hash: SCHEMA vec AS VEC VECTOR HNSW 6 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2
> FT.CREATE b ON HASH PREFIX 1 hash: SCHEMA vec AS VEC VECTOR HNSW 6 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2
> HSET hash:1 vec "\x00\x00\x00\x00\x00\x00\x00\x00"
> FT.INFO a
... 13) num_records
    14) "1"
> FT.INFO b
... 13) num_records
    14) "1"
```

Steps to reproduce the issue:
* Create index `a`
* Create indexes [c-z] until OOM
* Delete indexes [c-z]
* Create index `b`
* Insert element `i`
* => Observed that index `a` contains 0 elements, whereas index `b` contains 1 element
* Any new index also backfills existing elements correctly

```
> FT.CREATE a ON HASH PREFIX 1 hash: SCHEMA vec AS VEC VECTOR HNSW 6 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2
> [repeat FT.CREATE until OOM, FT.DROPINDEX except a ]
> FT._LIST
1) a
> FT.CREATE b ON HASH PREFIX 1 hash: SCHEMA vec AS VEC VECTOR HNSW 6 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2
> HSET hash:1 vec "\x00\x00\x00\x00\x00\x00\x00\x00"
> FT.INFO a
... 13) num_records
    14) "0"
> FT.INFO b
... 13) num_records
    14) "1"
```

Hope this description is sufficient to reproduce or root-cause the issue.

[...]

PR #166, opened 2025-06-04
Fix bug resulting in no ingestion to indexes with the same prefix as a dropped index

eifrah-aws:
In case multiple indexes are created for the same prefix the first index that will dropped - all of the subscribers associated with that prefix will be dropped.
Effectively making all these indexes "frozen"

This PR fixes issue: https://github.com/valkey-io/valkey-search/issues/124


murphyjacob4:
One thing: Since the PR title will be present in release notes, can we make the PR title more clearly reflect what bug is being fixed?

E.g. Fix bug resulting in no ingestion to indexes with the same prefix as a dropped index

yairgott (approved):

PR #177, opened 2025-06-11
Add coordinator port info metric

yanamolo:
The coordinator listening port is not clearly visible, with only a log message indicating this information that could be overlooked. 
This PR add an informational metric called `coordinator_server_listening_port` which is reported in the `search` section of the `INFO modules` command.
This is similar to what already exists for the engine port (with info metric `tcp_port`)


eifrah-aws:
LGTM

eifrah-aws (approved):

PR #178, opened 2025-06-11
The module can now be built on macOS

eifrah-aws:
The current state of the module is that it compile on Linux, but fails on macOS.

There are several reasons for this:

- `zlib` does not compile due to wrong detection of `fdopen` 
- Multiple defined symbols defined in header files (I addressed that by defining them as `__attribute__((weak))`) which is the closest thing to GCC's `__attribute__((common))`
- Some methods needed to be declared as `inline` to avoid ODR violatins
- Defined missing functions using macOS API:

```c++
#ifdef __APPLE__
#include <libkern/OSByteOrder.h>
#define htole64(x) OSSwapHostToLittleInt64(x)
#define le64toh(x) OSSwapLittleToHostInt64(x)
#endif
```

- Missing macOS linker flags: `--version-script`, `--start-group` and `--end-group`
- Update the `build.sh` script to be compatible with `bash-3` (which is the default on macOS)
- `std::numeric_limits::infinity()` does not work with `clang` (undefined behavior) when built with `-ffast-math` optimization. For now, when `clang` is used, we translate `-inf` / `+inf` into `std::numeric_limits::lowest()` & `std::numeric_limits::max()` respectively 
- Provide an explicit type conversion from `string_view` -> `unsigned long`
- Fixed `allocator_test`: on macOS, default page size is `16K`

This PR fixes issue: https://github.com/valkey-io/valkey-search/issues/179

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,17 @@ cmake_minimum_required(VERSION 3.16)
 
 # Project definition
 
-project(ValkeySearch VERSION 1.0.0 LANGUAGES C CXX)
+project(
+  ValkeySearch
+  VERSION 1.0.0
+  LANGUAGES C CXX)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+
+# This project can only compile macOS / Linux, we can use the UNIX predefined
+# variable to test this
+if(NOT UNIX)
+  message(FATAL_ERROR "valkey-search can only be built on Linux or macOS")
+endif()
 
 # Options
 option(BUILD_TESTS "Build valkey-search tests" ON)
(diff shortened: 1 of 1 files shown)


cmake/Modules/valkey_search.cmake
```diff
@@ -116,7 +116,9 @@ function(valkey_search_target_update_compile_flags TARGET)
   target_compile_options(${TARGET} PRIVATE -ffast-math)
   target_compile_options(${TARGET} PRIVATE -funroll-loops)
   target_compile_options(${TARGET} PRIVATE -ftree-vectorize)
-  target_compile_options(${TARGET} PRIVATE -fopenmp)
+  if(UNIX AND NOT APPLE)
```
yairgott, line 119: Should we just fail the build if it's not UNIX or APPLE?

eifrah-aws: Added a check in the main `CMake` file to abort the build early in the build stage

yairgott (approved):

PR #182, opened 2025-06-13
Limit FT.SEARCH query string parsing recursive depth by a module config

KarthikSubbarao:
Currently (without this PR), the query string is unbounded and since we are using recursive descent for parsing the query string's, it makes the valkey-search module susceptible to input related attacks.

This PR adds a limit for the `FT.SEARCH` command's query string parsing recursive depth by a new module config: `search.query-string-depth` which is set to 1000 by default. The new config is: "search.query-string-depth".

Example of Depth of 1:
```
127.0.0.1:6379> config set search.query-string-depth 1
OK
127.0.0.1:6379> FT.SEARCH my_index "@price:[10 20] =>[KNN 10 @doc_embedding $BLOB]" PARAMS 2 BLOB <your_vector_blob> RETURN 1 doc_embedding
1) (integer) 0
127.0.0.1:6379> FT.SEARCH my_index "@price:[10 20] | @category:{electronics|books} =>[KNN 10 @doc_embedding $BLOB]" PARAMS 2 BLOB <your_vector_blob> RETURN 1 doc_embedding
(error) Invalid filter expression: `@price:[10 20] | @category:{electronics|books}`. Query string recursive depth exceeded
```

Example of Depth of 10:
```
127.0.0.1:6379> config set search.query-string-depth 10
OK
127.0.0.1:6379> FT.SEARCH my_index "(((((((((@price:[10 20]))))))))) =>[KNN 10 @doc_embedding $BLOB]" PARAMS 2 BLOB <your_vector_blob> RETURN 1 doc_embedding
1) (integer) 0
127.0.0.1:6379> FT.SEARCH my_index "((((((((((@price:[10 20])))))))))) =>[KNN 10 @doc_embedding $BLOB]" PARAMS 2 BLOB <your_vector_blob> RETURN 1 doc_embedding
(error) Invalid filter expression: `((((((((((@price:[10 20]))))))))))`. Query string recursive depth exceeded
127.0.0.1:6379> FT.SEARCH my_index "((((((((@price:[10 20] | @category:{electronics|books})))))))) =>[KNN 10 @doc_embedding $BLOB]" PARAMS 2 BLOB <your_vector_blob> RETURN 1 doc_embedding
1) (integer) 0
```

Config ranges from 1 to 4294967295:
```
127.0.0.1:6379> config set search.query-string-depth 0
(error) ERR CONFIG SET failed (possibly related to argument 'search.query-string-depth') - argument must be between 1 and 4294967295 inclusive
```

Build and test
```
./build.sh
integration/run.sh
```


src/commands/filter_parser.cc
```diff
@@ -310,7 +311,11 @@ std::unique_ptr<query::Predicate> WrapPredicate(
 // 10. Numeric filters are inclusive. Exclusive min or max are expressed with (
 // prepended to the number, for example, [(100 (200].
 absl::StatusOr<std::unique_ptr<query::Predicate>>
-FilterParser::ParseExpression() {
+FilterParser::ParseExpression(int level) {
+  if (level++ == options::GetQueryStringDepth().GetValue()) {
```
allenss-amazon: \>= ?

src/commands/filter_parser.cc
```diff
@@ -310,7 +311,11 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+    return absl::InvalidArgumentError(
+        absl::StrCat("Query string recursive depth exceeded"));
```
allenss-amazon: StrCat for a constant string?

src/commands/filter_parser.h
```diff
@@ -61,7 +61,7 @@ class FilterParser {
   absl::flat_hash_set<std::string> filter_identifiers_;
 
   absl::StatusOr<bool> IsMatchAllExpression();
-  absl::StatusOr<std::unique_ptr<query::Predicate>> ParseExpression();
+  absl::StatusOr<std::unique_ptr<query::Predicate>> ParseExpression(int level = 0);
```
allenss-amazon: Defaults are dangerous.

src/valkey_search_options.cc
```diff
@@ -95,6 +98,15 @@ static const std::vector<int> kLogLevelValues = {
     static_cast<int>(LogLevel::kWarning), static_cast<int>(LogLevel::kNotice),
     static_cast<int>(LogLevel::kVerbose), static_cast<int>(LogLevel::kDebug)};
 
+/// Register the "--query-string-depth" flag. Controls the depth of the query
+/// string parsing from the FT.SEARCH cmd.
+static auto query_string_depth =
+    config::NumberBuilder(kQueryStringDepthConfig,   // name
+                          kDefaultQueryStringDepth,  // default size
+                          kMinimumQueryStringDepth,  // min size
+                          UINT_MAX)                  // max size
```
allenss-amazon, line 64: I really dislike this style of using static const things defined elsewhere that are used exactly once. All this does is require you to page back and forth in the code to understand what this does -- foolish cognitive load that obscures.. Let's push back against this and just put the strings and the constant values here (Brownie points for changing the other declarations similarly).

KarthikSubbarao: I Agree, I'll move the other config names & defaults down to where the config is actually created (both the new config and the existing configs)

src/commands/filter_parser.cc
KarthikSubbarao: I had also included the config value here in the message and was using the Cat for that. Will clean this up

src/valkey_search_options.cc
```diff
@@ -38,15 +38,6 @@ namespace options {
 
 constexpr uint32_t kHNSWDefaultBlockSize{10240};
 constexpr uint32_t kHNSWMinimumBlockSize{0};
```
KarthikSubbarao, line 18: These are needed in the namespace starting at line 42

src/valkey_search_options.cc
```diff
@@ -161,6 +172,10 @@ static auto log_level =
         .WithValidationCallback(ValidateLogLevel)
         .Build();
 
+vmsdk::config::Number& GetQueryStringDepth() {
+  return dynamic_cast<vmsdk::config::Number&>(*query_string_depth);
```
allenss-amazon: Why is a cast needed here? Also, why not just return the value itself, not the Number struct?

allenss-amazon (commented):
one nit and we're ready to merge.

src/valkey_search_options.cc
KarthikSubbarao: I was following the pattern from the configs below.

But it is not needed as you mentioned. I updated function to return the value itself

src/valkey_search_options.cc
```diff
@@ -38,15 +38,6 @@ namespace options {
...
 constexpr uint32_t kHNSWMinimumBlockSize{0};
-constexpr uint32_t kMaxThreadsCount{1024};
-
-constexpr absl::string_view kHNSWBlockSizeConfig{"hnsw-block-size"};
-constexpr absl::string_view kReaderThreadsConfig{"reader-threads"};
-constexpr absl::string_view kWriterThreadsConfig{"writer-threads"};
-constexpr absl::string_view kUseCoordinator{"use-coordinator"};
-constexpr absl::string_view kLogLevel{"log-level"};
-
-static const int64_t kDefaultThreadsCount = vmsdk::GetPhysicalCPUCoresCount();
```
eifrah-aws, line 27: I noticed that you removed these constants and defined them whenever they are needed. However, it is a matter of style, and this repo coding style defines them at the top of the file (at least this is how its done for other constants). Please place them back at the top

src/valkey_search_options.cc
eifrah-aws: This goes against the coding style (which I personally like...) in this repo...

src/commands/filter_parser.h
```diff
@@ -61,7 +61,7 @@ class FilterParser {
   absl::flat_hash_set<std::string> filter_identifiers_;
 
   absl::StatusOr<bool> IsMatchAllExpression();
-  absl::StatusOr<std::unique_ptr<query::Predicate>> ParseExpression();
+  absl::StatusOr<std::unique_ptr<query::Predicate>> ParseExpression(int level);
```
eifrah-aws: Can level be negative?

src/commands/filter_parser.cc
```diff
@@ -322,7 +323,10 @@ std::unique_ptr<query::Predicate> WrapPredicate(
 // 10. Numeric filters are inclusive. Exclusive min or max are expressed with (
 // prepended to the number, for example, [(100 (200].
 absl::StatusOr<std::unique_ptr<query::Predicate>>
-FilterParser::ParseExpression() {
+FilterParser::ParseExpression(int level) {
```
eifrah-aws: I am missing tests for this

src/valkey_search_options.cc
allenss-amazon: That was at my suggestion. IMO, adding a bunch of constant declarations that are only used once and are widely separated from their usage actually degrades the readability of the code. Here, these constants are used in a declaration where their semantics is obvious so having a name that duplicates the value of the string constant brings nothing helpful. Having the string constant be present miles away from it's usage in the declaration of a configurable increases the cognitive load the other reader -- pointlessly.

If you re-use the constant multiple times OR the text of the constant doesn't indicate it's semantics, then I think having a constant declaration is fine and helpful.

src/commands/filter_parser.cc
KarthikSubbarao: I was waiting on adding tests so that I can use the python integration framework if it is done before the PR is reviewed.

And I see it was added 2 days back :D Thank you, I will use this

src/commands/filter_parser.cc
```diff
@@ -322,7 +323,10 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+  if (level++ >= options::GetQueryStringDepth()) {
+    return absl::InvalidArgumentError("Query string recursive depth exceeded");
```
yairgott: It would be much clearer if the error states the `options::GetQueryStringDepth()` value.

src/valkey_search_options.cc
```diff
@@ -130,13 +127,27 @@ static auto writer_threads_count =
...
+    REDISMODULE_LOGLEVEL_VERBOSE,
+    REDISMODULE_LOGLEVEL_DEBUG,
+};
+
+static const std::vector<int> kLogLevelValues = {
+    static_cast<int>(LogLevel::kWarning), static_cast<int>(LogLevel::kNotice),
+    static_cast<int>(LogLevel::kVerbose), static_cast<int>(LogLevel::kDebug)};
+
 /// Control the modules log level verbosity
+constexpr absl::string_view kLogLevel{"log-level"};
```
yairgott: why do we need to duplicate: https://github.com/valkey-io/valkey-search/blob/2af5e6c02b67a4f4bdc742881f47ca08904c7f30/vmsdk/src/log.cc#L89 ?

KarthikSubbarao: This line of code was just refactored. This is not new code which I am adding in this PR

src/commands/filter_parser.cc
KarthikSubbarao: Updated the error message

src/commands/filter_parser.cc
KarthikSubbarao: Added a new python integration test for query parsing to test the limits. For the future work for query parsing updates, I planned to use this new file

src/commands/filter_parser.cc
```diff
@@ -322,7 +323,11 @@ std::unique_ptr<query::Predicate> WrapPredicate(
 // 10. Numeric filters are inclusive. Exclusive min or max are expressed with (
 // prepended to the number, for example, [(100 (200].
 absl::StatusOr<std::unique_ptr<query::Predicate>>
-FilterParser::ParseExpression() {
+FilterParser::ParseExpression(uint32_t level) {
+  if (level++ >= options::GetQueryStringDepth()) {
+    return absl::InvalidArgumentError(absl::StrCat("Query string recursive depth exceeded limit: ",
```
allenss-amazon: Not sure that "recursive depth" is really what we want to talk about here. It's hard to explain that the OR counts the same as the parenthesis. IMO, we should just say "expression too complex, please simplify" and not report the limit as it's hard to explain the connection.

integration/test_query_parser.py
```diff
@@ -0,0 +1,72 @@
...
+        try:
+            client.execute_command(
+                "FT.SEARCH", "my_index",
+                "((((((((((@price:[10 20])))))))))) =>[KNN 10 @doc_embedding $BLOB]",
+                "PARAMS", 2, "BLOB", "<your_vector_blob>",
+                "RETURN", 1, "doc_embedding"
+            )
+            assert False
+        except ResponseError as e:
+            assert str(e) == "Invalid filter expression: `((((((((((@price:[10 20]))))))))))`. Query string recursive depth exceeded limit: 10"
```
allenss-amazon: I don't think we should be repeating the query string in the error reply. That may not be code you changed, but I think we should remove it because if the string is large the actual error message gets lost.

allenss-amazon (approved):
I'm requesting some minor wording changes. which can be done in a subsequent PR -- because I think we also need to limit the size of the query string itself.

src/commands/filter_parser.cc
KarthikSubbarao: I can update this message to be "Query string is too complex" since documentation refers to this command argument as the query string.

@yairgott / @allenss-amazon - I will update this error message based on what we agree here

integration/test_query_parser.py
KarthikSubbarao: Yes, this is coming from an internal error builder in this Module, not introduced by PR specifically. I agree that it should not be repeated here

src/commands/filter_parser.cc
allenss-amazon: I'm good with that.

yairgott: Me too.

KarthikSubbarao: Thanks - I pushed a commit to address this

allenss-amazon (approved):

PR #194, opened 2025-06-26
Rebranding module to use Valkey instead of Redis

yanamolo:
As of today, the module uses module API commands and function names with Redis instead of Valkey.

Rebranded the module to use only Valkey.
For example, changed: `REDISMODULE_API` -> `VALKEYMODULE_API` etc.

### DIFF
--- a/src/acl.cc
+++ b/src/acl.cc
@@ -211,9 +211,9 @@ bool IsAllowedCommands(
   return allowed;
 }
 
-absl::string_view CallReplyStringToStringView(RedisModuleCallReply *r) {
+absl::string_view CallReplyStringToStringView(ValkeyModuleCallReply *r) {
   size_t l;
-  const char *r_c_str = RedisModule_CallReplyStringPtr(r, &l);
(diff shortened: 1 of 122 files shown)


yairgott:
> * Do we want to to move into SPDX licensing style?

Yes! This can be done either as part of this PR or a separated one.

madolson:
> Change that into Valkey (are we allowed?)?

If the license was originally from Redis, the guidance we got was to leave it as is. We should only be using SPDX for our licenses.

> Do we want to to move into SPDX licensing style?

Ideally yes. This is like the only ask from the Linux Foundation, (SPDX is one of their projects).

yairgott:
> If the license was originally from Redis, the guidance we got was to leave it as is. We should only be using SPDX for our licenses.

Just to be clear, none of valkey-search code is subject to Redis licensing, it was developed in-house while considering the Redis license restrictions. The mentioning of Redis in valkey-search license header in each source file is a bug. It was copied and pasted from one of the Valkey engine source file, which is subject to Redis licensing.

yairgott (approved):

PR #207, opened 2025-07-04
Add FT._DEBUG command

allenss-amazon:
* Initial implementation of FT._DEBUG command with subcommands
* Two initial subcommands are implemented to show info field metrics and metadata


eifrah-aws (changes requested):
3 general comments:

- Not related to this PR (we can fix this in a later PR): the `String` should not be aware of the `StringBuilder` class, in general the `StringBuilder` class should have `Build()` method that returns a `String`. This approach completely decouples the builder from the buildee (this is also true for the `Numeric` class)
- We are missing `ft._debug.json` file - I checked and other "debug" method like Valkey's `DEBUG` command does have a JSON file.
- Please run a `clang-format` on the code as it does not seem to match other files styling (e.g. 4 spaces vs 2 spaces)

src/commands/commands.h
```diff
@@ -35,6 +36,7 @@ constexpr absl::string_view kDropIndexCommand{"FT.DROPINDEX"};
 constexpr absl::string_view kInfoCommand{"FT.INFO"};
 constexpr absl::string_view kListCommand{"FT._LIST"};
 constexpr absl::string_view kSearchCommand{"FT.SEARCH"};
+constexpr absl::string_view kDebugCommand{"FT._DEBUG"};
```
yairgott, line 39: why not "FT.DEBUG"? I think that we should prefix commands with "_" to indicate a temporary command which is expected to the retire in the future.

vmsdk/src/info.cc
```diff
@@ -216,6 +217,24 @@ void Numeric<T>::Dump(ValkeyModuleInfoCtx *ctx) const {
...
+            char buffer[100];
+            size_t used = vmsdk::DisplayAsSIBytes(value, buffer, sizeof(buffer));
+            ValkeyModule_ReplyWithCString(ctx, buffer);
+        } else {
+            ValkeyModule_ReplyWithLongLong(ctx, value);
+        }
+    } else if constexpr (std::is_floating_point<T>::value) {
+        ValkeyModule_ReplyWithDouble(ctx, value);
+    } else {
+        CHECK(false);
```
yairgott: Should a CHECK be used here?

vmsdk/src/info.cc
```diff
@@ -248,9 +267,105 @@ void String::Dump(ValkeyModuleInfoCtx *ctx) const {
...
+    VMSDK_LOG(WARNING, ctx) << "Dumping Info Field: " << field->GetSection() << "/" << field->GetName();
+    ValkeyModule_ReplyWithCString(ctx, "Section");
+    ValkeyModule_ReplyWithCString(ctx, field->GetSection().data());
+    std::string external_name = options.name + "." + field->GetName();
+    ValkeyModule_ReplyWithCString(ctx, "Name");
+    ValkeyModule_ReplyWithCString(ctx, external_name.data());
+    return 4;
+}
+
+absl::Status DumpInfoMetaData(ValkeyModuleCtx *ctx, vmsdk::ArgsIterator itr, const vmsdk::module::Options& options) {
```
yairgott: Pass itr by reference

vmsdk/src/info.cc
```diff
@@ -248,9 +267,105 @@ void String::Dump(ValkeyModuleInfoCtx *ctx) const {
...
+                response_length += 2;
+            }
+            ValkeyModule_ReplySetArrayLength(ctx, response_length);
+        }
+    }
+    ValkeyModule_ReplySetArrayLength(ctx, num_infos);
+    return absl::OkStatus();
+}
+
+absl::Status DumpInfoValues(ValkeyModuleCtx *ctx, vmsdk::ArgsIterator itr, const vmsdk::module::Options& options) {
```
yairgott: Does it make sense to consolidate with `DumpInfoValues`, say with an additional flag to indicate wether to include the metadata?

src/commands/commands.h
allenss-amazon: Agreed, but I wanted to give notice that nothing in FT._DEBUG can be relied upon from release to release.

vmsdk/src/info.cc
allenss-amazon: will do.

eifrah-aws (approved):

PR #208, opened 2025-07-07
Add config hnsw-allow-replace-deleted

zvi-code:
This is to allow configure `allow_replace_deleted_` + add into ft.info `curr_deleted_vectors` and `curr_vectors`

### DIFF
--- a/integration/hnsw_allow_replace_deleted.py
+++ b/integration/hnsw_allow_replace_deleted.py
@@ -1,0 +1,415 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025, valkey-search contributors
+# All rights reserved.
+# SPDX-License-Identifier: BSD 3-Clause
+
+import numpy as np
+import struct
+import time
+from valkeytestframework.util.waiters import *
+from valkey import ResponseError
+from valkey.client import Valkey
+from valkey_search_test_case import ValkeySearchTestCaseBase
+from valkeytestframework.conftest import resource_port_tracker
+
+
+class TestHNSWAllowReplaceDeleted(ValkeySearchTestCaseBase):
+    """
+    Test suite for HNSW allow_replace_deleted configuration
+    
+    This test suite validates the HNSW vector index behavior when the allow_replace_deleted
+    feature is enabled or disabled. The feature controls whether deleted vector slots can be
+    reused for new insertions, which affects memory efficiency and index growth patterns.   
+    """
+    
+    def _create_vector_bytes(self, dim=128):
+        """Helper to create random normalized vector bytes"""
+        vec = np.random.rand(dim).astype(np.float32)
+        vec = vec / np.linalg.norm(vec)  # Normalize for cosine similarity
+        return struct.pack(f'{dim}f', *vec)
+    
+    def _extract_info_field(self, info_output, field_name):
+        """Helper to extract field from FT.INFO output"""
+        info_dict = dict(zip(info_output[::2], info_output[1::2]))
+        return info_dict.get(field_name)
+    
+    def _get_vector_counts(self, client, index_name):
+        """Helper to get current and deleted vector counts from FT.INFO"""
+        info = client.execute_command('FT.INFO', index_name)
+        
+        # The new fields are at the top level of FT.INFO response
+        info_dict = dict(zip(info[::2], info[1::2]))
+        
+        curr_vectors = int(info_dict.get(b'curr_vectors', 0))
+        deleted_vectors = int(info_dict.get(b'curr_deleted_vectors', 0))
+        
+        return curr_vectors, deleted_vectors
+
+    def test_memory_efficiency_with_replacement(self):
+        """Tests that enabling allow_replace_deleted prevents unnecessary memory growth by reusing
+            deleted vector slots instead of expanding the index capacity"""
+        client: Valkey = self.server.get_new_client()
+        
+        # Set the configuration at runtime
+        client.config_set('search.hnsw-allow-replace-deleted', 'yes')
+        
+        # Create index
+        response = client.execute_command(
+            'FT.CREATE', 'idx', 'SCHEMA', 'vec', 'VECTOR', 'HNSW', '6',
+            'TYPE', 'FLOAT32', 'DIM', '128', 'DISTANCE_METRIC', 'L2'
+        )
+        assert response == b'OK'
+        
+        # Add initial vectors
+        num_vectors = 10000
+        for i in range(num_vectors):
+            vec_bytes = self._create_vector_bytes(128)
+            client.hset(f'doc{i}', 'vec', vec_bytes)
+        
+        # Get HNSW block size configuration
+        block_size_config = client.config_get('search.hnsw-block-size')
+        block_size = int(block_size_config['search.hnsw-block-size'])
+        print(f"HNSW block size: {block_size}")
+        
+        # Calculate expected blocks based on initial vectors
+        initial_expected_blocks = (num_vectors + block_size - 1) // block_size  # Ceiling division
+        print(f"Expected initial blocks: {initial_expected_blocks} (for {num_vectors} vectors)")
+        
+        # Get initial vector counts
+        initial_curr, initial_deleted = self._get_vector_counts(client, 'idx')
+        print(f"Initial vector counts: curr={initial_curr}, deleted={initial_deleted}")
+        
+        # Delete half the vectors
+        deleted_count = 0
+        for i in range(0, num_vectors, 2):
+            result = client.delete(f'doc{i}')
+            if result == 1:
+                deleted_count += 1
+        
+        assert deleted_count == num_vectors // 2
+        
+        # Add new vectors (should reuse deleted slots)
+        new_vectors_count = num_vectors // 2
+        for i in range(num_vectors, num_vectors + new_vectors_count):
+            vec_bytes = self._create_vector_bytes(128)
+            client.hset(f'doc{i}', 'vec', vec_bytes)
+        
+        # Get final vector counts
+        final_curr, final_deleted = self._get_vector_counts(client, 'idx')
+        print(f"Final vector counts: curr={final_curr}, deleted={final_deleted}")
+        
+        # Calculate expected blocks with replacement (should stay same or minimal growth)
+        # With replacement enabled, total vectors should stay close to original
+        # because deleted slots are reused
+        total_vectors_with_replacement = num_vectors  # Should remain ~10k as slots are reused
+        final_expected_blocks = (total_vectors_with_replacement + block_size - 1) // block_size
+        print(f"Expected final blocks: {final_expected_blocks} (with replacement reusing slots)")
+        
+        # Make concrete assertions based on block size
+        # With replacement enabled, memory growth should be minimal
+        expected_growth = final_expected_blocks / initial_expected_blocks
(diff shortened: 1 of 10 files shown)


integration/hnsw_allow_replace_deleted.py
```diff
@@ -0,0 +1,415 @@
+#!/usr/bin/env python3
```
allenss-amazon, line 1: ? I don't think this is a standalone executable file.

integration/hnsw_allow_replace_deleted.py
```diff
@@ -0,0 +1,415 @@
...
+        return info_dict.get(field_name)
+    
+    def _get_vector_counts(self, client, index_name):
+        """Helper to get current and deleted vector counts from FT.INFO"""
+        info = client.execute_command('FT.INFO', index_name)
+        
+        # The new fields are at the top level of FT.INFO response
+        info_dict = dict(zip(info[::2], info[1::2]))
+        
+        curr_vectors = int(info_dict.get(b'curr_vectors', 0))
```
allenss-amazon, line 45: Isn't this duplicative of the "extra_info_field" helper function?

integration/hnsw_allow_replace_deleted.py
```diff
@@ -0,0 +1,415 @@
...
+            deleted vector slots instead of expanding the index capacity"""
+        client: Valkey = self.server.get_new_client()
+        
+        # Set the configuration at runtime
+        client.config_set('search.hnsw-allow-replace-deleted', 'yes')
+        
+        # Create index
+        response = client.execute_command(
+            'FT.CREATE', 'idx', 'SCHEMA', 'vec', 'VECTOR', 'HNSW', '6',
+            'TYPE', 'FLOAT32', 'DIM', '128', 'DISTANCE_METRIC', 'L2'
```
allenss-amazon, line 61: This would be more readable as a single string. execute_command will split it's first argument on spaces, since you don't have any embedded spaces in the string it's equivalent.

integration/hnsw_allow_replace_deleted.py
```diff
@@ -0,0 +1,415 @@
...
+        
+        # Get initial vector counts
+        initial_curr, initial_deleted = self._get_vector_counts(client, 'idx')
+        print(f"Initial vector counts: curr={initial_curr}, deleted={initial_deleted}")
+        
+        # Delete half the vectors
+        deleted_count = 0
+        for i in range(0, num_vectors, 2):
+            result = client.delete(f'doc{i}')
+            if result == 1:
```
allenss-amazon, line 88: does this ever return 0? Would that indicate a test failure? If so, let's just assert it == 1.

integration/hnsw_allow_replace_deleted.py
```diff
@@ -0,0 +1,415 @@
...
+        expected_growth = final_expected_blocks / initial_expected_blocks
+        print(f"Expected memory growth: {expected_growth:.2f}x (blocks: {initial_expected_blocks} -> {final_expected_blocks})")
+        
+        # Assert that memory growth is minimal (should be 1.0x or very close)
+        assert expected_growth <= 1.1, f"Expected minimal memory growth (<=1.1x) with replacement, got {expected_growth:.2f}x"
+        
+        # Verify vector counts match expectations - with replacement, deleted slots should be reused
+        assert final_curr == num_vectors, f"Expected {num_vectors} total vectors (reused slots), got {final_curr}"
+        assert final_deleted == 0, f"Expected 0 deleted vectors (slots reused), got {final_deleted}"
+
```
allenss-amazon, line 121: Is it correct that all of this block allocation stuff is driven by the # of vectors coming back through INFO ? If so, then I don't understand why all of the block computations are being done. 

Also, why not just load up N vectors. Read the memory usage. Then overwrite all of the N vectors M times and then read out the vector_count and memory usage -- the memory usage shouldn't be more than a few percent higher, right? Is that all that this test accomplishes?

integration/hnsw_allow_replace_deleted.py
```diff
@@ -0,0 +1,415 @@
...
+        )
+        
+        # Verify results
+        num_results = results[0]
+        assert num_results == 10, f"Expected 10 results, got {num_results}"
+        
+        # Verify no deleted documents in results
+        for i in range(1, len(results), 2):
+            doc_id = results[i].decode('utf-8')
+            assert doc_id not in deleted_docs, f"Deleted document {doc_id} in results"
```
allenss-amazon, line 241: This test isn't related to the "replace_deleted" right? It's just a functional test of vector delete, right?

integration/hnsw_allow_replace_deleted.py
```diff
@@ -0,0 +1,415 @@
...
+        # Get initial config value
+        result = client.config_get('search.hnsw-allow-replace-deleted')
+        assert result['search.hnsw-allow-replace-deleted'] == 'no'
+        
+        # Change at runtime
+        client.config_set('search.hnsw-allow-replace-deleted', 'yes')
+        
+        # Verify the change
+        result = client.config_get('search.hnsw-allow-replace-deleted')
+        assert result['search.hnsw-allow-replace-deleted'] == 'yes'
```
allenss-amazon, line 350: This seems duplicative of the tests above.

integration/hnsw_allow_replace_deleted.py
```diff
@@ -0,0 +1,415 @@
...
+    def test_performance_comparison(self):
+        """Measures and compares insertion and search performance between replacement enabled
+            and disabled modes to ensure the feature doesn't negatively impact performance."""
+        results = {}
+        results['no'] = {}
+        results['yes'] = {}
+        perf_cases = [('no',10000000), ('yes',20), ('yes',15), ('yes',10), ('yes',7), ('yes',5), ('yes',4), ('yes',3), ('yes',2)]
+        for allow_replace, part in perf_cases:
+            client: Valkey = self.server.get_new_client()
+
```
allenss-amazon, line 362: This kind of performance testing is inherently flaky. This is the wrong place for this kind of testing.

integration/valkey_search_test_case.py
```diff
@@ -11,7 +11,11 @@ class ValkeySearchTestCaseBase(ValkeyTestCase):
 
     @pytest.fixture(autouse=True)
     def setup_test(self, setup):
-        loadmodule = f"{os.getenv('MODULE_PATH')} --loadmodule {os.getenv('JSON_MODULE_PATH')}"
+        loadmodule = os.getenv('MODULE_PATH')
+        json_module_path = os.getenv('JSON_MODULE_PATH')
+        if json_module_path and json_module_path.strip():
+            loadmodule = f"{loadmodule} --loadmodule {json_module_path}"
+
```
allenss-amazon, line 18: Is there some reason this needs to be changed in the PR?

src/valkey_search_options.cc
```diff
@@ -139,6 +140,12 @@ static auto log_level =
         .WithValidationCallback(ValidateLogLevel)
         .Build();
 
+/// Register the "--hnsw-allow-replace-deleted" flag.
+static auto hnsw_allow_replace_deleted =
+    config::BooleanBuilder(kHNSWAllowReplaceDeleted, false)  // default false
```
allenss-amazon, line 145: Is False really the config we want?

testing/CMakeLists.txt
```diff
@@ -251,6 +251,23 @@ target_link_libraries(filter_test PUBLIC predicate)
...
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC index_schema_cc_proto)
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC index_base)
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC vector_base)
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC vector_flat)
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC vector_hnsw)
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC string_interning)
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC hnswlib_vmsdk)
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC managed_pointers)
+target_link_libraries(hnsw_allow_replace_deleted_test PUBLIC type_conversions)
+finalize_test_flags(hnsw_allow_replace_deleted_test)
```
allenss-amazon, line 269: Gawds, we desparately need to fix the unit test build files.

testing/hnsw_allow_replace_deleted_test.cc
```diff
@@ -0,0 +1,540 @@
...
+    std::cout.flush();
+    
+    std::cout << "About to init SchemaManager..." << std::endl;
+    std::cout.flush();
+    SchemaManager::InitInstance(std::make_unique<TestableSchemaManager>(
+        &fake_ctx_, []() { server_events::SubscribeToServerEvents(); }, nullptr,
+        false));
+    std::cout << "SchemaManager initialized" << std::endl;
+    std::cout.flush();
+
```
allenss-amazon, line 78: This kind of setup boiler plate stuff belongs in a common class, can we refactor this to make it more generically useful for future development?

testing/hnsw_allow_replace_deleted_test.cc
```diff
@@ -0,0 +1,540 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#include <atomic>
+#include <chrono>
```
allenss-amazon, line 9: How much of this test is duplicated in the python level tests? Seems like a lot to me.

zvi-code: a lot, but the python level tests have valkey-server and client. These are more of a unit\sub-system tests that hook internal api's. But specifically in the draft i didn't filter\clean a lot the tests, mostly cared about making sure i have tests that validate the functionality

src/valkey_search_options.cc
zvi-code: I don't know, but in any case i would't introduce the config change in this PR. It would require much deeper analysis of the implication. Current state the pro\con are clear, we should first get to the same "clarity of implications" state with enabled

integration/valkey_search_test_case.py
zvi-code: will remove. could be issue with framework handling module load with arguments

integration/hnsw_allow_replace_deleted.py
zvi-code: not sure, will simplify

integration/hnsw_allow_replace_deleted.py
zvi-code: will remove

integration/hnsw_allow_replace_deleted.py
zvi-code: will revise

integration/hnsw_allow_replace_deleted.py
zvi-code: Agree, I added it before i exposed the num deleted metric, so i used the "performance" to indicate the vector re-use. But now with the `curr_deleted_vectors` in ft.info, this is not the relevant way to test it. Will remove

zvi-code:
> I wonder if you measured the recall impact of enabling allow replace deleted?

Not systematically yet but it is part of my next steps on this. Will share the results when i have them

PR #209, opened 2025-07-07
Add config to skip loading index from rdb

zvi-code:
## ReIndex Vector Index RDB Loading

valkey-search includes an operational feature to mitigate issues in indexes. If set to true, index schema is loaded from rdb, but indexes are not loaded, they are created empty and will be re-filled by the backfill process.

### Configuration

```sh
valkey-server --loadmodule /path/to/libsearch.so --rdb-load-skip-index yes
```

```sh
CONFIG GET search.rdb-load-skip-index
```

### Benefits

- **Index consistency**: Recover from corrupted or inconsistent vector indexes
- **Deleted Memory reclaim**: If index has many deleted vectors rebuilding the index during startup will free this memory

When enabled, vector indexes are rebuilt via the backfill process while non-vector indexes (TAG, NUMERIC) continue to work immediately.


integration/valkey_search_test_case.py
```diff
@@ -11,7 +11,7 @@ class ValkeySearchTestCaseBase(ValkeyTestCase):
 
     @pytest.fixture(autouse=True)
     def setup_test(self, setup):
-        loadmodule = f"{os.getenv('MODULE_PATH')} --loadmodule {os.getenv('JSON_MODULE_PATH')}"
+        loadmodule = f"{os.getenv('MODULE_PATH')}"
```
allenss-amazon: this looks like a bad merge to me.

src/coordinator/search_converter.cc
```diff
@@ -145,9 +145,9 @@ GRPCSearchRequestToParameters(const SearchIndexPartitionRequest& request) {
             parameters->filter_parse_results.filter_identifiers));
   }
   for (auto& return_parameter : request.return_parameters()) {
-    parameters->return_attributes.emplace_back(query::ReturnAttribute(
-        vmsdk::MakeUniqueValkeyString(return_parameter.identifier()),
-        vmsdk::MakeUniqueValkeyString(return_parameter.alias())));
+    parameters->return_attributes.emplace_back(query::ReturnAttribute{
+        .identifier = vmsdk::MakeUniqueValkeyString(return_parameter.identifier()),
+        .alias = vmsdk::MakeUniqueValkeyString(return_parameter.alias())});
```
allenss-amazon: What changed here?

src/index_schema.cc
```diff
@@ -647,7 +655,20 @@ absl::string_view IndexSchema::GetStateForInfo() const {
...
+          has_vector_index = true;
+          break;
+        }
+      }
+      // TODO: This is not accurate in case we have backfill started due to new index (config set only has impact if index is loaded from rdb).
+      if (has_vector_index && options::GetReIndexVectorRDBLoad().GetValue()) {
+        return "vector_index_rebuilding";
+      } else {
+        return "backfill_in_progress";
+      }
```
allenss-amazon: Seems to me that having a virtual function for each index type would make this a lot cleaner. Then the index-schema wouldn't have to check what type of index it was. We can have default implementation of that function that's used for the non-vector types....

submodules/CMakeLists.txt
```diff
@@ -199,7 +199,16 @@ else()
...
+  # Force C++ standard for highwayhash before adding it as subdirectory
+  set(CMAKE_CXX_STANDARD_BACKUP ${CMAKE_CXX_STANDARD})
+  set(CMAKE_CXX_STANDARD 14)
   add_subdirectory(${MODULES_DIR}/highwayhash ${CMAKE_BINARY_DIR}/highwayhash)
+  set(CMAKE_CXX_STANDARD ${CMAKE_CXX_STANDARD_BACKUP})
+  # Set C++ standard for highwayhash target
+  if(TARGET highwayhash)
+    set_property(TARGET highwayhash PROPERTY CXX_STANDARD 14)
+    set_property(TARGET highwayhash PROPERTY CXX_STANDARD_REQUIRED ON)
+  endif()
```
allenss-amazon: why is this needed?

testing/force_reindex_vector_rdb_load.cc
```diff
@@ -0,0 +1,1197 @@
...
+                   static_cast<float>(i % 10) / 100.0f;
+        } else {
+          // Exponential decay pattern
+          vec[j] = std::exp(-static_cast<float>(j) / (kDimensions / 4.0f)) * (i + 1);
+        }
+      }
+      vectors.push_back(std::move(vec));
+    }
+    return vectors;
+  }
```
allenss-amazon: What purpose does this serve?

testing/force_reindex_vector_rdb_load.cc
```diff
@@ -0,0 +1,1197 @@
...
+            vec[j] = std::exp(-static_cast<float>(j) / (kDimensions / 3.0f)) * (i + 1) * 10.0f;
+          }
+          // Add unique component based on vector index to ensure uniqueness
+          vec[j] += static_cast<float>(i) * 0.01f + static_cast<float>(j) * 0.001f;
+        }
+        vectors.push_back(std::move(vec));
+      }
+      return vectors;
+    }
+
```
allenss-amazon: ??

README.md
```diff
@@ -48,6 +48,23 @@ valkey-search uses a **hybrid approach** with a query planner that selects the m
...
 - **Inline-filtering:** Filters results during the similarity search process.
 
+## ReIndex Vector Index RDB Loading
+
+valkey-search includes an operational feature to mitigate issues in the vector index.
+
+### Configuration
+
+```sh
+CONFIG SET reindex-vector-rdb-load true
```
yairgott: 1. I thought that by using config::BooleanBuilder the config should look like: `search.reindex-vector-rdb-load`. Am I missing something?
2. The name is a little confusing. WDYT about  `search.skip-rdb-load` ?

zvi-code: 1. Correct, will fix the search prefix
2. I started with this name actually (hence the pr name), but than thought of re-index because I think that down the road, the reindexing capabilities can\should be extended beyond rdb load. Even for rdb load, we might want to have threshold that if you have %deleted > X you re-index. Having said all this, I'm happy with skip for this functionality and we figure out the wider scope name when we implement it

src/coordinator/search_converter.cc
zvi-code: I had some build issues, not sure why, was running on my mac, will clean this up when converting to PR

integration/valkey_search_test_case.py
zvi-code: No It is intentional hack for Draft PR. I think the testing framework is no handling well multiple modules with args. will fix for the actual PR

src/index_schema.cc
zvi-code: yea, in any case, as i wrote in my TODO comment, this is not sufficient indication. We need to add a field that would indicate rebuild index was triggered. I agree about your comment, will fix it in the PR

submodules/CMakeLists.txt
zvi-code: build issues on my mac, will remove

testing/force_reindex_vector_rdb_load.cc
zvi-code: will remove, i think it was leftover from other experiments

testing/force_reindex_vector_rdb_load.cc
zvi-code: will remove

src/index_schema.cc
```diff
@@ -819,11 +820,30 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::LoadFromRDB(
...
+  if (ABSL_PREDICT_FALSE(skip_loading_index_data)) {
+    VMSDK_LOG(NOTICE, ctx) << "Creating empty indexes for schema '" 
+                           << index_schema_proto->name() 
+                           << "' - will rebuild via backfill";
+    while (supplemental_iter.HasNext()) {
+      VMSDK_ASSIGN_OR_RETURN(auto supplemental_content, supplemental_iter.Next());
+      auto chunk_it = supplemental_iter.IterateChunks();
+      while (chunk_it.HasNext()) {
+        VMSDK_ASSIGN_OR_RETURN(auto chunk_result, chunk_it.Next());
+        (void)chunk_result; // Mark as intentionally unused
```
eifrah-aws: use: `[[maybe_unused]]`

src/valkey_search_options.cc
```diff
@@ -108,7 +108,7 @@ static auto writer_threads_count =
 constexpr absl::string_view kUseCoordinator{"use-coordinator"};
 static auto use_coordinator =
     config::BooleanBuilder(kUseCoordinator, false)
-        .WithFlags(VALKEYMODULE_CONFIG_HIDDEN)  // can only be set during
+        .WithFlags(VALKEYMODULE_CONFIG_IMMUTABLE)  // can only be set during
```
eifrah-aws, line 111: Please run `clang-format` on your PR

src/valkey_search_options.cc
```diff
@@ -124,6 +124,13 @@ static const std::vector<int> kLogLevelValues = {
     static_cast<int>(LogLevel::kWarning), static_cast<int>(LogLevel::kNotice),
     static_cast<int>(LogLevel::kVerbose), static_cast<int>(LogLevel::kDebug)};
 
+
+/// Should this instance reindex vector index RDB loading?
+constexpr absl::string_view kReIndexVectorRDBLoad{"rdb-load-skip-index"};
+static auto rdb_load_skip_index =
```
eifrah-aws, line 129: nit: no need for the builder here (you are not using any of `With*` methods, just place a `static auto config::Boolean`

testing/common.cc
```diff
@@ -143,6 +143,19 @@ data_model::VectorIndex CreateFlatVectorIndexProto(
   return vector_index_proto;
 }
 
+data_model::NumericIndex CreateNumericIndexProto() {
+  data_model::NumericIndex numeric_index_proto;
+  return numeric_index_proto;
```
eifrah-aws: nit:

If the entire purpose of this method is to create a default numeric index, then it can be folded into:

```c++
data_model::NumericIndex CreateNumericIndexProto() {
  return {};
}
```

testing/index_schema_test.cc
```diff
@@ -1416,4 +1419,411 @@ TEST_F(IndexSchemaTest, ShouldBlockClient) {
...
+                            .value();
+
+    auto hnsw_index = indexes::VectorHNSW<float>::Create(
+        CreateHNSWVectorIndexProto(dimensions, distance_metric, initial_cap,
+                                   m, ef_construction, ef_runtime),
+        "embedding", data_model::AttributeDataType::ATTRIBUTE_DATA_TYPE_HASH)
+        .value();
+    VMSDK_EXPECT_OK(index_schema->AddIndex("embedding", "emb_id", hnsw_index));
+
+    // Add 1000 vectors
```
eifrah-aws, line 1456: nit:

```
// Add "num_vectors" vectors
```

eifrah-aws (commented):
We should add an integration test (python based) for this feature.

README.md
```diff
@@ -48,6 +48,28 @@ valkey-search uses a **hybrid approach** with a query planner that selects the m
 - **Pre-filtering**
 - **Inline-filtering:** Filters results during the similarity search process.
 
+## Skip Index on RDB Load
+
+valkey-search includes an operational feature to mitigate issues in indexes. If set to true, index schema is loaded from rdb, but indexes are not loaded, they are created empty and will be re-filled by the backfill process.
```
allenss-amazon, line 53: loaded from rdb and then discarded?

README.md
```diff
@@ -48,6 +48,28 @@ valkey-search uses a **hybrid approach** with a query planner that selects the m
...
+
+### Configuration
+
+The config is non-modifiable and should be provided to search module on load.
+```sh
+valkey-server --loadmodule /path/to/libsearch.so --rdb-load-skip-index yes
```
allenss-amazon: does this apply only at load time? Or for all LOAD commands?

integration/test_skip_index_load.py
```diff
@@ -0,0 +1,445 @@
...
+    # 00000060 | 3a 31 40 7d 7d 00 00 00   06 00 85 74 69 74 6c 65  | :1@}}......title
+    # 00000070 | 06 8f 53 61 6d 70 6c 65   20 44 6f 63 75 6d 65 6e  | ..Sample Documen
+    # 00000080 | 74 10 8b 64 65 73 63 72   69 70 74 69 6f 6e 0c b2  | t..description..
+    # 00000090 | 54 68 69 73 20 69 73 20   61 20 74 65 73 74 20 64  | This is a test d
+    # 000000a0 | 6f 63 75 6d 65 6e 74 20   77 69 74 68 20 61 6e 20  | ocument with an
+    # 000000b0 | 65 6d 62 65 64 64 69 6e   67 20 76 65 63 74 6f 72  | embedding vector
+    # 000000c0 | 20 31 33 89 65 6d 62 65   64 64 69 6e 67 0a 90 00  |  13.embedding...
+    # 000000d0 | 00 00 00 77 d6 88 3e 77   d6 08 3f b3 41 4d 3f 11  | ...w..>w..?.AM?.
+    # 000000e0 | ff f7 81 56 4f 92 79 aa   dc 84 01 02 02 02 80 00  | ...VO.y.........
+    # 000000f0 | 01 00 00 02 01 05 40 98   08 01 10 04 1a 91 01 0a  | ......@.........
```
allenss-amazon: Not sure what the value of having this in hex is providing. If there was a direct indicator of what's invalid or incorrect vs the correct values that MIGHT be interesting. But I'm skeptical.

integration/test_skip_index_load.py
```diff
@@ -0,0 +1,445 @@
...
+    # 00000310 | 20 01 28 80 50 32 07 08   10 10 c8 01 18 0a 05 0e  |  .(.P2..........
+    # 00000320 | 0a 0c 0a 05 64 6f 63 3a   31 1d 00 00 80 3f 05 00  | ....doc:1....?..
+    # 00000330 | 00 ff 9a ba e1 5f 26 29   2d da                    | ....._&)-.
+    #----------------------------------------------------------------------------
+    # spellchecker:on
+    """
+
+    # Corrupted RDB with vector index
+    RDB_BASE64 = (
+        "UkVESVMwMDEx+gp2YWxrZXktdmVyCzI1NS4yNTUuMjU1+gpyZWRpcy1iaXRzwED6BWN0aW1lwrvV"
```
allenss-amazon, line 20: For long-term maintainability, I think it's worth describing how this could be re-created. This would be useful in the future if we added some new formats for the RDB -- which seems likely.

integration/test_skip_index_load.py
```diff
@@ -0,0 +1,445 @@
...
+            dest_path = os.path.join(testdir, rdb_filename)
+            shutil.copy2(rdbfile, dest_path)
+            rdbfile = rdb_filename  # Use just the filename
+            logging.info(f"Copied RDB file from {rdbfile} to {dest_path}")
+
+        # Verify RDB file exists
+        rdb_path_in_testdir = os.path.join(testdir, rdbfile)
+        if os.path.exists(rdb_path_in_testdir):
+            logging.info(
+                f"RDB file exists at: {rdb_path_in_testdir}, size: {os.path.getsize(rdb_path_in_testdir)} bytes"
```
allenss-amazon, line 78: Why can't we use the existing test infrastructure instead of duplicating it? If that infrastructure is inadequate, let's enhance it.

integration/test_skip_index_load.py
```diff
@@ -0,0 +1,445 @@
...
+                                )
+                                log_found = True
+
+                                # Verify the failure was due to RDB corruption
+                                if (
+                                    "Failed to load ValkeySearch aux section from RDB"
+                                    in log_content
+                                ):
+                                    logging.info(
+                                        "✅ Confirmed: Server failed due to corrupted search index in RDB"
```
allenss-amazon, line 197: Scanning the logfile for a particular message is a common practice and should go into the test framework.

integration/test_skip_index_load.py
allenss-amazon: Also, there are tests later that depend on the metadata in this RDB, so having a description of how to build this would serve dual-purpose.

README.md
allenss-amazon: I think it's important to point out that there's a limit on the corruption that you can recover from.

integration/test_skip_index_load.py
```diff
@@ -0,0 +1,445 @@
...
+            assert (
+                b"index_to_test_skip_load" in indices
+            ), "Index not found in index list"
+
+            # Get index info to verify schema
+            info = client.execute_command("FT.INFO", "index_to_test_skip_load")
+            info_dict = {info[i]: info[i + 1] for i in range(0, len(info), 2)}
+
+            # Check that we have the expected fields (schema verification)
+            attributes = info_dict.get(b"attributes", [])
```
allenss-amazon, line 259: why the default value? If it's missing wouldn't want it to crash?

src/index_schema.cc
```diff
@@ -366,9 +368,11 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
   }
   if (added) {
     // Track key modifications by data type
-    if (attribute_data_type_->ToProto() == data_model::ATTRIBUTE_DATA_TYPE_HASH) {
+    if (attribute_data_type_->ToProto() ==
```
allenss-amazon: wouldn't a switch statement be clearer here?

src/valkey_search_options.cc
```diff
@@ -168,12 +173,21 @@ const vmsdk::config::Boolean& GetUseCoordinator() {
   return dynamic_cast<const vmsdk::config::Boolean&>(*use_coordinator);
 }
 
+const vmsdk::config::Boolean& GetSkipIndexLoad() {
+  return dynamic_cast<const vmsdk::config::Boolean&>(*rdb_load_skip_index);
```
allenss-amazon, line 177: why a dynamic cast?

README.md
zvi-code: That's true, you need a valid schema and chunks layout. This load of corrupted rdb test is mainly a validation that the config is indeed bypassing loading the index data and that backfill happens. The use case could be also for workaround to space reclamation or upgrade issues

integration/test_skip_index_load.py
zvi-code: The only goal was to provide some visibility into the rdb string below so it will not seem so opaque. If it's not serving this goal, I can remove

integration/test_skip_index_load.py
zvi-code: I have some scripts I created, but I don't know where would they fit. Also, for these scripts to serve as a general purpose tool, the scripts might require additional work beyond what I have right now

integration/test_skip_index_load.py
zvi-code: The infra is lacking capabilities, this can be migrated when they are added. I'll open a separate issue about the missing capabilities

src/index_schema.cc
```diff
@@ -851,11 +855,33 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::LoadFromRDB(
...
+  bool load_attributes_on_create = skip_loading_index_data;
+  // Attributes will be loaded from supplemental content. if
+  // !load_attributes_on_create
   VMSDK_ASSIGN_OR_RETURN(
       auto index_schema,
       IndexSchema::Create(ctx, *index_schema_proto, mutations_thread_pool,
-                          /*skip_attributes=*/true));
+                          !load_attributes_on_create));
+
+  if (ABSL_PREDICT_FALSE(skip_loading_index_data)) {
```
murphyjacob4: I'm not strongly opinionated, but have a slight preference to doing the check within each of the two branches below, so that we still log when there is unknown supplemental content.

The two types of supplemental content we have now (SUPPLEMENTAL_CONTENT_INDEX_CONTENT and SUPPLEMENTAL_CONTENT_KEY_TO_ID_MAP) are both related to the index contents, but that isn't necessarily guaranteed to be true forever.

README.md
murphyjacob4: Not necessarily a blocker for this change, but just thinking. The way that the current RDB format is setup, it looks like:

1. Minimum semantic version (e.g. 0.0.0 if any, 2.0.0 if you need 2.0 to load it)
2. Number of RDB sections
3. ... RDB sections ...

One thing I am thinking changes here is that the minimum semantic version relies on whether or not you care about the index contents or not. If we fundamentally change the index contents in version 2.0, but you don't care about it since you are loading in 1.0 with --rdb-load-skip-index, do we need to differentiate between the two?

integration/test_skip_index_load.py
allenss-amazon: I'm just trying to cut down on future maintenance. Assume that in the future we have an extended RDB V2.0, how will this test get updated? If you have some formula that a future maintainer can follow then it cuts down on that update activity. Without that formula, it seems likely that this test might not get updated.

integration/test_skip_index_load.py
allenss-amazon: I think using tests like this to drive the maturity of the infrastructure is what's going to happen. I'm just saying that this code should be put into the infrastructure, either as part of this PR or as a separate one.

README.md
zvi-code: This is a good point. Current code is minimal and should be thought of as operational tool[for memory reclamation and load failure cases]. I agree that we can enhance the capability to support upward compatibility. Wouldn't such capability require independant schema versioning? (do we have such?) Maybe related, do we have schema versioning to support mixed versions cluster, say we have gradual upgrade? [though this is a wider question]

README.md
```diff
@@ -56,11 +56,7 @@ valkey-search includes an operational feature to mitigate issues in indexes. If
 
 The config is non-modifiable and should be provided to search module on load.
 ```sh
-valkey-server --loadmodule /path/to/libsearch.so --rdb-load-skip-index yes
-```
-
-```sh
-CONFIG GET search.rdb-load-skip-index
+valkey-server --loadmodule /path/to/libsearch.so --skip-rdb-load yes
```
allenss-amazon, line 59: doesn't it have to be "search-skip-rdb-load" ?

src/index_schema.cc
zvi-code: @murphyjacob4, fixed, please review. I suspect existing code will fail load if unknown supplemental content, because it still needs to iterate over chunks. As part of the requested change i also made sure we iterate also over unknown type. please let me know if that's correct
```
  ~RDBSectionIter() {
    if (remaining_ > 0) {
      VMSDK_LOG(WARNING, nullptr)
          << "RDBSectionIter was not fully iterated. remaining_ == "
          << remaining_;
      DCHECK(remaining_ == 0);
    }
  }
```
Logs for when iteration is not completed:
```
19274:M 30 Jul 2025 12:02:37.808 # <search> [notice], tid: 254625352015904, /home/ubuntu/valkey-search/src/index_schema.cc:68: Starting backfill for index schema in DB 0: index_to_test_skip_load (size: 1)
719274:M 30 Jul 2025 12:02:37.808 # <search> [notice], tid: 254625352015904, /home/ubuntu/valkey-search/src/index_schema.cc:223: Index schema index_to_test_skip_load dropped from DB 0
719274:M 30 Jul 2025 12:02:37.808 # <module> [warning], tid: 254625352015904, /home/ubuntu/valkey-search/src/rdb_serialization.h:234: SupplementalContentIter was not fully iterated. remaining_ == 1
719274:M 30 Jul 2025 12:02:37.808 # <search> [warning], tid: 254625352015904, /home/ubuntu/valkey-search/src/rdb_serialization.cc:215: Failed to load ValkeySearch aux section from RDB: No more elements remaining; Failed to load index schema from RDB!
719274:M 30 Jul 2025 12:02:37.808 # The RDB file contains module AUX data for the module type 'Vk-Search', that the responsible module is not able to load. Check for modules log above for additional clues.
719274:M 30 Jul 2025 12:02:37.808 # Short read or OOM loading DB. Unrecoverable error, aborting now.
719274:M 30 Jul 2025 12:02:37.808 # Internal error in RDB reading offset 0, function at rdb.c:3425 -> Unexpected EOF reading RDB file
```

README.md
zvi-code: No, with config get you have the search. Prefix, when passing to the module as arg on load, it's without the prefix

PR #212, opened 2025-07-08
Add network byte counting to coordinator client

yulazariy:
Track network traffic metrics for SearchIndexPartition requests:
- Add coordinator_bytes_out and coordinator_bytes_in atomic counters
- Count outgoing bytes before sending requests
- Count incoming bytes only for successful responses
- Expose network traffic metrics through INFO command
- Add unit tests with real protocol buffer objects

This change enables monitoring of network traffic between coordinator clients and servers for better performance analysis and debugging.

This PR fixes this issue https://github.com/valkey-io/valkey-search/issues/213


eifrah-aws (approved):
LGTM

PR #218, opened 2025-07-11
Adding support for non vector queries

KarthikSubbarao:
This is a PR for adding support for non vector queries (e.g. numeric and tag based searches). Currently, I have tested with numeric/tag queries on Hash/JSON documents in CMD and CME. PR includes unit & integration tests.

Future work:
* Text query support - This will be implemented in a follow up PR in the fulltext branch
* QMA (query modifying attributes) support for Text queries - This will be implemented in a follow up PR in the fulltext branch

Questions / Considerations:

1. Scoring: Other search libraries use `inf` as the score for numeric/tag searches returning documents. valkey-search does not yet support `WITHSCORES`. However, when we do, we will have to remember to apply the same `inf` scoring for compatibility for non vector queries.
2. Sorting - valkey-search does not yet support `SORTBY`. Sorting will make sense for cases where non vector queries return a large number of documents. Currently, the order does not have a meaning behind it making this challenging for large responses.
3. Limit: We need to consider adding a limit on number of documents for non vector queries. However, this only makes sense if there is logic implemented for sorting

### DIFF
--- a/integration/test_non_vector.py
+++ b/integration/test_non_vector.py
@@ -28,7 +28,7 @@ expected_hash_value = {
     b'category': b"electronics"
 }
 
-numeric_tag_index_on_json = "FT.CREATE jsonproducts ON JSON PREFIX 1 jsonproduct: SCHEMA $.category TAG $.price NUMERIC $.rating NUMERIC"
+numeric_tag_index_on_json = "FT.CREATE jsonproducts ON JSON PREFIX 1 jsonproduct: SCHEMA $.category as category TAG $.price as price NUMERIC $.rating as rating NUMERIC"
 json_docs = [
     ['JSON.SET', 'jsonproduct:1', '$',
             '{"category":"electronics","name":"Laptop","price":999.99,"rating":4.5,"desc":"Great"}'],
@@ -41,7 +41,7 @@ json_docs = [
 ]
 numeric_query_on_json = [
     "FT.SEARCH", "jsonproducts",
-    "@$.price:[300 2000] @$.rating:[4.4 +inf]"
+    "@price:[300 2000] @rating:[4.4 +inf]"
 ]
 expected_numeric_json_key = b'jsonproduct:1'
 expected_numeric_json_value = {
@@ -53,7 +53,7 @@ expected_numeric_json_value = {
 }
 numeric_tag_query_on_json = [
     "FT.SEARCH", "jsonproducts",
-    "@$.category:{books} @$.price:[10 30] @$.rating:[4.7 +inf]"
+    "@category:{books} @price:[10 30] @rating:[4.7 +inf]"
 ]
 expected_numeric_tag_json_key = b'jsonproduct:4'
 expected_numeric_tag_json_value = {


src/commands/ft_search.cc
```diff
@@ -133,6 +133,27 @@ void SerializeNeighbors(ValkeyModuleCtx *ctx,
   }
 }
 
+// Handle non-vector queries by processing the neighbors and replying with the attribute contents.
+void SerializeNonVectorNeighbors(ValkeyModuleCtx *ctx,
```
yairgott, line 137: ctx and neighbors should be const.

KarthikSubbarao:
Thank you, @yairgott 

You and @allenss-amazon can let me know the plans for getting this change into main. From what I last heard, we want to address some of the consistency issues before supporting non vector queries.

Currently, this PR is set to be merged into the `fulltext` branch. I can switch this to `main` if needed

allenss-amazon:
Well, pushing this to Fulltext doesn't matter much.

I'm happy with pushing this into main.

KarthikSubbarao:
@yairgott  @allenss-amazon - I updated my remote branch to have `main` as the base branch.

You can take a look when you get a chance

integration/test_non_vector.py
```diff
@@ -0,0 +1,145 @@
...
+    ['JSON.SET', 'jsonproduct:2', '$',
+            '{"category":"electronics","name":"Tablet","price":499.00,"rating":4.0,"desc":"Good"}'],
+    ['JSON.SET', 'jsonproduct:3', '$',
+            '{"category":"electronics","name":"Phone","price":299.00,"rating":3.8,"desc":"Ok"}'],
+    ['JSON.SET', 'jsonproduct:4', '$',
+            '{"category":"books","name":"Book","price":19.99,"rating":4.8,"desc":"Excellent"}']
+]
+numeric_query_on_json = [
+    "FT.SEARCH", "jsonproducts",
+    "@$.price:[300 2000] @$.rating:[4.4 +inf]"
```
allenss-amazon: Don't you have to use the alias here? being able to provide a path suggests that we've indexed 100% of the JSON, then what's the point of a Schema?

integration/test_non_vector.py
```diff
@@ -0,0 +1,145 @@
...
+expected_numeric_json_value = {
+    "category": "electronics",
+    "name": "Laptop",
+    "price": 999.99,
+    "rating": 4.5,
+    "desc": "Great"
+}
+numeric_tag_query_on_json = [
+    "FT.SEARCH", "jsonproducts",
+    "@$.category:{books} @$.price:[10 30] @$.rating:[4.7 +inf]"
```
allenss-amazon: same here.

src/commands/ft_search_parser.cc
```diff
@@ -202,22 +202,25 @@ absl::Status ParseKNN(query::VectorSearchParameters &parameters,
...
-    auto max_ef_runtime_value = options::GetMaxEfRuntime().GetValue();
-    VMSDK_RETURN_IF_ERROR(
-        vmsdk::VerifyRange(parameters.ef.value(), 1, max_ef_runtime_value))
-        << "`EF_RUNTIME` must be a positive integer greater than 0 and cannot "
-           "exceed "
-        << max_ef_runtime_value << ".";
+  // Only verify the vector KNN parameters for vector based queries.
+  if (!parameters.attribute_alias.empty()) {
+    if (parameters.query.empty()) {
+      return absl::InvalidArgumentError("missing vector parameter");
```
allenss-amazon: "Invalid Query Syntax" ??

src/query/search.cc
```diff
@@ -327,6 +327,27 @@ absl::StatusOr<std::deque<indexes::Neighbor>> MaybeAddIndexedContent(
 
 absl::StatusOr<std::deque<indexes::Neighbor>> Search(
     const VectorSearchParameters &parameters, bool is_local_search) {
+  // Handle non vector queries first where attribute_alias is empty.
+  if (parameters.attribute_alias.empty()) {
```
allenss-amazon: Detecting vector/non-vector using this seems confusing. I'm fine with using it, but maybe we want a wrapper function with a more intuitive name.

integration/test_non_vector.py
KarthikSubbarao: We cannot use an alias in CME mode until this bug is fixed: https://github.com/valkey-io/valkey-search/issues/219

Otherwise, we get back the following error:
```
Index field `$.price` not exists
```

src/query/search.cc
KarthikSubbarao: A utility function would be nice. It needs to be used across the `commands`, `coordinator`, and `query` modules. So, I still need to find a common place for this

KarthikSubbarao: Adding it to `VectorSearchParameters` itself

```
  bool IsNonVectorQuery() const {
    return attribute_alias.empty();
  }
```

integration/test_non_vector.py
KarthikSubbarao: We cannot use an alias in CME mode until this bug is fixed: https://github.com/valkey-io/valkey-search/issues/219

Otherwise, we get back the following error:
```
Index field `$.price` not exists
```

src/query/search.cc
KarthikSubbarao: Addressed here: https://github.com/valkey-io/valkey-search/pull/218/commits/b9dd5f7b43e6b1464f6cf6e0605f479aa5176467

src/commands/ft_search_parser.cc
KarthikSubbarao: Addressed here: https://github.com/valkey-io/valkey-search/pull/218/commits/b9dd5f7b43e6b1464f6cf6e0605f479aa5176467

integration/test_non_vector.py
allenss-amazon: I think the PR for that fix is nearly ready to merge.

integration/test_non_vector.py
KarthikSubbarao: The fix was merged in. And I updated the tests to use the `alias` option

integration/test_non_vector.py
KarthikSubbarao: The fix was merged in. And I updated the tests to use the `alias` option

Issue #206, opened 2025-07-03
Unit Test Failure in ThreadPoolTest


allenss-amazon:
A recent CI run generated this failure:

See https://github.com/allenss-amazon/valkey-search/actions/runs/16057699611/job/45316098939

Attempts to recreate the failure haven't been successful. 

```
nning main() from /home/ubuntu/devl/valkey-search/submodules/.build-release-asan/.src/googletest/googletest/src/gtest_main.cc
[==========] Running 19 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 3 tests from ThreadPoolTest
[ RUN      ] ThreadPoolTest.StartAndJoin
[       OK ] ThreadPoolTest.StartAndJoin (8 ms)
[ RUN      ] ThreadPoolTest.priority
/workspace/vmsdk/testing/thread_pool_test.cc:269: Failure
Expected equality of these values:
  pending_run_high_priority
    Which is: 1
  0

/workspace/vmsdk/testing/thread_pool_test.cc:269: Failure
Expected equality of these values:
  pending_run_high_priority
    Which is: 1
  0

/workspace/vmsdk/testing/thread_pool_test.cc:269: Failure
Expected equality of these values:
  pending_run_high_priority
    Which is: 1
  0

/workspace/vmsdk/testing/thread_pool_test.cc:269: Failure
Expected equality of these values:
  pending_run_high_priority
    Which is: 1
  0

/workspace/vmsdk/testing/thread_pool_test.cc:269: Failure
Expected equality of these values:
  pending_run_high_priority
    Which is: 1
  0

/workspace/vmsdk/testing/thread_pool_test.cc:269: Failure
Expected equality of these values:
  pending_run_high_priority
    Which is: 1
  0

/workspace/vmsdk/testing/thread_pool_test.cc:269: Failure
Expected equality of these values:
  pending_run_high_priority
    Which is: 1
  0

/workspace/vmsdk/testing/thread_pool_test.cc:282: Failure
Expected equality of these values:
  pending_run_low_priority
    Which is: 3
  tasks
    Which is: 10

[  FAILED  ] ThreadPoolTest.priority (3 ms)
[ RUN      ] ThreadPoolTest.DynamicSizing
[       OK ] ThreadPoolTest.DynamicSizing (1009 ms)
[----------] 3 tests from ThreadPoolTest (1022 ms total)
```

allenss-amazon:
[...]

PR #221, opened 2025-07-14
Cripple broken unit test

allenss-amazon:
Disable a unit test that's been shown to be flaky. See #206


allenss-amazon:
this also cripples the unit test for #225.

PR #223, opened 2025-07-14
Merge updates from main to fulltext as of 7/14/2025

allenss-amazon:


allenss-amazon:
This found an unrelated flaky unit test, which is being fixed in PR #221. Once that's merged we'll redo this one.

Issue #183, opened 2025-06-17
Augment FT.INFO command to provide whole-cluster index information


allenss-amazon:
Today the FT.INFO command is nodal. This is a poor CX.

I propose that the FT.INFO command take an additional keyword argument to indicate whether the information is per-node or whole cluster. In the whole cluster case, gRPC messages are broadcast to primaries/replicas according to the same rules as FT.SEARCH and FT.AGGREGATE. The results are integrated together to present a single response.

allenss-amazon:
Reading the description of #112. This command should be the vehicle to implement the proposed FT.INDEXSTATUS. In other words, don't use a new command, use the proposed machinery of #112 but deliver the synthesized result in a field of the FT.INFO command result.

Thus to summarize. the proposed new syntax is:

```
FT.INFO <index> [LOCAL | GLOBAL]
```

The default for LOCAL vs GLOBAL can be specified in a configuration variable.

When decorated with LOCAL. This command uses the current behavior, i.e., produces statistics solely for the current node -- the current implementation.

When decorated with GLOBAL, the command gets broadcast to all shards -- using the routing rules that are described in #139. The command is decorated with the checksum of the current index and a checksum of the target slot map, allowing the command to detect the various consistency/backfill status of the index -- see #112 for details. This status gets reported as a new field in the FT.INFO results "STATUS". If the STATUS is not in the BACKFILLING or READY, then no aggregated statistics are reported, only that the index is unknown or in an inconsistent state.

PR #224, opened 2025-07-14
Create Info Fanout Proto Definition & Implement Server Info Index Partition Function

boda26:
1. Create info fanout proto definition
2. Implement server info index partition function
3. Create a template for get search target function


src/commands/ft_info.cc
```diff
@@ -29,6 +35,34 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
+  if (argc > 2) {  // Only parse scope if there's a third argument
+    itr.Next();
+    VMSDK_ASSIGN_OR_RETURN(auto scope_arg, itr.Get());
+    auto scope = vmsdk::ToStringView(scope_arg);
+
+    // default parameter is LOCAL; GLOBAL must be specified to retrieve global info
+    if (absl::EqualsIgnoreCase(scope, "GLOBAL")) {
+      is_global = true;
+      // std::cout << "==========Using Global Scope==========" << std::endl;
+      VMSDK_LOG(NOTICE, ctx) << "==========Using Global Scope==========";
```
allenss-amazon, line 62: Let's demote this to DEBUG.

src/commands/ft_info.cc
```diff
@@ -29,6 +35,34 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
+      
+      // Test GetInfoTargetsForFanout
+      auto targets = query::info_fanout::GetInfoTargetsForFanout(ctx);
+      // std::cout << "Found " << targets.size() << " fanout targets:" << std::endl;
+      VMSDK_LOG(NOTICE, ctx) << "Found " << targets.size() << " fanout targets:";
```
allenss-amazon, line 66: DEBUG

src/commands/ft_info.cc
```diff
@@ -29,6 +35,34 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
+      for (const auto& target : targets) {
+        // std::cout << "  Target type: " << (target.type == query::fanout::FanoutSearchTarget::Type::kLocal ? "LOCAL" : "REMOTE")
+        //           << ", address: " << target.address << std::endl;
+        VMSDK_LOG(NOTICE, ctx)
```
allenss-amazon, line 69: DEBUG

src/coordinator/coordinator.proto
```diff
@@ -133,3 +137,21 @@ message GetGlobalMetadataRequest {}
...
+
+message InfoIndexPartitionRequest {
+  string index_name = 1;
+}
+
+message InfoIndexPartitionResponse {
+  bool exists = 1;
+  string index_name = 2;
+  uint64 num_docs = 3;
+  uint64 num_records = 4;
```
allenss-amazon, line 149: What's the difference between num_docs and num_records?

src/coordinator/server.cc
```diff
@@ -164,6 +165,45 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
 
+grpc::ServerUnaryReactor* Service::InfoIndexPartition(
+    grpc::CallbackServerContext* context,
+    const InfoIndexPartitionRequest* request,
+    InfoIndexPartitionResponse* response) {
+  GRPCSuspensionGuard guard(GRPCSuspender::Instance());
+  auto latency_sample = SAMPLE_EVERY_N(100);
+  grpc::ServerUnaryReactor* reactor = context->DefaultReactor();
+
+  vmsdk::RunByMain(
```
allenss-amazon: This is a deadlock/timeout issue, is it possible to answer this query without involving the main thread? This needs some thought -- let's dicuss.

src/query/fanout.cc
```diff
@@ -219,54 +221,70 @@ absl::Status PerformSearchFanoutAsync(
...
-              absl::StrCat(ip, ":", coordinator::GetCoordinatorPort(port))});
-    }
-  }
-  absl::BitGen gen;
-  for (const auto &[shard_id, targets] : shard_id_to_target) {
-    size_t index = absl::Uniform(gen, 0u, targets.size());
-    selected_targets.push_back(targets.at(index));
-  }
-  return selected_targets;
+  std::cout << "Using get target template" << std::endl;
```
allenss-amazon: remove.

src/query/fanout_template.h
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }
+    
+    absl::BitGen gen;
+    for (const auto &[shard_id, targets] : shard_id_to_target) {
+      size_t index = absl::Uniform(gen, 0u, targets.size());
+      selected_targets.push_back(targets.at(index));
+    }
+    return selected_targets;
+  }
```
allenss-amazon, line 110: This seems fairly inefficient, we're creating Target objects for all possible targets and then discarding all but the selected ones. Can't this be inverted? 

On a different point, seems like this is the ideal place to control selection between primary/replicas, right? Can't this routine be parameterized to control for that?

Also, it seems like we can plan for a future optimization. This code starts by fetching the cluster-map and then crawling through it to create a shard-oriented map. Let's abstract that part into a separate function on the expectation that we'll be able to cache this data structure which will only rarely change.

testing/coordinator/info_converter_test.cc
```diff
@@ -0,0 +1,15 @@
...
+namespace valkey_search::coordinator {
+
+TEST(InfoConverterTest, CreateInfoIndexRequest) {
+  std::string index_name = "test_index";
+  
+  auto request = CreateInfoIndexPartitionRequest(index_name);
+  
+  ASSERT_NE(request, nullptr);
+  EXPECT_EQ(request->index_name(), "test_index");
+}
```
allenss-amazon: This seems purpose-less.

src/coordinator/coordinator.proto
boda26: I created these fields by referring to IndexSchema::RespondWithInfo() function in index_schema.cc, which defines the info gathered by FT.INFO command. Num_docs represents the count of unique docs that are indexed by the search schema, and num_records is the total count of __index entries__ across all indexes in the schema.

src/coordinator/server.cc
boda26: I referred to the SearchIndexPartition() defined for the search grpc server function, which also utilized main thread. Would the search also have a deadlock/timeout issue, or it is ok for search to use main thread but not for info?

testing/coordinator/info_converter_test.cc
boda26: this is just a very simple test for making the grpc request, i can remove it.

src/query/fanout_template.h
boda26: Yes, you are right. We can make it more efficient by grouping and selecting one node from each shard first, and then create the Target object. The second advice (parameterized control selection between primary/replica) is also a good idea. Since this function works now, can I optimize it after I finish the development of FT.INFO fanout?

src/coordinator/coordinator.proto
allenss-amazon: So, does that mean that num_records ~= # of fields * # of docs?

src/coordinator/server.cc
allenss-amazon: Search has the same problem. Search will get fixed. 

First rule of holes: quit digging. So, let's not dig this mainthread hole any deeper if we can avoid it.

allenss-amazon: If it's too difficult, there are other temporary workarounds for this problem. But if we can move it all off of the mainthread, then this problem is permanently fixed.

src/query/fanout_template.h
allenss-amazon: Doing it later is ok, if you feel like it's not going to cascade the amount of work.

The parameterized selection is something that's considered a bug in the current code and needs fixing. We might think about elevating this PR into the current VSS release cycle.

allenss-amazon: If we got integrated FT.INFO into the current release cycle that would be good also.

src/coordinator/coordinator.proto
boda26: Based on my understanding yes. These parameters are defined in the info schema and I directly used them in my index grpc request.

src/query/fanout_template.h
boda26: I will fix the getTarget function later after I finish FT.INFO fanout.

src/coordinator/server.cc
```diff
@@ -164,6 +165,64 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
+    response->set_exists(false);
+    response->set_index_name(index_name);
+    response->set_error(status_or_schema.status().ToString());
+    reactor->Finish(grpc::Status::OK);
+    return reactor;
+  }
+
+  // Keep the shared_ptr alive for the entire operation
+  auto schema = std::move(status_or_schema.value());
+  if (!schema) {
```
allenss-amazon: Is this possible? Seems like the SchemaManager shouldn't ever return a "good" nullptr.

src/coordinator/server.cc
```diff
@@ -164,6 +165,64 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
+    response->set_mutation_queue_size(schema->GetMutationQueueSize());
+    response->set_recent_mutations_queue_delay(schema->GetRecentMutationsQueueDelay());
+    response->set_state(schema->GetStateForInfoNoMain());
+  } catch (const std::exception& e) {
+    response->set_exists(false);
+    response->set_index_name(index_name);
+    response->set_error(std::string("Error collecting schema info: ") + e.what());
+  }
+
+  reactor->Finish(grpc::Status::OK);
```
allenss-amazon: Is it correct to still hold the Index guard during the gRPC call?

.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
     "clangd.path": "/usr/bin/clangd",
     "C_Cpp.codeAnalysis.runAutomatically": true,
     "editor.formatOnSave": false,
-    "C_Cpp.intelliSenseEngine": "disabled",
```
yairgott, line 8: Why do you want to change this? Today vscode uses clang-tidy for indexing which is considered more advanced.

src/commands/ft_info.cc
```diff
@@ -6,16 +6,21 @@
...
+#include "src/query/info_fanout.h"
 #include "vmsdk/src/command_parser.h"
 #include "vmsdk/src/status/status_macros.h"
 #include "vmsdk/src/type_conversions.h"
 #include "vmsdk/src/utils.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
+#include "vmsdk/src/log.h"
+
+
```
yairgott: nit: redundant newline

src/commands/ft_info.cc
```diff
@@ -29,6 +34,32 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
 
+  // Parse optional LOCAL/GLOBAL parameter
+  bool is_global = false;
+  if (argc > 2) {  // Only parse scope if there's a third argument
+    itr.Next();
+    VMSDK_ASSIGN_OR_RETURN(auto scope_arg, itr.Get());
+    auto scope = vmsdk::ToStringView(scope_arg);
+
+    // default parameter is LOCAL; GLOBAL must be specified to retrieve global info
+    if (absl::EqualsIgnoreCase(scope, "GLOBAL")) {
```
yairgott: Return error if the scope not match `global` or `local`.

src/commands/ft_info.cc
```diff
@@ -29,6 +34,32 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
+      is_global = true;
```
yairgott, line 60: It seems that you haven't publish yet the relevant code which uses the variables `is_global` and `targets`

src/coordinator/coordinator.proto
yairgott: > So, does that mean that num_records ~= # of fields * # of docs?

Yes, that is correct.

src/coordinator/server.cc
yairgott: +1, An error is returned if not found

src/coordinator/server.cc
```diff
@@ -164,6 +165,64 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
+    response->set_hash_indexing_failures(schema->GetHashIndexingFailures());
+    response->set_backfill_scanned_count(schema->GetBackfillScannedCount());
+    response->set_backfill_db_size(schema->GetBackfillDbSize());
+    response->set_backfill_inqueue_tasks(schema->GetBackfillInqueueTasks());
+    response->set_backfill_complete_percent(schema->GetBackfillPercentNoMain());
+    response->set_backfill_in_progress(schema->IsBackfillInProgressNoMain());
+    response->set_mutation_queue_size(schema->GetMutationQueueSize());
+    response->set_recent_mutations_queue_delay(schema->GetRecentMutationsQueueDelay());
+    response->set_state(schema->GetStateForInfoNoMain());
+  } catch (const std::exception& e) {
```
yairgott: why may an exception be thrown?

src/coordinator/server.cc
boda26: You are right, the lock can be released earlier after finishing copying the data, so releasing it before grpc call can be more efficient.

src/coordinator/server.cc
```diff
@@ -164,6 +165,64 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
+  if (!schema) {
+    response->set_exists(false);
+    response->set_index_name(index_name);
+    response->set_error("Schema is null");
+    reactor->Finish(grpc::Status::OK);
+    return reactor;
+  }
+
+  // Acquire read guard and ensure it stays alive
+  auto read_guard = schema->AcquireReadGuard();
```
yairgott: The code can be simplified a lot by using `vmsdk::RunByMain`, to eliminate most, if not all,  synchronization needs. Since this info command is not on the critical path, the overhead shouldn't be a concern.
Additionally, it would be cleaner if index_schema exposed a single interface that returns a statistics struct.

src/commands/ft_info.cc
boda26: yes this PR is a part, I am still developing the rest

yairgott:
Missing unittests.

src/coordinator/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set(SRCS_SERVER ${CMAKE_CURRENT_LIST_DIR}/server.cc
 
 valkey_search_add_static_library(server "${SRCS_SERVER}")
 target_include_directories(server PUBLIC ${CMAKE_CURRENT_LIST_DIR})
-target_include_directories(server PRIVATE ${CMAKE_BINARY_DIR})
+target_include_directories(server PRIVATE ${CMAKE_BINARY_DIR}/src/coordinator)
```
allenss-amazon: ? Why is this change needed?

src/coordinator/CMakeLists.txt
```diff
@@ -99,6 +99,13 @@ target_link_libraries(search_converter PUBLIC managed_pointers)
 target_link_libraries(search_converter PUBLIC type_conversions)
 target_link_libraries(search_converter PUBLIC status_macros)
 
+set(SRCS_INFO_CONVERTER ${CMAKE_CURRENT_LIST_DIR}/info_converter.cc
+                        ${CMAKE_CURRENT_LIST_DIR}/info_converter.h)
+
+valkey_search_add_static_library(info_converter "${SRCS_INFO_CONVERTER}")
+target_include_directories(info_converter PUBLIC ${CMAKE_CURRENT_LIST_DIR})
+target_link_libraries(info_converter PUBLIC coordinator_cc_proto)
+
```
allenss-amazon: Let's just add these to an existing library and end the madness of needless library proliferation.

src/index_schema.h
```diff
@@ -150,6 +150,40 @@ class IndexSchema : public KeyspaceEventSubscription,
...
+  }
+
+  // how many skip‐on‐add (hash_indexing_failures)
+  uint64_t GetHashIndexingFailures() const {
+    return stats_.subscription_add.skipped_cnt;
+  }
+
+  // current size of the mutation queue
+  uint64_t GetMutationQueueSize() const {
+    absl::MutexLock lock(&stats_.mutex_);
```
allenss-amazon: Why is this one locked but the others aren't?

src/query/CMakeLists.txt
```diff
@@ -94,6 +94,18 @@ target_link_libraries(response_generator PUBLIC type_conversions)
...
+
+valkey_search_add_static_library(info_fanout "${SRCS_INFO_FANOUT}")
+target_include_directories(info_fanout PUBLIC ${CMAKE_CURRENT_LIST_DIR})
+target_link_libraries(info_fanout PUBLIC fanout)
+target_link_libraries(info_fanout PUBLIC client_pool)
+target_link_libraries(info_fanout PUBLIC managed_pointers)
+target_link_libraries(info_fanout PUBLIC thread_pool)
+target_link_libraries(info_fanout PUBLIC status_macros)
+target_link_libraries(info_fanout PUBLIC valkey_module)
+
```
allenss-amazon: again, let's add to an existing library

src/index_schema.h
boda26: mutation_queue_size variable is guarded by the mutex lock, so the lock should be applied when accessing its value, while other variables are not.

.vscode/settings.json
```diff
@@ -17,5 +17,8 @@
         "--query-driver=**"
     ],
     "clangd.inactiveRegions.useBackgroundHighlight": true,
-    "clangd.enableCodeCompletion": true
+    "clangd.enableCodeCompletion": true,
```
allenss-amazon, line 20: This looks like accidental cruft, maybe we should remove it.

src/index_schema.cc
```diff
@@ -366,9 +366,11 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
   }
   if (added) {
     // Track key modifications by data type
-    if (attribute_data_type_->ToProto() == data_model::ATTRIBUTE_DATA_TYPE_HASH) {
+    if (attribute_data_type_->ToProto() ==
```
allenss-amazon, line 369: let's use a switch statement here.

src/index_schema.cc
```diff
@@ -1077,4 +1081,41 @@ void IndexSchema::VectorExternalizer(const InternedStringPtr &key,
...
+  absl::MutexLock lock(&mutex_);
+  return InfoIndexPartitionData{
+      .num_docs = document_cnt,
+      .hash_indexing_failures = subscription_add.skipped_cnt,
+      .backfill_inqueue_tasks = backfill_inqueue_tasks,
+      .mutation_queue_size = mutation_queue_size_,
+      .recent_mutations_queue_delay =
+          mutation_queue_size_ > 0
+              ? static_cast<uint64_t>(mutations_queue_delay_ / absl::Seconds(1))
+              : 0};
```
allenss-amazon, line 1094: This is being quantized to 1 second. That doesn't seem very useful. I think this should get moved to milliseconds. If that's breaking backward compatibility we can introduce an additional return value.

.vscode/settings.json
boda26: I am not sure what is the problem here, I never modified this line locally

src/index_schema.cc
boda26: here I am using the same calculation logic as
```  ValkeyModule_ReplyWithCString(
      ctx, absl::StrFormat("%lu sec", (stats_.mutation_queue_size_ > 0
                                           ? stats_.mutations_queue_delay_ /
                                                 absl::Seconds(1)
                                           : 0))
```
in ReplyWithInfo(). Should we modify this part in old code as well?

BCathcart:
This PR partially addresses #183

src/query/fanout_template.h
```diff
@@ -0,0 +1,113 @@
...
+    kLocal,
+    kRemote,
+  };
+  Type type;
+  // Empty string if type is kLocal.
+  std::string address;
+
+  bool operator==(const FanoutSearchTarget& other) const {
+    return type == other.type && address == other.address;
+  }
```
eifrah-aws, line 31: Probably:

```c++
bool operator==(const FanoutSearchTarget&) const = default;
```

is enough

src/query/fanout_template.h
```diff
@@ -0,0 +1,113 @@
...
+namespace valkey_search::query::fanout {
+
+struct FanoutSearchTarget {
+  enum Type {
+    kLocal,
+    kRemote,
+  };
+  Type type;
+  // Empty string if type is kLocal.
+  std::string address;
```
eifrah-aws, line 27: Wrap this in `std::optional`

src/query/fanout_template.h
```diff
@@ -0,0 +1,113 @@
...
+#include "absl/strings/str_cat.h"
+#include "src/coordinator/util.h"
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+#include "vmsdk/src/log.h"
+#include "vmsdk/src/managed_pointers.h"
+
+namespace valkey_search::query::fanout {
+
+struct FanoutSearchTarget {
+  enum Type {
```
eifrah-aws, line 21: Prefer `enum class`

src/query/fanout_template.h
```diff
@@ -0,0 +1,113 @@
...
+  }
+};
+
+// Template class for fanout operations across cluster nodes
+class FanoutTemplate {
+ public:
+  template<typename TargetType>
+  static std::vector<TargetType> GetTargets(
+      ValkeyModuleCtx *ctx,
+      std::function<TargetType()> create_local_target,
```
eifrah-aws, line 47: Use: `absl::AnyInvocable`

src/query/fanout_template.h
```diff
@@ -0,0 +1,113 @@
...
+      std::function<TargetType(const std::string&)> create_remote_target) {
+    size_t num_nodes;
+    auto nodes = vmsdk::MakeUniqueValkeyClusterNodesList(ctx, &num_nodes);
+    
+    // Store only node indices for random selection
+    absl::flat_hash_map<std::string, std::vector<size_t>> shard_id_to_node_indices;
+    std::vector<TargetType> selected_targets;
+    
+    for (size_t i = 0; i < num_nodes; ++i) {
+      std::string node_id(nodes.get()[i], VALKEYMODULE_NODE_ID_LEN);
```
eifrah-aws, line 57: Here and everywhere: prefer `{..}` initialization

src/coordinator/CMakeLists.txt
```diff
@@ -1,111 +1,108 @@
...
-target_link_libraries(server PUBLIC search)
-target_link_libraries(server PUBLIC latency_sampler)
-target_link_libraries(server PUBLIC log)
-target_link_libraries(server PUBLIC managed_pointers)
-target_link_libraries(server PUBLIC thread_pool)
-target_link_libraries(server PUBLIC type_conversions)
-target_link_libraries(server PUBLIC utils)
-target_link_libraries(server PUBLIC valkey_module)
-target_link_libraries(server PUBLIC coordinator_grpc_proto)
+# Single base library that consolidates ALL common dependencies
```
eifrah-aws, line 10: We need to merge all these into a single static library, I fail to see the point of having so many small static libraries ...

PR #235, opened 2025-07-15
valkey-test-framework: support creating cluster

eifrah-aws:
With this change:

- Added support for CME in valkey-test-framework
- Each test creates its own test folder under `/tmp/valkey-search-clusters/<TEST_NAME>/<PORT>`

### DIFF
--- a/integration/valkey_search_test_case.py
+++ b/integration/valkey_search_test_case.py
@@ -8,6 +8,7 @@ from valkey.connection import Connection
 import random
 import string
 import logging
+import shutil
 
 
 class ValkeySearchTestCaseBase(ValkeyTestCase):
@@ -60,17 +61,11 @@ class ValkeySearchTestCaseBase(ValkeyTestCase):
 
 
 class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
-    def dict_to_space_delimited(self, d):
-        result = []
-        for key, value in d.items():
-            result.append(f"--{key}")
-            result.append(str(value))
-        return " ".join(result)
-
-    def _start_server(self, port):
+    def _start_server(self, port, test_name):
         server_path = os.getenv("VALKEY_SERVER_PATH")
-        pid = os.getpid()
-        testdir = f"/tmp/valkey-search-clusters/{pid}/shard-{port}"
+        testdir_base = f"/tmp/valkey-search-clusters/{test_name}"
+        testdir = f"{testdir_base}/{port}"
+
         os.makedirs(testdir, exist_ok=True)
         curdir = os.getcwd()
         os.chdir(testdir)
@@ -81,7 +76,9 @@ class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
             f"dir {testdir}",
             "cluster-enabled yes",
         ]
-        with open(f"{testdir}/valkey.conf", "w+") as f:
+
+        conf_file = f"{testdir}/valkey.conf"
+        with open(conf_file, "w+") as f:
             for line in lines:
                 f.write(f"{line}\n")
             f.write("\n")
@@ -92,7 +89,7 @@ class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
             server_path=server_path,
             args="",
             port=port,
-            conf_file=f"{testdir}/valkey.conf",
+            conf_file=conf_file,
         )
         os.chdir(curdir)
         logfile = f"{testdir}/logfile_{port}"
@@ -116,9 +113,6 @@ class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
         for node_to_meet in range(0, primaries_count):
             if node_to_meet == node_idx:
                 continue
-            meet_command = (
-                f"MEET 127.0.0.1 {self.get_primary_port(node_to_meet)}"
-            )
             try:
                 client.execute_command(
                     " ".join(
@@ -143,7 +137,7 @@ class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
                 )
 
     @pytest.fixture(autouse=True)
-    def setup_test(self, setup):
+    def setup_test(self, request):
         self.servers = []
         ports = [
             self.get_bind_port(),
@@ -151,8 +145,14 @@ class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
             self.get_bind_port(),
         ]
 
+        testdir_base = f"/tmp/valkey-search-clusters/{request.node.name}"
+        if os.path.exists(testdir_base):
+            shutil.rmtree(testdir_base)
+
         for port in ports:
-            server, client, logfile = self._start_server(port)
+            server, client, logfile = self._start_server(
+                port, request.node.name
+            )
             self.servers.append(
                 {
                     "server": server,
@@ -174,12 +174,12 @@ class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
             self.cluster_meet(node_idx, 3)
 
         # Wait for the cluster to be up
-        logging.info("Waiting for cluster to change state...")
         for server in self.servers:
+            logging.info(f"Waiting for cluster to change state...{logfile}")
             self.wait_for_logfile(
                 server["logfile"], "Cluster state changed: ok"
             )
-        logging.info("Cluster is up and running!")
+            logging.info("Cluster is up and running!")
 
     def get_primary(self, index):
         return self.servers[index]["server"]
--- a/integration/valkey_search_test_case.py
+++ b/integration/valkey_search_test_case.py
@@ -2,6 +2,9 @@ import os
 import pytest
 from valkeytestframework.valkey_test_case import ValkeyTestCase
 from valkey import ResponseError
+from valkey.cluster import ValkeyCluster, ClusterNode
+from valkey.client import Valkey
+from valkey.connection import Connection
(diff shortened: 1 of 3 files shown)


integration/valkey_search_test_case.py
```diff
@@ -54,3 +57,157 @@ def parse_valkey_info(self, section):
                 key, value = line.split(":", 1)
                 stats_dict[key.strip()] = value.strip()
         return stats_dict
+
+
+class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
```
KarthikSubbarao, line 63: Would it be possible to have the common cluster setup code in the valkey-test-framework package?

I understand that the CME specific aspect for the valkey-search module includes the coordinator argument. But besides this, a lot of this is re-usable for other modules that need CME support in testing

integration/valkey_search_test_case.py
```diff
@@ -54,3 +57,157 @@ def parse_valkey_info(self, section):
...
+                            "CLUSTER",
+                            "MEET",
+                            "127.0.0.1",
+                            f"{self.get_primary_port(node_to_meet)}",
+                        ]
+                    )
+                )
+
+    @pytest.fixture(autouse=True)
+    def setup_test(self, setup):
```
KarthikSubbarao: Instead of using a fixture with predefined configurations (in this case, singe shard, one primary, no replicas), it would be good if this interface allows configurability in terms of number of shards and number of replicas per shard

integration/valkey_search_test_case.py
eifrah-aws: I will move it after we merge in a separate PR so we can unblock our work

integration/valkey_search_test_case.py
eifrah-aws: Yes, this is on the TODO list

integration/valkey_search_test_case.py
```diff
@@ -54,3 +58,156 @@ def parse_valkey_info(self, section):
...
+        client: Valkey = self.client_for_primary(node_idx)
+
+        for node_to_meet in range(0, primaries_count):
+            if node_to_meet == node_idx:
+                continue
+            try:
+                client.execute_command(
+                    " ".join(
+                        [
+                            "CLUSTERADMIN",
```
KarthikSubbarao: There is a typo here likely 

```suggestion
                            "CLUSTER",
```

integration/valkey_search_test_case.py
```diff
@@ -54,3 +58,156 @@ def parse_valkey_info(self, section):
...
+                client.execute_command(
+                    " ".join(
+                        [
+                            "CLUSTER",
+                            "MEET",
+                            "127.0.0.1",
+                            f"{self.get_primary_port(node_to_meet)}",
+                        ]
+                    )
+                )
```
KarthikSubbarao: If we use `CLUSTER MEET` the first time, it should succeed. This would not be needed

integration/valkey_search_test_case.py
```diff
@@ -54,3 +58,156 @@ def parse_valkey_info(self, section):
...
+
+    def _split_range_pairs(self, start, end, n):
+        points = [start + i * (end - start) // n for i in range(n + 1)]
+        return list(zip(points[:-1], points[1:]))
+
+    def add_slots(self, node_idx, first_slot, last_slot):
+        client: Valkey = self.client_for_primary(node_idx)
+        slots_to_add = list(range(int(first_slot), int(last_slot)))
+        try:
+            client.execute_command("CLUSTERADMIN ADDSLOTS", *slots_to_add)
```
KarthikSubbarao: Do you mean CLUSTER ADDSLOTS?

integration/valkey_search_test_case.py
```diff
@@ -54,3 +58,156 @@ def parse_valkey_info(self, section):
...
+        os.makedirs(testdir, exist_ok=True)
+        curdir = os.getcwd()
+        os.chdir(testdir)
+        lines = [
+            "enable-debug-command yes",
+            f"loadmodule {os.getenv('JSON_MODULE_PATH')}",
+            f"loadmodule {os.getenv('MODULE_PATH')} --use-coordinator",
+            f"dir {testdir}",
+            "cluster-enabled yes",
+        ]
```
KarthikSubbarao: Nit - Right now, we are generating a conf file and then using that to startup. I understand this is because the valkey-test-framework needs some updates to support the `--loadmodule` directive. Using a conf file is ok, but we should be able to easily update the utility functions in valkey-test-framework to handle the requirements (loading multiple modules and supporting module start up args, e.g. coordinator) and avoid generating a conf file just to bypass this issue

integration/valkey_search_test_case.py
```diff
@@ -54,3 +58,156 @@ def parse_valkey_info(self, section):
                 key, value = line.split(":", 1)
                 stats_dict[key.strip()] = value.strip()
         return stats_dict
+
+
+class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
+    def _start_server(self, port, test_name):
+        server_path = os.getenv("VALKEY_SERVER_PATH")
+        testdir_base = f"/tmp/valkey-search-clusters/{test_name}"
+        testdir = f"{testdir_base}/{port}"
```
KarthikSubbarao: Question: Why do we need  a separate directory for each server in the cluster?

If it is for the `node.conf` files, we can change the `cluster-config-file` to use a unique name. Example: concatenate the bind_ip and the server port.

integration/valkey_search_test_case.py
KarthikSubbarao: Ack. Thank you. Some of the issues seen so far are during fanned out requests to remote nodes - So this testing will be useful :D

integration/valkey_search_test_case.py
```diff
@@ -54,3 +58,156 @@ def parse_valkey_info(self, section):
                 key, value = line.split(":", 1)
                 stats_dict[key.strip()] = value.strip()
         return stats_dict
+
+
+class ValkeySearchClusterTestCase(ValkeySearchTestCaseBase):
+    def _start_server(self, port, test_name):
```
KarthikSubbarao, line 66: Would it be possible to pass start up arguments to the existing startup method in the `ValkeyServerHandle` class [here](https://github.com/valkey-io/valkey-test-framework/blob/unstable/src/valkey_test_case.py#L64C7-L64C25) ? It would be good to try and re-use existing methods

integration/valkey_search_test_case.py
eifrah-aws: We don't. We can do that as well. Matter of "taste" I guess. I have no strong opinion on this, will change it.

integration/valkey_search_test_case.py
eifrah-aws: Your comment will fix the issues that we know, **today**. Using configuration file basically gives the user more flexibility to basically do whatever he needs to do - without changing the framework to add more APIs. 

If we are talking about changes to the framework, I noticed that there are some hidden hard-coded stuff in the framework that should be removed (like: forcing log file format, forcing rdb and other hard coded command line arguments that we should move outside of the framework and let the user control them)

integration/valkey_search_test_case.py
eifrah-aws: I used this internally for testing, will remove this.

integration/valkey_search_test_case.py
eifrah-aws: Same as previous comment.

integration/valkey_search_test_case.py
eifrah-aws: This way I don't rely on the cluster bus and the cluster is ready (state: ok) faster.

integration/valkey_search_test_case.py
eifrah-aws: Fixed

integration/valkey_search_test_case.py
```diff
@@ -54,3 +58,142 @@ def parse_valkey_info(self, section):
...
+            server, client, logfile = self._start_server(
+                port, request.node.name
+            )
+            self.servers.append(
+                {
+                    "server": server,
+                    "client": client,
+                    "port": port,
+                    "logfile": logfile,
+                }
```
KarthikSubbarao, line 148: If it is possible to apply my suggestion from the comment above and use the `ValkeyServerHandle` object, it addresses this issue since we can access all of these (server, client, port, logfile) from the existing `ValkeyServerHandle` object. We will not have to manually set these

integration/valkey_search_test_case.py
```diff
@@ -54,3 +58,142 @@ def parse_valkey_info(self, section):
...
+
+    def client_for_primary(self, index):
+        return self.servers[index]["client"]
+
+    def new_cluster_client(self):
+        """Return a cluster client"""
+        startup_nodes = []
+        for index in range(0, self.CLUSTER_SIZE):
+            startup_nodes.append(
+                ClusterNode("127.0.0.1", self.get_primary_port(index))
```
KarthikSubbarao, line 187: If we are using the `ValkeyServerHandle` object, we can use the bind_ip address from there instead of hardcoding here. The other CLUSTER MEET function can also use the bind_ip address if we do this

integration/valkey_search_test_case.py
eifrah-aws: It uses the internal method

allenss-amazon (approved):

Issue #219, opened 2025-07-11
FT.SEARCH does not work when using an alias for attributes in JSON/HASH documents in CME (coordinator enabled) during fanouts


KarthikSubbarao:
There appears to be a bug in valkey-search in FT.SEARCH in CME mode (search-coordinator enabled) with JSON & Hash documents when we use an alias for fields. This bug exists with hybrid vector queries today.

Index Creation using `AS` for defining the alias:
```
    v.execute_command('FT.CREATE', 'json_idx1', 'ON', 'JSON', 'PREFIX', '1', 'json:', 
                 'SCHEMA', '$.n1', 'AS', 'n1', 'NUMERIC',
                 '$.n2', 'AS', 'n2', 'NUMERIC',
                 '$.n3', 'AS', 'n3', 'NUMERIC',
                 '$.t1', 'AS', 't1', 'TAG',
                 '$.t2', 'AS', 't2', 'TAG',
                 '$.t3', 'AS', 't3', 'TAG',
                 'v1', 'VECTOR', 'HNSW', '6', 'DIM', '3', 'TYPE', 'FLOAT32', 'DISTANCE_METRIC', 'IP')
```

FT.SEARCH command using the alias `n1`:
```
cmd = ['FT.SEARCH', 'json_idx1', 
       '(@n1:[0 5])=>[KNN 3 @v1 $BLOB]',
       'PARAMS', '2', 'BLOB', query_vector.tobytes(),
       'DIALECT', '3', 'LIMIT', '0', '1000']
```

Incorrect Error Response:
```
Index field `$.n1` not exists
```

This occurs in CME mode when using the alias of fields of a JSON/Hash document. I dug into why this is happening and it looks like we are sending the identifier (not the alias) to the remote nodes. The remote nodes are trying to handle the fanout using the identifier and cannot find the field (but they are aware of the alias).
Currently, if customers/users change the index creation to not use any alias for JSON and Hash documents, the FT.SEARCH works fine.

yairgott:
Kudos on the finding! I'm working on a fix.

yairgott:
The issue was addressed as part of the following [PR](https://github.com/valkey-io/valkey-search/pull/243).

PR #243, opened 2025-07-18
Fix attribute alias handling in filters when running in CME mode

yairgott:
Addressing reported [issue](https://github.com/valkey-io/valkey-search/issues/219).

The bug was that we used to serializing the filter predicate attribute identifier rather than the attribute alias.

Added integration tests to cover the usage of attribute aliasing and filtering.


yairgott:
cc: @KarthikSubbarao

src/index_schema.cc
```diff
@@ -243,7 +243,8 @@ absl::StatusOr<std::string> IndexSchema::GetIdentifier(
     absl::string_view attribute_alias) const {
   auto itr = attributes_.find(std::string{attribute_alias});
   if (itr == attributes_.end()) {
-    return absl::NotFoundError("Index field not exists");
+    return absl::NotFoundError(
+        absl::StrCat("Index field `", attribute_alias, "` not exists"));
```
allenss-amazon: does not exist.

src/index_schema.cc
```diff
@@ -366,9 +367,11 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
   }
   if (added) {
     // Track key modifications by data type
-    if (attribute_data_type_->ToProto() == data_model::ATTRIBUTE_DATA_TYPE_HASH) {
+    if (attribute_data_type_->ToProto() ==
```
allenss-amazon: switch?

allenss-amazon (approved):

src/index_schema.cc
yairgott: Sorry, what do mean? FWIW, this is tidy formatter change...

allenss-amazon: Why not use a 'switch' than a cascading if/else ?

src/index_schema.cc
```diff
@@ -234,7 +234,7 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexSchema::GetIndex(
   auto itr = attributes_.find(std::string{attribute_alias});
   if (ABSL_PREDICT_FALSE(itr == attributes_.end())) {
     return absl::NotFoundError(
-        absl::StrCat("Index field `", attribute_alias, "` not exists"));
+        absl::StrCat("Index field `", attribute_alias, "` does not exists"));
```
allenss-amazon, line 237: "does not exist", not "does not exists".

allenss-amazon (commented):
Let's fix the typo and merge it.

PR #245, opened 2025-07-21
Posting Object Implementation

Aksha1812:
### Basic Posting Object 

The Postings object maps a single stemmed word to the list of all (Key, Field, Position) triples where that word appears in the corpus. For each entry in the inverted term index, there is an instance of this structure which is used to contain the key/field/position information for each word.
```
Postings
├── std::map<Key, PositionMap> key_to_positions_
│
├── Key="document1"
│   └── PositionMap: std::map<Position, FieldMask*>
│       ├── Position=5  → FieldMask[fields 0,2] = 0b00000101
│       ├── Position=12 → FieldMask[field 1]    = 0b00000010  
│       └── Position=18 → FieldMask[fields 0,1,3] = 0b00001011
│
├── Key="document2"
│   └── PositionMap: std::map<Position, FieldMask*>
│       ├── Position=0  → FieldMask[fields 0,1] = 0b00000011
│       └── Position=3  → FieldMask[field 2]    = 0b00000100

```

> with `save_positions` mode can be used to turn off saving positions, in that case we will just have key and field mask info with a dummy position.

> This object is NOT multi-thread safe, it's expected that the caller performs
locking for mutation operations.

> Suffix and prefix tree will have shared ownership of this object

#### FieldMask Adaptive Storage

```
Field Count | Storage Type    | Memory Usage | Max Fields
------------|----------------|--------------|------------
1           | EmptyFieldMask | 0 bytes      | 1
2-8         | uint8_t        | 1 byte       | 8
9-64        | uint64_t       | 8 bytes      | 64
```

> Can add support for fields > 64 based on requirement

#### Key and Position Iterators

```
Iterator Usage Flow
├── postings.GetKeyIterator()
│
├── key_iter.GetPositionIterator()
│
└── pos_iter.GetPosition()  |  pos_iter.GetFieldMask()
    └── Result: 2 (position 32 bit integer)   |   Result: 5 (binary: 0b00000101) → Fields 0 and 2 set
```
Object usage can be referred from the unit tests written.

Supersedes https://github.com/Aksha1812/valkey-search/pull/2. Initial commits have been discussed here. This PR targets the correct repository with an updated base.

### DIFF
--- a/src/index_schema.h
+++ b/src/index_schema.h
@@ -97,6 +97,14 @@ class IndexSchema : public KeyspaceEventSubscription,
   inline const std::string &GetName() const { return name_; }
   inline std::uint32_t GetDBNum() const { return db_num_; }
 
+  bool GetSavePositions() const { return save_positions_; }
+  size_t GetNumTextFields() const { return num_text_fields_; }
+  // TODO: Change this after query support is added for full text search
+  void SetTextConfiguration(bool save_positions, size_t num_text_fields) {
+    save_positions_ = save_positions;
+    num_text_fields_ = num_text_fields;
+  }
+
   void OnKeyspaceNotification(ValkeyModuleCtx *ctx, int type, const char *event,
                               ValkeyModuleString *key) override;
 
@@ -166,6 +174,8 @@ class IndexSchema : public KeyspaceEventSubscription,
   std::unique_ptr<AttributeDataType> attribute_data_type_;
   std::string name_;
   uint32_t db_num_{0};
+  bool save_positions_{true};
+  size_t num_text_fields_{3};
 
   vmsdk::ThreadPool *mutations_thread_pool_{nullptr};
   InternedStringMap<DocumentMutation> tracked_mutated_records_
--- a/src/index_schema.h
+++ b/src/index_schema.h
@@ -175,7 +175,7 @@ class IndexSchema : public KeyspaceEventSubscription,
   std::string name_;
   uint32_t db_num_{0};
   bool save_positions_{true};
-  size_t num_text_fields_{3};
+  size_t num_text_fields_{0};
 
   vmsdk::ThreadPool *mutations_thread_pool_{nullptr};
   InternedStringMap<DocumentMutation> tracked_mutated_records_
--- a/src/indexes/text/posting.cc
+++ b/src/indexes/text/posting.cc
@@ -458,3 +458,5 @@ uint64_t Postings::PositionIterator::GetFieldMask() const {
   if (impl->current == impl->end) return 0;
   return impl->current->second->AsUint64();
 }
+
+}  // namespace valkey_search::text
--- a/src/indexes/text/posting.cc
+++ b/src/indexes/text/posting.cc
@@ -168,16 +168,6 @@ uint64_t FieldMaskImpl<MaskType, MAX_FIELDS>::AsUint64() const {
   }
 }
 
-// Create deep copy of field mask
-template<typename MaskType, size_t MAX_FIELDS>
-std::unique_ptr<FieldMask> FieldMaskImpl<MaskType, MAX_FIELDS>::Clone() const {
-  auto clone = std::make_unique<FieldMaskImpl<MaskType, MAX_FIELDS>>(num_fields_);
-  if constexpr (!std::is_same_v<MaskType, EmptyFieldMask>) {
-    clone->mask_ = mask_;
-  }
-  return clone;
-}
-
 // Explicit template instantiations
 template class FieldMaskImpl<EmptyFieldMask, 1>;
 template class FieldMaskImpl<uint8_t, 8>;
--- a/src/indexes/text/posting.cc
+++ b/src/indexes/text/posting.cc
@@ -1,0 +1,412 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#include "src/indexes/text/posting.h"
+
+#include <cstdint>
+#include <map>
+#include <memory>
+#include <stdexcept>
+#include <type_traits>
+
+namespace valkey_search::text {
+
+// Field Mask Implementation
+  
+// Factory method to create optimal field mask based on field count
+std::unique_ptr<FieldMask> FieldMask::Create(size_t num_fields) {
+  if (num_fields == 0) {
+    throw std::invalid_argument("num_fields must be greater than 0");
+  }
+  
+  // Select most memory-efficient implementation
+  if (num_fields <= 1) {
+    return std::make_unique<SingleFieldMask>();  // bool (1 byte)
+  } else if (num_fields <= 8) {
+    return std::make_unique<ByteFieldMask>(num_fields);  // uint8_t (1 byte)
+  } else if (num_fields <= 64) {
+    return std::make_unique<Uint64FieldMask>(num_fields);  // uint64_t (8 bytes)
+  } else {
+    throw std::invalid_argument("Too many text fields (max 64 supported)");
+  }
+}
+
+// Initialize field mask with specified field count
+template<typename MaskType, size_t MAX_FIELDS>
+FieldMaskImpl<MaskType, MAX_FIELDS>::FieldMaskImpl(size_t num_fields) 
+    : mask_(MaskType{}), num_fields_(num_fields) {
+  if (num_fields > MAX_FIELDS) {
+    throw std::invalid_argument("Field count exceeds maximum for this mask type");
+  }
+}
+
+// Set a specific field bit to true
+template<typename MaskType, size_t MAX_FIELDS>
+void FieldMaskImpl<MaskType, MAX_FIELDS>::SetField(size_t field_index) {
+  if (field_index >= num_fields_) {
+    throw std::out_of_range("Field index out of range");
+  }
+  
+  if constexpr (std::is_same_v<MaskType, bool>) {
+    mask_ = true;
+  } else {
+    mask_ |= (MaskType(1) << field_index);
+  }
+}
+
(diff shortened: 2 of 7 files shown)


src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+  Impl& operator=(const Impl& other) {
+    if (this != &other) {
+      save_positions_ = other.save_positions_;
+      num_text_fields_ = other.num_text_fields_;
+      key_to_positions_.clear();
+      
+      for (const auto& [key, pos_map] : other.key_to_positions_) {
+        PositionMap cloned_pos_map;
+        for (const auto& [pos, field_mask] : pos_map) {
+          cloned_pos_map[pos] = field_mask->Clone();
```
allenss-amazon: I doubt we'd use these. I'd just mark it as not moveable/copyeable.

src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+Postings::Postings(const Postings& other) 
+    : impl_(std::make_unique<Impl>(*other.impl_)) {}
+
+// Safe replacement with deep copy
+Postings& Postings::operator=(const Postings& other) {
+  if (this != &other) {
+    impl_ = std::make_unique<Impl>(*other.impl_);
+  }
+  return *this;
+}
```
allenss-amazon: same, I don't think these will be needed.

src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+
+// Basic Postings Object Implementation
+
+// Position map type alias - maps position to optimized FieldMask objects
+using PositionMap = std::map<Position, std::unique_ptr<FieldMask>>;
+
+class Postings::Impl {
+public:
+  bool save_positions_;
+  size_t num_text_fields_;
```
allenss-amazon, line 182: Long term, these won't get stored in the Postings objects, rather they'll be in the IndexSchema object we can plan ahead for that in the interfaces.

src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+// KeyIterator implementations
+bool Postings::KeyIterator::IsValid() const {
+  if (!impl_data_) return false;
+  auto* impl = static_cast<KeyIteratorImpl*>(impl_data_);
+  return impl->current != impl->end;
+}
+
+void Postings::KeyIterator::NextKey() {
+  if (!impl_data_) return;
+  auto* impl = static_cast<KeyIteratorImpl*>(impl_data_);
```
allenss-amazon: why is a cast needed here?

src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+const Key& Postings::KeyIterator::GetKey() const {
+  if (!impl_data_) {
+    static Key dummy_key;
+    return dummy_key;
+  }
+  auto* impl = static_cast<KeyIteratorImpl*>(impl_data_);
+  if (impl->current == impl->end) {
+    static Key dummy_key;
+    return dummy_key;
+  }
```
allenss-amazon: I think you should assert here.

src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+  return impl->current->first;
+}
+
+Postings::PositionIterator Postings::KeyIterator::GetPositionIterator() const {
+  PositionIterator pos_iterator;
+  if (!impl_data_) {
+    pos_iterator.impl_data_ = nullptr;
```
allenss-amazon: When would this be valid?

src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+    return pos_iterator;
+  }
+  
+  auto* impl = static_cast<KeyIteratorImpl*>(impl_data_);
```
allenss-amazon: same question here.

src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+  
+  // Return true if we landed on exact position match
+  return (impl->current != impl->end && impl->current->first == position);
+}
+
+const Position& Postings::PositionIterator::GetPosition() const {
+  if (!impl_data_) {
+    static Position dummy_position = 0;
+    return dummy_position;
+  }
```
allenss-amazon: assert?

src/indexes/text/posting.cc
```diff
@@ -0,0 +1,462 @@
...
+  auto* impl = static_cast<PositionIteratorImpl*>(impl_data_);
+  if (impl->current == impl->end) {
+    static Position dummy_position = 0;
+    return dummy_position;
+  }
```
allenss-amazon: assert?

src/indexes/text/posting.h
```diff
@@ -45,12 +63,21 @@ struct Postings {
   // The "num_text_fields" entry identifies how many bits of the field-mask are required
   // and is used to select the representation.
   Postings(bool save_positions, size_t num_text_fields);
+  
+  // Destructor
+  ~Postings();
+  
+  // Copy constructor and assignment operator
+  Postings(const Postings& other);
+  Postings& operator=(const Postings& other);
```
allenss-amazon: I think we don't need these. mark them deleted.

src/indexes/text/posting.cc
Aksha1812: Iterator Implementation is stored as void* to hide internal implementation details (pointer nodes used to iterate)  of the iterators. so this casts the postings data back to KeyIterator.

src/indexes/text/posting.cc
Aksha1812: Iterator Implementation is stored as void* to hide internal implementation details (pointer nodes used to iterate) of the iterators. so this casts the postings data back to KeyIterator.

src/indexes/text/posting.cc
Aksha1812: mostly when someone tries to create Keyiterator seperately without using GetKeyIterator or if someone expliciltly set keyiterator to null, nothing in the current code that does that, but this is just a sanity check.

src/indexes/text/posting.cc
Aksha1812: got it. made changes for this

src/indexes/text/posting.cc
Aksha1812: done

src/indexes/text/posting.cc
Aksha1812: done

src/indexes/text/posting.cc
Aksha1812: ok done.

src/indexes/text/posting.cc
Aksha1812: ok. done.

src/indexes/text/posting.h
Aksha1812: done

Aksha1812 (commented):
responded/ added changes for all comments

src/indexes/text/posting.cc
BCathcart: Can you explain this to me a bit more? My C++ is very rusty but I'm not following why making the fields private in the `KeyIterator` struct isn't enough to hide the implementation details

Aksha1812: even if i used private : it would be still in header file. internal implementation details would still be hidden. it would look like
```
private:
  const std::map<Key, PositionMap>* key_map_;
  std::map<Key, PositionMap>::const_iterator current_;
  std::map<Key, PositionMap>::const_iterator end_;
};
```
I think it would be fine to keep this out of the header file since no need to show what variable we use to iterate over the position map. but wouldn't matter much if i keep it in. also read that using *void saves some compile time and casting doesn't add much overhead.

BCathcart: What's the concern with it being in the header? It doesn't change the public APIs. I would have thought it adds compile time having an additional struct.

Aksha1812: Initial thought was in case there is some error here stack trace might lead to std::map<Key, PositionMap>::const_iterator current_ and std::map<Key, PositionMap>::const_iterator end_; which might be confusing for user or debugger. But yeah i agree it doesn't matter much. can move it to private for sake of simplicity / avoid extra steps.

src/index_schema.h
```diff
@@ -166,6 +174,8 @@ class IndexSchema : public KeyspaceEventSubscription,
   std::unique_ptr<AttributeDataType> attribute_data_type_;
   std::string name_;
   uint32_t db_num_{0};
+  bool save_positions_{true};
+  size_t num_text_fields_{3};
```
allenss-amazon: 3?

src/indexes/text/posting.cc
allenss-amazon: void * is not the right way to hide internal implementation details.

The right way is to use a "class" or "struct" forward definition and then use the correct name. No casting should be needed.

allenss-amazon: Why does the iterator need an xxxImpl object?

src/indexes/text/posting.cc
```diff
@@ -384,31 +346,18 @@ bool Postings::KeyIterator::SkipForwardKey(const Key& key) {
...
-    return pos_iterator;
-  }
-  
+  assert(impl_data_ && "Iterator has no implementation data");
   auto* impl = static_cast<KeyIteratorImpl*>(impl_data_);
-  if (impl->current == impl->end) {
-    pos_iterator.impl_data_ = nullptr;
-    return pos_iterator;
-  }
+  assert(impl->current != impl->end && "Iterator has reached end - no positions available");
```
allenss-amazon: Let's use the CHECK macro, it allows full C++ I/O expressions to construct the failure message.

allenss-amazon (approved):

PR #250, opened 2025-07-24
Ft create parsing and protobuff filling

Nivesh-01:
CLEANER AND NEW PULL REQUEST FOR THE SAME

The parser is ready for FT_CREATE Function. It parses the create command and populates the protobuff with appropriately. There are unit tests covering both the parser and the whether the protobuff is correctly being populated.

An example of the correct FT.CREATE command would be as follows: FT.CREATE fullindex ON HASH PREFIX 1 doc: PUNCTUATION ",.;:!?" WITHOFFSETS STOPWORDS 2 the and LANGUAGE ENGLISH SCHEMA title TEXT WITHSUFFIXTRIE MINSTEMSIZE 3 body TEXT NOSTEM
and what we get return would be OK

If there is an incorrect command like: FT.CREATE badindex ON HASH SCHEMA content TEXT MINSTEMSIZE -1
We get the error tell what went wrong as follows: (error) Invalid field type for field `content`: Error parsing value for the parameter `MINSTEMSIZE` - MINSTEMSIZE must be positive

Currently we are testing the functionality using unit tests which is covering all different test cases.


.vscode/settings.json
```diff
@@ -17,5 +17,92 @@
         "--query-driver=**"
     ],
     "clangd.inactiveRegions.useBackgroundHighlight": true,
-    "clangd.enableCodeCompletion": true
+    "clangd.enableCodeCompletion": true,
```
Aksha1812, line 20: I think this is not needed. files.associations configuration in VS Code's settings tells the editor to treat files with specific names or extensions as if they were written in a particular language. vs code auto adds stl files here, but it is only needed if you open stl header file frequently

src/indexes/text/text.cc
```diff
@@ -0,0 +1,61 @@
+/*
```
Aksha1812, line 1: Avoid changing / editing other parts of code which is not related to your change. including spelling changes, indentation changes etc..

.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
     "clangd.path": "/usr/bin/clangd",
     "C_Cpp.codeAnalysis.runAutomatically": true,
     "editor.formatOnSave": false,
-    "C_Cpp.intelliSenseEngine": "disabled",
+    "C_Cpp.intelliSenseEngine": "default",
```
yairgott, line 8: We're using clang-tidy, why do you want to change this?

.vscode/settings.json
```diff
@@ -17,5 +17,92 @@
         "--query-driver=**"
     ],
     "clangd.inactiveRegions.useBackgroundHighlight": true,
-    "clangd.enableCodeCompletion": true
+    "clangd.enableCodeCompletion": true,
+    "files.associations": {
```
yairgott, line 21: why this is needed?

src/indexes/text/text.cc
VoletiRam: +1

VoletiRam (commented):
Lets use the template added by Allen for the PR description: https://github.com/valkey-io/valkey-search/blob/main/rfc/TEMPLATE.md

.vscode/settings.json
Nivesh-01: Changes to that file was made mostly for the launch file, to allow vscode gdb debugging. I could remove all of that would that be a bit better?

PR #251, opened 2025-07-25
Updated parsing and protobuff population with a clean git push

Nivesh-01:
This is a cleaner version of the code with using and changing of classes that are only affecting my work.

The parser is ready for FT_CREATE Function. It parses the create command and populates the protobuff with appropriately. There are unit tests covering both the parser and the whether the protobuff is correctly being populated.

An example of the correct FT.CREATE command would be as follows: FT.CREATE fullindex ON HASH PREFIX 1 doc: PUNCTUATION ",.;:!?" WITHOFFSETS STOPWORDS 2 the and LANGUAGE ENGLISH SCHEMA title TEXT WITHSUFFIXTRIE MINSTEMSIZE 3 body TEXT NOSTEM
and what we get return would be OK

If there is an incorrect command like: FT.CREATE badindex ON HASH SCHEMA content TEXT MINSTEMSIZE -1
We get the error tell what went wrong as follows: (error) Invalid field type for field content: Error parsing value for the parameter MINSTEMSIZE - MINSTEMSIZE must be positive

Currently we are testing the functionality using unit tests which is covering all different test cases.

### DIFF
--- a/src/commands/ft_create_parser.cc
+++ b/src/commands/ft_create_parser.cc
@@ -11,6 +11,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <mutex>  // For std::once_flag
 #include <set>
 #include <string>
 
@@ -36,6 +37,14 @@
 
 namespace valkey_search {
 namespace {
+
+// Default stop words set
+const std::vector<std::string> kDefaultStopWords{
+    "a", "is", "the", "an", "and", "are", "as", "at", "be", "but", "by", "for",
+    "if", "in", "into", "it", "no", "not", "of", "on", "or", "such", "that", "their",
+    "then", "there", "these", "they", "this", "to", "was", "will", "with"
+};
+
 constexpr absl::string_view kInitialCapParam{"INITIAL_CAP"};
 constexpr absl::string_view kBlockSizeParam{"BLOCK_SIZE"};
 constexpr absl::string_view kMParam{"M"};
@@ -80,6 +89,18 @@ constexpr absl::string_view kMaxMConfig{"max-vector-m"};
 constexpr absl::string_view kMaxEfConstructionConfig{
     "max-vector-ef-construction"};
 constexpr absl::string_view kMaxEfRuntimeConfig{"max-vector-ef-runtime"};
+
+// FullText variables
+constexpr absl::string_view kTextParam{"TEXT"};
+constexpr absl::string_view kPunctuationParam{"PUNCTUATION"};
+constexpr absl::string_view kWithOffsetsParam{"WITHOFFSETS"};
+constexpr absl::string_view kNoOffsetsParam{"NOOFFSETS"};
+constexpr absl::string_view kWithSuffixTrieParam{"WITHSUFFIXTRIE"};
+constexpr absl::string_view kNoSuffixTrieParam{"NOSUFFIXTRIE"};
+constexpr absl::string_view kNoStopWordsParam{"NOSTOPWORDS"};
+constexpr absl::string_view kStopWordsParam{"STOPWORDS"};
+constexpr absl::string_view kNoStemParam{"NOSTEM"};
+constexpr absl::string_view kMinStemSizeParam{"MINSTEMSIZE"};
 
 /// Register the "--max-prefixes" flag. Controls the max number of prefixes per
 /// index.
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
   index_proto.set_allocated_tag_index(tag_index_proto.release());
   return absl::OkStatus();
 }
+
+
+vmsdk::KeyValueParser<FTCreateTextParameters> CreateTextFieldParser() {
+  vmsdk::KeyValueParser<FTCreateTextParameters> parser;
+  // Field-level parameters only: WITHSUFFIXTRIE, NOSUFFIXTRIE, NOSTEM, MINSTEMSIZE
+  parser.AddParamParser(
+      kWithSuffixTrieParam, GENERATE_FLAG_PARSER(FTCreateTextParameters, with_suffix_trie));
+  parser.AddParamParser(
+      kNoSuffixTrieParam, 
+      std::make_unique<vmsdk::ParamParser<FTCreateTextParameters>>(
+          [](FTCreateTextParameters &params, vmsdk::ArgsIterator &itr) -> absl::Status {
+            params.with_suffix_trie = false;
+            return absl::OkStatus();
+          }));
+  parser.AddParamParser(
+      kNoStemParam, GENERATE_FLAG_PARSER(FTCreateTextParameters, no_stem));
+  parser.AddParamParser(
+      kMinStemSizeParam, 
+      std::make_unique<vmsdk::ParamParser<FTCreateTextParameters>>(
+          [](FTCreateTextParameters &params, vmsdk::ArgsIterator &itr) -> absl::Status {
+            int value;
+            VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, value));
+            if (value <= 0) {
+              return absl::InvalidArgumentError("MINSTEMSIZE must be positive");
+            }
+            params.min_stem_size = value;
+            return absl::OkStatus();
+          }));
+  return parser;
+}
+
+absl::Status ParseStopWords(vmsdk::ArgsIterator &itr, GlobalTextParameters &params) {
+  uint32_t count;
+  VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, count));
+  if (count == 0) {
+    params.no_stop_words = true;
+    return absl::OkStatus();
+  }
+  
+  // Check if we have enough arguments remaining
+  if (static_cast<uint32_t>(itr.DistanceEnd()) < count) {
+    return absl::OutOfRangeError("Missing argument");
+  }
+  
+  params.stop_words.clear();
+  for (uint32_t i = 0; i < count; ++i) {
+    if (!itr.HasNext()) {
+      return absl::OutOfRangeError("Missing argument");
+    }
+    
+    // Check if the next argument is a reserved keyword that should not be consumed
+    VMSDK_ASSIGN_OR_RETURN(auto next_arg, itr.Get());
+    absl::string_view next_word = vmsdk::ToStringView(next_arg);
+    if (absl::EqualsIgnoreCase(next_word, kSchemaParam)) {
+      return absl::OutOfRangeError("Missing argument");
+    }
+    
+    std::string word;
+    VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, word));
+    params.stop_words.push_back(word);
+  }
+  return absl::OkStatus();
+}
+
+absl::Status ParseText(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
+                       const GlobalTextParameters &global_defaults) {
+  // Start with field-specific defaults, then parse field-level parameters
+  FTCreateTextParameters field_params;
+  field_params.with_suffix_trie = false;
+  field_params.no_stem = global_defaults.no_stem;  // Can be overridden
+  field_params.min_stem_size = global_defaults.min_stem_size;  // Can be overridden
+  
+  // Parse field-level parameters (WITHSUFFIXTRIE, NOSUFFIXTRIE, NOSTEM, MINSTEMSIZE)
+  static auto field_parser = CreateTextFieldParser();
+  VMSDK_RETURN_IF_ERROR(field_parser.Parse(field_params, itr, false));
+  
+  // Create and populate the TextIndex object (field-specific parameters only)
+  auto text_index_proto = std::make_unique<data_model::TextIndex>();
+  text_index_proto->set_with_suffix_trie(field_params.with_suffix_trie);
+  text_index_proto->set_no_stem(field_params.no_stem);
+  text_index_proto->set_min_stem_size(field_params.min_stem_size);
+  
+  // Set the text_index in the index_proto
+  index_proto.set_allocated_text_index(text_index_proto.release());
+  
+  return absl::OkStatus();
+}
+
+vmsdk::KeyValueParser<GlobalTextParameters> CreateGlobalTextParser() {
+  vmsdk::KeyValueParser<GlobalTextParameters> parser;
+  parser.AddParamParser(
+      kPunctuationParam, GENERATE_VALUE_PARSER(GlobalTextParameters, punctuation));
+  parser.AddParamParser(
+      kWithOffsetsParam, GENERATE_FLAG_PARSER(GlobalTextParameters, with_offsets));
+  parser.AddParamParser(
+      kNoOffsetsParam, 
+      std::make_unique<vmsdk::ParamParser<GlobalTextParameters>>(
+          [](GlobalTextParameters &params, vmsdk::ArgsIterator &itr) -> absl::Status {
+            params.with_offsets = false;
+            return absl::OkStatus();
+          }));
+  parser.AddParamParser(
+      kNoStemParam, GENERATE_FLAG_PARSER(GlobalTextParameters, no_stem));
+  parser.AddParamParser(
+      kNoStopWordsParam, GENERATE_FLAG_PARSER(GlobalTextParameters, no_stop_words));
+  return parser;
+}
+
+absl::Status ParseGlobalTextDefaults(vmsdk::ArgsIterator &itr, GlobalTextParameters &defaults) {
+  static auto parser = CreateGlobalTextParser();
+  VMSDK_RETURN_IF_ERROR(parser.Parse(defaults, itr, false));
+  
+  VMSDK_ASSIGN_OR_RETURN(auto res, vmsdk::IsParamKeyMatch(kStopWordsParam, false, itr));
+  if (res) {
+    VMSDK_RETURN_IF_ERROR(ParseStopWords(itr, defaults));
+  }
+  
+  // Handle LANGUAGE parameter for global text defaults
+  VMSDK_ASSIGN_OR_RETURN(res, vmsdk::ParseParam(kLanguageParam, false, itr,
+                                               defaults.language, *kLanguageByStr));
+  
+  // Validate punctuation is not empty after parsing
+  if (defaults.punctuation.length() >= 2 && 
+      ((defaults.punctuation.front() == '"' && defaults.punctuation.back() == '"') ||
+       (defaults.punctuation.front() == '\'' && defaults.punctuation.back() == '\''))) {
+    defaults.punctuation = defaults.punctuation.substr(1, defaults.punctuation.length() - 2);
+  }
+  
+  // Check if punctuation is empty after quote removal
+  if (defaults.punctuation.empty()) {
+    return absl::InvalidArgumentError("PUNCTUATION string cannot be empty");
+  }
+  
+  return absl::OkStatus();
+}
 absl::StatusOr<indexes::IndexerType> ParseIndexerType(
     vmsdk::ArgsIterator &itr) {
   absl::string_view index_type_str;
@@ -370,7 +526,8 @@ absl::StatusOr<indexes::IndexerType> ParseIndexerType(
 }
 absl::StatusOr<data_model::Attribute *> ParseAttributeArgs(
     vmsdk::ArgsIterator &itr, absl::string_view attribute_identifier,
-    data_model::IndexSchema &index_schema_proto) {
+    data_model::IndexSchema &index_schema_proto,
+    const GlobalTextParameters &global_text_defaults) {
   auto attribute_proto = index_schema_proto.add_attributes();
   attribute_proto->set_identifier(attribute_identifier);
   VMSDK_ASSIGN_OR_RETURN(auto res,
@@ -388,6 +545,8 @@ absl::StatusOr<data_model::Attribute *> ParseAttributeArgs(
   } else if (index_type == indexes::IndexerType::kNumeric) {
     VMSDK_RETURN_IF_ERROR(
         ParseNumeric(itr, *index_proto, attribute_identifier));
+  } else if (index_type == indexes::IndexerType::kText) {
+    VMSDK_RETURN_IF_ERROR(ParseText(itr, *index_proto, global_text_defaults));
   } else {
     CHECK(false);
   }
@@ -400,6 +559,17 @@ bool HasVectorIndex(const data_model::IndexSchema &index_schema_proto) {
     const auto &index = attribute.index();
     if (index.index_type_case() ==
         data_model::Index::IndexTypeCase::kVectorIndex) {
+      return true;
+    }
+  }
+  return false;
+}
+
+bool HasTextIndex(const data_model::IndexSchema &index_schema_proto) {
+  for (const auto &attribute : index_schema_proto.attributes()) {
+    const auto &index = attribute.index();
+    if (index.index_type_case() ==
+        data_model::Index::IndexTypeCase::kTextIndex) {
       return true;
     }
   }
@@ -441,6 +611,33 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
     return absl::InvalidArgumentError(
         NotSupportedParamErrorMsg(kPayloadFieldParam));
   }
+  
+  // Parse global text parameters before SCHEMA
+  GlobalTextParameters global_text_defaults;
+  // Initialize with defaults for each parse call
+  global_text_defaults.punctuation = ",.<>{}[]\"':;!@#$%^&*()-+=~/\\|";
+  global_text_defaults.min_stem_size = 4;
+  global_text_defaults.with_offsets = true;
+  global_text_defaults.no_stem = false;
+  global_text_defaults.no_stop_words = false;
+  global_text_defaults.language = data_model::LANGUAGE_ENGLISH;
+  global_text_defaults.stop_words = kDefaultStopWords;
+  
+  VMSDK_RETURN_IF_ERROR(ParseGlobalTextDefaults(itr, global_text_defaults));
+  
+  // Set global text parameters in IndexSchema
+  index_schema_proto.set_language(global_text_defaults.language);
+  index_schema_proto.set_punctuation(global_text_defaults.punctuation);
+  index_schema_proto.set_with_offsets(global_text_defaults.with_offsets);
+  index_schema_proto.set_no_stop_words(global_text_defaults.no_stop_words);
+  index_schema_proto.set_nostem(global_text_defaults.no_stem);
+  index_schema_proto.set_min_stem_size(global_text_defaults.min_stem_size);
+  
+  // Add stop words to the schema
+  for (const auto& word : global_text_defaults.stop_words) {
+    index_schema_proto.add_stop_words(word);
+  }
+  
   absl::string_view schema;
   VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, schema));
   if (!absl::EqualsIgnoreCase(schema, kSchemaParam)) {
@@ -457,7 +654,7 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
     VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, attribute_identifier));
     VMSDK_ASSIGN_OR_RETURN(
         auto attribute,
-        ParseAttributeArgs(itr, attribute_identifier, index_schema_proto),
+        ParseAttributeArgs(itr, attribute_identifier, index_schema_proto, global_text_defaults),
         _.SetPrepend() << "Invalid field type for field `"
                        << attribute_identifier << "`: ");
     if (identifier_names.find(attribute->identifier()) !=
@@ -472,9 +669,14 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
 
     identifier_names.insert(attribute->identifier());
   }
-  if (!HasVectorIndex(index_schema_proto)) {
+  
+  // Check if the schema has at least one required index type (vector or text)
+  bool has_vector = HasVectorIndex(index_schema_proto);
+  bool has_text = HasTextIndex(index_schema_proto);
+  
+  if (!has_vector && !has_text) {
     return absl::InvalidArgumentError(
-        "At least one attribute must be indexed as a vector");
+        "At least one attribute must be indexed as a vector or text field");
   }
   return index_schema_proto;
 }
--- a/src/commands/ft_create_parser.h
+++ b/src/commands/ft_create_parser.h
@@ -12,6 +12,8 @@
 #include <cstdint>
 #include <memory>
 #include <optional>
+#include <string>
+#include <vector>
 
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
@@ -39,6 +41,25 @@ struct FTCreateVectorParameters {
   absl::Status Verify() const;
   std::unique_ptr<data_model::VectorIndex> ToProto() const;
 };
+
+// Global text parameters (per-index) - populated in IndexSchema
+struct GlobalTextParameters {
+  std::string punctuation{",.<>{}[]\"':;!@#$%^&*()-+=~/\\\\"};
+  bool with_offsets{true};
+  bool no_stem{false};
+  std::vector<std::string> stop_words;
+  bool no_stop_words{false};
+  data_model::Language language{data_model::LANGUAGE_ENGLISH};
+  int min_stem_size{4};
+};
+
+// Field-specific text parameters (per text field) - populated in TextIndex
+struct FTCreateTextParameters {
+  bool with_suffix_trie{false};
+  bool no_stem{false};  // Can be overridden per field
+  int min_stem_size{4};
+};
+
 
 constexpr int kDefaultBlockSize{1024};
 constexpr int kDefaultM{16};
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -36,6 +36,7 @@
 #include "src/indexes/index_base.h"
 #include "src/indexes/numeric.h"
 #include "src/indexes/tag.h"
+#include "src/indexes/text.h"
 #include "src/indexes/vector_base.h"
 #include "src/indexes/vector_flat.h"
 #include "src/indexes/vector_hnsw.h"
@@ -79,6 +80,9 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
     }
     case data_model::Index::IndexTypeCase::kNumericIndex: {
       return std::make_shared<indexes::Numeric>(index.numeric_index());
+    }
+    case data_model::Index::IndexTypeCase::kTextIndex: {
+      return std::make_shared<indexes::Text>(index.text_index());
     }
     case data_model::Index::IndexTypeCase::kVectorIndex: {
       switch (index.vector_index().algorithm_case()) {
--- a/src/index_schema.proto
+++ b/src/index_schema.proto
@@ -29,7 +29,22 @@ message IndexSchema {
   repeated Attribute attributes = 6;
   Stats stats = 7;
   optional uint32 db_num = 8;
+
+  string punctuation = 9;
+  bool with_offsets = 10;
+  repeated string stop_words = 11;
+  bool no_stop_words = 12;
+  bool nostem = 13;
+  int32 min_stem_size = 14;
 }
+
+
+message TextIndex {
+  bool with_suffix_trie = 1;
+  bool no_stem = 2;
+  int32 min_stem_size = 3;
+}
+
 
 enum AttributeIndexType {
   ATTRIBUTE_INDEX_TYPE_UNSPECIFIED = 0;
@@ -47,6 +62,7 @@ message Index {
     VectorIndex vector_index = 1;
     NumericIndex numeric_index = 2;
     TagIndex tag_index = 3;
+    TextIndex text_index = 4;
   }
 }
 
--- a/src/indexes/text.h
+++ b/src/indexes/text.h
@@ -30,65 +30,142 @@
 #ifndef VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 #define VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 
-#include "src/text/text.h"
+#include <cstdint>
+#include <memory>
+#include <string>
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/base/thread_annotations.h"
+#include "absl/functional/any_invocable.h"
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "absl/synchronization/mutex.h"
+#include "src/index_schema.pb.h"
+#include "src/indexes/index_base.h"
+#include "src/rdb_serialization.h"
+#include "src/utils/string_interning.h"
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+
+// Forward declarations
+namespace valkey_search::query {
+class TextPredicate;
+}
 
 namespace valkey_search::indexes {
 
 class Text : public IndexBase {
  public:
-  explicit Text(const data_model::TextIndex& text_index_proto);
+  explicit Text(const data_model::TextIndex& text_index_proto) : IndexBase(IndexerType::kText) {
+    // TODO: Parse configuration from text_index_proto
+  }
   absl::StatusOr<bool> AddRecord(const InternedStringPtr& key,
                                  absl::string_view data) override
-      ABSL_LOCKS_EXCLUDED(index_mutex_);
+      ABSL_LOCKS_EXCLUDED(index_mutex_) {
+    absl::MutexLock lock(&index_mutex_);
+    tracked_keys_[key] = std::string(data);
+    return true;
+  }
   absl::StatusOr<bool> RemoveRecord(
       const InternedStringPtr& key,
       DeletionType deletion_type = DeletionType::kNone) override
-      ABSL_LOCKS_EXCLUDED(index_mutex_);
+      ABSL_LOCKS_EXCLUDED(index_mutex_) {
+    absl::MutexLock lock(&index_mutex_);
+    auto it = tracked_keys_.find(key);
+    if (it == tracked_keys_.end()) {
+      return false;
+    }
+    tracked_keys_.erase(it);
+    return true;
+  }
   absl::StatusOr<bool> ModifyRecord(const InternedStringPtr& key,
                                     absl::string_view data) override
-      ABSL_LOCKS_EXCLUDED(index_mutex_);
-  int RespondWithInfo(RedisModuleCtx* ctx) const override;
-  bool IsTracked(const InternedStringPtr& key) const override;
-  absl::Status SaveIndex(RDBOutputStream& rdb_stream) const override {
+      ABSL_LOCKS_EXCLUDED(index_mutex_) {
+    absl::MutexLock lock(&index_mutex_);
+    auto it = tracked_keys_.find(key);
+    if (it == tracked_keys_.end()) {
+      return false;
+    }
+    it->second = std::string(data);
+    return true;
+  }
+  int RespondWithInfo(ValkeyModuleCtx* ctx) const override {
+    ValkeyModule_ReplyWithSimpleString(ctx, "index_type");
+    ValkeyModule_ReplyWithSimpleString(ctx, "TEXT");
+    return 2;
+  }
+  bool IsTracked(const InternedStringPtr& key) const override {
+    absl::MutexLock lock(&index_mutex_);
+    return tracked_keys_.find(key) != tracked_keys_.end();
+  }
+  absl::Status SaveIndex(RDBChunkOutputStream chunked_out) const override {
     return absl::OkStatus();
   }
 
   inline void ForEachTrackedKey(
       absl::AnyInvocable<void(const InternedStringPtr&)> fn) const override {
     absl::MutexLock lock(&index_mutex_);
-    for (const auto& [key, _] : tracked_tags_by_keys_) {
+    for (const auto& [key, _] : tracked_keys_) {
       fn(key);
     }
   }
-  uint64_t GetRecordCount() const override;
-  std::unique_ptr<data_model::Index> ToProto() const override;
+  uint64_t GetRecordCount() const override {
+    absl::MutexLock lock(&index_mutex_);
+    return tracked_keys_.size();
+  }
+  std::unique_ptr<data_model::Index> ToProto() const override {
+    auto index = std::make_unique<data_model::Index>();
+    auto text_index = std::make_unique<data_model::TextIndex>();
+    index->set_allocated_text_index(text_index.release());
+    return index;
+  }
 
   InternedStringPtr GetRawValue(const InternedStringPtr& key) const
-      ABSL_NO_THREAD_SAFETY_ANALYSIS;
+      ABSL_NO_THREAD_SAFETY_ANALYSIS {
+    auto it = tracked_keys_.find(key);
+    if (it == tracked_keys_.end()) {
+      return nullptr;
+    }
+    return StringInternStore::Intern(it->second);
+  }
 
   class EntriesFetcherIterator : public EntriesFetcherIteratorBase {
    public:
-    bool Done() const override;
-    void Next() override;
-    const InternedStringPtr& operator*() const override;
+    bool Done() const override {
+      return true;
+    }
+    void Next() override {
+      // TODO: Implement iterator logic
+    }
+    const InternedStringPtr& operator*() const override {
+      static InternedStringPtr empty_ptr;
+      return empty_ptr;
+    }
 
    private:
   };
 
   class EntriesFetcher : public EntriesFetcherBase {
    public:
-    size_t Size() const override;
-    std::unique_ptr<EntriesFetcherIteratorBase> Begin() override;
+    size_t Size() const override {
+      return 0;
+    }
+    std::unique_ptr<EntriesFetcherIteratorBase> Begin() override {
+      return std::make_unique<EntriesFetcherIterator>();
+    }
 
    private:
   };
 
   virtual std::unique_ptr<EntriesFetcher> Search(
       const query::TextPredicate& predicate,
-      bool negate) const ABSL_NO_THREAD_SAFETY_ANALYSIS;
+      bool negate) const ABSL_NO_THREAD_SAFETY_ANALYSIS {
+    return std::make_unique<EntriesFetcher>();
+  }
 
  private:
   mutable absl::Mutex index_mutex_;
+  absl::flat_hash_map<InternedStringPtr, std::string> tracked_keys_;
 };
 }  // namespace valkey_search::indexes
 
--- a/src/indexes/text/text.cc
+++ b/src/indexes/text/text.cc
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2025, ValkeySearch contributors
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the name of Redis nor the names of its contributors may be used
+ *     to endorse or promote products derived from this software without
+ *     specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+// Basic stub implementations for text subsystem types
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "src/utils/string_interning.h"
+#include <memory>
+
+namespace valkey_search {
+namespace text {
+
+// Forward declare types used in text.h
+template<typename T, bool suffix>
+class RadixTree {
+public:
+    RadixTree() = default;
+    ~RadixTree() = default;
+};
+
+class Postings {
+public:
+    Postings() = default;
+    ~Postings() = default;
+};
+
+// Basic stub classes to satisfy linker
+class TextIndex {
+public:
+    TextIndex() = default;
+    ~TextIndex() = default;
+};
+
+}  // namespace text
+}  // namespace valkey_search
--- a/src/indexes/text/text.h
+++ b/src/indexes/text/text.h
@@ -1,5 +1,5 @@
 #ifndef _VALKEY_SEARCH_INDEXES_TEXT_TEXT_H
-#define _VALKSY_SEARCH_INDEXES_TEXT_TEXT_H
+#define _VALKEY_SEARCH_INDEXES_TEXT_TEXT_H
 
 /*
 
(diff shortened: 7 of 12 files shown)


src/commands/ft_create_parser.cc
```diff
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+            return absl::OkStatus();
+          }));
+  return parser;
+}
+
+absl::Status ParseStopWords(vmsdk::ArgsIterator &itr, GlobalTextParameters &params) {
+  uint32_t count;
+  VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, count));
+  if (count == 0) {
+    params.no_stop_words = true;
```
allenss-amazon: Why have a separate boolean? just clear params.stop_words.

src/commands/ft_create_parser.cc
```diff
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+  if (static_cast<uint32_t>(itr.DistanceEnd()) < count) {
+    return absl::OutOfRangeError("Missing argument");
+  }
+  
+  params.stop_words.clear();
+  for (uint32_t i = 0; i < count; ++i) {
+    if (!itr.HasNext()) {
+      return absl::OutOfRangeError("Missing argument");
+    }
+
```
allenss-amazon: Isn't this redundant with the explicit distance check above?

src/commands/ft_create_parser.cc
```diff
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+  for (uint32_t i = 0; i < count; ++i) {
+    if (!itr.HasNext()) {
+      return absl::OutOfRangeError("Missing argument");
+    }
+    
+    // Check if the next argument is a reserved keyword that should not be consumed
+    VMSDK_ASSIGN_OR_RETURN(auto next_arg, itr.Get());
+    absl::string_view next_word = vmsdk::ToStringView(next_arg);
+    if (absl::EqualsIgnoreCase(next_word, kSchemaParam)) {
+      return absl::OutOfRangeError("Missing argument");
```
allenss-amazon: why disable this particular stop word?

src/commands/ft_create_parser.cc
```diff
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+  // Parse field-level parameters (WITHSUFFIXTRIE, NOSUFFIXTRIE, NOSTEM, MINSTEMSIZE)
+  static auto field_parser = CreateTextFieldParser();
+  VMSDK_RETURN_IF_ERROR(field_parser.Parse(field_params, itr, false));
+  
+  // Create and populate the TextIndex object (field-specific parameters only)
+  auto text_index_proto = std::make_unique<data_model::TextIndex>();
+  text_index_proto->set_with_suffix_trie(field_params.with_suffix_trie);
+  text_index_proto->set_no_stem(field_params.no_stem);
+  text_index_proto->set_min_stem_size(field_params.min_stem_size);
+
```
allenss-amazon, line 463: shouldn't the stop words get copied over?

src/commands/ft_create_parser.cc
```diff
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+      kPunctuationParam, GENERATE_VALUE_PARSER(GlobalTextParameters, punctuation));
+  parser.AddParamParser(
+      kWithOffsetsParam, GENERATE_FLAG_PARSER(GlobalTextParameters, with_offsets));
+  parser.AddParamParser(
+      kNoOffsetsParam, 
+      std::make_unique<vmsdk::ParamParser<GlobalTextParameters>>(
+          [](GlobalTextParameters &params, vmsdk::ArgsIterator &itr) -> absl::Status {
+            params.with_offsets = false;
+            return absl::OkStatus();
+          }));
```
allenss-amazon: The "negative" flag parser shows up several times, let's just make a GENERATE_NEGATIVE_FLAG_PARSER.

src/commands/ft_create_parser.cc
```diff
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+  
+  // Handle LANGUAGE parameter for global text defaults
+  VMSDK_ASSIGN_OR_RETURN(res, vmsdk::ParseParam(kLanguageParam, false, itr,
+                                               defaults.language, *kLanguageByStr));
+  
+  // Validate punctuation is not empty after parsing
+  if (defaults.punctuation.length() >= 2 && 
+      ((defaults.punctuation.front() == '"' && defaults.punctuation.back() == '"') ||
+       (defaults.punctuation.front() == '\'' && defaults.punctuation.back() == '\''))) {
+    defaults.punctuation = defaults.punctuation.substr(1, defaults.punctuation.length() - 2);
```
allenss-amazon: The code to strip pre- and post- quotation marks is wrong. Those won't show up "on the wire", generally in documentation we're showing syntax that's sent to a tool like valkey-cli, which will strip those of before sending it to the server.

src/commands/ft_create_parser.cc
```diff
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+  text_index_proto->set_no_stem(field_params.no_stem);
+  text_index_proto->set_min_stem_size(field_params.min_stem_size);
+  
+  // Set the text_index in the index_proto
+  index_proto.set_allocated_text_index(text_index_proto.release());
+  
+  return absl::OkStatus();
+}
+
+vmsdk::KeyValueParser<GlobalTextParameters> CreateGlobalTextParser() {
```
allenss-amazon: Nit: Would "SchemaTextParameters" be more descriptive?

src/commands/ft_create_parser.cc
```diff
@@ -472,9 +669,14 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
 
     identifier_names.insert(attribute->identifier());
   }
-  if (!HasVectorIndex(index_schema_proto)) {
+  
+  // Check if the schema has at least one required index type (vector or text)
+  bool has_vector = HasVectorIndex(index_schema_proto);
+  bool has_text = HasTextIndex(index_schema_proto);
+  
+  if (!has_vector && !has_text) {
```
allenss-amazon: there's no requirement to have a text | vector field.

src/indexes/text.h
```diff
@@ -30,65 +30,142 @@
...
+    absl::MutexLock lock(&index_mutex_);
+    auto it = tracked_keys_.find(key);
+    if (it == tracked_keys_.end()) {
+      return false;
+    }
+    it->second = std::string(data);
+    return true;
+  }
+  int RespondWithInfo(ValkeyModuleCtx* ctx) const override {
+    ValkeyModule_ReplyWithSimpleString(ctx, "index_type");
```
allenss-amazon: should we include the other configurable per-index attributes?

src/indexes/text.h
```diff
@@ -30,65 +30,142 @@
...
   virtual std::unique_ptr<EntriesFetcher> Search(
       const query::TextPredicate& predicate,
-      bool negate) const ABSL_NO_THREAD_SAFETY_ANALYSIS;
+      bool negate) const ABSL_NO_THREAD_SAFETY_ANALYSIS {
+    return std::make_unique<EntriesFetcher>();
+  }
 
  private:
   mutable absl::Mutex index_mutex_;
+  absl::flat_hash_map<InternedStringPtr, std::string> tracked_keys_;
```
allenss-amazon: We will need to discuss this, I hadn't planned on having another copy of the data. I do see that the current logic assumes it can fetch an indexed value from an index, but Text is case where we might not want to support that.

testing/ft_create_parser_test.cc
```diff
@@ -556,7 +569,7 @@ INSTANTIATE_TEST_SUITE_P(
              .command_str = "idx1 on HASH SChema hash_field1 as "
                             "hash_field11 numeric ",
              .expected_error_message =
-                 "At least one attribute must be indexed as a vector",
+                 "At least one attribute must be indexed as a vector or text field",
```
allenss-amazon: As described above, there shouldn't be any "required" field types.

testing/ft_create_parser_test.cc
```diff
@@ -910,6 +923,588 @@ INSTANTIATE_TEST_SUITE_P(
...
+                 .min_stem_size = 4,
+             }},
+             .expected = {
+                 .index_schema_name = "idx1",
+                 .on_data_type = data_model::ATTRIBUTE_DATA_TYPE_HASH,
+                 .attributes = {{
+                     .identifier = "text_field",
+                     .attribute_alias = "text_field",
+                     .indexer_type = indexes::IndexerType::kText,
+                 }}
```
allenss-amazon: Shouldn't there be testing for the schema-level text parameters: stopwords, punctuation, etc.

testing/ft_create_parser_test.cc
```diff
@@ -910,6 +923,588 @@ INSTANTIATE_TEST_SUITE_P(
...
+             .test_name = "invalid_text_empty_punctuation_global",
+             .success = false,
+             .command_str = "idx1 on HASH PUNCTUATION \"\" SCHEMA text_field TEXT",
+             .too_many_attributes = false,
+             .hnsw_parameters = {},
+             .flat_parameters = {},
+             .tag_parameters = {},
+             .text_parameters = {},
+             .expected = {},
+             .expected_error_message = "PUNCTUATION string cannot be empty",
```
allenss-amazon, line 1195: Clearly, this is the case that lead you to adding the code to "strip" leading/trailing quotes. But that's a failure in this test machinery. The quote stripping code should be added to the code that breaks "command_str" into argc/argv. In other words, just make a pass over argv[*] and if you see a string with leading&trailing quotes, then strip them off there.

testing/ft_create_parser_test.cc
```diff
@@ -910,6 +923,588 @@ INSTANTIATE_TEST_SUITE_P(
...
+             .tag_parameters = {},
+             .text_parameters = {},
+             .expected = {},
+             .expected_error_message = "Missing argument",
+         },
+         // Additional TEXT field edge case tests
+         {
+             .test_name = "text_punctuation_single_quote",
+             .success = false,
+             .command_str = "idx1 on HASH SCHEMA text_field TEXT PUNCTUATION '.,;'",
```
allenss-amazon, line 1249: same problem as above....

src/commands/ft_create_parser.cc
```diff
@@ -359,6 +380,141 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+  uint32_t count;
+  VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, count));
+  if (count == 0) {
+    params.no_stop_words = true;
+    return absl::OkStatus();
+  }
+  
+  // Check if we have enough arguments remaining
+  if (static_cast<uint32_t>(itr.DistanceEnd()) < count) {
+    return absl::OutOfRangeError("Missing argument");
```
allenss-amazon: "Missing argument" could be more descriptive, maybe "missing count after STOPWORDS keyword" or something like that.

src/commands/ft_create_parser.cc
Nivesh-01: you are right, made the change.

src/commands/ft_create_parser.cc
Nivesh-01: I thought the same way, but kept another boolean since it was how we initially thought about it but since you feel the same as way I will remove no_stop_words altogether from the schema, it is redundant and does not add any value.

src/commands/ft_create_parser.cc
Nivesh-01: This is like a fail safe check. This is to stop the case where we consume one of the words in the schema. Like the if we have two punctuation in the FT.create command or if there are fewer words than specified (even though we already check for that). It is just an added layer of extra check that could be removed, but I think it keeps the code more robust.

src/commands/ft_create_parser.cc
Nivesh-01: we are only copying over text level parameters here, stop words is a global parameter and is being copied directly in the ParseFTCreateArgs (starts there).

src/commands/ft_create_parser.cc
Nivesh-01: ok, i will fix this. Should we throw an error if there is something that is not vector, text, numeric, or tag? Or should I just remove it and not check for it at all.

src/indexes/text.h
Nivesh-01: oh ok, I will comment it out for now then.

testing/ft_create_parser_test.cc
Nivesh-01: I could do that but was not sure if I should. If I add in test for that it would make each test very wordy instead of being compact and easy to read. It would also break away from the way the vector parsing test have been done. For the schema level parameters we have quite a few test which implicitly tests for them. For example in the following test:
{
             .test_name = "invalid_text_empty_punctuation_global",
             .success = false,
             .command_str = "idx1 on HASH PUNCTUATION \"\" SCHEMA text_field TEXT",
             .too_many_attributes = false,
             .hnsw_parameters = {},
             .flat_parameters = {},
             .tag_parameters = {},
             .text_parameters = {},
             .expected = {},
             .expected_error_message = "PUNCTUATION string cannot be empty",
         },
         
If my test passes it shows that the punctuation field is being checked to see if the parsing is proper.

But I could also add it explicitly if need be, it should not be hard. Please let me know what you think.

src/commands/ft_create_parser.cc
Nivesh-01: I will add it in the next commit.

src/commands/ft_create_parser.cc
Nivesh-01: I named it this way instead of something like SchemaTextParameters because initially when I was implementing I was getting confused in the naming since all the parameters are a part of schema in the protobuff they do not come inside (after) schema. But if it confusing I could change it.

src/commands/ft_create_parser.cc
allenss-amazon: I understand that you're trying to put out better error messages for one class of syntax errors. However, this code requires us to document that "SCHEMA" can't be used as a stop word, right? I don't think this is path we want to go down.

testing/ft_create_parser_test.cc
allenss-amazon: I'm fine with creating a different parsing mechanism to make schema-level parsing tests easier to understand and to write. But we should have tests for those.

src/commands/ft_create_parser.cc
```diff
@@ -36,6 +37,14 @@
 
 namespace valkey_search {
 namespace {
+
+// Default stop words set
+const std::vector<std::string> kDefaultStopWords{
+    "a", "is", "the", "an", "and", "are", "as", "at", "be", "but", "by", "for",
+    "if", "in", "into", "it", "no", "not", "of", "on", "or", "such", "that", "their",
+    "then", "there", "these", "they", "this", "to", "was", "will", "with"
```
BCathcart, line 45: Just curious how did we decide on these stop words?

src/commands/ft_create_parser.cc
BCathcart: I assume the schema can't have anything other than the field types we support. We should be able to support any combination of them barring any upper limits we may set and probably enforce that the schema has at least one field. We should get @KarthikSubbarao's changes from main into the full text branch if they aren't already since I believe he would have already changed this code.

src/indexes/text.h
```diff
@@ -30,65 +30,142 @@
...
+    }
+    return StringInternStore::Intern(it->second);
+  }
 
   class EntriesFetcherIterator : public EntriesFetcherIteratorBase {
    public:
-    bool Done() const override;
-    void Next() override;
-    const InternedStringPtr& operator*() const override;
+    bool Done() const override {
```
BCathcart: Do these methods need to be implemented with stubs right now? I'd rather leave them unimplemented or at least with clear TODOs so that it's easier to identify what still needs to be implemented at a quick glance

src/indexes/text/text.cc
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the name of Redis nor the names of its contributors may be used
```
BCathcart: Looks like this copyright notice is outdated since it mentions Redis?

src/indexes/text/text.cc
```diff
@@ -0,0 +1,61 @@
...
+// Basic stub implementations for text subsystem types
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "src/utils/string_interning.h"
+#include <memory>
+
+namespace valkey_search {
+namespace text {
+
+// Forward declare types used in text.h
```
BCathcart: These are empty class definitions for classes we already have in other files. We probably don't need this file

src/indexes/text/text.h
BCathcart, line 1: This duplicates some of what `src/indexes/text.h` already has. I expect this is an outdated file we can remove, since `TextIndex` and `IndexSchemaText` look more like things that should be associated with the entire schema and not individual text fields. Not specific to your PR but just fyi since you're touching them.

src/indexes/text_index.h
```diff
@@ -1,19 +1,56 @@
...
+public:
+  RadixTree() = default;
+};
+}
+
+using Key = InternedStringPtr;
+
 namespace indexes {
 
 struct Text : public IndexBase {
```
BCathcart: I don't think this needs to inherit IndexBase. This seems like the schema-level text index object we'll need. It doesn't have to be part of this PR, but I wouldn't worry about touching these files in the meantime if you can avoid it.

src/indexes/text_index.h
```diff
@@ -1,19 +1,56 @@
...
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+#include "src/attribute_data_type.h"
+#include "src/index_schema.pb.h"
+#include "src/indexes/index_base.h"
+#include "src/rdb_serialization.h"
+#include "src/utils/string_interning.h"
 
 namespace valkey_search {
+
+// Forward declare the types
```
BCathcart: Can you just include `radix_tree.h`?

src/commands/ft_create_parser.cc
Nivesh-01: oh yes I see that. I will remove that as well.

src/commands/ft_create_parser.cc
Nivesh-01: @KarthikSubbarao from my conversation with him I believe removed the snippet itself from there so there is no checking going on over there which I can do as well if need be otherwise I can have a check for everything whichever works

src/indexes/text/text.cc
Nivesh-01: I just took it out from another file, I can remove it or change it whichever is better.

src/indexes/text_index.h
Nivesh-01: done

src/indexes/text_index.h
Nivesh-01: The inheriting of the base class is something that was already there in the skeleton. I added the other changes because the whole file war error ridden and adding these lines mitigated that to some extent. Should I revert it back to the original.

src/indexes/text/text.h
Nivesh-01: I have only corrected the lines of code that had an error in the syntax, which was causing an error when I was trying to build. I could revert it back to the original to see if it would work now. I also thought the same thing when I was editing the file that why are there two files which are kind of the same.

src/indexes/text/text.cc
Nivesh-01: You are right, it is deleted now and will not be visible in further iterations of the PR

src/indexes/text.h
Nivesh-01: Added TODO Comments for functions also added it to the top as a reminder the certain functions need to be implemented later.

src/commands/ft_create_parser.cc
Nivesh-01: image

Got it from the RFC. Attached is the image of the place. (Attached the picture so that it is easier to see)

src/commands/ft_create_parser.cc
Nivesh-01: I have removed the above lines of code.

src/commands/ft_create_parser.cc
allenss-amazon: They are from Redis documentation.

src/commands/ft_create_parser.cc
allenss-amazon: I like the "NOSTOPWORDS" keyword, I just think that the internal implementation shouldn't have a boolean that's equivalent to setting the array length to zero.

src/commands/ft_create_parser.cc
allenss-amazon: Global usually means "single" or "everywhere", here you really mean "global to the schema".

src/commands/ft_create_parser.cc
Nivesh-01: done

src/indexes/text.h
Nivesh-01: I am not sure. This will be used by FT.INFO from what I understand. I am not sure if it is useful to give other configurable attributes.

testing/ft_create_parser_test.cc
Nivesh-01: done

src/indexes/text.h
Nivesh-01: I am not sure. This will be used by FT.INFO from what I understand. I am not sure if it is useful to give other configurable attributes.

src/indexes/text.h
Nivesh-01: Are we doing no tracking in text.h for now. Asking so that I can completely remove tracked keys for text.h

src/commands/ft_create_parser.cc
Nivesh-01: I am commenting this method out, but the error message that is given out is not clean at all. Since we are not explicitly checking for this case the error that is being thrown would be all over the place. For example if we give a command like the following FT.CREATE idx1 ON HASH STOPWORDS 3 the and PUNCTUATION ".:," SCHEMA text_field TEXT we will not get the missing argument error instead we will get an error for ".:," as bad arguments. What will happen in this case is that PUNCTUATION will be used as a Stopword and there will be no declaration for punctuation. The error will keep changing based on our create.
Will this not be a problem.

src/commands/ft_create_parser.cc
Nivesh-01: done

.github/workflows/unittests-tsan.yml
Aksha1812, line 1: what are these ASAN build changes. was this requested as part of this PR?

src/indexes/text.h
```diff
@@ -30,65 +30,142 @@
...
-  explicit Text(const data_model::TextIndex& text_index_proto);
+  explicit Text(const data_model::TextIndex& text_index_proto) : IndexBase(IndexerType::kText) {
+    // TODO: Parse configuration from text_index_proto
+  }
   absl::StatusOr<bool> AddRecord(const InternedStringPtr& key,
                                  absl::string_view data) override
-      ABSL_LOCKS_EXCLUDED(index_mutex_);
+      ABSL_LOCKS_EXCLUDED(index_mutex_) {
+    absl::MutexLock lock(&index_mutex_);
+    tracked_keys_[key] = std::string(data);
```
Aksha1812: what does this key tracking do. is it in the scope of this task? Locking mechanism should be separately be reviewed and submitted imo.

.github/workflows/unittests-tsan.yml
Aksha1812: looks like still not rebased properly. this is part of changes submitted by yair 2 weeks ago

src/indexes/text.h
Nivesh-01: With the new commit from a different PR it should be changed

testing/ft_create_parser_test.cc
Nivesh-01: Added the tests for those as well.

src/commands/ft_create_parser.cc
Nivesh-01: Changed ParseGlobalTextDefaults to ParseSchemaGlobalTextDefaults and CreateGlobalTextParser to SchemaGlobTextParameters, which I think will be more descriptive. Please let me know if you would like me to change anything else as well.
I also went ahead and changed the GlobalTextParameters to SchemaGlobTextParams which represented the schema params in ft_create_parser

src/commands/ft_create_parser.cc
Nivesh-01: I understand. I have removed all the places where this stripping was taking place. For testing what I did was add the stripping mechanism to utils.cc in testing_infra file.

testing/ft_create_parser_test.cc
Nivesh-01: I think this should be solved with the new change in testing_infra

.gitignore
```diff
@@ -11,3 +11,4 @@ build-*
 build/
 .build-*
 debs/*.deb
+.vscode/
```
Aksha1812: Who requested this. Shouldn’t do this in my opinion.

.vscode/settings.json
Aksha1812, line 1: Don’t do this. I think ask was just to remove your changes from this file. Keep it as is.

src/indexes/text_index.h
```diff
@@ -60,4 +60,4 @@ struct IndexSchemaText{
 }  // namespace indexes
 }  // namespace valkey_search
 
-#endif
+#endif
```
Aksha1812: ?

src/indexes/text.h
```diff
@@ -98,4 +98,4 @@ class Text : public IndexBase {
 };
 }  // namespace valkey_search::indexes
 
-#endif  // VALKEYSEARCH_SRC_INDEXES_TEXT_H_
+#endif  // VALKEYSEARCH_SRC_INDEXES_TEXT_H_
```
Aksha1812: ?

src/indexes/text/text.cc
```diff
@@ -0,0 +1,61 @@
+/*
```
Aksha1812: remove file

src/commands/ft_create_parser.cc
```diff
@@ -472,6 +642,7 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
 
     identifier_names.insert(attribute->identifier());
   }
+
```
Aksha1812, line 697: ?

src/commands/ft_create_parser.cc
```diff
@@ -388,23 +520,36 @@ absl::StatusOr<data_model::Attribute *> ParseAttributeArgs(
...
-  for (const auto &attribute : index_schema_proto.attributes()) {
-    const auto &index = attribute.index();
-    if (index.index_type_case() ==
-        data_model::Index::IndexTypeCase::kVectorIndex) {
-      return true;
-    }
-  }
-  return false;
-}
+// bool HasVectorIndex(const data_model::IndexSchema &index_schema_proto) {
```
Aksha1812: I assume you are going to update this

src/indexes/text_index.h
Nivesh-01: there will be a commented added there in the next iteration. Here it is just deleted but then added back again

.gitignore
Nivesh-01: I think there was some talk about it previously in the meeting and this change would essentially remove the need to always check and also prevent accidental pushing of personal changes.
Lets discuss this with everyone before I change this.

src/commands/ft_create_parser.cc
Nivesh-01: No, this was code that was there previously and is not currently required since we are removing this check. Commented it instead of removing it because it might be of use later on.

Aksha1812: Which check is removed. why are we touching vector indexing code?

src/indexes/text_index.h
Aksha1812: I don't think you need to change anything here for your task

.gitignore
Aksha1812: even if it is needed. shouldn't be part of this change.

src/commands/ft_create_parser.cc
Nivesh-01: This was to see if we ftCreateFunction had a vector in its command or not. This was also addressed in one of the other comments by allen

Aksha1812:
@Nivesh-01,  I would recommend to remove all other changes unrelated to current scope of this PR. Changes only pertaining to Updated parsing and protobuff population should be part of this PR.

Nivesh-01:
I have to make changes to other files for my changes to build. If I do not add these changes my work will not build. That is the reason I pushed changes yesterday, those were changes for my work only but that was not building. The push today builds and works just as we expect it to work but it has additional files.

src/indexes/text/radix_tree.h
VoletiRam, line 1: You don't need to touch this file. Please keep changes relevant to the scope of your work.

src/indexes/text_index.h
```diff
@@ -35,15 +36,13 @@ struct TextIndex {
   // thus this object becomes responsible for cross-tree locking issues.
   // Multiple locking strategies are possible. TBD (a shared-ed word lock table should work well)
   //
-  std::shared_ptr<RadixTree<std::unique_ptr<Postings *>, false>>> prefix_;
-  std::optional<text::RadixTree> suffix_;
-
-  absl::hashmap<Key, text::RadixTree> reverse_;
```
VoletiRam: why do you remove this?

src/indexes/vector_base.cc
```diff
@@ -99,6 +99,13 @@ bool InlineVectorEvaluator::EvaluateNumeric(
   return predicate.Evaluate(value);
 }
 
+bool InlineVectorEvaluator::EvaluateText(
+    const query::TextPredicate &predicate) {
+  // TODO: Implement text evaluation logic
+  // For now, return true as a placeholder
+  return true;
+}
+
```
VoletiRam: Not sure why you need to evaluate text in vector file

src/query/predicate.cc
```diff
@@ -109,6 +110,18 @@ ComposedPredicate::ComposedPredicate(std::unique_ptr<Predicate> lhs_predicate,
...
+    : Predicate(PredicateType::kText),
+      index_(index),
+      alias_(alias),
+      identifier_(vmsdk::MakeUniqueValkeyString(identifier)),
+      query_text_(query_text) {}
+
+bool TextPredicate::Evaluate(Evaluator& evaluator) const {
+  return evaluator.EvaluateText(*this);
+}
+
```
VoletiRam: We can leave the extension of text evaluation to query parser work. This is beyond the scope of the PR.

src/query/predicate.h
```diff
@@ -127,6 +131,29 @@ class TagPredicate : public Predicate {
...
+  }
+  const std::string& GetQueryText() const { return query_text_; }
+
+ private:
+  const indexes::Text* index_;
+  std::string alias_;
+  vmsdk::UniqueValkeyString identifier_;
+  std::string query_text_;
+};
+
```
VoletiRam: We can leave the extension of text evaluation to query parser work. This is beyond the scope of the PR.

src/query/response_generator.cc
```diff
@@ -109,6 +109,12 @@ class PredicateEvaluator : public query::Evaluator {
     return predicate.Evaluate(&out_numeric.value());
   }
 
+  bool EvaluateText(const query::TextPredicate &predicate) override {
+    // TODO: Implement text evaluation logic
+    // For now, return true as a placeholder
+    return true;
+  }
+
```
VoletiRam: We can leave the extension of text evaluation to query parser work. This is beyond the scope of the PR.

.gitignore
VoletiRam: +1, limit your change to your scope of the task

src/indexes/vector_base.cc
Nivesh-01: It is for what we had discussed in the meeting when Allen was in Seattle. It is inline vector parsing for an hybrid query with text

Aksha1812:
https://github.com/valkey-io/valkey-search/pull/251#issuecomment-3141454167 I don't see lexer.h, radixtree.h etc.. in CmakeLists.txt so i think would be safe to remove your changes from them as they are not being compiled. and if there are any other build failures you can keep those files out of CmakeLists.txt as we don't want to worry about building yet. so that you can keep only the changes in scope

src/query/predicate.cc
Nivesh-01: Agreed but this is also one of those things that was being flagged when I was trying to build after rebase.

src/query/predicate.h
Nivesh-01: same as above

src/query/response_generator.cc
Nivesh-01: same reason as above

.gitignore
Nivesh-01: Changed it. Will be visible in the next push

src/indexes/text/radix_tree.h
Nivesh-01: updated this to only change code to build everything up.

src/indexes/text_index.h
Nivesh-01: Some of it is syntax, other is redundant code

src/commands/ft_create_parser.cc
KarthikSubbarao: If there any issues observed with the valkey-search code and they are not related to FTS / your PR, I would suggest leaving them untouched in the fulltext branch.

If we would like to clean up unrelated code, we can do this in the main branch first (not in the fulltext) branch to avoid merge conflicts as much as possible. These changes will eventually be merged into the fulltext branch

src/commands/ft_create_parser.cc
```diff
@@ -401,23 +497,36 @@ absl::StatusOr<data_model::Attribute *> ParseAttributeArgs(
...
+//     const auto &index = attribute.index();
+//     if (index.index_type_case() ==
+//         data_model::Index::IndexTypeCase::kVectorIndex) {
+//       return true;
+//     }
+//   }
+//   return false;
+// }
+
+// bool HasTextIndex(const data_model::IndexSchema &index_schema_proto) {
```
KarthikSubbarao: What is this used for? Is it needed for any current or upcoming PR?

src/commands/ft_create_parser.cc
KarthikSubbarao: Thanks. It is missing from words which we would consider as "stop words", but we can follow this for compatibility

src/indexes/text.cc
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the name of Redis nor the names of its contributors may be used
```
BCathcart: I think we need to establish a consistent copyright header. This seems outdated since it mentions "Redis".

Most other files look to use the following so let's use this here as well:
```
/*
 * Copyright (c) 2025, valkey-search contributors
 * All rights reserved.
 * SPDX-License-Identifier: BSD 3-Clause
 *
 */
```

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +555,137 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
     return absl::InvalidArgumentError(NotSupportedParamErrorMsg(kFilterParam));
   }
-  VMSDK_RETURN_IF_ERROR(ParseLanguage(itr, index_schema_proto));
-  VMSDK_RETURN_IF_ERROR(ParseScore(itr, index_schema_proto));
-  VMSDK_ASSIGN_OR_RETURN(
-      res, vmsdk::IsParamKeyMatch(kPayloadFieldParam, false, itr));
-  if (res) {
-    return absl::InvalidArgumentError(
-        NotSupportedParamErrorMsg(kPayloadFieldParam));
+  // Parse global text parameters before SCHEMA
```
KarthikSubbarao, line 547: Did we test out FT.CREATE in an integration test as we talked about?

Example: https://github.com/valkey-io/valkey-search/blob/main/integration/test_query_parser.py

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +555,137 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
+  SchemaGlobTextParams global_text_defaults;
+  // Initialize with defaults for each parse call
+  global_text_defaults.punctuation = ",.<>{}[]\"':;!@#$%^&*()-+=~/\\|";
```
KarthikSubbarao: `",.<>{}[]\"':;!@#$%^&*()-+=~/\\|"` would be good to store as a static constant instead of just inside this function

src/commands/ft_create_parser.cc
Nivesh-01: No, this did the same work as the HasVectorIndex function, I was using it before but not anymore. I kept it just in case it was needed later

src/commands/ft_create_parser.h
```diff
@@ -40,6 +42,24 @@ struct FTCreateVectorParameters {
   std::unique_ptr<data_model::VectorIndex> ToProto() const;
 };
 
+// Global text parameters (per-index) - populated in IndexSchema
+struct SchemaGlobTextParams {
+  std::string punctuation{",.<>{}[]\"':;!@#$%^&*()-+=~/\\\\"};
```
KarthikSubbarao: Seems like this is duplicated.

There is another line which holds this: `,.<>{}[]\"':;!@#$%^&*()-+=~/\\|`.

Why are they different? Can we use the a single static constant please?

src/commands/ft_create_parser.cc
Nivesh-01: done

src/indexes/text.cc
Nivesh-01: Done in all of my files. There are some License files that still have the old one and might need to be changed later.

src/commands/ft_create_parser.cc
KarthikSubbarao: Let us not add code if it is not used in the PR (if we don't forsee a use)

src/commands/ft_create_parser.h
```diff
@@ -40,6 +42,24 @@ struct FTCreateVectorParameters {
   std::unique_ptr<data_model::VectorIndex> ToProto() const;
 };
 
+// Global text parameters (per-index) - populated in IndexSchema
+struct SchemaGlobTextParams {
```
KarthikSubbarao: Question: Why do we call it global? This applies to a single Index, right?

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +555,137 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
+    }
+    
+    // Track current position to detect if no parameter was consumed
+    auto initial_distance = itr.DistanceEnd();
+    
+    // Try SCORE parameter
+    auto score_status = ParseScore(itr, index_schema_proto);
+    if (!score_status.ok()) {
+      return score_status;  // Return the score error immediately
+    }
```
allenss-amazon, line 575: VMSDK_RETURN_IF_ERROR ?

.vscode/settings.json
Nivesh-01: done

src/index_schema.proto
```diff
@@ -29,8 +29,22 @@ message IndexSchema {
   repeated Attribute attributes = 6;
   Stats stats = 7;
   optional uint32 db_num = 8;
+
+  string punctuation = 9;
+  bool with_offsets = 10;
+  repeated string stop_words = 11;
+  bool nostem = 12;
+  int32 min_stem_size = 13;
```
KarthikSubbarao: 1. Some of these are duplicated with the TextIndex
2. Any reason why these are not added per `TextIndex`?

Nivesh-01: 1- It is meant to be like that. So the ones here would be for all indexes so if there are 3 text indexes they will all have this in common while textIndex options can vary. So if we have set punctuation it will be the same for all indexes while minstem size can either be set for all the indexes or it can be overridden for one index.

2- I think this is how we designed it initially. we wanted some to be for each text index while others for all of them. Some like language will be for all indexes - vector, numeric, text ...

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +555,137 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
+    
+    VMSDK_ASSIGN_OR_RETURN(res, vmsdk::IsParamKeyMatch(kLanguageFieldParam, false, itr));
+    if (res) {
+      return absl::InvalidArgumentError(NotSupportedParamErrorMsg(kLanguageFieldParam));
+    }
+    
+    VMSDK_ASSIGN_OR_RETURN(res, vmsdk::IsParamKeyMatch(kScoreFieldParam, false, itr));
+    if (res) {
+      return absl::InvalidArgumentError(NotSupportedParamErrorMsg(kScoreFieldParam));
+    }
```
allenss-amazon: Hmm. As a philosophical issue, I'm not sure I want to generate a different error message for a keyword we recognize that we might never implement vs. a keyword that we don't recognize.

src/commands/ft_create_parser.h
Nivesh-01: this will be for all fields in the schema. So if there are 5 text fields in the schema then they will be for all of them hence global (global for all fields) while the other one will be per text field. I can change the name but at that time this made more sense to distinguish between the two.

src/commands/ft_create_parser.h
Nivesh-01: yes done , will push the changes once there are no more comments to be added

src/index_schema.proto
Nivesh-01: You can also see @VoletiRam's discussion with allen in the group. We also discussed about this in the meeting when allen was in seattle

testing/ft_create_parser_test.cc
```diff
@@ -556,29 +617,12 @@ INSTANTIATE_TEST_SUITE_P(
...
-                 .separator = ",",
-                 .case_sensitive = false,
-             }},
-             .expected = {.index_schema_name = "idx1",
-                        .on_data_type = data_model::ATTRIBUTE_DATA_TYPE_HASH,
-                        .attributes = {{
-                            .identifier = "hash_field1",
-                            .attribute_alias = "hash_field11",
-                            .indexer_type = indexes::IndexerType::kTag,
-                        }}},
```
allenss-amazon: Was the deletion of this test case intentional?

testing/ft_create_parser_test.cc
```diff
@@ -932,6 +976,683 @@ INSTANTIATE_TEST_SUITE_P(
...
+                 .no_stem = false,
+                 .min_stem_size = 4,
+             }},
+             .expected = {
+                 .index_schema_name = "idx1",
+                 .on_data_type = data_model::ATTRIBUTE_DATA_TYPE_HASH,
+                 .attributes = {{
+                     .identifier = "text_field",
+                     .attribute_alias = "text_field",
+                     .indexer_type = indexes::IndexerType::kText,
```
allenss-amazon: No expected value for punctuation?

testing/ft_create_parser_test.cc
allenss-amazon: No expected value for punctuation?

testing/ft_create_parser_test.cc
```diff
@@ -932,6 +976,683 @@ INSTANTIATE_TEST_SUITE_P(
...
+                    .attribute_alias = "text_field",
+                    .indexer_type = indexes::IndexerType::kText,
+                }},
+                .global_text_params = {
+                    .punctuation = ",.<>{}[]\"':;!@#$%^&*()-+=~/\\|",
+                    .stop_words = {
+                        "a", "is", "the", "an", "and", "are", "as", "at", "be", "but", "by", "for",
+                        "if", "in", "into", "it", "no", "not", "of", "on", "or", "such", "that", "their",
+                        "then", "there", "these", "they", "this", "to", "was", "will", "with"},
+                    .language = data_model::Language::LANGUAGE_ENGLISH,
```
allenss-amazon: You mean the default stop_words here, right? Maybe you can arrange to not have to repeat them when they are the default value.

testing/ft_create_parser_test.cc
```diff
@@ -932,6 +976,683 @@ INSTANTIATE_TEST_SUITE_P(
...
+                 .min_stem_size = 4,
+             }},
+             .expected = {
+                 .index_schema_name = "idx1",
+                 .on_data_type = data_model::ATTRIBUTE_DATA_TYPE_HASH,
+                 .attributes = {{
+                     .identifier = "text_field",
+                     .attribute_alias = "text_field",
+                     .indexer_type = indexes::IndexerType::kText,
+                 }}
```
allenss-amazon: No check for the correctly parser stopword list?

testing/ft_create_parser_test.cc
```diff
@@ -932,6 +976,683 @@ INSTANTIATE_TEST_SUITE_P(
...
```
allenss-amazon: NO check for correctly parsed Punctionation?

testing/ft_create_parser_test.cc
Nivesh-01: This test is going to fail and the punctuation will not be parsed since it is inside the schema which is wrong. Since it is failing having expected does not make sense. The expected is left from one of the older tests where punctuation being inside was not a problem. I will remove .expected from here since it will never be checked at all

testing/ft_create_parser_test.cc
Nivesh-01: same as above. When I changed code to not support these tests I did not remove that part. Some of the tests are redundant like this one since essentially punctuation with or without quotes when inside the schema will not be parsed same as the test before so I will remove the test

src/commands/ft_create_parser.h
allenss-amazon: Naming is difficult, it's global to the schema. I agree that some kind of naming differentiation is good. Why not just drop the "Glob" part, SchemaTextParams seems OK to me.

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +555,137 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
+      return score_status;  // Return the score error immediately
+    }
+    if (itr.DistanceEnd() < initial_distance) {
+      continue; // Parameter was consumed, continue to next iteration
+    }
+    
+    // Try LANGUAGE parameter
+    auto language_status = ParseLanguage(itr, index_schema_proto);
+    if (!language_status.ok()) {
+      return language_status;  // Return the language error immediately
```
KarthikSubbarao, line 583: Should we be returning an error here?

Nivesh-01: Yes so essentially we are catching the error in ParseLanguage but we need to still check and return the error from here, cause if we do not do that the loop will still continue and give weird errors.

src/index_schema.proto
VoletiRam: change to uint32 instead of int32 for min_stem_size.

src/indexes/text.cc
```diff
@@ -0,0 +1,138 @@
...
+#include "src/indexes/text.h"
+#include "src/indexes/text_index.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "src/indexes/text/posting.h"
+
+namespace valkey_search::indexes {
+
+// TextIndex constructor implementation
+TextIndex::TextIndex(const data_model::TextIndex& text_index_proto)
```
BCathcart: Why is this here?

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +555,137 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
+    // Try unsupported field parameters
+    VMSDK_ASSIGN_OR_RETURN(res, vmsdk::IsParamKeyMatch(kPayloadFieldParam, false, itr));
+    if (res) {
+      return absl::InvalidArgumentError(NotSupportedParamErrorMsg(kPayloadFieldParam));
+    }
+    
+    VMSDK_ASSIGN_OR_RETURN(res, vmsdk::IsParamKeyMatch(kLanguageFieldParam, false, itr));
+    if (res) {
+      return absl::InvalidArgumentError(NotSupportedParamErrorMsg(kLanguageFieldParam));
+    }
```
KarthikSubbarao: If this already covered by `ParseLanguage(`, let's remove this

src/commands/ft_create_parser.cc
KarthikSubbarao: This is already covered

KarthikSubbarao (approved):

src/index_schema.proto
```diff
@@ -29,8 +29,20 @@ message IndexSchema {
...
+  string punctuation = 9;
+  bool with_offsets = 10;
+  repeated string stop_words = 11;
 }
 
+
+message TextIndex {
+  bool with_suffix_trie = 1;
+  bool no_stem = 2;
+  int32 min_stem_size = 3;
```
VoletiRam, line 42: Change this uint32

integration/test_fulltext.py
```diff
@@ -0,0 +1,129 @@
+import pytest
+from valkey import ResponseError
+from valkey.client import Valkey
+from valkey_search_test_case import ValkeySearchTestCaseBase
+from valkeytestframework.conftest import resource_port_tracker
+
+
+class TestFullText(ValkeySearchTestCaseBase):
+
+    def test_basic_text_create_with_stopwords_and_punctuation(self):
```
KarthikSubbarao: Would suggest a single test since each test has its own server by default.

src/commands/ft_create_parser.cc
```diff
@@ -356,22 +378,95 @@ absl::Status ParseTag(vmsdk::ArgsIterator &itr, data_model::Index &index_proto,
...
+  }
+  
+  params.stop_words.clear();
+  for (uint32_t i = 0; i < count; ++i) {
+    // Check if the next argument is a reserved keyword that should not be consumed
+    // VMSDK_ASSIGN_OR_RETURN(auto next_arg, itr.Get());
+    // absl::string_view next_word = vmsdk::ToStringView(next_arg);
+    // if (absl::EqualsIgnoreCase(next_word, kSchemaParam)) {
+    //   return absl::OutOfRangeError("Missing argument for STOPWORDS. The count does not match the number of arguments provided for STOPWORDS");
+    // }
```
VoletiRam, line 437: why was this commented? We can remove the lines if not necessary

src/index_schema.cc
```diff
@@ -82,6 +83,10 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
     case data_model::Index::IndexTypeCase::kNumericIndex: {
       return std::make_shared<indexes::Numeric>(index.numeric_index());
     }
+    case data_model::Index::IndexTypeCase::kTextIndex: {
+      // return std::make_shared<indexes::Text>(index.text_index());
+      return absl::InvalidArgumentError("Text Index is yet to be defined");
+    }
```
VoletiRam, line 89: Why are we not creating index? Isn't it the part of the task to create text index and pass relevant protobuf to constructors?

KarthikSubbarao: The index is not ready to be created yet and the text.cc/.h files don't compile yet.

It would be better for @Aksha1812 to uncomment the 1 line above since he is working on getting the Text Index functional 

This is much better than trying to write 100s of lines of code in this PR just to be replaced in the upcoming PR

src/commands/ft_create_parser.cc
KarthikSubbarao: +1 - @Nivesh-01 Can you check if this is needed or uncomment it if it is required?

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +544,125 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
     return absl::InvalidArgumentError(NotSupportedParamErrorMsg(kFilterParam));
   }
-  VMSDK_RETURN_IF_ERROR(ParseLanguage(itr, index_schema_proto));
-  VMSDK_RETURN_IF_ERROR(ParseScore(itr, index_schema_proto));
-  VMSDK_ASSIGN_OR_RETURN(
-      res, vmsdk::IsParamKeyMatch(kPayloadFieldParam, false, itr));
-  if (res) {
-    return absl::InvalidArgumentError(
-        NotSupportedParamErrorMsg(kPayloadFieldParam));
+  // Parse global text parameters before SCHEMA
```
KarthikSubbarao, line 547: Can we use consistent terminology? "global" -> "index schema"

src/commands/ft_create_parser.cc
```diff
@@ -11,6 +11,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <mutex>  // For std::once_flag
```
KarthikSubbarao, line 14: We can remove this, right?

src/commands/ft_create_parser.cc
```diff
@@ -485,6 +694,7 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
 
     identifier_names.insert(attribute->identifier());
   }
+
```
KarthikSubbarao, line 697: You can remove this new line

src/index_schema.proto
```diff
@@ -29,8 +29,20 @@ message IndexSchema {
   repeated Attribute attributes = 6;
   Stats stats = 7;
   optional uint32 db_num = 8;
+
```
Aksha1812, line 32: @KarthikSubbarao inorder , slop , save_positions these properties need to be added too right? maybe not with this pr but later as needed

src/commands/ft_create_parser.cc
allenss-amazon: Stopwords shouldn't be restricted from using "reserved" words, So removing this code is correct.

src/commands/ft_create_parser.cc
```diff
@@ -401,6 +497,8 @@ absl::StatusOr<data_model::Attribute *> ParseAttributeArgs(
   } else if (index_type == indexes::IndexerType::kNumeric) {
     VMSDK_RETURN_IF_ERROR(
         ParseNumeric(itr, *index_proto, attribute_identifier));
+  } else if (index_type == indexes::IndexerType::kText) {
```
allenss-amazon, line 500: This should be converted to a switch statement.

src/commands/ft_create_parser.cc
allenss-amazon: Here, global could be replaced with per-schema or schema-level.

src/commands/ft_create_parser.cc
allenss-amazon: An unrecognized Language would be an error. So this makes sense. Though we should have used VMSDK_RETURN_IF_ERROR.

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +544,125 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
+    
+    // Track current position to detect if no parameter was consumed
+    auto initial_distance = itr.DistanceEnd();
+    
+    // Try SCORE parameter
+    auto score_status = ParseScore(itr, index_schema_proto);
+    if (!score_status.ok()) {
+      return score_status;  // Return the score error immediately
+    }
+    if (itr.DistanceEnd() < initial_distance) {
```
allenss-amazon, line 576: != instead of < ??

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +544,125 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
-    return absl::InvalidArgumentError(
-        NotSupportedParamErrorMsg(kPayloadFieldParam));
+  // Parse global text parameters before SCHEMA
+  PerIndexTextParams per_index_text_defaults;
+  // Initialize with defaults for each parse call
+  per_index_text_defaults.punctuation = kDefaultPunctuation;
+  per_index_text_defaults.min_stem_size = 4;
+  per_index_text_defaults.with_offsets = true;
+  per_index_text_defaults.no_stem = false;
+  per_index_text_defaults.language = data_model::LANGUAGE_ENGLISH;
```
allenss-amazon, line 554: These are the schema-level defaults, right? I think we should shy away from using "Index" as its so overloaded as to be confusing. Maybe schema_text_defaults would be a better name?

src/commands/ft_create_parser.cc
```diff
@@ -446,14 +544,125 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
...
+    VMSDK_ASSIGN_OR_RETURN(res, vmsdk::IsParamKeyMatch(kNoStopWordsParam, false, itr));
+    if (res) {
+      per_index_text_defaults.stop_words.clear();
+      continue;
+    }
+    
+    VMSDK_ASSIGN_OR_RETURN(res, vmsdk::IsParamKeyMatch(kStopWordsParam, false, itr));
+    if (res) {
+      VMSDK_RETURN_IF_ERROR(ParseStopWords(itr, per_index_text_defaults));
+      continue;
```
allenss-amazon, line 630: Not sure why this sequence of compares isn't redone as a KeyValue parser?

allenss-amazon (changes requested):
Most of my comments are style/nits. There are some things that I'd like to see addressed, but they are primarily internal to this subsystem. I could support merging this version if it's blocking other people and starting a separate CR to address the issues that I've cited.

src/commands/ft_create_parser.cc
Aksha1812: +1

PR #252, opened 2025-07-25
Augment FT.INFO command to provide whole-cluster index information

boda26:
Augment FT.INFO command to provide whole-cluster index information


src/commands/ft_info.cc
```diff
@@ -6,38 +6,180 @@
...
+
+  return VALKEYMODULE_OK;
+}
+
+void Free(ValkeyModuleCtx *ctx, void *privdata) {
+  delete static_cast<InfoAsyncResult *>(privdata);
+}
+
+int Timeout(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
+  return ValkeyModule_ReplyWithSimpleString(ctx, "Request timed out");
```
allenss-amazon: This should be a ReplyWithError.

src/commands/ft_info.cc
```diff
@@ -6,38 +6,180 @@
...
+      return absl::OkStatus();
+    }
+  } else {
+    // More than 2 parameters provided
+    ValkeyModule_ReplyWithError(ctx, vmsdk::WrongArity(kInfoCommand).c_str());
+    return absl::OkStatus();
+  }
+
+  if (is_global) {
+    VMSDK_LOG(NOTICE, ctx) << "==========Using Global Scope==========";
```
allenss-amazon: DEBUG?

src/commands/ft_info.cc
```diff
@@ -6,38 +6,180 @@
...
+    // Global info fanout - aggregate info from all cluster nodes
+    auto parameters = std::make_unique<query::info_fanout::InfoParameters>();
+    parameters->index_name = std::string(index_schema_name);
+
+    auto targets = query::info_fanout::GetInfoTargetsForFanout(ctx);
+    VMSDK_LOG(NOTICE, ctx) << "Found " << targets.size() << " fanout targets:";
```
allenss-amazon: DEBUG

src/commands/ft_info.cc
```diff
@@ -6,38 +6,180 @@
...
+
+    for (const auto &target : targets) {
+      VMSDK_LOG(NOTICE, ctx)
```
allenss-amazon: DEBUG

src/commands/ft_info.cc
```diff
@@ -6,38 +6,180 @@
...
+      blocked_client.SetReplyPrivateData(payload.release());
+    };
+
+    return query::info_fanout::PerformInfoFanoutAsync(
+        ctx, targets, ValkeySearch::Instance().GetCoordinatorClientPool(),
+        std::move(parameters), ValkeySearch::Instance().GetReaderThreadPool(),
+        std::move(on_done));
+
+  } else {
+    VMSDK_LOG(NOTICE, ctx) << "==========Using Local Scope==========";
```
allenss-amazon: DEBUG

src/commands/ft_info.cc
```diff
@@ -6,38 +6,180 @@
...
+    VMSDK_ASSIGN_OR_RETURN(
+        auto index_schema,
+        SchemaManager::Instance().GetIndexSchema(
+            ValkeyModule_GetSelectedDb(ctx), index_schema_name));
+    static const auto permissions =
+        PrefixACLPermissions(kInfoCmdPermissions, kInfoCommand);
+    VMSDK_RETURN_IF_ERROR(
+        AclPrefixCheck(ctx, permissions, index_schema->GetKeyPrefixes()));
```
allenss-amazon: Seems like this should be on both local and global paths.

src/coordinator/client.cc
```diff
@@ -165,4 +165,41 @@ void ClientImpl::SearchIndexPartition(
...
+  struct InfoIndexPartitionArgs {
+    ::grpc::ClientContext context;
+    std::unique_ptr<InfoIndexPartitionRequest> request;
+    InfoIndexPartitionResponse response;
+    InfoIndexPartitionCallback callback;
+    std::unique_ptr<vmsdk::StopWatch> latency_sample;
+  };
+  auto args = std::make_unique<InfoIndexPartitionArgs>();
+  // Use a 5-second deadline (matches InfoParameters::timeout_ms default)
+  args->context.set_deadline(
```
allenss-amazon: let's make this configurable.

src/coordinator/coordinator.proto
```diff
@@ -133,3 +137,26 @@ message GetGlobalMetadataRequest {}
...
+  uint64 backfill_scanned_count = 6;
+  uint64 backfill_db_size = 7;
+  uint64 backfill_inqueue_tasks = 8;
+  float backfill_complete_percent = 9;
+  bool backfill_in_progress = 10;
+  uint64 mutation_queue_size = 11;
+  uint64 recent_mutations_queue_delay = 12;
+  string state = 13;
+  string error = 14;
+  uint64 schema_fingerprint = 15;
```
allenss-amazon, line 161: Shouldn't the version also be part of this?

src/commands/ft_info.cc
```diff
@@ -6,38 +6,180 @@
...
+
+  if (!info.exists) {
+    std::string error_msg =
+        absl::StrFormat("Index with name '%s' not found", index_name);
+    return ValkeyModule_ReplyWithError(ctx, error_msg.c_str());
+  }
+
+  if (info.has_schema_mismatch) {
+    return ValkeyModule_ReplyWithError(
+        ctx, "ERR found index schema inconsistency in the cluster");
```
allenss-amazon: "Index is in inconsistent state"

src/query/fanout_template.h
```diff
@@ -0,0 +1,150 @@
...
+        int flags;
+        if (ValkeyModule_GetClusterNodeInfo(ctx, node_id.c_str(), ip, master_id,
+                                            &port, &flags) != VALKEYMODULE_OK) {
+          VMSDK_LOG_EVERY_N_SEC(WARNING, ctx, 1)
+              << "Failed to get node info for node " << node_id
+              << ", skipping node...";
+          continue;
+        }
+        
+        if (flags & VALKEYMODULE_NODE_PFAIL || flags & VALKEYMODULE_NODE_FAIL) {
```
allenss-amazon: flags & (VALKEYMODULE_NODE_PFAIL | VALKEYMODULE_NODE_FAIL)

src/query/fanout_template.h
```diff
@@ -0,0 +1,150 @@
...
+          VMSDK_LOG_EVERY_N_SEC(WARNING, ctx, 1)
+              << "Node " << node_id << " (" << ip
+              << ") is failing, skipping for fanout...";
+          continue;
+        }
+        
+        // Only select master nodes
+        if (flags & VALKEYMODULE_NODE_MASTER) {
+          if (flags & VALKEYMODULE_NODE_MYSELF) {
```
allenss-amazon, line 104: flags & (VALKEMODULE_NODE_MASTER | VALKEYMODULE_NODE_MYSELF)

src/query/fanout_template.h
```diff
@@ -0,0 +1,150 @@
...
+        int flags;
+        if (ValkeyModule_GetClusterNodeInfo(ctx, node_id.c_str(), ip, master_id,
+                                            &port, &flags) != VALKEYMODULE_OK) {
+          VMSDK_LOG_EVERY_N_SEC(WARNING, ctx, 1)
+              << "Failed to get node info for node " << node_id
+              << ", skipping node...";
+          continue;
+        }
+        auto master_id_str = std::string(master_id, VALKEYMODULE_NODE_ID_LEN);
+        if (flags & VALKEYMODULE_NODE_PFAIL || flags & VALKEYMODULE_NODE_FAIL) {
```
allenss-amazon: same as above

src/query/fanout_template.h
```diff
@@ -0,0 +1,150 @@
...
+        size_t index = absl::Uniform(gen, 0u, node_indices.size());
+        size_t selected_node_index = node_indices.at(index);
+        
+        // Re-fetch node info only for the selected node
+        std::string node_id(nodes.get()[selected_node_index], VALKEYMODULE_NODE_ID_LEN);
+        char ip[INET6_ADDRSTRLEN] = "";
+        char master_id[VALKEYMODULE_NODE_ID_LEN] = "";
+        int port;
+        int flags;
+        if (ValkeyModule_GetClusterNodeInfo(ctx, node_id.c_str(), ip, master_id,
```
allenss-amazon, line 190: given the testing above, how is this possible?

src/query/info_fanout.cc
```diff
@@ -0,0 +1,385 @@
...
+      : outstanding_requests(outstanding_requests),
+        callback(std::move(callback)),
+        parameters(std::move(parameters)) {}
+
+  void AddResults(coordinator::InfoIndexPartitionResponse& response) {
+    absl::MutexLock lock(&mutex);
+
+    if (response.exists()) {
+      // Log fingerprint for debugging
+      VMSDK_LOG(NOTICE, nullptr)
```
allenss-amazon: DEBUG

src/query/info_fanout.cc
```diff
@@ -0,0 +1,385 @@
...
+          << "Remote node fingerprint for index '" << response.index_name()
+          << "': " << response.schema_fingerprint();
+
+      // Check fingerprint consistency first
+      if (aggregated_result.schema_fingerprint == 0) {
+        aggregated_result.schema_fingerprint = response.schema_fingerprint();
+        VMSDK_LOG(NOTICE, nullptr) << "Set reference fingerprint to: "
+                                   << aggregated_result.schema_fingerprint;
+      } else if (aggregated_result.schema_fingerprint !=
```
allenss-amazon: This treats a fingerprint of zero as a special case. What cases are you referring to? BTW, a fingerprint of 0 is a completely valid value -- right?

I think you're not properly recognizing the "first" result from the "next" result. The first result deserves special handling -- and it will simplify and clarify the code below.

src/query/info_fanout.cc
```diff
@@ -0,0 +1,385 @@
...
+                              InfoResponseCallback callback,
+                              std::unique_ptr<InfoParameters> parameters)
+      : outstanding_requests(outstanding_requests),
+        callback(std::move(callback)),
+        parameters(std::move(parameters)) {}
+
+  void AddResults(coordinator::InfoIndexPartitionResponse& response) {
+    absl::MutexLock lock(&mutex);
+
+    if (response.exists()) {
```
allenss-amazon: What does it mean if exists is false? If I'm reading the code right, you silently ignore this response. What conditions could that happen in?

src/query/info_fanout.cc
```diff
@@ -0,0 +1,385 @@
...
+        } else {
+          aggregated_result.error += ";" + response.error();
+        }
+      }
+    }
+  }
+
+  // Handle local result (similar to how search handles
+  // std::deque<indexes::Neighbor>)
+  void AddResults(InfoResult& local_result) {
```
allenss-amazon: I think it would be simpler to convert an InfoResult into a coordinator::InfoIndexPartitionResponse (which should have the same fields with the same semantics --right?). And then call the code above. Wouldn't that same code be used to generate the returned result?

src/query/info_fanout.cc
```diff
@@ -0,0 +1,385 @@
...
+      aggregated_result.error += ";" + error_message;
+    }
+  }
+
+  ~InfoPartitionResultsTracker() {
+    absl::MutexLock lock(&mutex);
+    absl::StatusOr<InfoResult> result = aggregated_result;
+    callback(result, std::move(parameters));
+  }
+};
```
allenss-amazon: How does this destructor get triggered? Is there a specific place in the code where you're explicitly deleting it? If so, I think that stylistically it's cleaner to have a specific routine to deliver the result callback rather than to piggyback that action on the destructor.

src/query/info_fanout.cc
```diff
@@ -0,0 +1,385 @@
...
+    result.hash_indexing_failures = data.hash_indexing_failures;
+    result.backfill_scanned_count = data.backfill_scanned_count;
+    result.backfill_db_size = data.backfill_db_size;
+    result.backfill_inqueue_tasks = data.backfill_inqueue_tasks;
+    result.mutation_queue_size = data.mutation_queue_size;
+    result.recent_mutations_queue_delay = data.recent_mutations_queue_delay;
+    result.backfill_in_progress = data.backfill_in_progress;
+    result.backfill_complete_percent = data.backfill_complete_percent;
+    result.state = data.state;
+    result.error = "";
```
allenss-amazon: Feels like you should be using absl::StatusOr<InfoResult>, not putting the error message directly into the InfoResult object.

src/query/info_fanout.cc
```diff
@@ -0,0 +1,385 @@
...
+
+InfoResult GetLocalInfoResult(ValkeyModuleCtx* ctx,
+                              const std::string& index_name) {
+  InfoResult result;
+
+  // Get local index schema and extract info
+  auto index_schema_result = SchemaManager::Instance().GetIndexSchema(
+      ValkeyModule_GetSelectedDb(ctx), index_name);
+
+  if (index_schema_result.ok()) {
```
allenss-amazon: Stylistically, it's usually better if the "shorter" part of the "IF" comes before the "longer" part. in other words, instead of: 
```
if (condition) {
   // Large block of code
} else {
   // Small block of code
}
```
You should prefer:
```
if (!condition) {
   // Small block of code
} else {
   // Large block of code
}
```
The reason for the preference is that it's easier to see the "condition" when you're reading the "else" code.

Another minor style preference is for functions that quickly return an error.
```
absl::StatusOr<xxx> function() {
   ...
   if (Validation) {
      return absl::StatusNotOkError(...);
   }
   ... // Normal path processing
```
Here you're keeping the indent level shorter for the main processing.

src/query/info_fanout.cc
```diff
@@ -0,0 +1,385 @@
...
+    if (info_targets.size() >= 30 && thread_pool->Size() > 1) {
+      PerformRemoteInfoRequestAsync(std::move(request_copy), node.address,
+                                    coordinator_client_pool, tracker,
+                                    thread_pool);
+    } else {
+      PerformRemoteInfoRequest(std::move(request_copy), node.address,
+                               coordinator_client_pool, tracker);
+    }
+  }
+
```
allenss-amazon: This command isn't time-critical, so this particular optimization is premature. Pick one path or the other.

src/schema_manager.cc
```diff
@@ -64,6 +68,21 @@ vmsdk::config::Number &GetMaxIndexes() {
...
+                                 std::numeric_limits<int32_t>::max())
+        .WithValidationCallback(
+            CHECK_RANGE(1, std::numeric_limits<int32_t>::max(),
+                        kIndexSchemaBackfillBatchSizeConfig))
+        .Build();
+
+vmsdk::config::Number &GetBackfillBatchSize() {
+  return dynamic_cast<vmsdk::config::Number &>(*backfill_batch_size);
+}
+
```
allenss-amazon, line 85: feels like this isn't part of the PR, perhaps it needs to be rebased.

testing/coordinator/info_client_server_test.cc
```diff
@@ -0,0 +1,354 @@
...
+      return grpc::Status::OK;
+    }
+  }
+  
+  // We need to implement other methods even if we don't use them
+  grpc::Status GetGlobalMetadata(
+      grpc::ServerContext* context,
+      const GetGlobalMetadataRequest* request,
+      GetGlobalMetadataResponse* response) override {
+    return grpc::Status(grpc::StatusCode::UNIMPLEMENTED, "Not implemented in fake service");
```
allenss-amazon: assert.

testing/coordinator/info_client_server_test.cc
```diff
@@ -0,0 +1,354 @@
...
+    server_address_ = "localhost:" + std::to_string(selected_port_);
+    
+    // Create client using MakeInsecureClient with proper context
+    client_ = ClientImpl::MakeInsecureClient(
+        vmsdk::MakeUniqueValkeyDetachedThreadSafeContext(&fake_ctx_),
+        server_address_);
+    ASSERT_NE(client_, nullptr) << "Failed to create client";
+    
+    // Give server time to start
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
allenss-amazon: Why are you waiting? More specifically, what are you waiting for? If there's some specific event that needs to happen before terminating this function, then wait for it explicitly. If there isn't then remove the wait.

testing/coordinator/info_client_server_test.cc
```diff
@@ -0,0 +1,354 @@
...
+  // Wait for callback with timeout
+  auto status_future = status_promise.get_future();
+  auto response_future = response_promise.get_future();
+  
+  ASSERT_EQ(status_future.wait_for(std::chrono::seconds(5)), std::future_status::ready);
+  ASSERT_EQ(response_future.wait_for(std::chrono::seconds(5)), std::future_status::ready);
+  
+  grpc::Status status = status_future.get();
+  InfoIndexPartitionResponse response = response_future.get();
+
```
allenss-amazon: Seems like this is cut/paste, perhaps making it a shared function would make this code clearer.

testing/coordinator/info_client_server_test.cc
```diff
@@ -0,0 +1,354 @@
...
+    EXPECT_TRUE(status.ok()) << "Request " << i << " failed: " << status.error_message();
+    EXPECT_TRUE(response.exists());
+    EXPECT_EQ(response.index_name(), "existing_index");
+    EXPECT_EQ(response.num_docs(), 42);
+    EXPECT_EQ(response.num_records(), 100);
+  }
+  
+  // Verify bytes were counted for all requests
+  EXPECT_GT(Metrics::GetStats().coordinator_bytes_out.load(), num_requests * 10);
+  EXPECT_GT(Metrics::GetStats().coordinator_bytes_in.load(), num_requests * 10);
```
allenss-amazon: why 10?

testing/coordinator/info_client_test.cc
```diff
@@ -0,0 +1,146 @@
...
+      });
+      
+  EXPECT_TRUE(callback_called);
+  
+  // Verify byte counters match the actual sizes of the protobuf objects
+  EXPECT_EQ(Metrics::GetStats().coordinator_bytes_out.load(), actual_request_size);
+  EXPECT_EQ(Metrics::GetStats().coordinator_bytes_in.load(), actual_response_size);
+}
+
+// Test that we don't count incoming bytes for error responses
```
allenss-amazon: Is it required that not count the bytes for the error responses?

src/query/fanout_template.h
boda26: This is not equivalent, I need to first detect if it is a master node, then classify based on whether it is a local node or not.

src/commands/ft_info.cc
```diff
@@ -90,6 +91,64 @@ int Timeout(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
...
+        absl::StrFormat("Primary index with name '%s' not found", index_name);
+    return ValkeyModule_ReplyWithError(ctx, error_msg.c_str());
+  }
+
+  if (info.has_schema_mismatch) {
+    return ValkeyModule_ReplyWithError(
+        ctx, "ERR found primary index schema inconsistency in the cluster");
+  }
+
+  ValkeyModule_ReplyWithArray(ctx, 10);
```
allenss-amazon: Let's use the DEFERRED array length thing, that way you can edit the list without having to take off your shoes to count entries.

src/commands/ft_info.cc
```diff
@@ -90,6 +91,64 @@ int Timeout(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
...
+  ValkeyModule_ReplyWithSimpleString(ctx, "global");
+  ValkeyModule_ReplyWithSimpleString(ctx, "true");
+  ValkeyModule_ReplyWithSimpleString(ctx, "index_name");
+  ValkeyModule_ReplyWithSimpleString(ctx, index_name);
```
allenss-amazon: Repeating the input parameters isn't the Valkey way.

src/commands/ft_info.cc
```diff
@@ -121,6 +181,15 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
         return absl::OkStatus();
       }
       is_global = true;
+    } else if (absl::EqualsIgnoreCase(scope, "PRIMARY")) {
+      if (!ValkeySearch::Instance().IsCluster() ||
+          !ValkeySearch::Instance().UsingCoordinator()) {
+        ValkeyModule_ReplyWithError(ctx,
+                                    "ERR PRIMARY keyword requires cluster mode "
```
allenss-amazon: let's leave out the "coordinator" part, that's an under-the-hood implementation detail.

allenss-amazon: You could just say not valid in this configuration or mode.

src/commands/ft_info.cc
```diff
@@ -121,6 +181,15 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
+                                    "with coordinator enabled");
+        return absl::OkStatus();
+      }
+      is_primary = true;
     } else {
       ValkeyModule_ReplyWithError(
           ctx, "ERR Invalid scope parameter. Must be LOCAL or GLOBAL");
```
allenss-amazon: I think this is out of date now.

src/commands/ft_info.cc
```diff
@@ -178,6 +247,43 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
+      VMSDK_LOG(DEBUG, ctx)
+          << "  Target type: "
+          << (target.type == query::fanout::FanoutSearchTarget::Type::kLocal
+                  ? "LOCAL"
+                  : "REMOTE")
+          << ", address: " << target.address;
+    }
+    vmsdk::BlockedClient blocked_client(ctx, primary_info_async::Reply,
+                                        primary_info_async::Timeout,
+                                        primary_info_async::Free, 5000);
```
allenss-amazon: Please make the 5000 a configuable.

src/commands/ft_info.cc
```diff
@@ -178,6 +247,43 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
+    blocked_client.MeasureTimeStart();
+    auto on_done =
+        [blocked_client = std::move(blocked_client)](
+            absl::StatusOr<query::primary_info_fanout::PrimaryInfoResult>
```
allenss-amazon: maybe a using directive is in order here, it would shorten some of these names and make it more readable.

src/query/info_fanout.cc
```diff
@@ -386,7 +386,7 @@ absl::Status PerformInfoFanoutAsync(
 
 std::vector<fanout::FanoutSearchTarget> GetInfoTargetsForFanout(
     ValkeyModuleCtx* ctx) {
-  return fanout::GetSearchTargetsForFanout(ctx, true);
+  return fanout::FanoutTemplate::GetTargets(ctx, true);
```
allenss-amazon: A convenience routine that's one line calling another convenience routine?? Too much indirection like that actually distracts the reader from understanding the code.

src/query/primary_info_fanout.cc
```diff
@@ -0,0 +1,285 @@
...
+    absl::MutexLock lock(&mutex);
+
+    if (response.exists()) {
+      // Log fingerprint for debugging
+      VMSDK_LOG(DEBUG, nullptr)
+          << "Remote node fingerprint for index '" << response.index_name()
+          << "': " << response.schema_fingerprint();
+
+      // Check fingerprint consistency first
+      if (aggregated_result.schema_fingerprint == 0) {
```
allenss-amazon: this is wrong, if you have a legal fingerprint of 0 then the next value is accepted unconditionally and that's exactly the case that should be caught as an error.

You need to special case the fingerprint handling for the first one. I'd recommend std::optional.

src/query/primary_info_fanout.cc
```diff
@@ -0,0 +1,285 @@
...
+        return;
+      }
+      aggregated_result.exists = true;
+      aggregated_result.index_name = response.index_name();
+      aggregated_result.num_docs += response.num_docs();
+      aggregated_result.num_records += response.num_records();
+      aggregated_result.hash_indexing_failures +=
+          response.hash_indexing_failures();
+
+      if (!response.error().empty()) {
```
allenss-amazon: When would it be the case that response.error() was not empty but all of the other fields contained good information? Seems to me that checking for the error should be at the start and once you see an error, either in this message or any previous message, you should just bail as none of the accumulated values are meaningful.

src/query/primary_info_fanout.cc
```diff
@@ -0,0 +1,285 @@
...
+      aggregated_result.hash_indexing_failures +=
+          local_result.hash_indexing_failures;
+
+      if (!local_result.error.empty()) {
+        if (aggregated_result.error.empty()) {
+          aggregated_result.error = local_result.error;
+        } else {
+          aggregated_result.error += ";" + local_result.error;
+        }
+      }
```
allenss-amazon: Again, when do you have an error but good data elsewhere in the message? Wouldn't that be more of a "warning" ?

src/query/primary_info_fanout.cc
```diff
@@ -0,0 +1,285 @@
...
+
+    VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+        << "Error during info fanout: " << error_message;
+
+    if (aggregated_result.error.empty()) {
+      aggregated_result.error = error_message;
+    } else {
+      aggregated_result.error += ";" + error_message;
+    }
+  }
```
allenss-amazon: this code looks a lot like previous code, can be be centralized?

src/query/primary_info_fanout.cc
```diff
@@ -0,0 +1,285 @@
...
+    const std::string& address,
+    coordinator::ClientPool* coordinator_client_pool,
+    std::shared_ptr<PrimaryInfoPartitionResultsTracker> tracker) {
+  auto client = coordinator_client_pool->GetClient(address);
+
+  int timeout_ms;
+  {
+    absl::MutexLock lock(&tracker->mutex);
+    timeout_ms = tracker->parameters->timeout_ms;
+  }
```
allenss-amazon: Is this really variable across all of the different requests?

src/query/info_fanout.cc
```diff
@@ -47,15 +47,15 @@ struct InfoPartitionResultsTracker {
...
+      if (!aggregated_result.schema_fingerprint.has_value()) {
         aggregated_result.schema_fingerprint = response.schema_fingerprint();
         VMSDK_LOG(NOTICE, nullptr) << "Set reference fingerprint to: "
-                                   << aggregated_result.schema_fingerprint;
-      } else if (aggregated_result.schema_fingerprint !=
+                                   << aggregated_result.schema_fingerprint.value();
+      } else if (aggregated_result.schema_fingerprint.value() !=
                  response.schema_fingerprint()) {
         VMSDK_LOG(WARNING, nullptr)
             << "Schema fingerprint mismatch detected! Reference: "
```
allenss-amazon: Hmm, not sure why this needs to be in the log. Maybe this should be prefixed by EVERY_N_SECONDS to avoid spamming the log.

src/query/info_fanout.cc
```diff
@@ -136,21 +136,30 @@ struct InfoPartitionResultsTracker {
 
     if (local_result.exists) {
       // Log fingerprint for debugging
-      VMSDK_LOG(NOTICE, nullptr)
-          << "Local node fingerprint for index '" << local_result.index_name
-          << "': " << local_result.schema_fingerprint;
+      if (local_result.schema_fingerprint.has_value()) {
+        VMSDK_LOG(NOTICE, nullptr)
```
allenss-amazon: DEBUG

src/query/info_fanout.cc
```diff
@@ -136,21 +136,30 @@ struct InfoPartitionResultsTracker {
...
+            << "Local node fingerprint for index '" << local_result.index_name
+            << "': " << local_result.schema_fingerprint.value();
+      } else {
+        VMSDK_LOG(NOTICE, nullptr)
```
allenss-amazon: DEBUG

src/query/primary_info_fanout.cc
```diff
@@ -90,21 +90,30 @@ struct PrimaryInfoPartitionResultsTracker {
...
-          << "': " << local_result.schema_fingerprint;
+      if (local_result.schema_fingerprint.has_value()) {
+        VMSDK_LOG(DEBUG, nullptr)
+            << "Local node fingerprint for index '" << local_result.index_name
+            << "': " << local_result.schema_fingerprint.value();
+      } else {
+        VMSDK_LOG(DEBUG, nullptr)
+            << "Local node fingerprint for index '" << local_result.index_name
+            << "': <not set>";
+      }
```
allenss-amazon: I think you can just print it, if it's not present I believe it will print as nullopt.

src/query/primary_info_fanout.cc
```diff
@@ -90,21 +90,30 @@ struct PrimaryInfoPartitionResultsTracker {
...
-      } else if (aggregated_result.schema_fingerprint !=
-                 local_result.schema_fingerprint) {
+        if (local_result.schema_fingerprint.has_value()) {
+          VMSDK_LOG(DEBUG, nullptr) << "Set reference fingerprint to: "
+                                     << local_result.schema_fingerprint.value();
+        }
+      } else if (local_result.schema_fingerprint.has_value() &&
+                 aggregated_result.schema_fingerprint.value() !=
+                 local_result.schema_fingerprint.value()) {
         VMSDK_LOG(WARNING, nullptr)
```
allenss-amazon: DEBUG

src/query/info_fanout.cc
boda26: I changed to std::optional for initial value

src/query/info_fanout.cc
boda26: When I wrote this code, I haven't completed consistency check yet, so when an index might not exist on some nodes, it would simply return, and do not modify any aggregated results. Now with consistency check applied, I think we can delete this case, since it would directly report an error

src/schema_manager.cc
boda26:
```
void SchemaManager::OnServerCronCallback(ValkeyModuleCtx *ctx,
                                         [[maybe_unused]] ValkeyModuleEvent eid,
                                         [[maybe_unused]] uint64_t subevent,
                                         [[maybe_unused]] void *data) {
  SchemaManager::Instance().PerformBackfill(ctx, options::GetBackfillBatchSize().GetValue());
  ```
  This getter is needed for calling PerformBackfill in this function. backfill_batch_size lives in options namespace and is private, so we need this getter to call it outside the namespace.

src/query/primary_info_fanout.cc
boda26: I will add a configurable timeout for ft.info command which can be set by CONFIG SET

PR #256, opened 2025-07-27
Preliminary radix tree and word iterator

BCathcart:
This is the first iteration of RadixTree and WordIterator that will be used to unblock some early integrations. These will be heavily optimized in the future. Unit testing was done with ✨vibes✨ and will need to be iterated on as well.

### DIFF
--- a/src/indexes/text/radix_tree.h
+++ b/src/indexes/text/radix_tree.h
@@ -47,17 +47,36 @@ into the codebase efficiently enough to be deployed in production code.
 
 */
 
+#include <iostream>
+#include <map>
 #include <memory>
+#include <optional>
 #include <span>
+#include <stack>
+#include <variant>
 
+#include "absl/functional/function_ref.h"
 #include "absl/strings/string_view.h"
-#include "src/text/text.h"
+
+namespace valkey_search {
+namespace text {
+
+// Needed for std::visit
+template <class... Ts>
+struct overloaded : Ts... {
+  using Ts::operator()...;
+};
+template <class... Ts>
+overloaded(Ts...) -> overloaded<Ts...>;
+
+using Byte = uint8_t;
+using BytePath = std::string;
 
 template <typename Target, bool reverse>
 struct RadixTree {
   struct WordIterator;
   struct PathIterator;
-  RadixTree();
+  RadixTree() = default;
 
   //
   // This function is the only way to mutate the RadixTree, all other functions
@@ -76,9 +95,9 @@ struct RadixTree {
   // return value is nullopt then this word is deleted from the RadixTree.
   //
   //
-  void Mutate(absl::string_view word,
-              absl::invokeable<std::option<Target>>(std::option<Target>) >
-                  mutate);
+  void Mutate(
+      absl::string_view word,
+      absl::FunctionRef<std::optional<Target>(std::optional<Target>)> mutate);
 
   // Get the number of words that have the specified prefix in O(len(prefix))
   // time.
@@ -96,9 +115,71 @@ struct RadixTree {
   // Create a Path iterator at a specific starting prefix
   PathIterator GetPathIterator(absl::string_view prefix) const;
 
+ private:
+  /*
+   * This is the first iteration of a RadixTree. It will be optimized in the
+   * future, likely with multiple different representations.
+   *
+   * Right now there are three types of nodes:
+   *    1) Leaf node that has a target and no children.
+   *    2) Branching node that has between 2 and 256 children and may or may not
+   *       have a target.
+   *    3) Compressed node that has a single child of one or more bytes and may
+   *       or may not have a target.
+   *
+   * Differentiating nodes that have multiple children vs a single child takes
+   * inspiration from Rax in the Valkey core. An alternative would be to merge
+   * compressed and branching nodes into one:
+   *
+   * using NodeChildren = std::variant<
+   *     std::monostate,                            // Leaf node
+   *     std::map<BytePath, std::unique_ptr<Node>>  // Internal node
+   * >;
+   *
+   * For example,
+   *
+   *                  [compressed]
+   *                  "te" |
+   *                   [branching]
+   *                "s" /     \ "a"
+   *          [compressed]   [compressed]
+   *          "ting" /           \ "m"
+   *   Target <- [leaf]           [leaf] -> Target
+   *
+   *  would become...
+   *
+   *                     [node]
+   *                  "te" |
+   *                     [node]
+   *             "sting" /     \ "am"
+   *       Target <- [leaf]    [leaf] -> Target
+   *
+   * There is one less level to the graph, but the complexity at the internal
+   * nodes has increased and will be tricky to compress into a performant,
+   * compact format given the varying sized outgoing edges. We'll consider the
+   * implementations carefully when we return to optimize.
+   *
+   */
+  struct Node;
+  using NodeChildren =
+      std::variant<std::monostate,                             // Leaf node
+                   std::map<Byte, std::unique_ptr<Node>>,      // Branching node
+                   std::pair<BytePath, std::unique_ptr<Node>>  // Compressed
+                                                               // node
+                   >;
+  struct Node {
+    uint32_t sub_tree_count;
+    std::optional<Target> target;
+    NodeChildren children;
+  };
+
+  Node root_;
+
+ public:
   //
   // The Word Iterator provides access to sequences of Words and the associated
-  // Postings Object in lexical order.
+  // Postings Object in lexical order. Currently the word iterator assumes the
+  // radix tree is not mutated for the life of the iterator.
   //
   struct WordIterator {
     // Is iterator valid?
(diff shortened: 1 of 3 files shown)


src/indexes/text/radix_tree.h
```diff
@@ -116,7 +197,25 @@ struct RadixTree {
...
     // Access the current location, asserts if !IsValid()
     absl::string_view GetWord() const;
-    Postings& GetPostings() const;
+    Target& GetTarget() const;
+
+   private:
+    friend struct RadixTree;
+    explicit WordIterator(Node* node, absl::string_view prefix)
+        : curr_(node), word_(prefix) {
+      if (curr_ && !curr_->target.has_value()) {
```
allenss-amazon: is curr_ == nullptr ever a valid state?

src/indexes/text/radix_tree.h
```diff
@@ -116,7 +197,25 @@ struct RadixTree {
 
     // Access the current location, asserts if !IsValid()
     absl::string_view GetWord() const;
-    Postings& GetPostings() const;
+    Target& GetTarget() const;
+
+   private:
+    friend struct RadixTree;
+    explicit WordIterator(Node* node, absl::string_view prefix)
```
allenss-amazon: const Node* node ??

src/indexes/text/radix_tree.h
```diff
@@ -116,7 +197,25 @@ struct RadixTree {
...
+    }
+
+    using MapIterator = std::map<Byte, std::unique_ptr<Node>>::iterator;
+    std::stack<std::pair<uint32_t,               // depth
+                         std::pair<MapIterator,  // next sibling iterator
+                                   MapIterator   // end iterator
+                                   >>>
+        stack_;
+    Node* curr_;
+    std::string word_;
```
allenss-amazon, line 215: Do you need to copy or is a string_view sufficient?

src/indexes/text/radix_tree.h
```diff
@@ -149,13 +248,283 @@ struct RadixTree {
     // get current Path. If IsWord is true, then there's a word here....
     absl::string_view GetPath();
 
-    // Get Postings for this word, will assert if !IsWord()
-    Postings& GetPostings() const;
+    // Get the target for this word, will assert if !IsWord()
+    Target& GetTarget() const;
```
allenss-amazon: const Target?

src/indexes/text/radix_tree.h
```diff
@@ -149,13 +248,283 @@ struct RadixTree {
...
     void Defrag();
-  }
+  };
 };
 
+template <typename Target, bool reverse>
+void RadixTree<Target, reverse>::Mutate(
+    absl::string_view word,
+    absl::FunctionRef<std::optional<Target>(std::optional<Target>)> mutate) {
+  assert(!word.empty());
```
allenss-amazon: Let's use CHECK. it has the advantage that you can use inserters (<<) and variables to create better failure messages.

src/indexes/text/radix_tree.h
```diff
@@ -149,13 +248,283 @@ struct RadixTree {
...
+                  new_branches[path[0]] = std::move(child.second);
+                } else {
+                  // Create an intermediate compressed node to the leaf
+                  new_branches[path[0]] = std::make_unique<Node>(
+                      Node{0, std::nullopt,
+                           std::pair{path.substr(1), std::move(child.second)}});
+                }
+                n->children = std::move(new_branches);
+                // Next iteration will hit the branching path for the same node
+                next = n;
```
allenss-amazon, line 320: doesn't this branch need to manipulate "remaining" ??

src/indexes/text/radix_tree.h
```diff
@@ -149,13 +248,283 @@ struct RadixTree {
...
+  while (!remaining.empty()) {
+    std::visit(
+        overloaded{
+            [&](const std::monostate&) { exit = true; },
+            [&](const std::map<Byte, std::unique_ptr<Node>>& children) {
+              const auto it = children.find(remaining[0]);
+              if (it != children.end()) {
+                n = it->second.get();
+              } else {
+                exit = true;
```
allenss-amazon: This suggests there's no state in the WordIterator to distinguish whether it's on a valid word or has hit "end". So, you're planning to check for this dynamically? I think that's fine now, but long term I suspect we want ensure that a test for "end" is cheap as things like the accessor functions etc. are going to assert on that condition.

src/indexes/text/radix_tree.h
```diff
@@ -149,13 +248,283 @@ struct RadixTree {
...
+              }
+            }},
+        n->children);
+    if (exit) {
+      n = nullptr;
+      break;
+    }
+  }
+  return WordIterator(const_cast<Node*>(n), prefix);
```
allenss-amazon: The const_cast shouldn't be required if you properly declare the node pointer.

src/indexes/text/radix_tree.h
```diff
@@ -116,7 +197,25 @@ struct RadixTree {
...
+      if (curr_ && !curr_->target.has_value()) {
+        Next();
+      }
+    }
+
+    using MapIterator = std::map<Byte, std::unique_ptr<Node>>::iterator;
+    std::stack<std::pair<uint32_t,               // depth
+                         std::pair<MapIterator,  // next sibling iterator
+                                   MapIterator   // end iterator
+                                   >>>
```
allenss-amazon: tuple?

allenss-amazon: Also, the default container for stack is a deque. I believe we'll want to use absl::InlinedVector instead with a default size of something that's large enough to avoid memory allocations for normal length words (32?)

src/indexes/text/radix_tree.h
```diff
@@ -76,9 +95,9 @@ struct RadixTree {
   // return value is nullopt then this word is deleted from the RadixTree.
   //
   //
-  void Mutate(absl::string_view word,
-              absl::invokeable<std::option<Target>>(std::option<Target>) >
-                  mutate);
+  void Mutate(
```
Aksha1812, line 100: Does this Mutate() function fill in for Add(), Modify(), Remove() as in https://github.com/valkey-io/valkey-search/blob/main/src/indexes/numeric.h#L44. This would be used in AddRecord()/ ModifyRecord()/ RemoveRecord() here https://github.com/valkey-io/valkey-search/blob/fulltext/src/indexes/text.h#L39 right. lmk if i am wrong. but i guess we need to go through TextIndexObj to do this as in RadixTree would be initialized in the TextIndex and AddRecord() etc.. APIS would call via the TextIndexObj to Mutate on the RadixTree

allenss-amazon: Yes. The optional input and output provide four combinations which are sufficient to implement Add, Modify and Remove. (The no input, no output case [the 4th] makes no sense here ;-)

Aksha1812: thank you. makes sense. @BCathcart in the tests below while testing basic operations. Can you also add a test which does Remove operations using Mutate(). would be nice to see usage before adding it to RemoveRecord() in text.cc file.

src/indexes/text/radix_tree.h
BCathcart: That means we're at the end of the iteration. We will construct a word iterator with `Done()` returning true immediately if `node` is passed as a nullptr. Alternatively `GetWordIterator()` could throw an exception if there are no words that match the prefix.

src/indexes/text/radix_tree.h
BCathcart: I add and subtract from it as I move about the tree so it's not really a view into an existing string

src/indexes/text/radix_tree.h
BCathcart: In this case we've come to a compressed node and realized no characters of this new word match the existing outgoing edge. We then convert the node to a branching node with one outgoing edge for the previously existing edge. Then instead of adding a second outgoing edge for the new word here, we loop again and hit the branching case code above which will create the edge and subtract from remaining.

src/indexes/text/radix_tree.h
BCathcart: I assumed the `Done()` API was how we would test that we have reached the end and it currently checks if `curr_ == nullptr`. `GetWordIterator()` traverses the tree until it finds the first node whose path matches the provided prefix. If no such node exists then `exit` (I could rename this to `no_match`) is set to true and at the end of the loop we break out with `n = nullptr`. The word iterator will then be constructed with `curr_ = nullptr` so that `Done()` returns true immediately. Does that approach make sense?

src/indexes/text/radix_tree.h
BCathcart: Removing a word from the tree isn't currently supported. I'll do it as a follow-up but it shouldn't matter for our first basic end-to-end integration goal.

BCathcart:
@allenss-amazon @yairgott I wasn't aware of this existing PatriciaTree class. At a quick glance there are a lot of similarities to what I have done so far. Since I've already started down the path of a new implementation, I wouldn't be opposed to continuing this one and organically optimizing it for the TEXT case, then later considering collapsing them into a single implementation as Allen mentions. I'm open to opinions on this though.

Side note.. are we using "Patricia Tree" as a synonym for "Radix Tree" here?

allenss-amazon:
The tree implementation for FT requires lexical ordering of iterators. For the existing PatriciaTree that would require changing the hashmap to a btree -- which probably isn't much of a change. It's also unclear to me if the bookkeeping requirements of the two use cases are identical. But clearly it's desirable to end up with a single implementation.

Perhaps my memory is playing tricks on me, but I believe that I had looked at the patricia tree implementation before doing the FT design docs and saw that it was very inefficient. The current implementation seems much better than what I remembered.

allenss-amazon (approved):

PR #261, opened 2025-07-29
Changes made to the file structure and inheritance for textIndex

Nivesh-01:
Deleted text.h from index/text dir because it was redundant
Text_index will have all text based implementations like the radix tree which will be called on by the TEXT.h. We will not have it inherit from indexBase.
Only Text.h in indexes dir inherits from indexBase


src/indexes/text_index.h
```diff
@@ -4,7 +4,7 @@
 namespace valkey_search {
 namespace indexes {
 
-struct Text : public IndexBase {
```
BCathcart, line 7: Should we rename this to "TextIndex"? I believe the class in `src/indexes/text.h` that actually implements `IndexBase` is already called "Text"

src/indexes/text/text.h
```diff
@@ -1,87 +0,0 @@
...
-  //
-  // The main query data structure maps Words into Postings objects. This
-  // is always done with a prefix tree. Optionally, a suffix tree can also be maintained.
-  // But in any case for the same word the two trees must point to the same Postings object,
-  // which is owned by this pair of trees. Plus, updates to these two trees need
-  // to be atomic when viewed externally. The locking provided by the RadixTree object
-  // is NOT quite sufficient to guarantee that the two trees are always in lock step.
-  // thus this object becomes responsible for cross-tree locking issues.
-  // Multiple locking strategies are possible. TBD (a shared-ed word lock table should work well)
-  //
```
BCathcart, line 62: We could preserve this description and the one for `by_key_` below in `text_index.h` for now

src/indexes/text/text.h
```diff
@@ -1,87 +0,0 @@
...
-  virtual std::unique_ptr<data_model::Index> ToProto() const override;
-  virtual void ForEachTrackedKey(
-      absl::AnyInvocable<void(const InternedStringPtr&)> fn) const override;
-
-  virtual uint64_t GetRecordCount() const override;
-
- private:
-  // Each text field is assigned a unique number within the containing index, this is used
-  // by the Postings object to identify fields.
-  size_t text_field_number;
```
BCathcart, line 47: We could add this field in the other `text.h` file we're keeping. Seems like something we'll need

src/indexes/text_index.h
Aksha1812: +1

src/indexes/text/text.h
```diff
@@ -1,87 +0,0 @@
...
-#include <concepts>
-#include <memory>
-
-#include "src/utils/string_interning.h"
-
-namespace valkey_search {
-namespace text {
-
-using Key = vmsdk::InternedStringPtr;
-using Position = uint32_t;
```
Aksha1812, line 19: yes some of this other stuff is missing in the src/indexes/text.h as well. make sure you safely move contents from this file to another

Aksha1812:
Also src/indexes/text.h file is importing `#include "src/text/text.h"` which needs to be removed instead import `#include "src/indexes/text_index.h"` also fix other occurences of importing `#include "src/text/text.h"` or `#include "src/indexes/text/text.h"` if any are remaining

src/indexes/text_index.h
Nivesh-01: done

src/indexes/text/text.h
Nivesh-01: done

src/indexes/text/text.h
Nivesh-01: done

src/indexes/text/text.h
Nivesh-01: done

src/indexes/text.h
```diff
@@ -53,6 +52,13 @@ class Text : public IndexBase {
     return absl::OkStatus();
   }
 
+  private:
+  // Each text field is assigned a unique number within the containing index, this is used
+  // by the Postings object to identify fields.
+  size_t text_field_number;
+  std::shared_ptr<TextIndex> text_;
```
Aksha1812, line 59: TextIndex is not imported in this file `#include "src/indexes/text_index.h. see if anything else is remaining like this i can see sources for IndexBase, InternedStringPtr, absl::Status classes are not imported. you can refer src/indexes/numeric.h. include classes needed here.

src/indexes/text.h
```diff
@@ -53,6 +52,13 @@ class Text : public IndexBase {
     return absl::OkStatus();
   }
 
+  private:
```
Aksha1812, line 55: there are two private sections in the Text class right now

src/indexes/text.h
```diff
@@ -30,7 +30,6 @@
 #ifndef VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 #define VALKEYSEARCH_SRC_INDEXES_TEXT_H_
```
Aksha1812, line 32: spdx copyright*

src/indexes/text.h
```diff
@@ -53,6 +52,13 @@ class Text : public IndexBase {
...
   }
 
+  private:
+  // Each text field is assigned a unique number within the containing index, this is used
+  // by the Postings object to identify fields.
+  size_t text_field_number;
+  std::shared_ptr<TextIndex> text_;
+
+
   inline void ForEachTrackedKey(
```
Aksha1812, line 62: i think it would be better if you can copy all the api calls from the file deleted. this file seems like it was referered directly from numeric.h and doesn't make much sense now. rather keep it empty as in the deleted file and let person who owns it implement it completely

src/indexes/text_index.h
```diff
@@ -1,21 +1,62 @@
 #ifndef VALKEY_SEARCH_INDEXES_TEXT_TEXT_H_
 #define VALKEY_SEARCH_INDEXES_TEXT_TEST_H_
```
Aksha1812, line 2: VALKEY_SEARCH_INDEXES_TEXTINDEX_H_ *

src/indexes/text_index.h
```diff
@@ -1,21 +1,62 @@
...
 namespace valkey_search {
 namespace indexes {
 
-struct Text : public IndexBase {
+  using Key = vmsdk::InternedStringPtr;
+  using Position = uint32_t;
+
+  // Each text field is assigned a unique number within the containing index, this is used
+  // by the Postings object to identify fields.
+  size_t text_field_number;
```
Aksha1812, line 17: why are these two statements both in text_index.h and text.h files.

src/indexes/text_index.h
```diff
@@ -1,21 +1,62 @@
 #ifndef VALKEY_SEARCH_INDEXES_TEXT_TEXT_H_
 #define VALKEY_SEARCH_INDEXES_TEXT_TEST_H_
 
+#include <concepts>
+#include <memory>
+
+#include "src/utils/string_interning.h"
+
 namespace valkey_search {
```
Aksha1812, line 9: can't we just do namespace valkey_search::indexes {

src/indexes/text_index.h
```diff
@@ -1,21 +1,62 @@
...
 namespace indexes {
 
-struct Text : public IndexBase {
+  using Key = vmsdk::InternedStringPtr;
+  using Position = uint32_t;
```
Aksha1812, line 13: did you deliberately did not add from the deleted file
 using Byte = uint8_t;
using Char = uint32_t; 

or is it already in the namespace somewhere?

src/indexes/text_index.h
```diff
@@ -1,21 +1,62 @@
...
+  // Each text field is assigned a unique number within the containing index, this is used
+  // by the Postings object to identify fields.
+  size_t text_field_number;
+  std::shared_ptr<TextIndex> text_
+
+
+
+struct TextIndex {
   // Constructor
   Text(const data_model& text_index_proto);
```
Aksha1812, line 24: constructor should be TextIndex right

src/indexes/text.h
Nivesh-01: done

src/indexes/text_index.h
```diff
@@ -1,21 +1,62 @@
...
+  // which is owned by this pair of trees. Plus, updates to these two trees need
+  // to be atomic when viewed externally. The locking provided by the RadixTree object
+  // is NOT quite sufficient to guarantee that the two trees are always in lock step.
+  // thus this object becomes responsible for cross-tree locking issues.
+  // Multiple locking strategies are possible. TBD (a shared-ed word lock table should work well)
+  //
+  std::shared_ptr<RadixTree<std::unique_ptr<Postings *>, false>>> prefix_;
   std::optional<text::RadixTree> suffix_;
 
   absl::hashmap<Key, text::RadixTree> reverse_;
```
allenss-amazon, line 41: this is probably wrong and should be deleted.

PR #265, opened 2025-07-29
Global memory metrics by type

uriyage:
Adds detailed memory usage tracking for different index types in valkey-search, for better visibility into memory usage patterns.

### Main Changes
• **Global Metrics System**: Tracks memory usage across vectors, HNSW nodes/edges, tags, numeric records, and interned strings
• **INFO Command Integration**: Metrics exposed through INFO SEARCH
• **Lifecycle Tracking**: Separates active vs marked-deleted memory

### Key Metrics Added
• Vector memory (active + marked deleted)
• HNSW nodes and edges (active + marked deleted) 
• Flat index nodes
• Tag memory and count
• Numeric records
• Interned strings memory
• Keys memory

### Technical Details
• Uses atomic counters for thread safety
• Updated all index implementations to report metrics
• Enhanced string interning with metadata tracking

### Files Modified
• Core: global_metrics.h, metric_types.h
• Indexes: vector, tag, numeric implementations
• Utils: string interning enhancements
• Tests: unit and integration coverage

### DIFF
--- a/src/indexes/global_metrics.h
+++ b/src/indexes/global_metrics.h
@@ -1,0 +1,197 @@
+ /*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#ifndef VALKEYSEARCH_SRC_INDEXES_GLOBAL_METRICS_H
+#define VALKEYSEARCH_SRC_INDEXES_GLOBAL_METRICS_H
+
+#include <atomic>
+#include <cassert>
+#include <cstdint>
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/strings/string_view.h"
+#include "src/utils/string_interning.h"
+#include "src/indexes/metric_types.h"
+#include "vmsdk/src/info.h"
+
+namespace valkey_search::indexes {
+
+// Packed metadata structure for InternedString (32 bits total)
+struct MetaData {
+  uint32_t metric_type : 7;          // 7 bits for metric type
+  uint32_t use_count : 16;           // 16 bits for use count
+  uint32_t ever_used : 1;            // 1 bit to track if ever been used
+  uint32_t reserved : 8;             // 8 bits reserved for future use
+};
+
+struct MetricData {
+  std::atomic<uint64_t> count{0};           
+};
+
+const absl::flat_hash_map<MetricType, absl::string_view> kMetricTypeToStr = {
+  {MetricType::kVectorsMemory, "vectors_memory"},
+  {MetricType::kVectorsMemoryMarkedDeleted, "vectors_memory_marked_deleted"},
+  {MetricType::kHnswNodes, "hnsw_nodes"},
+  {MetricType::kHnswNodesMarkedDeleted, "hnsw_nodes_marked_deleted"},
+  {MetricType::kHnswEdges, "hnsw_edges"},
+  {MetricType::kHnswEdgesMarkedDeleted, "hnsw_edges_marked_deleted"},
+  {MetricType::kFlatNodes, "flat_nodes"},
+  {MetricType::kTags, "tags"},
+  {MetricType::kTagsMemory, "tags_memory"},
+  {MetricType::kNumericRecords, "numeric_records"},
+  {MetricType::kInternedStrings, "interned_strings"},
+  {MetricType::kInternedStringsMarkedDeleted, "interned_strings_marked_deleted"},
+  {MetricType::kInternedStringsMemory, "interned_strings_memory"},
+  {MetricType::kKeysMemory, "keys_memory"}
+};
+
+class GlobalIndexStats {
+ public:
+  static GlobalIndexStats& Instance() {
+    static GlobalIndexStats instance;
+    return instance;
+  }
+  
+  void Incr(MetricType metric_type, uint64_t value = 1) {
+    GetOrCreateMetric(metric_type).count.fetch_add(value, std::memory_order_relaxed);
+  }
+  
+  void Decr(MetricType metric_type, uint64_t value = 1) {
+    GetOrCreateMetric(metric_type).count.fetch_sub(value, std::memory_order_relaxed);
+  }
+  
+  uint64_t GetCount(MetricType metric_type) const {
+    auto it = metrics_.find(metric_type);
+    return it != metrics_.end() ? it->second->count.load(std::memory_order_relaxed) : 0;
+  }
+  
+  absl::flat_hash_map<MetricType, uint64_t> GetAllMetrics() const {
+    absl::flat_hash_map<MetricType, uint64_t> result;
+    for (const auto& [type, metric] : metrics_) {
+      result[type] = metric->count.load(std::memory_order_relaxed);
+    }
+    return result;
+  }
+  
+ private:
+  GlobalIndexStats() = default;
+  
+  MetricData& GetOrCreateMetric(MetricType metric_type) {
+    auto [it, inserted] = metrics_.try_emplace(metric_type, nullptr);
+    if (inserted) {
+      it->second = std::make_unique<MetricData>();
+    }
+    return *it->second;
+  }
+
+  absl::flat_hash_map<MetricType, std::unique_ptr<MetricData>> metrics_;
+};
+
+inline void OnInternedStringAlloc(::valkey_search::InternedString* interned_str, MetricType metric_type) {
+  if (!interned_str) return;
+  size_t bytes = interned_str->Str().length();
+  GlobalIndexStats::Instance().Incr(MetricType::kInternedStringsMemory, bytes);
+  GlobalIndexStats::Instance().Incr(MetricType::kInternedStrings, 1);
+
+  if (metric_type == MetricType::kNone) return;
+
+  MetaData metadata{};
+  metadata.metric_type = static_cast<uint8_t>(metric_type);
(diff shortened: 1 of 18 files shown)


src/indexes/global_metrics.h
```diff
@@ -0,0 +1,197 @@
...
+  {MetricType::kHnswEdgesMarkedDeleted, "hnsw_edges_marked_deleted"},
+  {MetricType::kFlatNodes, "flat_nodes"},
+  {MetricType::kTags, "tags"},
+  {MetricType::kTagsMemory, "tags_memory"},
+  {MetricType::kNumericRecords, "numeric_records"},
+  {MetricType::kInternedStrings, "interned_strings"},
+  {MetricType::kInternedStringsMarkedDeleted, "interned_strings_marked_deleted"},
+  {MetricType::kInternedStringsMemory, "interned_strings_memory"},
+  {MetricType::kKeysMemory, "keys_memory"}
+};
```
allenss-amazon: Why not an array?

src/indexes/global_metrics.h
```diff
@@ -0,0 +1,197 @@
...
+#include "src/utils/string_interning.h"
+#include "src/indexes/metric_types.h"
+#include "vmsdk/src/info.h"
+
+namespace valkey_search::indexes {
+
+// Packed metadata structure for InternedString (32 bits total)
+struct MetaData {
+  uint32_t metric_type : 7;          // 7 bits for metric type
+  uint32_t use_count : 16;           // 16 bits for use count
```
allenss-amazon: Let's move use-count to the top, so it can be incremented as a u16

src/indexes/global_metrics.h
```diff
@@ -0,0 +1,197 @@
...
+  
+  MetricData& GetOrCreateMetric(MetricType metric_type) {
+    auto [it, inserted] = metrics_.try_emplace(metric_type, nullptr);
+    if (inserted) {
+      it->second = std::make_unique<MetricData>();
+    }
+    return *it->second;
+  }
+
+  absl::flat_hash_map<MetricType, std::unique_ptr<MetricData>> metrics_;
```
allenss-amazon: Why not an array? Why the extra level of indirection with std::unique_ptr?

src/indexes/global_metrics.h
```diff
@@ -0,0 +1,197 @@
...
+  if (!interned_str) return;
+  size_t bytes = interned_str->Str().length();
+  GlobalIndexStats::Instance().Incr(MetricType::kInternedStringsMemory, bytes);
+  GlobalIndexStats::Instance().Incr(MetricType::kInternedStrings, 1);
+
+  if (metric_type == MetricType::kNone) return;
+
+  MetaData metadata{};
+  metadata.metric_type = static_cast<uint8_t>(metric_type);
+  metadata.use_count = 0;             // Start with 0
```
allenss-amazon: why start with 0?

uriyage: Changed implementation to use `INIT_USE_COUNT` we need to differentiate between a temp vector string that was never added to an index to a vector that was used in and index.

src/indexes/global_metrics.h
uriyage: Done.

src/indexes/global_metrics.h
uriyage: Done.

src/indexes/global_metrics.h
uriyage: Fixed.

src/indexes/global_metrics.h
```diff
@@ -12,43 +12,31 @@
...
+#define METRIC_ENTRY(name, str) str,
+  METRIC_TYPES_TABLE
+#undef METRIC_ENTRY
 };
 
+inline absl::string_view GetMetricTypeString(MetricType metric_type) {
+  const auto index = static_cast<size_t>(metric_type);
+  if (index >= static_cast<size_t>(MetricType::kMetricTypeCount)) {
+    return "";  // Invalid metric type
+  }
```
allenss-amazon, line 36: Feels like this should be an assert.

src/indexes/global_metrics.h
```diff
@@ -12,43 +12,31 @@
...
-  {MetricType::kTags, "tags"},
-  {MetricType::kTagsMemory, "tags_memory"},
-  {MetricType::kNumericRecords, "numeric_records"},
-  {MetricType::kInternedStrings, "interned_strings"},
-  {MetricType::kInternedStringsMarkedDeleted, "interned_strings_marked_deleted"},
-  {MetricType::kInternedStringsMemory, "interned_strings_memory"},
-  {MetricType::kKeysMemory, "keys_memory"}
+constexpr absl::string_view kMetricTypeStrings[] = {
+#define METRIC_ENTRY(name, str) str,
+  METRIC_TYPES_TABLE
```
allenss-amazon, line 28: This kind of macro with-in a macro is a very powerful thing. But I believe it would be clearer if you passed in the "METRIC_ENTRY" as a parameter to METRIC_TYPES_TABLE rather than having an externally known name that's not obvious from the usage. By passing in the "macro to invoke" you eliminate odd-ball global dependencies.

src/indexes/global_metrics.h
allenss-amazon: Also, feels like the body of this and the table should be in .cc file, isn't "kMetricStringTypes" really supposed to be hidden behind "GetMetricTypeString" ??

src/indexes/global_metrics.h
```diff
@@ -57,125 +45,66 @@ class GlobalIndexStats {
   }
   
   void Incr(MetricType metric_type, uint64_t value = 1) {
-    GetOrCreateMetric(metric_type).count.fetch_add(value, std::memory_order_relaxed);
+    GetMetric(metric_type).count.fetch_add(value, std::memory_order_relaxed);
   }
   
   void Decr(MetricType metric_type, uint64_t value = 1) {
-    GetOrCreateMetric(metric_type).count.fetch_sub(value, std::memory_order_relaxed);
+    GetMetric(metric_type).count.fetch_sub(value, std::memory_order_relaxed);
```
allenss-amazon, line 52: There should be some insulation against going negative (large positive). This is a common failure in these kinds of tracking operations. When you detect a negative counter, you should clamp it to zero and put out a log message -- be sure to use the LOG_EVERY_N .... macro to rate-limit the log messages, because you could get a lot of them here. Optionally, we might want to have a configurable to make this be a hard crash in order to help debug any account errors.

src/indexes/global_metrics.h
```diff
@@ -57,125 +45,66 @@ class GlobalIndexStats {
...
-  }
-  
-  absl::flat_hash_map<MetricType, uint64_t> GetAllMetrics() const {
-    absl::flat_hash_map<MetricType, uint64_t> result;
-    for (const auto& [type, metric] : metrics_) {
-      result[type] = metric->count.load(std::memory_order_relaxed);
+    auto& store = ::valkey_search::StringInternStore::Instance();
+    uint64_t total = 0;
+    if (metric_type == MetricType::kInternedStrings) {
+      for (size_t j = 0; j < static_cast<size_t>(::valkey_search::StringType::kStringTypeCount); ++j) {
```
allenss-amazon, line 59: Switch statement please.

src/indexes/global_metrics.h
```diff
@@ -57,125 +45,66 @@ class GlobalIndexStats {
...
+      return store.GetMarkedDeletedCounters().object_count;
+    } else if (metric_type == MetricType::kTagsMemory) {
+      return store.GetCounters(::valkey_search::StringType::TAG).memory_bytes;
+    } else if (metric_type == MetricType::kKeysMemory) {
+      return store.GetCounters(::valkey_search::StringType::KEY).memory_bytes;
+    } else {
+      const size_t index = static_cast<size_t>(metric_type);
+      if (index >= static_cast<size_t>(MetricType::kMetricTypeCount)) {
+        return 0;  // Invalid metric type
+      }
```
allenss-amazon, line 82: feels like an assert would be more appropriate here on the "default" label on the switch.

src/indexes/metric_types.h
```diff
@@ -10,29 +10,30 @@
 
 namespace valkey_search::indexes {
 
+#define METRIC_TYPES_TABLE \
```
allenss-amazon, line 13: as describe above make METRIC_ENTRY be a parameter to this macro

allenss-amazon (changes requested):
I have two issues with the overall design.

1. It seems like this design may fail whenever you have the same string with different type marks. KEY and TAG for example. Depending on the order of creation, the two different objects will either be classified as a TAG or as a KEY. And we charge differently for them, right? I believe there are two ways to fix this. a/ Instantiate different InternStringPools, one for each type. This would require changing every instance of InternStringPtr to the correctly typed one. 

There's a less intrusive hack which is b/ Incorporate the MetricType into intern table. In other words change:
```
absl::flat_hash_map<absl::string_view, std::weak_ptr<InternedString>>
      str_to_interned_ ABSL_GUARDED_BY(mutex_);
```
into

```absl::flat_hash_map<std::pair<MetricType, absl::string_view>, std::weak_ptr<InternedString>>
      str_to_interned_ ABSL_GUARDED_BY(mutex_);
```
This logically ends creating different pools for different MetricTypes, but without the need to change all references to InternStringPtrs.

4. Why is vector treated differently? If it has fundamentally different counting characteristics that suggests to me that perhaps it should be a different type.

PR #249, opened 2025-07-24
Reimplement timeout handling.

allenss-amazon:
Current handling of timeout is inadequate to avoid DOS. The problem is that when timeout is detected by the valkey core, the client is unblocked, but the background computation continues.

This PR solves that problem as well as implementing #201.

Now, timeout handling is no longer performed by the core. Timeout handling is done by having the major control loops (HNSW, FLAT and pre-filtered) check a "Cancellation Token". This token will check for timeout and if so, the major control loops will return their current results, i.e., "partial" results.

If the configurable "search_enable_partial_results" is true, then the FT.SEARCH command will complete and return the partial results. No indication is made whether the command actually completed or timeout out.

If the configurable is false, then an actual error is generated.

Regardless of the setting of the configurable, the background search query process terminates.

PR #268, opened 2025-07-30
Handle OOM during search command fanout

yanamolo:
As of today, search command is registered with `deny-oom` flag. So in case of OOM in the node, search command would be denied.
But when search command arrives through the coordinator, it is handled even though the node in maxmemory.
This PR adds OOM handling for commands coming from the coordinator - even if one of the nodes is OOM, error is returned to the client.


src/query/fanout.cc
```diff
@@ -213,6 +222,10 @@ absl::Status PerformSearchFanoutAsync(
           if (neighbors.ok()) {
             tracker->AddResults(*neighbors);
           } else {
+            if (neighbors.status().code() ==
```
eifrah-aws: Use `absl::IsResourceExhausted(..)`

src/query/search.cc
```diff
@@ -395,7 +395,17 @@ absl::Status SearchAsync(std::unique_ptr<VectorSearchParameters> parameters,
...
        is_local_search]() mutable {
-        auto res = Search(*parameters, is_local_search);
+        // Handle OOM for search requests, mainly defends against request
+        // coming from the coordinator
+        auto ctx = vmsdk::MakeUniqueValkeyThreadSafeContext(nullptr);
+        auto ctx_flags = ValkeyModule_GetContextFlags(ctx.get());
+        absl::StatusOr<std::deque<indexes::Neighbor>> res;
+        if (ctx_flags & VALKEYMODULE_CTX_FLAGS_OOM) {
+          res = absl::ResourceExhaustedError(
+              "OOM command not allowed when used memory > 'maxmemory'");
```
eifrah-aws: This string is repeating, lets make it a constant?

src/query/fanout.cc
```diff
@@ -119,10 +120,15 @@ struct SearchPartitionResultsTracker {
     absl::MutexLock lock(&mutex);
     absl::StatusOr<std::deque<indexes::Neighbor>> result =
         std::deque<indexes::Neighbor>();
-    while (!results.empty()) {
-      result->push_back(
-          std::move(const_cast<indexes::Neighbor &>(results.top())));
-      results.pop();
+    if (reached_oom) {
```
eifrah-aws, line 143: So, we decided that if one node fails, we drop everything?

eifrah-aws (changes requested):

src/query/fanout.cc
yanamolo: Yes, was decided to return an error if one of the shard is OOM

integration/test_oom_handling.py
```diff
@@ -0,0 +1,88 @@
...
+                b"\x00\x00\x00@\x00\x00@@\x00\x00\x80@",
+            ],
+            b"vec:1",
+            [
+                b"__vector_score",
+                b"1.19209289551e-07",
+                b"vector",
+                b"\x00\x00\x80?\x00\x00\x00@\x00\x00@@",
+            ],
+        ]
```
allenss-amazon: This seems over-specified. All you really care is that you get a 2 vector response vs an error, right? Let's just code that. Imagine a situation where there was a slight change in the order of computation for HNSW, where you get slightly different round-offs for the distance computations (which might also happen on different CPU architectures or different compiler optimizations, etc.). That would require changing this result here -- pointlessly.

integration/test_oom_handling.py
```diff
@@ -0,0 +1,88 @@
...
+from valkeytestframework.conftest import resource_port_tracker
+import struct
+import pytest
+from valkey.exceptions import OutOfMemoryError
+
+
+class TestSearchOOMHandling(ValkeySearchClusterTestCase):
+    """
+    Test suite for search command OOM handling. We expect that
+    when one node hits oom during fanout, whole command would be aborted.
```
allenss-amazon: The rework for timeout (See https://github.com/valkey-io/valkey-search/pull/249) introduces a configurable "EnablePartialResults" which controls how to handle fanout/in request when a shard takes too long. We might consider OOM to get handled like "too long", i.e., either return partial results OR generate a user error.

src/query/search.cc
allenss-amazon: Why is this needed here??? If the command is marked deny-oom, then you won't get here, right?

allenss-amazon (changes requested):
I believe this needs to get harmonized with the timeout patch. I think it gets easier to do. IMO, I think all that's required is to error-out a gRPC search request when the current node is OOM. Then in the fan-in logic I think you can treat OOM the same as timeout error.

allenss-amazon:
> @allenss-amazon we should have probably closed it if so. Anyway, will you add it in your pr?

I'd rather get my current PR merged and then add this as an additional PR. can you review my PR so we can merge it or ask Eran Ifrah to review it again?

src/query/search.cc
yanamolo: This is in a case where a client connects directly to the OOM shard. But we get to here also through the coordinator. When commands are arriving through coordinator server, we call `query::SearchAsync`, this is why I added the check here

integration/test_oom_handling.py
yanamolo: Yes, you are right, will be fixed.

src/query/fanout.cc
yanamolo: Will be fixed

integration/test_oom_handling.py
yanamolo: I think but in the OOM case it can cause inconsistency. If for example you have two primaries, one is OOM and one is not, if you connect to the OOM one - you will immediately get OOM error due to the `deny-oom` flag. But if you will connect to the primary that has enough memory, you will get partial results? It was discussed with Zvi and also with Jacob, who agreed that in case of OOM we will always want to return error.

src/query/search.cc
yanamolo: Will add constant

src/query/search.cc
```diff
@@ -399,7 +399,16 @@ absl::Status SearchAsync(std::unique_ptr<VectorSearchParameters> parameters,
...
+        // Handle OOM for search requests, mainly defends against request
+        // coming from the coordinator
+        auto ctx = vmsdk::MakeUniqueValkeyThreadSafeContext(nullptr);
+        auto ctx_flags = ValkeyModule_GetContextFlags(ctx.get());
+        absl::StatusOr<std::deque<indexes::Neighbor>> res;
+        if (ctx_flags & VALKEYMODULE_CTX_FLAGS_OOM) {
+          res = absl::ResourceExhaustedError(kOOMMsg);
+        } else {
+          res = Search(*parameters, is_local_search);
+        }
```
allenss-amazon: Why isn't this logic inside of Search itself? Putting it here means it will behave inconsistently. For example, if I wrap the search in a multi/exec it won't be aborted on an OOM.

integration/test_oom_handling.py
```diff
@@ -0,0 +1,72 @@
...
+        return client.execute_command(
+            "FT.SEARCH",
+            self.index_name,
+            "*=>[KNN 2 @vector $query_vector]",
+            "PARAMS",
+            "2",
+            "query_vector",
+            search_vector,
+        )
+
```
allenss-amazon: I'd like to see some CMD testing, the control flows are different.

Also, for CME testing, you need to test the case of where OOM is on the initiating node, again there are different flows in the code for this.

src/query/search.cc
yanamolo: Moved it into Search function

integration/test_oom_handling.py
yanamolo: Added the test cases

src/query/fanout.cc
```diff
@@ -76,10 +77,11 @@ struct SearchPartitionResultsTracker {
                       const grpc::Status &status) {
 
     if (!status.ok()) {
-      if (!options::GetEnablePartialResults().GetValue()) {
+      if (status.error_code() == grpc::RESOURCE_EXHAUSTED) {
+        reached_oom.store(true);
```
allenss-amazon, line 81: I think you should cancel on this path also. No reason for the local node to keep working on searching. add "parameters->cancellation_token->Cancel()"

src/query/search.cc
```diff
@@ -332,6 +332,14 @@ absl::StatusOr<std::deque<indexes::Neighbor>> MaybeAddIndexedContent(
 
 absl::StatusOr<std::deque<indexes::Neighbor>> Search(
     const VectorSearchParameters &parameters, bool is_local_search) {
+  // Handle OOM for search requests, mainly defends against request
+  // coming from the coordinator
+  auto ctx = vmsdk::MakeUniqueValkeyThreadSafeContext(nullptr);
+  auto ctx_flags = ValkeyModule_GetContextFlags(ctx.get());
+  absl::StatusOr<std::deque<indexes::Neighbor>> res;
```
allenss-amazon: What is this "res" variable? I don't see it being used anywhere.

allenss-amazon (changes requested):
Just a couple of nits and we're ready to merge.

src/query/search.cc
yanamolo: Was left there by mistake, will be removed

src/query/fanout.cc
yanamolo: Will make sure it is canceled

eifrah-aws (approved):

PR #272, opened 2025-07-31
Implement lexical scanner

VoletiRam:
Implemented initial lexical scanner for text search that handles UTF-8 text tokenization with configurable punctuation boundaries and whitespace normalization, establishing the foundation for the text indexing pipeline as outlined in the text search RFC.

## Implementation Details

- Added core UTF-8 text tokenization functionality in the `text` namespace
- Implemented configurable punctuation character handling for word boundary detection
- Created a pipeline for text processing with whitespace normalization
- Added support for escaped punctuation characters using backslash syntax
- Implemented token position tracking to support future offset-based features
- Added utility functions for lexical analysis
- Includes test coverage for various tokenization scenarios

## Note to Reviewers

⚠️ __IMPORTANT__: This PR depends on Nivesh's FT.CREATE support PR which has not been merged yet. As a result:

1. __The build will fail until the dependency is merged.__
2. This PR is provided for early review of just the lexical scanner implementation.
3. Once the dependency is merged, this PR will be rebased on top of it.

Please focus your review on the lexical scanner implementation, especially the tokenization approach, UTF-8 handling, and the overall design of the text processing pipeline. The integration with the text index implementation will be part of future PRs.

## Working Implementation for Review

While this PR depends on Nivesh's FT.CREATE support which hasn't been merged yet, I've created a fully working end-to-end implementation in my personal fork for reviewers who want to see the complete functionality:

__Full working implementation__: [](https://github.com/VoletiRam/valkey-search/tree/lexical_scanner)<https://github.com/VoletiRam/valkey-search/tree/lexical_scanner> Compare changes: [](https://github.com/valkey-io/valkey-search/compare/fulltext...VoletiRam:valkey-search:lexical_scanner)<https://github.com/valkey-io/valkey-search/compare/fulltext...VoletiRam:valkey-search:lexical_scanner>

This branch contains my implementation of both:

1. The necessary FT.CREATE foundation (temporary implementation until Nivesh's PR is merged)
2. The lexical scanner implementation (the focus of this PR)

Reviewers can check out this branch to see and test the full text search tokenization pipeline in action. This should provide better context for how the lexical scanner fits into the overall text search functionality.

### DIFF
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#include "src/indexes/text.h"
+#include "src/indexes/text_index.h"
+#include "src/indexes/text/lexer.h"
+#include "vmsdk/src/log.h"
+
+namespace valkey_search::indexes {
+
+Text::Text(const data_model::TextIndex& text_index_proto, 
+           const data_model::IndexSchema* index_schema_proto)
+    : IndexBase(IndexerType::kText),
+      text_impl_(std::make_unique<text::TextFieldIndex>(
+          text_index_proto,
+          index_schema_proto,
+          // Pass field identifier if available, otherwise empty string
+          (index_schema_proto && !index_schema_proto->name().empty()) 
+              ? index_schema_proto->name() : "")) {}
+
+absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
+                                     absl::string_view data) {
+  absl::MutexLock lock(&index_mutex_);
+  // TODO: stub
+  return text_impl_->AddRecord(key, data);
+}
+
+absl::StatusOr<bool> Text::RemoveRecord(const InternedStringPtr& key,
+                                        DeletionType deletion_type) {
+  absl::MutexLock lock(&index_mutex_);
+  // TODO: stub
+  return text_impl_->RemoveRecord(key, deletion_type);
+}
+
+absl::StatusOr<bool> Text::ModifyRecord(const InternedStringPtr& key,
+                                        absl::string_view data) {
+  absl::MutexLock lock(&index_mutex_);
+  // TODO: stub
+  return text_impl_->ModifyRecord(key, data);
+}
+
+int Text::RespondWithInfo(ValkeyModuleCtx* ctx) const {
+  // TODO: stub
+  return text_impl_->RespondWithInfo(ctx);
+}
+
+bool Text::IsTracked(const InternedStringPtr& key) const {
+  // TODO: stub
+  return text_impl_->IsTracked(key);
+}
+
+uint64_t Text::GetRecordCount() const {
+  // TODO: stub
+  return text_impl_->GetRecordCount();
+}
+
+std::unique_ptr<data_model::Index> Text::ToProto() const {
+  // TODO: stub
+  return text_impl_->ToProto();
+}
+
+InternedStringPtr Text::GetRawValue(const InternedStringPtr& key) const {
+  // TODO:stub
+  static InternedStringPtr empty;
+  return empty;
+}
+
+// EntriesFetcherIterator implementations
+bool Text::EntriesFetcherIterator::Done() const {
+  // TODO:stub
+  return true;
+}
+
+void Text::EntriesFetcherIterator::Next() {
+  // TODO:stub
+}
+
+const InternedStringPtr& Text::EntriesFetcherIterator::operator*() const {
+  // TODO:stub
+  static InternedStringPtr empty;
+  return empty;
+}
+
+// EntriesFetcher implementations
+size_t Text::EntriesFetcher::Size() const {
+  // TODO:stub
+  return 0;
+}
+
+std::unique_ptr<EntriesFetcherIteratorBase> Text::EntriesFetcher::Begin() {
+  // TODO:stub
+  return std::make_unique<EntriesFetcherIterator>();
+}
+
+std::unique_ptr<Text::EntriesFetcher> Text::Search(
+    const query::TextPredicate& predicate, bool negate) const {
+  // TODO: stub
+  return std::make_unique<EntriesFetcher>();
+}
+
+} // namespace valkey_search::indexes
+
+namespace valkey_search {
+namespace text {
+
+absl::StatusOr<bool> TextFieldIndex::AddRecord(const InternedStringPtr& key,
+                                               absl::string_view data) {
+  auto lexer_result = lexer_.ProcessString(data);
+  if (!lexer_result.ok()) {
+    // Use debug logging for non-parseable data (very rare, not actionable)
+    // Increased interval to once per minute to prevent log spam
+    VMSDK_LOG_EVERY_N_SEC(DEBUG, nullptr, 60)
+        << "Text field analysis failed for key " << key->Str() << ": " 
+        << lexer_result.status().message();
+    return lexer_result.status();
+  }
+  
+  // Get the processed terms
+  const auto& terms = lexer_result->GetProcessedTerms();
+  
+  // Use info-level logging for normal operation, with rate limiting
+  // TODO: This logging is for testing purpose, will be removed or updated to
+  // DEBUG level as we have other change below working as anticipated
+  VMSDK_LOG_EVERY_N_SEC(NOTICE, nullptr, 1) 
+      << "Text field [" 
+      << (field_identifier_.empty() ? "unknown" : field_identifier_)
+      // Access field configuration flags safely
+      << (text_index_proto_.nostem() ? " NOSTEM" : "")
+      << (text_index_proto_.suffix_tree() ? " SUFFIX" : "")
+      // Include language information if index_schema_proto_ is available and not using NOSTEM
+      << (!text_index_proto_.nostem() && index_schema_proto_ && index_schema_proto_->language() != data_model::LANGUAGE_UNSPECIFIED ? 
+            (" LANG:" + data_model::Language_Name(index_schema_proto_->language())) : "")
+      << "] processed document '" << key->Str() 
+      << "' into " << terms.size() << " terms";
+  
+  // TODO: Once RadixTree and Postings are implemented:
+  // 1. Store terms in prefix tree (text_->prefix_) with appropriate locking
+  // 2. Store terms in suffix tree if text_index_proto_.suffix_tree() is enabled with appropriate locking
+  // 3. Update postings for each term to track key and positions (if with_offsets_ is true)
+  // Note: Updates to prefix/suffix tree data structures must be coordinated
+  // with their respective locking mechanisms to ensure thread safety
+  
+  // For now, return success regardless of word count
+  // Empty documents and documents with no indexable words are valid cases
+  // This is consistent with other index types that can have empty records
+  return true;
+}
+
+}  // namespace text
+}  // namespace valkey_search
--- a/src/indexes/text.h
+++ b/src/indexes/text.h
@@ -30,6 +30,108 @@
 #ifndef VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 #define VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 
+#include "absl/base/thread_annotations.h"
+#include "absl/synchronization/mutex.h"
+#include <concepts>
+#include <memory>
+#include <optional>
+
+#include "absl/functional/any_invocable.h"
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "src/indexes/index_base.h"
+#include "src/index_schema.pb.h"
+#include "src/indexes/text_index.h"
+#include "src/rdb_serialization.h"
+#include "src/utils/string_interning.h"
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+#include "src/indexes/text/lexer.h"
+
+// Forward declaration
+namespace valkey_search::indexes {
+enum class DeletionType;
+enum class IndexerType;
+}
+
+namespace valkey_search {
+namespace text {
+
+using Byte = uint8_t;
+using Char = uint32_t;
+
+struct TextFieldIndex {
+    TextFieldIndex(const data_model::TextIndex& text_index_proto,
+                 const data_model::IndexSchema* index_schema_proto = nullptr,
+                 std::string field_identifier = "") 
+      : text_index_proto_(text_index_proto),
+        index_schema_proto_(index_schema_proto),
+        field_identifier_(field_identifier),
+        // Initialize lexer in the constructor's initialization list
+        lexer_(index_schema_proto_ ? 
+              Lexer(text_index_proto_, *index_schema_proto_) : 
+              Lexer(text_index_proto_, data_model::IndexSchema())) {
+    // Initialize text index structures
+    text_ = std::make_shared<TextIndex>();
+  }
+
+  // Return the field identifier
+  const std::string& GetFieldIdentifier() const {
+    return field_identifier_;
+  }
+  ~TextFieldIndex() = default;
+
+  absl::StatusOr<bool> AddRecord(const InternedStringPtr& key,
+                                 absl::string_view data);
+  absl::StatusOr<bool> RemoveRecord(const InternedStringPtr& key,
+                                    indexes::DeletionType deletion_type) {
+    return false; // Placeholder 
+  }
+  absl::StatusOr<bool> ModifyRecord(const InternedStringPtr& key,
+                                    absl::string_view data) {
+    return false; // Placeholder 
+  }
+  int RespondWithInfo(ValkeyModuleCtx* ctx) const {
+    return 0; // Placeholder 
+  }
+  bool IsTracked(const InternedStringPtr& key) const {
+    return false; // Placeholder 
+  }
+  absl::Status SaveIndex(RDBChunkOutputStream chunked_out) const {
+    return absl::OkStatus(); // Placeholder 
+  }
+
+  std::unique_ptr<data_model::Index> ToProto() const {
+    auto index_proto = std::make_unique<data_model::Index>();
+    index_proto->mutable_text_index();
+    return index_proto;
+  }
+  void ForEachTrackedKey(
+      absl::AnyInvocable<void(const InternedStringPtr&)> fn) const {
+    // Placeholder 
+  }
+
+  uint64_t GetRecordCount() const {
+    return 0; // Placeholder 
+  }
+
+ private:
+  // Each text field is assigned a unique number within the containing index
+  size_t text_field_number = 0;
+  // The per-index text index
+  std::shared_ptr<TextIndex> text_;
+  // Store references to configuration protos
+  const data_model::TextIndex& text_index_proto_;
+  const data_model::IndexSchema* index_schema_proto_;
+  // Field identifier (name or alias)
+  std::string field_identifier_;
+  // Lexer for text processing
+  text::Lexer lexer_;
+};
+
+
+}  // namespace text
+}  // namespace valkey_search
 
 namespace valkey_search::indexes {
 
--- a/src/indexes/text/lexer.cc
+++ b/src/indexes/text/lexer.cc
@@ -1,0 +1,352 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#include "src/indexes/text/lexer.h"
+
+#include <bitset>
+#include <cstddef>
+#include <string>
+#include <vector>
+
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/ascii.h"
+#include "absl/strings/string_view.h"
+
+namespace valkey_search::text {
+
+namespace {
+// ASCII character bitmap for fast punctuation lookup
+using PunctuationBitmap = std::bitset<256>;
+
+// Check if a character is valid UTF-8 continuation byte
+bool IsUtf8Continuation(char c) {
+  return (static_cast<unsigned char>(c) & 0xC0) == 0x80;
+}
+
+// Get the length of a UTF-8 character from its first byte
+int GetUtf8CharLength(char c) {
+  unsigned char byte = static_cast<unsigned char>(c);
+  if (byte < 0x80) return 1;  // ASCII
+  if ((byte & 0xE0) == 0xC0) return 2;  // 110xxxxx
+  if ((byte & 0xF0) == 0xE0) return 3;  // 1110xxxx
+  if ((byte & 0xF8) == 0xF0) return 4;  // 11110xxx
+  return -1;  // Invalid UTF-8
+}
+
+// Validate UTF-8 sequence starting at position
+bool IsValidUtf8Sequence(absl::string_view text, size_t pos, int expected_len) {
+  if (pos + expected_len > text.size()) return false;
+  
+  for (int i = 1; i < expected_len; ++i) {
+    if (!IsUtf8Continuation(text[pos + i])) {
+      return false;
+    }
+  }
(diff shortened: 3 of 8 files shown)


src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+ */
+
+#include "src/indexes/text.h"
+#include "src/indexes/text_index.h"
+#include "src/indexes/text/lexer.h"
+#include "vmsdk/src/log.h"
+
+namespace valkey_search::indexes {
+
+Text::Text(const data_model::TextIndex& text_index_proto,
```
allenss-amazon, line 15: What is this class going to be used for?

src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+  if (!schema_punctuation_.empty()) {
+    auto status = lexer.SetPunctuation(schema_punctuation_);
+    if (!status.ok()) {
+      // Log configuration errors (rate-limited to prevent spam)
+      VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+          << "Text field punctuation configuration error: " 
+          << status.message();
+      return status;
+    }
+  }
```
allenss-amazon: Why isn't the Lexer constructed when the TextFieldIndex is constructed and stored inside of it?

src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+  
+  // Process the input text through the lexer
+  auto lexer_result = lexer.ProcessString(data);
+  if (!lexer_result.ok()) {
+    // Log tokenization errors (rate-limited to prevent spam)
+    VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+        << "Text field tokenization failed: " 
+        << lexer_result.status().message();
+    return lexer_result.status();
```
allenss-amazon: A non-parseable data string is always possible, I don't think its something that we want to bother logging. However, it's definitely something that we want an info-counter for. Though I would ask the question, what conditions would cause the Lexer to return an error? I'm skeptical that it's possible for it to return an error. I suppose a non-UTF8 character MIGHT be considered an error but otherwise I'm having difficulty understanding when that's possible.

src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+  // Get the tokenized words
+  const auto& words = lexer_result->GetWords();
+  
+  // Log normal operation for each successful tokenization
+  VMSDK_LOG(NOTICE, nullptr) 
+      << "Text field tokenized document '" << key->Str() 
+      << "' into " << words.size() << " words";
+  
+  // Apply field-specific settings
+  bool no_stem = text_index_proto_.nostem() || schema_nostem_;
```
allenss-amazon: Why isn't this just in the TextField at construction time?

src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+  bool use_suffix = text_index_proto_.suffix_tree();
+  
+  // Use field-level setting if specified, otherwise fall back to schema-level setting
+  // For min_stem_size, assume field value of 0 means "use default"
+  uint32_t min_stem_size = text_index_proto_.min_stem_size() > 0 ?
```
allenss-amazon: Why isn't this in the TextField? Seems like this should be set properly at parse time and there's no reason to be computing the value here.

src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+  const auto& words = lexer_result->GetWords();
+  
+  // Log normal operation for each successful tokenization
+  VMSDK_LOG(NOTICE, nullptr) 
+      << "Text field tokenized document '" << key->Str() 
+      << "' into " << words.size() << " words";
+  
+  // Apply field-specific settings
+  bool no_stem = text_index_proto_.nostem() || schema_nostem_;
+  bool use_suffix = text_index_proto_.suffix_tree();
```
allenss-amazon: I don't think you care about this here.

src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+  // For min_stem_size, assume field value of 0 means "use default"
+  uint32_t min_stem_size = text_index_proto_.min_stem_size() > 0 ? 
+                          text_index_proto_.min_stem_size() : 
+                          schema_min_stem_size_;
+  
+  // TODO: Once RadixTree and Postings are implemented:
+  // 1. Apply stemming based on no_stem and min_stem_size settings
+  // 2. Filter stop words if configured (schema_stop_words_)
+  // 3. Store words in prefix tree (text_->prefix_)
+  // 4. Store words in suffix tree if use_suffix is enabled (text_->suffix_)
```
allenss-amazon: Storing words in the prefix/suffix tree needs to be coordinated with that data structure, It may require doing the prefix&suffix tree updates under its locking.

src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+  // 5. Update postings for each word to track key and positions
+  // 6. Handle word normalization based on schema language setting
```
allenss-amazon: this should get moved up.

src/indexes/text.cc
```diff
@@ -0,0 +1,172 @@
...
+  
+  // For now, successfully tokenize and indicate success
+  // This allows tests to pass, demonstrating working tokenization
+  // Empty data is valid (returns true), non-empty data with words also returns true
+  return words.size() > 0 || data.empty();
```
allenss-amazon: I don't think there's any error in ingesting a text field that has no words in it. It could be just a bunch of spaces.

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,189 @@
...
+}
+
+absl::StatusOr<LexerOutput> Lexer::ProcessString(absl::string_view s) const {
+  // Validate UTF-8
+  if (!IsValidUtf8(s)) {
+    return absl::InvalidArgumentError("Invalid UTF-8 sequence in input");
+  }
+  
+  LexerOutput output;
+  PunctuationBitmap punct_bitmap = CreatePunctuationBitmap(punctuation_);
```
allenss-amazon, line 217: Why isn't this in the TextFieldIndex object?

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,189 @@
...
+      }
+      
+      // Skip regular punctuation
+      if (punct_bitmap[static_cast<unsigned char>(current)]) {
+        pos++;
+        continue;
+      }
+      
+      // Found non-punctuation character - start of word
+      break;
```
allenss-amazon: wouldn't
```
 if (!punction.....) { break }
``` 
be cleaner?

allenss-amazon (changes requested):

src/indexes/text.cc
VoletiRam: The Text class is necessary because the core search system is designed to interact only with implementations of IndexBase. This class handles the boundary between the generic index interface and our text-specific implementation. This follows the same architecture as Tag, Numeric, and Vector indexes, where we separate the interface compliance layer from the specialized implementation. I can add a documentation explaining this.

src/indexes/text.cc
VoletiRam: Ah, you're right. It can be common to every document per each text index field instead of per document. Will update this.

Also, considering to implement shared component(TextAnalyzer) for lexical so that the configurations can be used by both indexing and querying

src/indexes/text.cc
VoletiRam: Agree. Didn't focus much on ft.create or parser related code as will rebase and refine properly anyway after Nivesh's PR gets merged.

src/indexes/text.cc
VoletiRam: Ack, will move this to TextField.

src/indexes/text.cc
VoletiRam: Ack.

src/indexes/text.cc
VoletiRam: The main potential error would be malformed UTF-8, but that's already handled at a lower level.

I'll remove the WARNING level logging and replace it with:

1. An info-counter for tracking occurrences without log spam
2. DEBUG level logs (if we need them for deep debugging)

src/indexes/text.cc
VoletiRam: Ack, Will update this comment with the information.

src/indexes/text.cc
VoletiRam: Aah, sure

src/indexes/text.cc
VoletiRam: Agreed. Will update the code to not treat empty word lists as an error case.

src/indexes/text/lexer.cc
VoletiRam: Will move it.

src/indexes/text/lexer.cc
VoletiRam: Agreed.

src/indexes/text.cc
allenss-amazon: That makes sense to me

src/indexes/text.cc
allenss-amazon: Well, if UTF-8 has already been handled, then you should probably assert here and you don't need a counter!

src/indexes/text.h
```diff
@@ -30,6 +30,108 @@
...
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "src/indexes/index_base.h"
+#include "src/index_schema.pb.h"
+#include "src/indexes/text_index.h"
+#include "src/rdb_serialization.h"
+#include "src/utils/string_interning.h"
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+#include "src/indexes/text/lexer.h"
```
allenss-amazon, line 49: Let's put these into canonical order.

src/indexes/text.h
```diff
@@ -30,6 +30,108 @@
...
+namespace valkey_search {
+namespace text {
+
+using Byte = uint8_t;
+using Char = uint32_t;
+
+struct TextFieldIndex {
+    TextFieldIndex(const data_model::TextIndex& text_index_proto,
+                 const data_model::IndexSchema* index_schema_proto = nullptr,
+                 std::string field_identifier = "")
```
allenss-amazon, line 66: Is there a reason to make these parameters have default values?

src/indexes/text.h
```diff
@@ -30,6 +30,108 @@
...
+enum class IndexerType;
+}
+
+namespace valkey_search {
+namespace text {
+
+using Byte = uint8_t;
+using Char = uint32_t;
+
+struct TextFieldIndex {
```
allenss-amazon, line 63: Maybe I'm confused, doesn't this get derived from IndexBase?

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
...
+  unsigned char byte = static_cast<unsigned char>(c);
+  if (byte < 0x80) return 1;  // ASCII
+  if ((byte & 0xE0) == 0xC0) return 2;  // 110xxxxx
+  if ((byte & 0xF0) == 0xE0) return 3;  // 1110xxxx
+  if ((byte & 0xF8) == 0xF0) return 4;  // 11110xxx
+  return -1;  // Invalid UTF-8
+}
+
+// Validate UTF-8 sequence starting at position
+bool IsValidUtf8Sequence(absl::string_view text, size_t pos, int expected_len) {
```
allenss-amazon, line 42: Wouldn't this make more sense to just have a string_view be passed in? The caller can easily do a string_view::substr to make the single string_view parameter.

src/indexes/text.h
allenss-amazon: Or is it Text that's derived from IndexBase? Not sure why we have both, or is this part of the "Needs to be committed after xxx PR" ??

src/indexes/text/lexer.cc
allenss-amazon: Or conversely, why not have this function do the construction of another string_view and call the next function IsValieUtf8 ?

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
...
+    if (char_len < 0 || !IsValidUtf8Sequence(text, pos, char_len)) {
+      return false;
+    }
+    pos += char_len;
+  }
+  return true;
+}
+
+// Create punctuation bitmap from string
+PunctuationBitmap CreatePunctuationBitmap(const std::string& punctuation) {
```
allenss-amazon, line 67: This seems like a constructor. Should it be a constructor?

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
...
+  PunctuationBitmap bitmap;
+  for (char c : punctuation) {
+    bitmap.set(static_cast<unsigned char>(c));
+  }
+  return bitmap;
+}
+
+// No defaults here - parser is the source of truth
```
allenss-amazon, line 75: ?

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
...
+}  // namespace
+
+// Initialize lexer from schema proto
+Lexer::Lexer(const data_model::IndexSchema& schema_proto) {
+  ConfigureFromSchemaProto(schema_proto);
+}
+
+// Initialize lexer from text index and schema protos
+Lexer::Lexer(const data_model::TextIndex& text_index_proto,
+             const data_model::IndexSchema& schema_proto) {
```
allenss-amazon, line 86: I think there's a chained constructor syntax that makes sense here.

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
...
+  
+  // Field-specific min stem size overrides schema if non-zero
+  if (text_index_proto.min_stem_size() > 0) {
+    min_stem_size_ = text_index_proto.min_stem_size();
+  }
+}
+      
+// TODO: Add these methods once we implement stop words functionality
+void Lexer::SetStopWords(const std::vector<std::string>& stop_words) {
+  // No-op for now, will be implemented in future
```
allenss-amazon, line 124: looks like it's implemented.

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
...
+  for (const auto& word : stop_words) {
+    stop_words_.insert(absl::AsciiStrToLower(word));
+  }
+}
+
+const std::unordered_set<std::string>& Lexer::GetStopWords() const {
+  return stop_words_;
+}
+
+// TODO: Add these methods once we implement stemming functionality
```
allenss-amazon, line 135: what more needs t obe done here?

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
+/*
```
Aksha1812, line 1: does the filename need to be changed now since everything is under lexer.cc now? maybe tokenizer.cc ?

src/indexes/text/lexer.h
```diff
@@ -69,12 +134,36 @@ struct LexerOutput {
   };
 
   const std::vector<Word>& GetWords() const;
+  
+  // Get the processed terms after all text processing steps
+  const std::vector<std::string>& GetProcessedTerms() const;
```
allenss-amazon, line 139: how would this be different from GetWords ?

allenss-amazon: and GetLexerOutput?

allenss-amazon: Not sure that the consumers of this care about any intermediate results.

src/indexes/text/lexer.h
```diff
@@ -3,28 +3,39 @@
...
 #include "absl/status/statusor.h"
-#include "src/text/text.h"
+#include "absl/strings/string_view.h"
+#include "src/index_schema.pb.h"
 
-namespace valkey_search::text {
+namespace valkey_search {
+namespace text {
+
+// Forward declaration
```
BCathcart, line 29: nit: I don't think this comment is necessary

src/indexes/text.cc
```diff
@@ -0,0 +1,154 @@
...
+  // TODO: stub
+  return std::make_unique<EntriesFetcher>();
+}
+
+} // namespace valkey_search::indexes
+
+namespace valkey_search {
+namespace text {
+
+absl::StatusOr<bool> TextFieldIndex::AddRecord(const InternedStringPtr& key,
```
Aksha1812, line 110: what is difference between above 
```
absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
                                     absl::string_view data) {
```

and this 
```

absl::StatusOr<bool> TextFieldIndex::AddRecord(const InternedStringPtr& key,
                                       absl::string_view data) {
```
 TextFieldIndex does not inherit IndexBase right?

src/indexes/text.h
```diff
@@ -30,6 +30,108 @@
...
+    text_ = std::make_shared<TextIndex>();
+  }
+
+  // Return the field identifier
+  const std::string& GetFieldIdentifier() const {
+    return field_identifier_;
+  }
+  ~TextFieldIndex() = default;
+
+  absl::StatusOr<bool> AddRecord(const InternedStringPtr& key,
```
Aksha1812, line 84: I guess we don't need all this for TextfieldIndex right. ? all these methods are from indexbase.

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
...
+  // Validate UTF-8
+  if (!IsValidUtf8(s)) {
+    return absl::InvalidArgumentError("Invalid UTF-8 sequence in input");
+  }
+  
+  LexerOutput output;
+  PunctuationBitmap punct_bitmap = CreatePunctuationBitmap(punctuation_);
+  
+  // Track positions for all words
+  output.has_positions_ = true;
```
BCathcart, line 220: Do we always track position?

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,352 @@
...
+}
+
+// Create punctuation bitmap from string
+PunctuationBitmap CreatePunctuationBitmap(const std::string& punctuation) {
+  PunctuationBitmap bitmap;
+  for (char c : punctuation) {
+    bitmap.set(static_cast<unsigned char>(c));
+  }
+  return bitmap;
+}
```
BCathcart, line 73: Just for my understanding what differentiates the helper functions you defined as private members of the class and these ones?

src/indexes/text.cc
```diff
@@ -0,0 +1,154 @@
...
+                                               absl::string_view data) {
+  auto lexer_result = lexer_.ProcessString(data);
+  if (!lexer_result.ok()) {
+    // Use debug logging for non-parseable data (very rare, not actionable)
```
BCathcart, line 114: Are there any restrictions the user can put on a text field when creating an index that might cause us to reject or is invalid utf-8 the only case?

I agree with Allen that a counter is probably better because a million rejected records in 60 seconds tells a lot different story than a single rejected record, but in either case all we'd know is one particular key was rejected

src/indexes/text.cc
```diff
@@ -0,0 +1,154 @@
...
+    // Increased interval to once per minute to prevent log spam
+    VMSDK_LOG_EVERY_N_SEC(DEBUG, nullptr, 60)
+        << "Text field analysis failed for key " << key->Str() << ": "
```
BCathcart, line 117: For managed services there could be some concerns around logging keys since they would be customer data. I'm wondering what the OSS policy is on this. Also, if we are going to log the key, we might as well specify the field name as well

src/indexes/text.h
```diff
@@ -30,6 +30,108 @@
...
+                 std::string field_identifier = "") 
+      : text_index_proto_(text_index_proto),
+        index_schema_proto_(index_schema_proto),
+        field_identifier_(field_identifier),
+        // Initialize lexer in the constructor's initialization list
+        lexer_(index_schema_proto_ ? 
+              Lexer(text_index_proto_, *index_schema_proto_) : 
+              Lexer(text_index_proto_, data_model::IndexSchema())) {
+    // Initialize text index structures
+    text_ = std::make_shared<TextIndex>();
```
BCathcart, line 75: Is this the actual text index object with the prefix and suffix tree? If so it should be initialized on FT.CREATE and associated with the schema rather than a particular text field

src/indexes/text/lexer.cc
BCathcart, line 75: What is this referring to?

src/indexes/text.h
Aksha1812, line 63: @allenss-amazon was talking about this class in the chat. Even if we need this I guess this wouldn't go in this file, as Text is per field/ attribute class. all this tokenization is schema level.

src/indexes/text/lexer.h
```diff
@@ -42,15 +53,69 @@ struct Lexer {
 
   std::string GetPunctuation() const;
 
-  // Instantiate with default Punctuation
-  Lexer();
+  // Enable/disable case conversion (enabled by default)
+  void SetCaseConversionEnabled(bool enabled);
```
BCathcart, line 57: Can you walk me a bit through the flow of how the lexer will be used? Will all these conditions that affect how we tokenize be specified during the index creation? If so, would it be better to set them during index creation and do away with these getters/setters (assuming these conditions can't be toggled afterwards)?

I assume the lexer instance per field index will be used for both the ingestion and query. Are there any conditions that can be specified at the query level to configure the tokenization? If there are then passing the configuration directly to `ProcessString()` might make more sense

src/indexes/text.h
BCathcart: I'm also wondering what is the motivation for this object. Seems like a duplication of `Text` below which implements `IndexBase` as expected

src/indexes/text.cc
```diff
@@ -0,0 +1,154 @@
...
+    // Use debug logging for non-parseable data (very rare, not actionable)
+    // Increased interval to once per minute to prevent log spam
+    VMSDK_LOG_EVERY_N_SEC(DEBUG, nullptr, 60)
+        << "Text field analysis failed for key " << key->Str() << ": " 
+        << lexer_result.status().message();
+    return lexer_result.status();
+  }
+  
+  // Get the processed terms
+  const auto& terms = lexer_result->GetProcessedTerms();
```
BCathcart, line 123: The return type of `ProcessString()` is `absl::StatusOr<LexerOutput>` which seems similar to a `Result` type in Rust where you're returning an error or an output object. Why not get the processed terms directly from the `lexer_result`. I also don't think you ever set the `processed_result_` data member in the object so this will probably always be empty

src/indexes/text/lexer.cc
Aksha1812: I guess, but we don't save it until save_positions is turned on, I guess it is just some metadata coming from original string. while optimizing maybe we can just remove pulling it if save_positions is turned off

src/indexes/text.h
VoletiRam: Yes, this is change is sort of ""Needs to be committed after xxx PR"". I built this in loose way as workaround since I have dependency. I don't want to spend too much on how indexing is gonna be implemented. May be I should've added a comment on what needs to be reviewed. The main change is on Lexical analysis which is in AddRecord() and please focus on reviewing it. I will update the PR to match with Nivesh change as his PR is in good rebase state now.

src/indexes/text.h
VoletiRam: Sure.

src/indexes/text/lexer.cc
VoletiRam: Lexer encompasses the full lexical analysis pipeline, not just tokenization. The term "lexer" correctly conveys that this module handles the complete process of breaking down and analyzing text at the lexical level, which includes but extends beyond simple tokenization.

src/indexes/text/lexer.cc
VoletiRam: Will remove this

src/indexes/text.cc
VoletiRam: It should've been Text, yes

src/indexes/text/lexer.cc
Aksha1812: Got it. thanks you. makes sense

PR #279, opened 2025-08-04
Supporting Text Query Parsing - And implementing Exact Term Search 

KarthikSubbarao:
Draft PR while we are integrating the FTS code - Supporting Text Query Parsing - And implementing Exact Term Search .

TODO:
- Assigning the exact term operator
- Search operation
- Coordinator fanout
- Testing

### DIFF
--- a/src/query/predicate.h
+++ b/src/query/predicate.h
@@ -19,6 +19,7 @@
 #include "vmsdk/src/type_conversions.h"
 
 namespace valkey_search::indexes {
+class Text;
 class Numeric;
 class Tag;
 }  // namespace valkey_search::indexes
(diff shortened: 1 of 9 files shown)


src/query/predicate.h
```diff
@@ -19,13 +19,15 @@
...
 namespace valkey_search::indexes {
+class Text;
 class Numeric;
 class Tag;
 }  // namespace valkey_search::indexes
 
 namespace valkey_search::query {
 
 enum class PredicateType {
+  kText,
```
allenss-amazon, line 30: Are these on the network? If so it might be better to move the new value to end to keep the existing numbers intact.

PR #282, opened 2025-08-05
Integrating snowball stemmer

Aksha1812:
## Integrating snowball stemmer

Adding third party support for snowball stemmer. Made a vendor copy, same as other third party libs added. [Reference](https://github.com/valkey-io/valkey-search/commit/ecc2b4e3b803c4ed69af00342292d52f83e327a7#diff-cb5c5e2931e06958148802e2622d94721ac084f953bf83a8af2955ff9c837ed8). Snowball has a set of .sbl files which are rules for stemming for each language which are converted to programming language binaries to define rules of stemming. have kept compiled c binary files which support the language and other utility files necessary to interpret these language binaries. 

Added add_language.sh , remove_language.sh , update_language.sh scripts. to manage supported languages. 

Currently used version specified in 'VERSION' file as done in other copies of third party libs. all new languages added will be of the same consistent version mentioned here. Update will update all language binaries with specified version, if not specified it will default to the latest stable version.

Current version : **v3.0.1**

### Usage
```bash
./add_language.sh french german spanish
```

```bash
./remove_language.sh french german
```

```bash
./update_languages.sh --version v2.2.0
```

PS :  Have added temporary stemmer files until rest of tokenization #272 is ready, to make sure snowball integration is working. Once #272 is ready will modify APIs.


allenss-amazon (approved):

src/indexes/text/lexer.cc
```diff
@@ -0,0 +1,55 @@
...
+  }
+  
+  stemmer_ = sb_stemmer_new(language.c_str(), "UTF_8");
+  CHECK(stemmer_) << "Failed to initialize stemmer for language: " + language;
+  
+  return absl::OkStatus();
+}
+
+// TODO : Add test for stemming after rest of lexical operations are added
+absl::StatusOr<std::string> Lexer::StemWord(const std::string& word) const {
```
allenss-amazon, line 36: This never fails, why StatusOr ?? Do you expect it to fail in the future?

Aksha1812: yes. likely scenarios are stemming failures, invalid utf-8 input, other unexpected failures from snowball library. will update this based on how Ram implements lexer and what all checks are covered by him. if those checks are good enough can remove statusOr later.

Issue #255, opened 2025-07-27
Existing JSON data in replica not backfilled into index


ichenhe:
## Environment

- The latest valkey-search code 526c4c5856e7a3fb4cb9acb80a993c9aa45d97ba.
- Cluster mode: enabled
- Replica num: >= 1

## Summary

In an cluster-mode-enabled cluster with replicas, when a new index is created or restored (for example via full sync), existing JSON data in replica is not backfilled into the index. As a result, searching with index on replica does not find existing JSON keys.

## Investigation

We observed that this issue impacts only JSON data and only the backfill on replica. We observed this issue when triggering a backfill after new index creation on cluster-mode enabled, and when triggering a backfill from RDB restore for both cluster-mode enabled and disabled.

From our initial investigation, we found that  `VectorExternalizer` does not return any data when retrieving JSON records when executed within the backfill scan context on the replica, while the same works correctly for HASH keys. We found that `VectorExternalizer` uses different APIs to retrieve data from Valkey for both data types - in HASH data is retrieved via module API VM_HashGet (see [here](https://github.com/valkey-io/valkey-search/blob/20ceefaeecd5fddca8fa763a16401851ceda6c57/src/attribute_data_type.cc#L32)) while in JSON data is retrieved via JSON.GET command (see [here](https://github.com/valkey-io/valkey-search/blob/20ceefaeecd5fddca8fa763a16401851ceda6c57/src/attribute_data_type.cc#L127-L128)).

During RDB restore, we observe that although the indexed JSON data is restored from RDB, data is removed from the index at the end of the restore process when applying staged data after RDB loading. Since the backfill doesn’t find the JSON keys associated with the indexed data, the data is removed from the index (see [here](https://github.com/valkey-io/valkey-search/blob/20ceefaeecd5fddca8fa763a16401851ceda6c57/src/index_schema.cc#L411)) in the ProcessAttributeMutation function.

As an experiment, we confirmed that replacing VM_HashGet with HGET results in the same error for HASH datatype during backfill scan context on replica. We also confirmed that this only impacts the backfill - when a key is processed in the backfill scan context - and not steady state. The VectorExternalizer retrieves JSON data correctly when processing key space notifications.

## Reproduce

Run on primary node:

```
JSON.SET json:0 $ '{"vec": [-1, 0]}'
[...]

PR #287, opened 2025-08-07
Add support to retrieving JSON values via the JSON shared API

yairgott:
This change addresses the issue reported in [issue #255](https://github.com/valkey-io/valkey-search/issues/255).

Support for the JSON shared API is WIP and introduced in [this PR](https://github.com/valkey-io/valkey-json/pull/71).

To avoid introducing a hard dependency on a specific valkey-json release, the implementation is backward compatible: it prefers the new shared API if available, and falls back to VM_Call otherwise.

Once the valkey-json changes are included in an official release, we plan to deprecate the fallback logic and enforce valkey-json version that includes the new JSON shared API.

### DIFF
--- a/src/attribute_data_type.cc
+++ b/src/attribute_data_type.cc
@@ -102,13 +102,6 @@ absl::StatusOr<RecordsMap> HashAttributeDataType::FetchAllRecords(
   return std::move(callback_data.key_value_content);
 }
 
-absl::string_view TrimBrackets(absl::string_view record) {
-  if (absl::ConsumePrefix(&record, "[")) {
-    absl::ConsumeSuffix(&record, "]");
-  }
-  return record;
-}
-
 absl::Status NormalizeJsonRecord(absl::string_view record,
                                  vmsdk::UniqueValkeyString &out_record) {
   if (!record.empty() && record[0] != '[') {
@@ -150,10 +143,10 @@ absl::Status GetJsonRecord(ValkeyModuleCtx *ctx, ValkeyModuleKey *open_key,
       return absl::NotFoundError(
           absl::StrCat("No such record with identifier: `", identifier, "`"));
     }
+    auto record_tmp = vmsdk::UniqueValkeyString(record_str);
     if (!record) {
       return absl::OkStatus();
     }
-    auto record_tmp = vmsdk::UniqueValkeyString(record_str);
     return NormalizeJsonRecord(vmsdk::ToStringView(record_tmp.get()), *record);
   }
   auto reply = vmsdk::UniquePtrValkeyCallReply(ValkeyModule_Call(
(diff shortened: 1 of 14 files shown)


src/attribute_data_type.cc
```diff
@@ -116,46 +123,77 @@ absl::StatusOr<vmsdk::UniqueValkeyString> NormalizeJsonRecord(
...
+    }
+    ValkeyModuleString *record_str = nullptr;
+    if (json_get(open_key, identifier.data(), &record_str) ==
+        VALKEYMODULE_ERR) {
+      return absl::NotFoundError(
+          absl::StrCat("No such record with identifier: `", identifier, "`"));
+    }
+    if (!record) {
+      return absl::OkStatus();
+    }
```
allenss-amazon, line 149: This will leak a string if it's true. Not sure why this is here, maybe move it up? Are there cases where you need to know that JSON data is present, but don't care about the actual values?

src/attribute_data_type.cc
```diff
@@ -166,7 +204,26 @@ absl::StatusOr<RecordsMap> JsonAttributeDataType::FetchAllRecords(
...
+  }
+  json_get = (int (*)(ValkeyModuleKey *, const char *, ValkeyModuleString **))
+      ValkeyModule_GetSharedAPI(ctx, "JSON_GetValue");
+  // Note: In cluster mode, replicas must have the JSON module loaded to access
+  // the JSON shared API. Otherwise, invoking commands via ValkeyModule_Call
+  // from a replica will result in a MOVED response.
+  if (!json_get && ValkeySearch::Instance().IsCluster()) {
+    VMSDK_LOG(WARNING, ctx)
+        << "Note: When cluster mode is enabled, valkey-search requires "
+           "valkey-json version 1.02 or higher for proper JSON support.";
```
allenss-amazon, line 218: Are we certain this is the right json version? Maybe just say that this JSON module doesn't support search?

testing/attribute_data_type_test.cc
```diff
@@ -272,12 +278,30 @@ INSTANTIATE_TEST_SUITE_P(
...
+  }
   ValkeyModuleCtx fake_ctx;
   JsonAttributeDataType json_attribute_data_type;
+  vmsdk::UniqueValkeyString key_str;
+  vmsdk::UniqueValkeyOpenKey key_obj;
+  const std::string query_attribute_name = "vector_json_path";
+  absl::flat_hash_map<std::string, std::string> json_path_results = {
+      {"$", "[res0]"},
+      {"json_path_results1", "[res1]"},
+      {"json_path_results2", "[res2]"}};
```
allenss-amazon: This isn't a valid JSON response. For a single array element with a string, you'll need double quotes around the interior string.

allenss-amazon (changes requested):
Need to fix the string leak, then i think we can merge it. But before we can close the Issue, we'll need to get the JSON code committed AND build an integration test.

src/attribute_data_type.cc
yairgott: Good catch!

testing/attribute_data_type_test.cc
yairgott: ack

src/attribute_data_type.cc
yairgott: The current valkey-json release version is 1.01.

allenss-amazon (approved):

Issue #227, opened 2025-07-15
Bug: FLUSHALL Does Not Reset Index `num_docs` and `capacity`


liatweissman:
## Description
When using `FLUSHALL` on a CME instance with Valkey-Search, the keys and indexes are deleted as expected. However, the automatically recreated index retains its previous `num_docs` value and `capacity`, instead of resetting to an empty state. This causes inconsistencies between the actual keyspace and the reported index metadata.

---

### **Expected Behavior**

* `FLUSHALL` should:

  * Delete all keys and associated indexes.
  * Recreate the index with:

    * `num_docs = 0`
    * Default capacity and metadata

---

### **Actual Behavior**

After running `FLUSHALL`:

* Keys and indexes are removed correctly.
* But the auto-recreated index has:

  * `num_docs = X` (same as before the flush, e.g., `"40000"`)
  * Retained capacity and metadata (same as pre-flush)

This implies the index is *not* starting fresh, even though no documents exist.

---

### **Reproduction Steps**

1. Create a vector index with documents
2. Run `FLUSHALL`
3. Run `FT.INFO <index_name>`
4. Observe `num_docs` and `capacity` values

---

### **Observed Output**

```
:6379> FLUSHALL
:6379> FT.INFO hash_idx:0:0

...

 9) num_docs
10) "40000"   <--- Should be 0
...
 8) index
    8) 1) capacity
       2) (integer) 40960   <--- Still retains old capacity
...
```

Inserting a single new document results in:

```
num_docs = "40001"
num_records = "1"
```

Which suggests the previous count is not cleared on FLUSH.

---

### **Relevant Logs**

```
flushall command received
Deleting index schema hash_idx:0:0 on FLUSHDB
Recreating index schema hash_idx:0:0 on FLUSHDB
Starting backfill for index schema in DB 0: hash_idx:0:0
Index schema hash_idx:0:0 dropped from DB 0
Index schema hash_idx:0:0 finished backfill. Scanned 0 keys in 166ms
```

allenss-amazon:
I believe this is fixed now, please re-test.

PR #288, opened 2025-08-07
Fix num_docs not getting reset across a flushall. Add flushall test

allenss-amazon:
As cited in issue #227 the "num_docs" field in FT.INFO isn't reset after a FLUSHALL. This only happens in CME because in CMD the indexes are deleted.


src/index_schema.cc
```diff
@@ -146,7 +146,7 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
 
 absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::Create(
     ValkeyModuleCtx *ctx, const data_model::IndexSchema &index_schema_proto,
-    vmsdk::ThreadPool *mutations_thread_pool, bool skip_attributes) {
+    vmsdk::ThreadPool *mutations_thread_pool, bool skip_attributes, bool reload) {
```
murphyjacob4, line 149: Would it make more sense to just zero out the stats explicitly here before? 

https://github.com/valkey-io/valkey-search/blob/8b53ed55a179ffb076f30cdfd03a4e63ef859eb5/src/schema_manager.cc#L432

The parameter `reload` isn't very intuitive to me - I guess it is to differentiate between new index schemas and index schemas from RDB?

allenss-amazon: Yes, I can certainly change the name, maybe to rdb_load?

The code you cite might not be the only path that this failure can occur. What happens on a metadata reconciliation? Is the document_count field from the wire going to replace the localized version? That's wrong. I believe the only case where you should be using the protobug.document_count is when it's coming from an RDB.

src/index_schema.cc
```diff
@@ -201,16 +202,22 @@ IndexSchema::IndexSchema(ValkeyModuleCtx *ctx,
...
+      }
     }
   }
-  stats_.document_cnt = index_schema_proto.stats().documents_count();
+  // The protobuf has volatile fields that get save/restores in the RDB. here we
+  // reconcile the source of the index_schema_proto (reload or not) and restore those fields
+  if (reload) {
+    stats_.document_cnt = index_schema_proto.stats().documents_count();
+  } else {
+    stats_.document_cnt = 0;
```
yairgott: This is not required, `document_cnt` has a default value of 0.

allenss-amazon:
Oh, and you need to fix the DCO.

PR #289, opened 2025-08-07
Rough code to kickstart basic text ingestion integration

BCathcart:
Some hygiene:
- moved all the files in the `text/` directory to `valkey_search::indexes::text` namespace
- updated the license headers in the `text/` folder
- have the postings object expect interned strings for the keys

Some very rudimentary ingestion pipeline work:
- Pass a "shared" `TextIndex` object to the `Text` attribute-level object. It's not really shared right now, so each text attribute will have a separate `TextIndex`

Next steps:
- Implement the mutation lambda that operates on the postings object. We also need to hook up `SetTextConfiguration()` and pass this down to the `Text` object so it can initialize postings objects correctly.
- Provide a text field identifier number to the `Text` objects on creation
- Make the `TextIndex` structure actually schema-level and shared amongst all text fields in a schema


src/indexes/text/text_index.h
```diff
@@ -0,0 +1,54 @@
+/*
```
Aksha1812, line 1: do we need to move it inside text? all other indexing classes are part of src/indexes. and this is schema level to , My understanding is src/indexes/text so more like a collection of utils / helper files.

src/index_schema.cc
```diff
@@ -884,19 +892,21 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::LoadFromRDB(
   // Supplemental content will include indices and any content for them
   while (supplemental_iter.HasNext()) {
     VMSDK_ASSIGN_OR_RETURN(auto supplemental_content, supplemental_iter.Next());
-    if (ABSL_PREDICT_TRUE(!skip_loading_index_data) && supplemental_content->type() ==
-        data_model::SupplementalContentType::
-            SUPPLEMENTAL_CONTENT_INDEX_CONTENT) {
+    if (ABSL_PREDICT_TRUE(!skip_loading_index_data) &&
```
Aksha1812, line 895: you can leave this unchanged too if clang passes on the earlier one

src/indexes/text/text_index.h
BCathcart: Idk it made a bit more sense to me this way but doesn't matter too much. I liked seeing one top level file that implements the standard IndexBase API. Then all the supporting structures can live in the text folder. I agree it's a bit weird that some of the supporting structures are owned at the schema level in the text case and it won't only be `text.h` code that uses this file. I have no problem shuffling things around in the future again if we agree it makes more sense

src/index_schema.cc
BCathcart: This is just standard clang formatting changes. I guess clang wasn't applied here before

allenss-amazon (approved):

PR #292, opened 2025-08-07
Bridge some gaps in ingestion pipeline

Aksha1812:
Building on top of #289 

```
class Text 
├── Inherits IndexBase
├── Implemented AddRecord API , placeholder for rest.
├── Text takes the whole index_schema_proto, text_index_proto, shared reference to the TextIndexSchema class, text field number as parameters

class TextIndex
├── Holds the prefix tree and optional suffix tree. (postings are shared between prefix and suffix tree
├── wrapper around the trees, we need to go via TextIndex to mutate on the tree , get word iterator, get path iterator
├── resides in the TextIndexSchema, One TextIndex per Schema. 

class TextIndexSchema
├── wrapper between IndexSchema and TextIndex. 
├── While initializing IndexBases per field/attribute if we encounter a text field and if TextIndexSchema reference doesn't already exist we initialize a Shared Instance of TextIndexSchema. This happens in Initialized in IndexSchema -> IndexFactory. We have one TextIndexSchema per IndexSchema. 
├── We have one TextIndex for one TextIndexSchema, num_text_fields and a hashmap<Key, TextIndex> to support the Delete record and the post-filtering case
├── this TextIndexSchema is passed in Text class as a parameter, help reach TextIndex which leads to prefix/suffix trees. also passes num_text_fields #295
```

- updated posting to take only properties needed, rather than whole index_schema
- Implemented ToProto() method as well

Note : Actively adding more to this PR, and open to suggestions. End goal to make ingestion pipeline work end-to-end

### DIFF
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -6,6 +6,7 @@
  */
 
 #include "src/index_schema.h"
+#include "src/indexes/text/text_index.h"
 
 #include <algorithm>
 #include <atomic>
@@ -84,13 +85,15 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
       return std::make_shared<indexes::Numeric>(index.numeric_index());
     }
     case data_model::Index::IndexTypeCase::kTextIndex: {
-      // TODO: make this TextIndex a proper schema-level object
-      std::shared_ptr<indexes::text::TextIndex> index_structure =
-          std::make_shared<indexes::text::TextIndex>(
-              indexes::text::TextIndex());
-      // TODO: pass in a unique text field ID number
+      // Create or reuse shared TextIndexSchema
+      if (!index_schema->GetTextIndexSchema()) {
+        index_schema->SetTextIndexSchema(std::make_shared<indexes::TextIndexSchema>());
+      }
+      //TODO : Increment logic 
+      //index_schema->text_index_schema_->num_text_fields_++;
+      // TODO: pass in a unique text field ID number 
       return std::make_shared<indexes::Text>(index.text_index(),
-                                             index_structure);
+                                             index_schema->GetTextIndexSchema(), 0);
     }
     case data_model::Index::IndexTypeCase::kVectorIndex: {
       switch (index.vector_index().algorithm_case()) {
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -87,7 +87,7 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
     case data_model::Index::IndexTypeCase::kTextIndex: {
       // Create or reuse shared TextIndexSchema
       if (!index_schema->GetTextIndexSchema()) {
-        index_schema->SetTextIndexSchema(std::make_shared<indexes::TextIndexSchema>());
+        index_schema->CreateTextIndexSchema();
       }
       //TODO : Increment logic 
       //index_schema->text_index_schema_->num_text_fields_++;
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -92,7 +92,9 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
       //TODO : Increment logic 
       //index_schema->text_index_schema_->num_text_fields_++;
       // TODO: pass in a unique text field ID number 
+      auto index_schema_proto = index_schema->ToProto();
       return std::make_shared<indexes::Text>(index.text_index(),
+                                             *index_schema_proto,
                                              index_schema->GetTextIndexSchema(), 0);
     }
     case data_model::Index::IndexTypeCase::kVectorIndex: {
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -87,8 +87,8 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
     case data_model::Index::IndexTypeCase::kTextIndex: {
       // Create or reuse shared TextIndexSchema
       if (!index_schema->GetTextIndexSchema()) {
-        auto index_schema_proto = index_schema->ToProto();
-        index_schema->CreateTextIndexSchema(*index_schema_proto);
+        // TODO : get the index_schema_proto the right way here to pass schema level properties
+        index_schema->CreateTextIndexSchema();
       }
       //TODO : Increment logic 
       //index_schema->text_index_schema_->num_text_fields_++;
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -87,14 +87,13 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
     case data_model::Index::IndexTypeCase::kTextIndex: {
       // Create or reuse shared TextIndexSchema
       if (!index_schema->GetTextIndexSchema()) {
-        index_schema->CreateTextIndexSchema();
+        auto index_schema_proto = index_schema->ToProto();
+        index_schema->CreateTextIndexSchema(*index_schema_proto);
       }
       //TODO : Increment logic 
       //index_schema->text_index_schema_->num_text_fields_++;
       // TODO: pass in a unique text field ID number 
-      auto index_schema_proto = index_schema->ToProto();
       return std::make_shared<indexes::Text>(index.text_index(),
-                                             *index_schema_proto,
                                              index_schema->GetTextIndexSchema(), 0);
     }
     case data_model::Index::IndexTypeCase::kVectorIndex: {
--- a/src/index_schema.h
+++ b/src/index_schema.h
@@ -31,6 +31,7 @@
 #include "src/index_schema.pb.h"
 #include "src/indexes/index_base.h"
 #include "src/indexes/vector_base.h"
+#include "src/indexes/text/text_index.h"
 #include "src/keyspace_event_manager.h"
 #include "src/rdb_serialization.h"
 #include "src/utils/string_interning.h"
@@ -97,13 +98,8 @@ class IndexSchema : public KeyspaceEventSubscription,
   inline const std::string &GetName() const { return name_; }
   inline std::uint32_t GetDBNum() const { return db_num_; }
 
-  bool GetSavePositions() const { return save_positions_; }
-  size_t GetNumTextFields() const { return num_text_fields_; }
-  // TODO: Change this after query support is added for full text search
-  void SetTextConfiguration(bool save_positions, size_t num_text_fields) {
-    save_positions_ = save_positions;
-    num_text_fields_ = num_text_fields;
-  }
+  std::shared_ptr<indexes::TextIndexSchema> GetTextIndexSchema() const { return text_index_schema_; }
+  void SetTextIndexSchema(std::shared_ptr<indexes::TextIndexSchema> schema) { text_index_schema_ = schema; }
 
   void OnKeyspaceNotification(ValkeyModuleCtx *ctx, int type, const char *event,
                               ValkeyModuleString *key) override;
@@ -176,6 +172,7 @@ class IndexSchema : public KeyspaceEventSubscription,
   uint32_t db_num_{0};
   bool save_positions_{true};
   size_t num_text_fields_{0};
+  std::shared_ptr<indexes::TextIndexSchema> text_index_schema_;
 
   vmsdk::ThreadPool *mutations_thread_pool_{nullptr};
   InternedStringMap<DocumentMutation> tracked_mutated_records_
--- a/src/index_schema.h
+++ b/src/index_schema.h
@@ -99,8 +99,12 @@ class IndexSchema : public KeyspaceEventSubscription,
   inline std::uint32_t GetDBNum() const { return db_num_; }
 
   std::shared_ptr<indexes::text::TextIndexSchema> GetTextIndexSchema() const { return text_index_schema_; }
-  void CreateTextIndexSchema(const data_model::IndexSchema& index_schema_proto) { 
-    text_index_schema_ = std::make_shared<indexes::text::TextIndexSchema>(index_schema_proto); 
+  void CreateTextIndexSchema(const data_model::IndexSchema* index_schema_proto = nullptr) { 
+    if (index_schema_proto) {
+      text_index_schema_ = std::make_shared<indexes::text::TextIndexSchema>(*index_schema_proto); 
+    } else {
+      text_index_schema_ = std::make_shared<indexes::text::TextIndexSchema>(); 
+    }
   }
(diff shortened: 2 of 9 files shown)


src/indexes/text.h
```diff
@@ -8,19 +8,33 @@
...
-#include "src/query/predicate.h"
+#include "src/index_schema.pb.h"
+#include "src/utils/string_interning.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
+// Forward declarations
+namespace valkey_search::query {
+class TextPredicate;
+}
+
```
VoletiRam: We don't need this yet

src/indexes/text/text_index.h
```diff
@@ -30,15 +36,24 @@ struct TextIndex {
...
+  std::optional<text::RadixTree<std::shared_ptr<text::Postings>, true>> suffix_;
 };
 
-struct IndexSchemaText {
+class TextIndexSchema {
+ public:
   //
   // This is the main index of all Text fields in this index schema
   //
-  std::shared_ptr<TextIndex> corpus_;
```
VoletiRam, line 41: Why change corpus?

Aksha1812: just changed the name `std::shared_ptr<indexes::TextIndex> text_index_;`

src/indexes/text.h
Aksha1812: removed

src/indexes/text/text_index.h
VoletiRam: Is this for text field index? What are we gonna store here?

VoletiRam (approved):

src/indexes/text/text_index.h
VoletiRam: This is for prefix and suffix tree Text index.

src/index_schema.cc
```diff
@@ -84,13 +85,15 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
...
     }
     case data_model::Index::IndexTypeCase::kTextIndex: {
-      // TODO: make this TextIndex a proper schema-level object
-      std::shared_ptr<indexes::text::TextIndex> index_structure =
-          std::make_shared<indexes::text::TextIndex>(
-              indexes::text::TextIndex());
-      // TODO: pass in a unique text field ID number
+      // Create or reuse shared TextIndexSchema
+      if (!index_schema->GetTextIndexSchema()) {
+        index_schema->SetTextIndexSchema(std::make_shared<indexes::TextIndexSchema>());
```
BCathcart: Very minor change but maybe we can have the function called `CreateTextIndexSchema` and construct the shared pointer in it?

src/index_schema.cc
```diff
@@ -84,13 +85,15 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
...
+      }
+      //TODO : Increment logic 
+      //index_schema->text_index_schema_->num_text_fields_++;
```
BCathcart, line 94: This will be done in @VoletiRam's follow up?

src/indexes/CMakeLists.txt
```diff
@@ -94,14 +94,17 @@ target_link_libraries(vector_flat PUBLIC memory_allocation_overrides)
 target_link_libraries(vector_flat PUBLIC status_macros)
 target_link_libraries(vector_flat PUBLIC valkey_module)
 
-set(SRCS_TEXT ${CMAKE_CURRENT_LIST_DIR}/text.cc
+set(SRCS_TEXT ${CMAKE_CURRENT_LIST_DIR}/text/text_index.h
+              ${CMAKE_CURRENT_LIST_DIR}/text.cc
               ${CMAKE_CURRENT_LIST_DIR}/text.h
+              ${CMAKE_CURRENT_LIST_DIR}/text/radix_tree.h
```
BCathcart: Just curious for my own understanding if we actually need to include `radix_tree.h` here for things to compile?

src/indexes/text.h
```diff
@@ -81,6 +84,17 @@ class Text : public IndexBase {
       ABSL_NO_THREAD_SAFETY_ANALYSIS;
 
  private:
+  // Each text field is assigned a unique number within the containing index, this is used
```
BCathcart: "Each text field index within the schema is assigned a unique number"?  Might be a bit more clear than "containing index"

src/indexes/text.h
```diff
@@ -81,6 +84,17 @@ class Text : public IndexBase {
...
+  // by the Postings object to identify fields.
+  size_t text_field_number_;
+  
+  // Reference to the shared text index schema
+  std::shared_ptr<TextIndexSchema> text_index_schema_;
+  
+  // Map to track which keys are indexed and their raw data
+  absl::flat_hash_map<InternedStringPtr, InternedStringPtr> tracked_tags_by_keys_
+      ABSL_GUARDED_BY(index_mutex_);
```
BCathcart: If this isn't used right now I'd just remove it and add an item to Karthik's doc tracking follow-up items. I think there were some questions around how we want to implement this

src/indexes/text/text_index.h
```diff
@@ -1,24 +1,30 @@
-/*
- * Copyright (c) 2025, valkey-search contributors
- * All rights reserved.
- * SPDX-License-Identifier: BSD 3-Clause
- *
- */
```
BCathcart, line 6: Good catch on the misspelling in the #define but let's put the copyright notice back

src/indexes/text/text_index.h
```diff
@@ -1,24 +1,30 @@
...
+#include "absl/strings/string_view.h"
+#include "absl/functional/function_ref.h"
+#include "absl/container/flat_hash_map.h"
 #include "src/indexes/text/radix_tree.h"
-#include "src/utils/string_interning.h"
+#include "src/indexes/text/posting.h"
+#include "src/index_schema.pb.h"
 
-namespace valkey_search::indexes::text {
+namespace valkey_search::indexes {
```
BCathcart: Why change the namespace? I was basing the namespace off the file structure and you can access the RadixTree and Postings objects without prepending `text::` if you keep the namespace

src/indexes/text/text_index.h
```diff
@@ -30,15 +36,24 @@ struct TextIndex {
...
+  // Prefix tree
+  text::RadixTree<std::shared_ptr<text::Postings>, false> prefix_;
+  
+  // Suffix tree
+  std::optional<text::RadixTree<std::shared_ptr<text::Postings>, true>> suffix_;
 };
 
-struct IndexSchemaText {
+class TextIndexSchema {
+ public:
```
BCathcart: I think changing these to public and making them classes is kind of redundant. Struct members are public by default.

src/indexes/text/text_index.h
```diff
@@ -30,15 +36,24 @@ struct TextIndex {
...
   //
   // This is the main index of all Text fields in this index schema
   //
-  std::shared_ptr<TextIndex> corpus_;
+  TextIndexSchema() : num_text_fields_(0), text_index_(std::make_shared<TextIndex>()) {}
+  ~TextIndexSchema() = default;
+
+  uint8_t num_text_fields_;
+  std::shared_ptr<indexes::TextIndex> text_index_;
```
BCathcart: Why remove the comment about text_index_ here? I don't think it hurts to be a bit more explicit unless we have a good reason to remove it

BCathcart (commented):
Looks good overall. Some minor comments. In particular I'd suggest putting some things back in `text_index.h` to the way they were

src/indexes/CMakeLists.txt
Aksha1812: I guess no since it is #included in text_index file and text_index has been added here

src/index_schema.cc
Aksha1812: yes

src/index_schema.cc
Aksha1812: changed

src/indexes/text.h
Aksha1812: changed

src/indexes/text.h
Aksha1812: changed

src/indexes/text/text_index.h
Aksha1812: changed

src/indexes/text/text_index.h
Aksha1812: changed

src/indexes/text/text_index.h
Aksha1812: changed

src/indexes/text/text_index.h
Aksha1812: changed

src/indexes/text.h
```diff
@@ -37,13 +47,6 @@ class Text : public IndexBase {
...
   }
 
- private:
-  // Each text field is assigned a unique number within the containing index,
-  // this is used by the Postings object to identify fields.
-  // size_t text_field_number_; - TODO
-  // 
-  std::shared_ptr<text::TextIndex> text_index_;
-
   inline void ForEachTrackedKey(
```
BCathcart, line 50: Maybe we shouldn't change this to public without a good reason but this can be done in a follow-up

src/indexes/text/text_index.h
```diff
@@ -49,6 +64,6 @@ struct IndexSchemaText {
   absl::flat_hash_map<Key, TextIndex> by_key_;
 };
 
-}  // namespace valkey_search::indexes::text
+}  // namespace valkey_search::indexes
```
BCathcart: Unimportant but this could be left alone too

src/indexes/text.h
Aksha1812: haven't changed it to public. replaced by text_index_schema_. if i understand correctly. lmk if i am talking about something else

BCathcart: You moved `private:` below and it's public up to that point

src/indexes/text.cc
```diff
@@ -17,37 +17,44 @@
...
+                postings = existing.value();
+              } else {
+                // Create new Postings object with schema configuration
+                // TODO: Get save_positions from IndexSchema, for now assume true
+                bool save_positions = true;
+                uint8_t num_text_fields = text_index_schema_->num_text_fields_;
+                postings = std::make_shared<text::Postings>(save_positions, num_text_fields);
+              }
+              
+              // Add the key and position to postings using correct API
```
BCathcart: This sounds like a TODO. I assume we're already using the correct API?

src/indexes/text.cc
```diff
@@ -17,37 +17,52 @@
...
+    : IndexBase(IndexerType::kText), 
+      text_field_number_(text_field_number),
+      text_index_schema_(text_index_schema),
+      with_suffix_trie_(text_index_proto.with_suffix_trie()),
+      no_stem_(text_index_proto.no_stem()),
+      min_stem_size_(text_index_proto.min_stem_size()),
+      language_(index_schema_proto.language()),
+      punctuation_(index_schema_proto.punctuation()),
+      with_offsets_(index_schema_proto.with_offsets()),
+      stop_words_(index_schema_proto.stop_words().begin(), index_schema_proto.stop_words().end()) {
```
BCathcart: What do you think about storing these inside the TextIndexSchema instead since they are schema-level? Then we wouldn't have to have a separate copy for every text field index

VoletiRam: Not all values are schema-level. We have values of both per-field and schema level here. I think it makes sense to store them together in each text field index or at Lexer class as we use these values over there.

If we store the schema level configurations in Text-schema, we need to pass down Text schema to Lexer. I would suggest to pass down entire index reference upto lexer where we need them and not assign the values here at text field

Aksha1812: Sorry, didn't follow @VoletiRam 's comment. are saying the same thing as Brennan?

src/index_schema.h
```diff
@@ -176,6 +172,7 @@ class IndexSchema : public KeyspaceEventSubscription,
   uint32_t db_num_{0};
   bool save_positions_{true};
   size_t num_text_fields_{0};
```
VoletiRam: You forget to delete these two

src/indexes/text.cc
VoletiRam: No. I am saying we cannot store `text_index_proto` configurations in TextIndexSchema as they are per-field level.

src/index_schema.cc
```diff
@@ -84,13 +85,16 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
...
     case data_model::Index::IndexTypeCase::kTextIndex: {
-      // TODO: make this TextIndex a proper schema-level object
-      std::shared_ptr<indexes::text::TextIndex> index_structure =
-          std::make_shared<indexes::text::TextIndex>(
-              indexes::text::TextIndex());
-      // TODO: pass in a unique text field ID number
+      // Create or reuse shared TextIndexSchema
+      if (!index_schema->GetTextIndexSchema()) {
+        auto index_schema_proto = index_schema->ToProto();
+        index_schema->CreateTextIndexSchema(*index_schema_proto);
```
VoletiRam: You are essential creating new copy of index schema proto here. Can we avoid this?

Aksha1812: we don't get the index_schema_proto in the index factory. we take the index_schema and get the proto from there. if we don't want to use this copy we would need to pass it as parameter to index_schema_factory. wanted to avoid changing that

src/indexes/text/text_index.h
```diff
@@ -46,9 +46,9 @@ struct TextIndex {
 };
 
 struct TextIndexSchema {
-  TextIndexSchema() : num_text_fields_(0), text_index_(std::make_shared<TextIndex>()) {}
+  TextIndexSchema() : num_text_fields_(1), text_index_(std::make_shared<TextIndex>()) {}
   TextIndexSchema(const data_model::IndexSchema& index_schema_proto) 
-      : num_text_fields_(0), 
+      : num_text_fields_(1),
```
KarthikSubbarao, line 51: We can address this in Ram's PR

src/indexes/text.cc
```diff
@@ -70,15 +80,21 @@ int Text::RespondWithInfo(ValkeyModuleCtx* ctx) const {
 }
 
 bool Text::IsTracked(const InternedStringPtr& key) const {
-  throw std::runtime_error("Text::IsTracked not implemented");
+  return false;
```
KarthikSubbarao, line 83: We will need the mapping for updates. Right now, this works for `Add` operations which is good for the initial implementation. It needs to support Modify operations next

KarthikSubbarao (approved):

PR #294, opened 2025-08-07
Supporting Text Query Parsing and search logic for Exact Term Matching

KarthikSubbarao:
The PR adds support for Text Query Parsing and implementing exact term search logic using word & key iterators.

This is an early layout of code with exact term search logic functional. Query and Search code paths will evolve in follow up PRs as we support more advanced text queries.

Added Integration tests for exact term matching.

```
$VALKEY_CLI -p 7001 FT.CREATE products ON HASH PREFIX 1 "product:" SCHEMA category TAG  price NUMERIC rating NUMERIC desc TEXT desc2 TEXT 
$VALKEY_CLI -p 7001 HSET product:1 category "electronics" name "Laptop" price 999.99 rating 4.5 desc "hello" desc2 "hello2"
$VALKEY_CLI -p 7001 HSET product:2 category "electronics" name "Tablet" price 499.00 rating 4.0 desc "hello" desc2 "hello2"
$VALKEY_CLI -p 7001 HSET product:6 category "electronics" name "Tablet" price 499.00 rating 4.0 desc "hello" desc2 "hello2"
$VALKEY_CLI -p 7001 HSET product:3 category "electronics" name  "Phone" price 299.00 rating 3.8 desc "byte" desc2 "hello"
$VALKEY_CLI -p 7001 HSET product:4 category "books"  name "Book" price 19.99 rating 4.8 desc "world" desc2 "hello2"
~/valkey/src/valkey-cli -p 7001 FT.SEARCH products '@desc:"world"
```

Will be handled in follow up PRs:
1. Exact Phrase (sentence) - Follow up
2. CME support
3. Default field search (when a field is not specified)
4. Filtering of fields (attributes) in the Search path. Right now, the Key Iterator of the Postings object does not expose field info and also does not allow filtering of specific fields. This will be needed for correctness.


src/commands/filter_parser.cc
```diff
@@ -217,6 +218,45 @@ absl::StatusOr<absl::string_view> FilterParser::ParseTagString() {
   return expression_.substr(pos, stop_pos);
 }
 
+absl::StatusOr<std::unique_ptr<query::TextPredicate>> FilterParser::ParseTextPredicate(
+    const std::string& field_name) {
+  auto index = index_schema_.GetIndex(field_name);
+  if (!index.ok() ||
+      index.value()->GetIndexerType() != indexes::IndexerType::kText) {
+    return absl::InvalidArgumentError(absl::StrCat(
+        "`", field_name, "` is not indexed as a numeric field"));
```
allenss-amazon: Text Field ?

src/commands/filter_parser.cc
```diff
@@ -217,6 +218,45 @@ absl::StatusOr<absl::string_view> FilterParser::ParseTagString() {
...
+    pos_++;
+  }
+  if (in_quotes && text_value.empty()) {
+    return absl::InvalidArgumentError("Empty quoted string");
+  }
+  if (text_value.empty()) {
+    return absl::InvalidArgumentError("Empty text predicate");
+  }
+  auto text_index = dynamic_cast<const indexes::Text*>(index.value().get());
+  if (!text_index) {
```
allenss-amazon: assert? Seems like this was tested above.

src/indexes/text.cc
```diff
@@ -96,9 +96,54 @@ std::unique_ptr<data_model::Index> Text::ToProto() const {
...
 }
 
+size_t Text::CalculateSize(const query::TextPredicate& predicate) const {
+  switch (predicate.GetOperation()) {
+    case query::TextPredicate::Operation::kExact: {
+      // TODO: Handle phrase matching.
+      auto word = predicate.GetTextString();
+      if (word.empty()) return 0;
+      auto iter = text_index_schema_->text_index_->prefix_.GetWordIterator(word);
+      auto target_posting = iter.GetTarget();
```
allenss-amazon, line 106: We need to get aligned on this point. My thinking is that the tree's include the counts from the Postings, so you shouldn't need to go to the postings objects. That way, your size of something with a wildcard is just as cheap.

src/query/predicate.h
```diff
@@ -127,6 +131,53 @@ class TagPredicate : public Predicate {
...
+class TextPredicate : public Predicate {
+ public:
+  enum class Operation {
+    kExact,      // Exact term / exact phrase match.
+    kPrefix,     // Prefix Wildcard match
+    kSuffix,     // kSuffix Wildcard match
+    kInfix,      // kInfix Wildcard match
+    kFuzzy,      // Fuzzy match
+  };
+
```
allenss-amazon, line 144: I'm wondering if having separate objects for each operation class makes more sense. They can all have TextPredicate as a  base class.

src/indexes/text.cc
KarthikSubbarao: Ideally, yes. However, the count returned here is the number of matched documents / keys. This depends on the type of search performed (by the relevant iterators). So, it feels like this should be an API exposed by the iterator instead

src/query/predicate.h
KarthikSubbarao: I will need to spend some time looking at how different search operations will work how best to construct the new Text Predicates. It would help to look into this more prior to switching fully to a specific Hierarchy

src/commands/filter_parser.cc
KarthikSubbarao:
```suggestion
        "`", field_name, "` is not indexed as a text field"));
```

src/commands/filter_parser.cc
```diff
@@ -217,6 +218,45 @@ absl::StatusOr<absl::string_view> FilterParser::ParseTagString() {
...
+    return absl::InvalidArgumentError(
+        absl::StrCat("Field '", field_name, "' is not a text field"));
+  }
```
KarthikSubbarao:
```suggestion
```

PR #296, opened 2025-08-08
Adding suffix 'field' to classes inheriting IndexBase

Aksha1812:
Proposal to update nomenclature : Classes inheriting IndexBase handle a single field/attribute. This is not apparent from the name of the classes inheriting IndexBase. e.g Tag, Text etc.. . proposing to add 'Field' suffix to these classes.

```
 IndexBase
├── Tag
├── Text  
├── Numeric
└── VectorBase
    ├── VectorFlat
    └── VectorHNSW

```

changes to 

```
IndexBase
├── TagField
├── TextField  
├── NumericField
└── VectorBaseField
    ├── VectorFlatField
    └── VectorHNSWField

```

### DIFF
--- a/src/indexes/text/posting.h
+++ b/src/indexes/text/posting.h
@@ -19,7 +19,7 @@ Conceptually, this object holds an ordered list of Keys and for each Key there i
 an ordered list of Positions. Each position is tagged with a bitmask of fields.
 
 A KeyIterator is provided to iterate over the keys within this object.
-A PositionIterator is provided to iterate over the positions of an individual Key.
+A PositionIterator is provided to iterate over the positions of an individual Key. 
 
 */
 


yairgott:
A couple of points:
1. The terminology that we're using is `Attribute` rather than `Field` to generalize both JSON and Hash. I believe that we're using this terminology in our public documentations (if not, we should).
2. I think that the it makes sense to also include the following changes:
2.1. IndexBase -> AttributeBase , note that it's already using the namespace `valkey_search::indexes`
2.2.  [IndexerType](https://github.com/valkey-io/valkey-search/blob/3d27b2cdb69b685c59dc9c692781001e0d1396ef/src/indexes/index_base.h#L28C12-L28C23) , kIndexerTypeByStr, etc -> `AttributeType`, ...

WDYT?

allenss-amazon:
Well, I'm 100% in favor of consistency. The Redis documents also use the term Attribute in many places. While I personally find field more usable than attribute, IDC which we end up with as long as it's consistently applied.

yairgott:
>  Keeping that in mind are you saying it should rather be called 'TagAttribute' or 'NumericAttribute' rather than using 'TagField' or 'NumericField' to be in line with terminology currently being used.

Correct.

> 2. Agreed will add these changes as well.

SG!

allenss-amazon:
Did we abandon this? If so, let's close it

PR #301, opened 2025-08-10
Add a configuration for resuming workers after fork

### LINKED ISSUE #277: writes are blocked for 60 sec during fork
It seems we suspend the writer threads for the entire duration of the fork?
```
<search> [warning], tid: 269520427946016, /home/ubuntu/valkey-search/src/valkey_search.cc:933: Fork child died notification received. Resuming writer worker thread pool returned message:  Suspend duration: 19.095796798s
```

Looking at the code at https://github.com/valkey-io/valkey-search/blob/1be979bbbddb74685430779caea20b40c3ab69ee/src/valkey_search.cc#L808
```
// Beside the thread which initiates the fork, no other threads are present
// in the forked child process. This could lead to full sync corruption as the
// fork system-call may occur in the middle of mutating the index. In addition,
// vector insertion may lead to high amount of dirty pages which increases the
// chances to OOM during full sync. Addressing these by temporary suspending the
// writer thread pool during full sync. The writer thread pool resumes once the
// child process dies or suspension time exceeds 60 seconds. Suspending the
// workers guarantees that no thread is mutating the index while the fork is
// happening. For more details see:
// https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_atfork.html
```

I don't think this is acceptable for some use cases. For index consistency a much shorter, suspect is required, by just resuming after fork.
Regarding the CoW, worth understanding if this really helps. I would suspect that in some cases it's the opposite.

This requires further clarification to figure best path forward

eranfrie:
Add a `max-worker-suspension` configuration for resuming workers after fork

The behavior of the new `max-worker-suspension` configuration is as follows:
- If max-worker-suspension > 0, the worker thread pool is resumed either after the fork is died or after `max-worker-suspension` seconds passed.
- If max-worker-suspension <= 0, the worker thread pool is resumed after fork is born.

Fixes #277

## Testing

Added a test - `OnForkChildBornCallback`: tests that when `max_worker_suspension_secs` is set to 0, the workers are resumed on the "Fork born" event.

Also, existing tests pass:
1. `FullSyncFork`.
2. `OnForkChildCallback` (renamed to `OnForkChildDiedCallback`).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -18,11 +18,8 @@ INTEG_RETRIES=1
 
 echo "Root directory: ${ROOT_DIR}"
 
-# Import our functions
-. ${ROOT_DIR}/scripts/common.rc
-
 function print_usage() {
-cat<<EOF
+    cat <<EOF
 Usage: build.sh [options...]
 
     --help | -h                       Print this help message and exit.
@@ -51,8 +48,7 @@ EOF
 }
 
 ## Parse command line arguments
-while [ $# -gt 0 ]
-do
+while [ $# -gt 0 ]; do
     arg=$1
     case $arg in
     --clean)
@@ -116,18 +112,18 @@ do
         shift || true
         echo "Using extra cmake arguments: ${CMAKE_EXTRA_ARGS}"
         ;;
-     --tsan)
+    --tsan)
         CMAKE_EXTRA_ARGS="${CMAKE_EXTRA_ARGS} -DSAN_BUILD=thread"
         SAN_BUILD="thread"
         shift || true
         echo "Using extra cmake arguments: ${CMAKE_EXTRA_ARGS}"
         ;;
-    --verbose|-v)
+    --verbose | -v)
         shift || true
         VERBOSE_ARGS="-v"
         echo "Verbose build: true"
         ;;
-    --help|-h)
+    --help | -h)
         print_usage
         exit 0
         ;;
@@ -137,6 +133,10 @@ do
         ;;
     esac
 done
+
+# Import our functions, needs to be done after parsing the command line arguments
+export SAN_BUILD
+. ${ROOT_DIR}/scripts/common.rc
 
 function configure() {
     printf "${BOLD_PINK}Running cmake...${RESET}\n"
@@ -205,7 +205,7 @@ function check_tool() {
     local tool_name=$1
     local message=$2
     printf "Checking for ${tool_name}..."
-    command -v ${tool_name} > /dev/null || \
+    command -v ${tool_name} >/dev/null ||
         (printf "${RED}failed${RESET}.\n${RED}ERROR${RESET} - could not locate tool '${tool_name}'. ${message}\n" && exit 1)
     printf "${GREEN}ok${RESET}\n"
 }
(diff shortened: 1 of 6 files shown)


src/valkey_search_options.cc
```diff
@@ -104,6 +104,20 @@ static auto writer_threads_count =
             })
         .Build();
 
+/// Register the "--max-worker-suspension" flag.
+/// Controls the resumption of the worker thread pool:
+///   - If max-worker-suspension > 0, resume the workers either when the
+///     fork is died or after max-worker-suspension secs passed.
+///   - If max-worker-suspension <= 0, resume the workers when the fork
+///     is borned.
+constexpr absl::string_view kMaxWorkerSuspension{"max-worker-suspension"};
```
eifrah-aws: For clarity, please add the units here: `max-worker-suspension-secs`

src/valkey_search_options.cc
```diff
@@ -104,6 +104,20 @@ static auto writer_threads_count =
...
+static auto max_worker_suspension =
+    config::NumberBuilder(kMaxWorkerSuspension,  // name
+                          60,                    // default value
+                          0,                     // min value
+                          3600)                  // max value
+        .Build();
```
eifrah-aws: No need for builder here:

```c++
static auto max_worker_suspension 
    = config::Number(kMaxWorkerSuspension,  // name
                          60,               // default value
                          0,                // min value
                          3600);            // max value
```

src/valkey_search_options.cc
```diff
@@ -173,6 +187,10 @@ vmsdk::config::Number& GetWriterThreadCount() {
   return dynamic_cast<vmsdk::config::Number&>(*writer_threads_count);
 }
 
+vmsdk::config::Number& GetMaxWorkerSuspension() {
+      return dynamic_cast<vmsdk::config::Number&>(*max_worker_suspension);
```
eifrah-aws: Once you apply my suggestion above, this becomes:

```c++
return max_worker_suspension;
```

src/valkey_search.cc
```diff
@@ -890,8 +884,9 @@ void ValkeySearch::OnServerCronCallback(ValkeyModuleCtx *ctx,
   // Resume worker thread pool if suspension time exceeds the max allowed
   // duration
   if (writer_thread_pool_suspend_watch_.has_value() &&
+      options::GetMaxWorkerSuspensionSecs().GetValue() > 0 &&
```
allenss-amazon, line 901: Isn't this redundant?

allenss-amazon (changes requested):
DCO Failure.

src/valkey_search.cc
eranfrie: `max-worker-suspension-config` can be <= 0. when it's, the workers are resumed when the fork is born, and not based on based on a timeout

allenss-amazon: Ah, ok, then please fix the DCO and we can merge it.

eranfrie: fixed

src/valkey_search_options.cc
```diff
@@ -104,6 +104,19 @@ static auto writer_threads_count =
             })
         .Build();
 
+/// Register the "--max-worker-suspension-secs" flag.
+/// Controls the resumption of the worker thread pool:
+///   - If max-worker-suspension-secs > 0, resume the workers either when the
+///     fork is died or after max-worker-suspension-secs seconds passed.
+///   - If max-worker-suspension-secs <= 0, resume the workers when the fork
```
eifrah-aws, line 111: nit: the configuration (min=0, max=3600, default=60) does not allow the value to drop below `0`

eifrah-aws (approved):

PR #302, opened 2025-08-10
Fix command-based ACL prefix check

### LINKED ISSUE #300: [BUG] ACL command allowing doesn't work
**Describe the bug**

ACL user with access string `on ~* resetchannels -@all +ft.search` can not execute `FT.SEARCH` command and will get error:

```
(error) The user doesn't have a permission to execute a command
```

Other commands `FT.CREATE, FT.DROPINDEX, FT.INFO` that call `AclPrefixCheck` should have the same issue.

**To Reproduce**

```
# create index and data
FT.CREATE idx_vector_hash_hnsw ON HASH PREFIX 1 hash: SCHEMA vec VECTOR HNSW 6 DIM 2 TYPE FLOAT32 DISTANCE_METRIC COSINE
hset hash:0 vec "\x00\x00\x00\x00\x00\x00\x00\x00"

# create ACL user
ACL SETUSER u1 on >password ~* resetchannels -@all +ft.search

# change user and test
AUTH u1 password
FT.SEARCH idx_vector_hash_hnsw "*=>[KNN 20 @vec $query_vec]" PARAMS 2 query_vec "\x00\x00\x00\x00\x00\x00\x00\x00"
```

**Expected behavior**

The ACL user with correct permission should be able to execute commands.

**Environment**

- Valkey Search Module Version: current latest commit `ca45eeb0a11c6731eed45f1a1dd8d70eb13c2403`
- Valkey Core Version: current Valkey core docker image `valkey/valkey:unstable`

**Investigation**

The required permissions are defined in capital:

https://github.com/valkey-io/valkey-search/blob/ca45eeb0a11c6731eed45f1a1dd8d70eb13c2403/src/commands/commands.h#L34-L39

However, `ACL GETUSER`'s response is lowercase:

https://github.com/valkey-io/valkey-search/blob/ca45eeb0a11c6731eed45f1a1dd8d70eb13c2403/src/acl.cc#L336-L338

Which causes mismatch in `GetKeysWithAllowedCommands() -> IsAllowedCommands()`

https://github.com/valkey-io/valkey-search/blob/ca45eeb0a11c6731eed45f1a1dd8d70eb13c2403/src/acl.cc#L198-L199

as an opposite `+@search` does work because it was defined in lowercase.

ichenhe:
Fix the issue by converting ACL commands from `ACL GETUSER` to uppercase before matching.

Fixed #300


allenss-amazon (approved):

ichenhe:
The ACL prefix check is removed from the search module in #387

PR #304, opened 2025-08-11
Add ACL integration tests

yanamolo:
Added ACL integration tests. The tests check the following:
- Check correctness for commands flags and categories (sanity check)
- Test with various users and verifies correct permissions
- Test for index with several prefixes and user that has access only to one of the prefixes. Test makes sure user has no access to that index.

This integration test class can be expanded in the future for more cases.

### DIFF
--- a/integration/test_valkey_search_acl.py
+++ b/integration/test_valkey_search_acl.py
@@ -1,0 +1,194 @@
+from typing import List
+from valkey.client import Valkey
+from valkey_search_test_case import ValkeySearchTestCaseBase
+from valkeytestframework.conftest import resource_port_tracker
+import struct
+import pytest
+from valkey.exceptions import ResponseError
+
+INDEX_NAME = "myIndex"
+
+
+class TestCommandsACLs(ValkeySearchTestCaseBase):
+
+    def _verify_user_permissions(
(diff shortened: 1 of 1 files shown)


integration/test_valkey_search_acl.py
```diff
@@ -0,0 +1,194 @@
...
+            ("ACL SETUSER user1 on >search_pass ~* &* -@all +@write", False, True),
+            ("ACL SETUSER user1 on >search_pass ~* &* -@all +@read", True, False),
+        ],
+    )
+    def test_acl_category_permissions(
+        self, user, should_access_read, should_access_write
+    ):
+        search_vector = struct.pack("<3f", *[1.0, 2.0, 3.0])
+        # List of search commands
+        valkey_search_commands = [
```
eifrah-aws, line 50: There a new class `Index` that you can use instead of this.

For example, to create an index:

```python
client: Valkey = self.server.get_new_client()
hnsw_index = Index(INDEX_NAME, [Vector("v", 3, type="HNSW", m=2, efc=1), Numeric("n")], prefixes=...)
hnsw_index.create(client)
```

yanamolo: Moved to use this class

PR #309, opened 2025-08-12
Fix crash due to Valkey Timeout before Cancellation timeout

allenss-amazon:
Timeout cancellation is a race between the Timeout operation of the main thread (initialied via the BlockClient call) and the self-cancellation of the background thread. Normally, the background thread wins but when the main thread wins, it would crash because the private data of the blocked client hadn't been set yet. This PR adds machinery to force the "main thread wins" path and eliminates the machinery to cancel the background thread.

Also, the mainthread timeout is setup to be slightly slower (10 mSec) than the background self-cancellation in order to allow the partial results machinery to operate.

### DIFF
--- a/src/utils/cancel.cc
+++ b/src/utils/cancel.cc
@@ -18,12 +18,16 @@ static vmsdk::config::Number PollFrequency(
     "timeout-poll-frequency", 100, 1, std::numeric_limits<long long>::max());
 static vmsdk::config::Boolean TestForceTimeout("test-force-timeout", false);
 
+static vmsdk::config::Boolean TestForcePausePoint("test-force-pause", false);
+
 static vmsdk::info_field::Integer Timeouts(
     "timeouts", "cancel-timeouts", vmsdk::info_field::IntegerBuilder().Dev());
 static vmsdk::info_field::Integer gRPCCancels(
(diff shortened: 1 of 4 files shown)


src/utils/cancel.cc
```diff
@@ -18,12 +18,16 @@ static vmsdk::config::Number PollFrequency(
     "timeout-poll-frequency", 100, 1, std::numeric_limits<long long>::max());
 static vmsdk::config::Boolean TestForceTimeout("test-force-timeout", false);
 
+static vmsdk::config::Boolean TestForcePausePoint("test-force-pause", false);
```
eifrah-aws, line 21: We should be able to hide this flag from users, similar to the `Dev()` in the INFO machinery. 

My proposal:

- Add new command-line-option only switch to allow "Debug" flags.
- Add new `Debug()` option to the `config` builders.
- Move this flag to be `Debug()`
- Flags that are marked as "Debug" can only be modified if the global flag: `search.enable-debug-configs` is enabled.
- The new global flag `search.enable-debug-configs` is a command line option **only** flag, which defaults to `no`.

This ensures that only the people that started the module can use `Debug` flags.

Since we don't have the mechanism for this, we can do this in a separate PR.

eifrah-aws (approved):

PR #310, opened 2025-08-12
Fanout operation base and cluster-wide primary and cluster fanout operation modes

boda26:
Fanout operation base and cluster-wide primary and cluster fanout operation modes


.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
     "clangd.path": "/usr/bin/clangd",
     "C_Cpp.codeAnalysis.runAutomatically": true,
     "editor.formatOnSave": false,
-    "C_Cpp.intelliSenseEngine": "default",
+    "C_Cpp.intelliSenseEngine": "disabled",
```
allenss-amazon: ?

.vscode/settings.json
```diff
@@ -17,8 +17,4 @@
         "--query-driver=**"
     ],
     "clangd.inactiveRegions.useBackgroundHighlight": true,
-    "clangd.enableCodeCompletion": true,
-    "files.associations": {
-        "optional": "cpp"
-    }
```
allenss-amazon: ?

integration/test_info_cluster.py
```diff
@@ -0,0 +1,60 @@
...
+from valkey.cluster import ValkeyCluster
+from valkey.client import Valkey
+from valkeytestframework.conftest import resource_port_tracker
+
+def _parse_info_kv_list(reply):
+    it = iter(reply)
+    out = {}
+    for k in it:
+        v = next(it, None)
+        out[k.decode() if isinstance(k, bytes) else k] = v
```
allenss-amazon: If the value was bytes, why not decode it here also, it clears up a lot of logic below.

integration/test_info_cluster.py
```diff
@@ -0,0 +1,60 @@
...
+        assert info is not None
+        mode = info.get("mode")
+        index_name = info.get("index_name")
+        assert (mode in (b"cluster", "cluster"))
+        assert (index_name in (b"index1", "index1"))
+
+        backfill_in_progress = int(info["backfill_in_progress"].decode() if isinstance(info["backfill_in_progress"], bytes) else info["backfill_in_progress"])
+        backfill_complete_percent_max = float(info["backfill_complete_percent_max"].decode() if isinstance(info["backfill_complete_percent_max"], bytes) else info["backfill_complete_percent_max"])
+        backfill_complete_percent_min = float(info["backfill_complete_percent_min"].decode() if isinstance(info["backfill_complete_percent_min"], bytes) else info["backfill_complete_percent_min"])
+        state = info["state"].decode() if isinstance(info["state"], bytes) else info["state"]
```
allenss-amazon: fixing parse_info_kv_list would make this cleaner.

integration/test_info_cluster.py
```diff
@@ -0,0 +1,60 @@
...
+            cluster.execute_command("HSET", f"doc:{i}", "price", str(10 + i))
+
+        assert node0.execute_command(
+            "FT.CREATE", "index1",
+            "ON", "HASH",
+            "PREFIX", "1", "doc:",
+            "SCHEMA", "price", "NUMERIC"
+        ) == b"OK"
+
+        time.sleep(0.2)
```
allenss-amazon: Sleeps are bad. They make tests flaky. What condition are you waiting for?

integration/test_info_primary.py
```diff
@@ -0,0 +1,68 @@
...
+            raw = node0.execute_command("FT.INFO", "index1", "PRIMARY")
+            info = _parse_info_kv_list(raw)
+            last = info
+
+            if "num_docs" in info and "num_records" in info:
+                num_docs = int(info["num_docs"].decode() if isinstance(info["num_docs"], bytes) else info["num_docs"])
+                num_records = int(info["num_records"].decode() if isinstance(info["num_records"], bytes) else info["num_records"])
+                if num_docs >= N and num_records >= N:
+                    break
+            time.sleep(0.2)
```
allenss-amazon: Looks like this should use the "wait_for" logic.

src/commands/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ target_link_libraries(commands PUBLIC schema_manager)
 target_link_libraries(commands PUBLIC valkey_search)
 target_link_libraries(commands PUBLIC vector_base)
 target_link_libraries(commands PUBLIC fanout)
+target_link_libraries(commands PUBLIC primary_info_fanout_operation)
+target_link_libraries(commands PUBLIC cluster_info_fanout_operation)
 target_link_libraries(commands PUBLIC response_generator)
```
allenss-amazon: Let's just add these executables to an existing library.

src/coordinator/CMakeLists.txt
```diff
@@ -1,111 +1,108 @@
...
+target_link_libraries(coordinator_base INTERFACE numeric)
+target_link_libraries(coordinator_base INTERFACE tag)
+target_link_libraries(coordinator_base INTERFACE search)
+target_link_libraries(coordinator_base INTERFACE predicate_header)
+target_link_libraries(coordinator_base INTERFACE schema_manager)
+target_link_libraries(coordinator_base INTERFACE rdb_serialization)
+target_link_libraries(coordinator_base INTERFACE utils)
+target_link_libraries(coordinator_base INTERFACE highwayhash)
+target_link_libraries(coordinator_base INTERFACE vector_base)
+target_link_libraries(coordinator_base INTERFACE response_generator)
```
allenss-amazon: Not sure what this is accomplishing. I'm in favor of reducing the complexity of our build files (what we have now is a huge mess), but let's not mix purposes in the same PR. 

To be clear, let's break the build stuff change into a separate PR -- it'll be easier to review.

src/index_schema.cc
```diff
@@ -373,6 +373,16 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
...
+    // switch (attribute_data_type_->ToProto()) {
+    //   case data_model::ATTRIBUTE_DATA_TYPE_HASH:
+    //     Metrics::GetStats().ingest_hash_keys++;
+    //     break;
+    //   case data_model::ATTRIBUTE_DATA_TYPE_JSON:
+    //     Metrics::GetStats().ingest_json_keys++;
+    //     break;
+    //   default:
+    //     break;
+    // }
```
allenss-amazon: ?

src/query/cluster_info_fanout_operation.cc
```diff
@@ -0,0 +1,209 @@
...
+                                              int argc) {
+  if (!errors_.empty()) {
+    std::string all_errors;
+    for (const auto& error : errors_) {
+      if (!all_errors.empty()) {
+        all_errors += "\n";
+      }
+      all_errors += error;
+    }
+    return ValkeyModule_ReplyWithError(ctx, all_errors.c_str());
```
allenss-amazon: This form of error reporting is really totally different from typical Valkey handling and probably isn't really that helpful to customers. From a customer perspective, what error responses make sense?

1) Unknown index name. (Which you caught earlier, so that can't happen here)
2) Inconsistent cluster state (But I was able to communicate to all cluster members)
3) Unable to communicate with other cluster members.

I think that's really all that it makes sense to report.

What I do think is valuable is to have a more fine-grained view of the failure in the log. In case the operator needs to have a better understanding of what's wrong. So, I'd recommend that when you get any of the (2) or (3) class of error that you put a message into the log with the target IP address (so we know who the culprit is). But be sure to use the VMSDK_LOG_EVERY_N_SEC to rate-limit the log messages.

In short, I think all you need is a couple of flags or an enum-like variable to track the error results. But at the point where you detect a (2) or (3) error you can spit out a log message.

allenss-amazon (changes requested):

integration/test_info_cluster.py
```diff
@@ -3,35 +3,42 @@
...
+    def is_indexing_complete(self, node, index_name):
+        try:
+            raw = node.execute_command("FT.INFO", index_name, "CLUSTER")
+            info = _parse_info_kv_list(raw)
+            if not info:
+                return False
+            backfill_in_progress = int(info.get("backfill_in_progress", 1))
+            state = info.get("state", "")
+            return backfill_in_progress == 0 and state == "ready"
+        except:
```
allenss-amazon: This kind of general exception isn't good practice as it can hide failures that aren't expected. If you really are expecting an exception, then you should the type and compare the detailed string to ensure that you're catching the exception your expecting. If you aren't expecting an exception, then don't use the try/except, let it crash -- you have bug somewhere.

integration/test_info_primary.py
```diff
@@ -3,23 +3,37 @@
...
+            raw = node.execute_command("FT.INFO", index_name, "PRIMARY")
+            info = _parse_info_kv_list(raw)
+            if not info:
+                return False
+            num_docs = int(info.get("num_docs", 0))
+            num_records = int(info.get("num_records", 0))
+            return num_docs >= N and num_records >= N
+        except:
+            return False
+
```
allenss-amazon, line 35: same story about exceptions.

src/query/cluster_info_fanout_operation.cc
```diff
@@ -177,17 +170,10 @@ void ClusterInfoFanoutOperation::InvokeRemoteRpc(
...
-    std::string all_errors;
-    for (const auto& error : errors_) {
-      if (!all_errors.empty()) {
-        all_errors += "\n";
-      }
-      all_errors += error;
-    }
-    return ValkeyModule_ReplyWithError(ctx, all_errors.c_str());
+  if (!index_name_error_nodes.empty() || !communication_error_nodes.empty() ||
+      !inconsisitent_state_error_nodes.empty()) {
```
allenss-amazon: inconsistent_state_error_nodes (spelling)

src/query/fanout_operation_base.h
```diff
@@ -80,21 +80,80 @@ class FanoutOperationBase {
...
+      communication_error_nodes.push_back(target);
+    }
+  }
+
+  virtual int GenerateErrorReply(ValkeyModuleCtx* ctx) {
+    absl::MutexLock lock(&mutex_);
+    std::string error_message;
+    // Log index name errors
+    if (!index_name_error_nodes.empty()) {
+      error_message += "Index name not found.\n";
```
allenss-amazon: =, not +=

src/query/fanout_operation_base.h
```diff
@@ -80,21 +80,80 @@ class FanoutOperationBase {
...
+      for (const FanoutSearchTarget& target : index_name_error_nodes) {
+        if (target.type == FanoutSearchTarget::Type::kLocal) {
+          VMSDK_LOG_EVERY_N_SEC(WARNING, ctx, 5) << "Index name error on local node";
+        } else {
+          VMSDK_LOG_EVERY_N_SEC(WARNING, ctx, 5)
+              << "Index name error on remote node: " << target.address;
```
allenss-amazon: For the log messages. It's good to have some kind of prefix that you can "grep" for. Applying that consistently to your error messages will make them easy to grab all at once. Here, I might use something like "FANOUTERROR:" as a prefix to the error messages.

src/query/fanout_operation_base.h
```diff
@@ -80,21 +80,80 @@ class FanoutOperationBase {
...
-        if (!resp.error().empty()) {
-          this->OnError(grpc::Status(grpc::StatusCode::INTERNAL, resp.error()),
+        if (resp.error_type() ==
+            coordinator::FanoutErrorType::INDEX_NAME_ERROR) {
+          this->OnError(grpc::Status(grpc::StatusCode::INTERNAL, ""),
+                        coordinator::FanoutErrorType::INDEX_NAME_ERROR, target);
+        } else if (resp.error_type() ==
+                   coordinator::FanoutErrorType::INCONSISTENT_STATE_ERROR) {
+          this->OnError(grpc::Status(grpc::StatusCode::INTERNAL, ""),
+                        coordinator::FanoutErrorType::INCONSISTENT_STATE_ERROR,
```
allenss-amazon: I'd recommend a specific value of the FanoutErrorType that indicates no error, typically something like OK. Then you can make a switch statement on that error type. This code is susceptible to long-term maintenance failures, because if in the future a new failure code is added AND the future programmer misses this place in the code, it will assume there wasn't an error. If you use a switch statement, then you can explicitly handle all of the cases you define now and provide a "default:" case that asserts. This is best practice for this kind of code as it future-proofs it.

allenss-amazon (changes requested):

src/commands/ft_info.cc
```diff
@@ -6,29 +6,103 @@
...
+}
+
+}  // namespace options
+
 absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
                        int argc) {
   if (argc < 2) {
-    return absl::InvalidArgumentError(vmsdk::WrongArity(kInfoCommand));
+    ValkeyModule_ReplyWithError(ctx, vmsdk::WrongArity(kInfoCommand).c_str());
+    return absl::OkStatus();
```
allenss-amazon, line 55: Why was this changed?

src/commands/ft_info.cc
```diff
@@ -37,8 +111,19 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
       PrefixACLPermissions(kInfoCmdPermissions, kInfoCommand);
   VMSDK_RETURN_IF_ERROR(
       AclPrefixCheck(ctx, permissions, index_schema->GetKeyPrefixes()));
-  index_schema->RespondWithInfo(ctx);
 
+  if (is_primary) {
+    auto op = new query::primary_info_fanout::PrimaryInfoFanoutOperation(
+        std::string(index_schema_name), timeout_ms);
```
allenss-amazon, line 117: A comment here would be good. the casual reader would look at this sequence and think that the newly allocated object was "leaked". Your comment should point out that the object blocks the current client and eventually takes care of it's own destruction.

PR #312, opened 2025-08-13
Update Vector HNSW output field format

hwware:


src/indexes/vector_base.cc
```diff
@@ -373,27 +373,17 @@ absl::StatusOr<bool> VectorBase::UpdateMetadata(
...
   ValkeyModule_ReplyWithSimpleString(ctx, "type");
   ValkeyModule_ReplyWithSimpleString(ctx, "VECTOR");
-  ValkeyModule_ReplyWithSimpleString(ctx, "index");
-
-  ValkeyModule_ReplyWithArray(ctx, VALKEYMODULE_POSTPONED_ARRAY_LEN);
+  int array_len = 2;
+  array_len += RespondWithInfoImpl(ctx);
   ValkeyModule_ReplyWithSimpleString(ctx, "capacity");
   ValkeyModule_ReplyWithLongLong(ctx, GetCapacity());
-  ValkeyModule_ReplyWithSimpleString(ctx, "dimensions");
```
yairgott, line 381: what is the reason for extracting the dimensions and the distance_metric from VectorBase and adding it to each of the vector types?

hwware: The only reason is to align with the returned information of FT.INFO from RedisSearch, so that the clients will not be broken when they migrate from RedisSearch to ValkeySearch.   The following is part of the returned information from RedisSearch:

image

hwware:
Update the test case and make CI happy now.

yairgott (approved):

PR #320, opened 2025-08-15
End to end test for JSON backfill replication. Plus enhances to pytest environment.

allenss-amazon:
1. Implement test for JSON Backfill bug. (CME test is marked as XFail until it gets fixed)
2. Enhance integration/indexes.py to work for JSON
3. Implement ability to force cross-shard search connections to only hit replicas.


src/commands/ft_search.cc
```diff
@@ -242,6 +242,8 @@ void Free([[maybe_unused]] ValkeyModuleCtx *ctx, void *privdata) {
 
 }  // namespace async
 
+vmsdk::config::Boolean ForceReplicasOnly("test-force-replicas-only", false);
```
eifrah-aws, line 246: question: does the `test-` prefix suggests that this is a "debug" param?

eifrah-aws (approved):

PR #321, opened 2025-08-16
Update clang tidy/format github job and apply formatting

BCathcart:
I updated `clang_tidy_format.yml` so that hopefully now it will always generate the diff and find the files to check the formatting of. In past PRs the main branch wasn't fetched, so the diff failed and the job concluded that no .cc or .h files were modified.

e.g.
```
Run echo "Finding modified or added .cc and .h files..."
Finding modified or added .cc and .h files...
fatal: ambiguous argument 'main': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
No .cc or .h files modified.
```

I also applied clang-format to all .h/.cc files under:
- src/**
- testing/**
- vmsdk/src/**
- vmsdk/testing/**

I ran the `ci/check_changes.sh` script locally against all the files afterwards and it seemed happy, but I think it's just giving false positives. Looks like clang-tidy actually spits out tons of errors. I'm disabling the clang-tidy check in the github job until we get the existing code to a tidied state.

### DIFF
--- a/.github/workflows/clang_tidy_format.yml
+++ b/.github/workflows/clang_tidy_format.yml
@@ -25,7 +25,32 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2  # Relevant for push case
+
+      - name: Determine base branch
+        id: determine-base
+        run: |
+          set -euo pipefail
+          # If it's a pull request, compare against the base branch.
+          # If it's a direct push, compare against the previous commit.
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            git fetch origin ${{ github.event.pull_request.base.ref }}
+            BASE_BRANCH="origin/${{ github.event.pull_request.base.ref }}"
+            echo "PR detected. Comparing with base branch: $BASE_BRANCH"
+          else
+            BASE_BRANCH=$(git rev-parse HEAD~1)
+            echo "Push detected. Comparing with previous commit: $BASE_BRANCH"
+          fi
+          
+          # Verify the base branch/commit exists
+          if ! git rev-parse --verify "$BASE_BRANCH" >/dev/null 2>&1; then
+            echo "Error: Base branch/commit '$BASE_BRANCH' not found"
+            exit 1
+          fi
+          
+          echo "BASE_BRANCH=$BASE_BRANCH" >> $GITHUB_ENV
 
       # Build the Docker image from the Dockerfile in your repo
       - name: Build Docker Image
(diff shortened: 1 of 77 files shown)


allenss-amazon:
Thanks for the work here. Let's discuss a path forward to get this fixed in the next OSS meeting. I'll put it on the calendar.

eifrah-aws:
LGTM.

I would also extend the `build.sh` script to run `clang-format` checks AND FIX any issues found as part of the build itself.

In addition, can we also run formatter on the python files? (e.g. `black`)

An example for bash function that does that:

```bash
function random_string() {
  randstr=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 8 | head -n 1)
  echo ${randstr}
}

function fix_fromatting() {
  if [[ "${IS_LOCAL_BUILD}" == "no" ]]; then
    # Avoid this when running as part of a CI process
    return
  fi
  pushd ${ROOT_DIR}
  # Building locally
  local files=$(git diff --name-only | grep -E "(\.cc|\.h)")

  # Add files that are committed but not pushed yet
  local already_committed_files=$(git diff HEAD origin/main --name-only)

  if [ -z "${files}" ] && [ -z "${already_committed_files}" ]; then
    printf "${BOLD_PINK}No files to format.${file}\n${RESET}"
    return
  fi

  files="${files} ${already_committed_files}"
  files=$(echo $files | tr ' ' '\n' | sort | uniq)

  local format_tmp_dir="/tmp/$(random_string)/"
  mkdir -p ${format_tmp_dir}
  for file in ${files}; do
    file_name=$(basename ${file})
    printf "${BOLD_PINK}Checking format for${RESET}: ${file}"
    clang-format ${file} >${format_tmp_dir}/${file_name}
    local result=$(diff -u ${file} ${format_tmp_dir}/${file_name})
    if [ ! -z "${result}" ]; then
      clang-format -i ${file}
      printf " ... ${YELLOW}fixed${RESET}\n"
    else
      printf " ... ${GREEN}ok${RESET}\n"
    fi
  done
  rm -fr ${format_tmp_dir}
  popd
}
```

BCathcart:
I agree we should at a minimum have a build target (`build.sh --format`) that applies clang-format to all .h/.cc files. We can consider failing a release build if the formatting hasn't been applied to help developers catch it before submitting a PR. I'm mostly on vacation right now but I'll do this as a follow-up when I'm back and have some time. If someone wants to pick it up in the meantime though, feel free.

Formatting python code is probably a separate conversation. Afaik we haven't been enforcing any formatting practices so it's pretty likely to obfuscate the git history for the majority of the code. Maybe we should start with standardizing it in [valkey-test-framework](https://github.com/valkey-io/valkey-test-framework) and then work towards applying it across all valkey repos in one effort.

Last thing I want to mention is that if anyone isn't developing in the container environment, you may be surprised by some formatting issues in your PR if your clang-format version does not match `Ubuntu clang-format version 18.1.3 (1ubuntu1)`. I originally applied clang-format v20 and there were a few lines of difference across the code base.

eifrah-aws:
>Last thing I want to mention is that if anyone isn't developing in the container environment, you may be surprised by some formatting issues in your PR if your clang-format version does not match Ubuntu clang-format version 18.1.3 (1ubuntu1). I originally applied clang-format v20 and there were a few lines of difference across the code base.

I would have guess that it is a matter of defaults?

Our current `.clang-format` file contains:

```yaml
BasedOnStyle: Google
IndentWidth: 2
ColumnLimit: 80
```

The way I usually do this is by creating a complete configuration file by running: `clang-format --dump-config > .clang-format` and then I edit the settings I want to change. This usually proves to be a more resilient approach when switching `clang-format` versions.

PR #322, opened 2025-08-17
fix: tracks only stored interned string

proost:
@allenss-amazon 

I found missed part. there are interned string, but not stored in the store. but deallocation track all interned strings. so we should track only shared interned strings which are stored in the store.


allenss-amazon (changes requested):
I think this is the wrong strategy. We should less clever, not more clever. The underlying problem is that you're not including the owned, but not shared strings in stringpool. Why not? I think if you just wrap the constructor and destructor of the InternString object with another IsolatedMemoryScope, then you'll solve the problem. It will means that InternString object which aren't entered in the pool get counted in the overall stat. But I don't see that as a problem.

PR #324, opened 2025-08-18
Add Text Properties to Schema Proto Reconstruction

BCathcart:
Includes the text-specific properties stored in the `TextIndexSchema` instance when reconstructing the schema proto in the `ToProto()` call.

The diff includes unrelated formatting changes, so it'll be easier to review once #321 is merged into the full text branch.


src/indexes/text/text_index.cc
```diff
@@ -12,26 +12,63 @@
...
+namespace {
+
+bool IsWhitespace(unsigned char c) {
+  return std::isspace(c) || std::iscntrl(c);
 }
 
-sb_stemmer* TextIndexSchema::GetStemmer() const {
-  if (!stemmer_) {
-    stemmer_ = sb_stemmer_new(GetLanguageString().c_str(), "UTF_8");
+PunctuationBitmap BuildPunctuationBitmap(const std::string& punctuation) {
```
Aksha1812, line 21: Is all this part of this PR, or is it just not rebased correctly?

src/index_schema.cc
```diff
@@ -859,20 +855,25 @@ std::unique_ptr<data_model::IndexSchema> IndexSchema::ToProto() const {
...
-                                                   stop_words_.end());
-
   auto stats = index_schema_proto->mutable_stats();
   stats->set_documents_count(stats_.document_cnt);
   std::transform(
       attributes_.begin(), attributes_.end(),
       google::protobuf::RepeatedPtrFieldBackInserter(
           index_schema_proto->mutable_attributes()),
       [](const auto &attribute) { return *attribute.second.ToProto(); });
+
```
Aksha1812, line 910: This is the main change right? Does this only set the proto fields if there is a text field present? I guess language proto field existed before full text parsing changes by Nivesh and might be in use or might be used later for Tag fields as well if I am not wrong.

BCathcart: All these were added by Ram in his lexical scanner commit which ended up overlapping with this PR: https://github.com/valkey-io/valkey-search/commit/fc70f307801c4b79f12a2a8ba166d339f2c11efe.

When I was rebasing, I kept the flow I had originally envisioned. It's a bit cleaner in my mind but I'm open to opinions.

src/indexes/text/text_index.cc
BCathcart: Yes I refactored `text_index.h/.cc`

src/index_schema.cc
```diff
@@ -859,20 +855,25 @@ std::unique_ptr<data_model::IndexSchema> IndexSchema::ToProto() const {
...
+  // Text-specific properties
+  if (text_index_schema_) {
```
VoletiRam: We have chance of losing these index level configs if there is no text field in ft.create command. Assume we create an index without text field, then we don't store language in any class members with the current change and skip the initialization of text index schema (because no text field). During proto reconstruction,  we lose language field . Language is used by other indexes. Though it would look optimized to move text index fields to text index schema, the current structure was designed considering future support of ft.alter where we update existing index to add new field types. 

Btw Allen didn't like conditional treatment for optimization of few bits of memory, please check here: https://github.com/valkey-io/valkey-search/pull/342#discussion_r2294970268

src/indexes/text/text_index.cc
VoletiRam: This refactor makes sense.BuildPunctuationBitmap wrote in header by accident. Making it inline unecessary as it would increase compilation time, though not significant.

VoletiRam (commented):
I will explain the existing architecture of lexer
```
 LEXICAL SCANNER CONFIGURATION STORAGE ARCHITECTURE

┌─────────────────────────────────────────────────────────────────┐
│                          IndexSchema                            │
│  ┌─────────────────────────────────────────────────────────────┐│
│  │                  IndexSchema Protobuf                      ││
│  │  • language: LANGUAGE_ENGLISH                               ││
│  │  • punctuation: ".,!?;:"                                    ││
│  │  • stop_words: ["the", "and", "or"]                        ││
│  │  • with_offsets: true                                       ││
│  └─────────────────────────────────────────────────────────────┘│
│                              │                                  │
│                              ▼ (direct copy during Create())   │
│  ┌─────────────────────────────────────────────────────────────┐│
│  │                  IndexSchema Class                         ││
│  │  • name_: string                       (index metadata)     ││
│  │  • db_num_: uint32_t                   (index metadata)     ││
│  │  • language_: LANGUAGE_ENGLISH         (RAW storage)        ││
│  │  • punctuation_: ".,!?;:"              (RAW storage)        ││
│  │  • stop_words_: ["the", "and", "or"]   (RAW storage)        ││
│  │  • with_offsets_: true                 (RAW storage)        ││
│  │  • text_index_schema_: shared_ptr      (created lazily)     ││
│  └─────────────────────────────────────────────────────────────┘│
│                              │                                  │
│                              ▼ (CreateTextIndexSchema())        │
│  ┌─────────────────────────────────────────────────────────────┐│
│  │                  TextIndexSchema                            ││
│  │  • num_text_fields_: uint8_t          (field allocator)     ││
│  │  • text_index_: shared<TextIndex>     (shared RadixTrees)   ││
│  │  • language_: LANGUAGE_ENGLISH        (STORED - for stemmer)││
│  │  • with_offsets_: true                (STORED - for postings)││
│  │  • punct_bitmap_: std::bitset<256>    (OPTIMIZED - shared)  ││
│  │  • stop_words_set_: flat_hash_set     (OPTIMIZED - shared)  ││
│  │  • stemmer_: sb_stemmer* (mutable)    (LAZY - shared)       ││
│  └─────────────────────────────────────────────────────────────┘│
└─────────────────────────────────────────────────────────────────┘
                               │
                               ▼ (shared reference)
┌─────────────────────────────────────────────────────────────────┐
│                         Text Class                              │
│  ┌─────────────────────────────────────────────────────────────┐│
│  │                   TextIndex Protobuf                       ││
│  │  • with_suffix_trie: true                                   ││
│  │  • no_stem: false                                           ││
│  │  • min_stem_size: 3                                         ││
│  └─────────────────────────────────────────────────────────────┘│
│                              │                                  │
│                              ▼ (extract during Text() ctor)    │
│  ┌─────────────────────────────────────────────────────────────┐│
│  │                   Text Class Members                       ││
│  │  • text_field_number_: size_t         (from TextIndexSchema) ││
│  │  • with_suffix_trie_: true            (from TextIndex proto)││
│  │  • no_stem_: false                    (from TextIndex proto)││
│  │  • min_stem_size_: int32_t            (from TextIndex proto)││
│  │  • text_index_schema_: shared_ptr     (shared reference)    ││
│  │  • untracked_keys_: InternedStringSet (field instance data) ││
│  └─────────────────────────────────────────────────────────────┘│
└─────────────────────────────────────────────────────────────────┘
                               │
                               ▼ (parameter passing - NO STORAGE)
┌─────────────────────────────────────────────────────────────────┐
│                         Lexer Struct                            │
│  ┌─────────────────────────────────────────────────────────────┐│
│  │                   STATELESS PROCESSOR                       ││
│  │  • NO CONFIG STORAGE - PURE FUNCTIONS                       ││
│  │                                                             ││
│  │  • Tokenize(text, punct_bitmap, stemmer, stemming_enabled,  ││
│  │             min_stem_size, stop_words_set) const            ││
│  │                                                             ││
│  │  Parameters come from:                                      ││
│  │  • punct_bitmap ← TextIndexSchema::punct_bitmap_            ││
│  │  • stemmer ← TextIndexSchema::GetStemmer()                  ││
│  │  • stemming_enabled ← Text::!no_stem_                       ││
│  │  • min_stem_size ← Text::min_stem_size_                     ││
│  │  • stop_words_set ← TextIndexSchema::stop_words_set_        ││
│  └─────────────────────────────────────────────────────────────┘│
└─────────────────────────────────────────────────────────────────┘
```

The current design avoids conditional checks for text index schema.  During RDB save, we reconstruct proto by copying from index schema class members. We can discuss if you think your refactoring has other advantages.

src/index_schema.cc
BCathcart: Thanks for the background. Ft.alter would be an interesting use case that I didn't consider. 

> What should be different is that these should be marked "optional" in the protobuf and if they aren't present, then we should initialize them to a default value -- so that the remainder of the code doesn't treat it as a special case. Even if there aren't any text fields for this Schema, I don't care about burning a few extra bytes in the protobuf if it simplifes my life later.
>
> We might think about making the protobuf have a struct named something like SchemaExtensionV2 that has the extra fields and then the entire struct would be marked optional. Seems a bit cleaner from a documentation perspective and avoids creating failure modes because one of these fields was specified but the others weren't.

This would be for backward compatibility with an older RDB right? Did we do something around this?

src/index_schema.cc
BCathcart: Reverted back to what is was after insight from Ram

src/index_schema.proto
```diff
@@ -30,6 +30,7 @@ message IndexSchema {
   Stats stats = 7;
   optional uint32 db_num = 8;
 
+  // TODO: make these optional so that we can give them default values if loading an old version RDB
```
VoletiRam, line 33: We can apply default values when these fields are not there during RDB load.

BCathcart: Yeah the idea here is to be able to differentiate when a full text field didn't exist at the time of the snapshot. Right now on load, I believe these fields are given some default value based on their type if they don't exist in the RDB. For example, stop words would be an empty array and we want to differentiate if the user explicitly specified no stop words with their FT.CREATE or the concept of stop words simply didn't exist at the time. If it's the former we would want to leave the empty array but if it's the latter we probably want to give the index the default stop words list.

PR #326, opened 2025-08-18
Address KeyIterator Gaps

Aksha1812:
* KeyIterator can now take field specifications to filter on certain fields
* Updated PhraseIterator to search on only field specified in the query
* Use WITHOFFSETS as input for save_positions.
* Add integration test for full text search per field when word is present in different fields

### DIFF
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -115,11 +115,16 @@ size_t Text::CalculateSize(const query::TextPredicate& predicate) const {
 std::unique_ptr<Text::EntriesFetcher> Text::Search(
     const query::TextPredicate& predicate,
     bool negate) const {
+  // TODO : Create FieldMask with multiple bits set when multiple fields queried in Text Predicate
+  // Create FieldMask with the specific text field bit set
+  auto field_mask = text::FieldMask::Create(text_index_schema_->num_text_fields_);
+  field_mask->SetField(text_field_number_);
+  
   auto fetcher = std::make_unique<EntriesFetcher>(
     CalculateSize(predicate),
     text_index_schema_->text_index_,
     negate ? &untracked_keys_ : nullptr,
-    text_field_number_);
+    std::move(field_mask));
   fetcher->operation_ = predicate.GetOperation();
   // Currently, we support a single word (exact term) match.
   fetcher->data_ = predicate.GetTextString();
@@ -136,7 +141,7 @@ std::unique_ptr<EntriesFetcherIteratorBase> Text::EntriesFetcher::Begin() {
       std::vector<WordIterator> iterVec = {iter};
       bool slop = 0;
       bool in_order = true;
-      auto itr = std::make_unique<text::PhraseIterator>(iterVec, slop, in_order, untracked_keys_, text_field_number_);
+      auto itr = std::make_unique<text::PhraseIterator>(iterVec, slop, in_order, untracked_keys_, std::move(field_mask_));
       itr->Next();
       return itr;
     }
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -116,9 +116,13 @@ std::unique_ptr<Text::EntriesFetcher> Text::Search(
     const query::TextPredicate& predicate,
     bool negate) const {
   // TODO : Create FieldMask with multiple bits set when multiple fields queried in Text Predicate
-  // Create FieldMask with the specific text field bit set
+  // Create FieldMask with default all fields set, then clear and set specific field
   auto field_mask = text::FieldMask::Create(text_index_schema_->num_text_fields_);
-  field_mask->SetField(text_field_number_);
+  field_mask->SetAllFields();  // Initialize with default case all fields set (0xffffffffffff)
+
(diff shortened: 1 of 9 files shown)


src/indexes/text/phrase.cc
```diff
@@ -5,16 +5,18 @@ namespace valkey_search::indexes::text {
 PhraseIterator::PhraseIterator(const std::vector<WordIterator>& words,
                               size_t slop,
                               bool in_order,
-                              const InternedStringSet* untracked_keys)
+                              const InternedStringSet* untracked_keys,
+                              std::optional<size_t> text_field_number)
```
allenss-amazon: When would this be omitted -- or is this temporary?

src/indexes/text.cc
```diff
@@ -115,10 +115,19 @@ size_t Text::CalculateSize(const query::TextPredicate& predicate) const {
 std::unique_ptr<Text::EntriesFetcher> Text::Search(
     const query::TextPredicate& predicate,
     bool negate) const {
+  // Create FieldMask with default all fields set,
+  auto field_mask = text::FieldMask::Create(text_index_schema_->num_text_fields_);
+  field_mask->SetAllFields();  // Initialize with default case all fields set (0xffffffffffff)
+
+  // TODO : Modify this to handle fields based on input query passed down from the text predicate. when no field specified default field mask used
+  field_mask->ClearAllFields(); // Clear all fields to 0
```
allenss-amazon: Set all then clear all?

src/indexes/text.h
```diff
@@ -93,6 +95,7 @@ class Text : public IndexBase {
     query::TextPredicate::Operation operation_;
     absl::string_view data_;
     bool no_field_{false};
+    std::unique_ptr<text::FieldMask> field_mask_;
```
allenss-amazon: FieldMask is cheap, (it's a word or two). Just copy it....

src/indexes/text/phrase.cc
```diff
@@ -5,16 +5,18 @@ namespace valkey_search::indexes::text {
 PhraseIterator::PhraseIterator(const std::vector<WordIterator>& words,
                               size_t slop,
                               bool in_order,
+                              std::unique_ptr<FieldMask> field_mask,
```
allenss-amazon: pass by value.

src/indexes/text/phrase.h
```diff
@@ -76,15 +77,15 @@ class PhraseIterator : public indexes::EntriesFetcherIteratorBase {
   std::vector<WordIterator> words_;
   std::shared_ptr<Postings> target_posting_;
   Postings::KeyIterator key_iter_;
-  uint32_t current_idx_ = 0;
   bool begin_ = true;  // Used to track if we are at the beginning of the iterator.
   size_t slop_;
   bool in_order_;
   const InternedStringSet* untracked_keys_;
   InternedStringPtr current_key_;
+  std::unique_ptr<FieldMask> field_mask_;
```
allenss-amazon: by value.

src/indexes/text/posting.cc
```diff
@@ -42,11 +28,12 @@ class FieldMaskImpl : public FieldMask {
   void SetField(size_t field_index) override;
   void ClearField(size_t field_index) override;
   bool HasField(size_t field_index) const override;
+  bool HasFields(const FieldMask& field_mask) const override;
```
allenss-amazon: A base class with virtual methods is overkill here. For V1, we'll be limiting # of TEXT FIELD to 64, so a simple struct with a size_t that's passed around by value is enough now.

allenss-amazon: Imagine a future update with say, 4 64-bit words (256 field limit). It's waaaaay cheaper to copy around 4 words than it is to malloc/free an object.

allenss-amazon: Oddly, anding and oring of 4 words sequentially is cheaper than a virtual function call (in a modern super-scalar processor)

allenss-amazon: But I wouldn't spend time on fixing this now.

src/indexes/text.cc
Aksha1812: this is temporary, default value is -1 so basically all fields.  ideally we would only clear and set again if there is a field specified in the query. currently query doesn't support searching on all fields aka without field specified

src/indexes/text.h
```diff
@@ -76,10 +76,12 @@ class Text : public IndexBase {
    public:
     EntriesFetcher(size_t size,
                 const std::shared_ptr<text::TextIndex>& text_index,
-                const InternedStringSet* untracked_keys = nullptr)
+                const InternedStringSet* untracked_keys = nullptr,
```
allenss-amazon, line 79: Let's shy away from optional parameters.

allenss-amazon:
Merged
Sent from my iPhone, please excuse all typos and awful spelling corrects.

On Aug 20, 2025, at 4:18 PM, AkshaThakkar1812 ***@***.***> wrote:

﻿

@Aksha1812<https://github.com/Aksha1812> requested your review on: #326<https://github.com/valkey-io/valkey-search/pull/326> Address KeyIterator Gaps.

PR #327, opened 2025-08-19
Rough delete logic for the radix tree

BCathcart:
Throwing out some very rough delete logic for the radix tree to unblock people if needed. I think the delete works but there is a bug in the word iterator causing `DeleteStructuralCases` to fail, but I ran out of time to look into it. I'll be back online by the weekend.

### DIFF
--- a/src/indexes/text/radix_tree.h
+++ b/src/indexes/text/radix_tree.h
@@ -123,10 +123,8 @@ struct RadixTree {
   // Create a Path iterator at a specific starting prefix
   PathIterator GetPathIterator(absl::string_view prefix) const;
 
-  // Debug function to print tree structure
+  // Prints tree structure
   void DebugPrintTree(const std::string& label = "") const;
-
- private:
 
  private:
   /*
@@ -188,7 +186,6 @@ struct RadixTree {
 
   Node root_;
 
-  // Debug helper function
   void DebugPrintNode(const Node* node, const std::string& path, int depth) const {
     std::string indent(depth * 2, ' ');
     std::cout << indent << "Node[" << path << "]";
@@ -500,7 +497,7 @@ void RadixTree<Target, reverse>::Mutate(
             [&](std::pair<BytePath, std::unique_ptr<Node>>& child) {
               // The target node is a compressed node. If its parent is a
               // compressed node, we can modify the parent to have an edge
-              // directly to target node's child.
+              // directly to its child and remove the target node.
               const auto& parent = node_path.back();
               if (std::holds_alternative<
                       std::pair<BytePath, std::unique_ptr<Node>>>(
--- a/src/indexes/text/radix_tree.h
+++ b/src/indexes/text/radix_tree.h
@@ -60,6 +60,7 @@ into the codebase efficiently enough to be deployed in production code.
 #include <memory>
 #include <optional>
 #include <span>
+#include <sstream>
 #include <tuple>
 #include <variant>
 
@@ -122,6 +123,9 @@ struct RadixTree {
 
   // Create a Path iterator at a specific starting prefix
   PathIterator GetPathIterator(absl::string_view prefix) const;
+
+  // Returns tree structure as vector of strings
+  std::vector<std::string> DebugGetTreeStrings() const;
 
   // Prints tree structure
   void DebugPrintTree(const std::string& label = "") const;
@@ -208,9 +212,11 @@ struct RadixTree {
    */
   void TrimBranchFromTree(absl::string_view word, std::deque<Node*>& node_path);
 
-  void DebugPrintNode(const Node* node, const std::string& path, int depth,
-                      bool is_last = true,
-                      const std::string& prefix = "") const;
+  // Recursive helper for DebugGetTreeStrings
+  std::vector<std::string> DebugGetTreeString(const Node* node,
+                                              const std::string& path,
+                                              int depth, bool is_last,
+                                              const std::string& prefix) const;
 
  public:
   //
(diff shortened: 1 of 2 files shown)


src/indexes/text/radix_tree.h
```diff
@@ -339,16 +372,148 @@ void RadixTree<Target, reverse>::Mutate(
...
   }
 
   std::optional<Target> new_target = mutate(n->target);
 
   if (new_target) {
     n->target = new_target;
   } else {
-    // TODO: delete word from the tree
-    assert(false);
+    // Delete the word from the tree
```
Aksha1812, line 393: so brennan's implementation handles only deleting a word from a tree, In RemoveRecord API when we delete a key, I would have to find the all the words in that keys value using the word iterator, for each word i would perform Mutate with lambda function -- if it is existing I would have to go to its Target (Posting) and remove that key using RemoveKey() from Posting. if Posting empty we delete word (return std::nullopt; as per brennan's implementation) otherwise just return modified Target. Does that sound right @allenss-amazon @BCathcart

testing/radix_test.cc
```diff
@@ -123,21 +123,115 @@ TEST_F(RadixTreeTest, EdgeCases) {
...
+    EXPECT_FALSE(existing.has_value());
+    return std::nullopt;
+  });
 
   // TODO: Enable when assertion handling is clarified
   EXPECT_DEATH(prefix_tree_->Mutate("", [](auto) { return TestTarget(1); }),
                ".*");
 }
 
+TEST_F(RadixTreeTest, DeleteStructuralCases) {
```
Aksha1812: this test is failing, investigating

Aksha1812: Seems like issue not with the RadixTree itself but maybe the WordIterator . I can RadixTree has deleted the word correctly, but WordIterator return old deleted word

```
=== Tree Structure - After adding testing/test ===
Node[] BRANCH(2)
  Node[c] COMPRESSED[a]
    Node[ca] BRANCH(2)
      Node[can] TARGET LEAF
      Node[cat] TARGET LEAF
  Node[t] COMPRESSED[est]
    Node[test] TARGET COMPRESSED[ing]
      Node[testing] TARGET LEAF
=== End Structure ===


=== Tree Structure - After deleting 'test' ===
Node[] BRANCH(2)
  Node[c] COMPRESSED[a]
    Node[ca] BRANCH(2)
      Node[can] TARGET LEAF
      Node[cat] TARGET LEAF
  Node[t] COMPRESSED[esting]
    Node[testing] TARGET LEAF
=== End Structure ===

/workplace/thaakb/valkey-search/valkey-search/testing/radix_test.cc:183: Failure
Expected equality of these values:
  test_iter.GetWord()
    Which is: "test"
  "testing"
    Which is: 0x683ec2
```

allenss-amazon: Iterators don't need to be valid across mutations. Is this just a test error?

Aksha1812: Added https://github.com/valkey-io/valkey-search/pull/340. this should fix the error

Aksha1812:
@BCathcart  regarding the radix tree related build warning AI suggested. 

## Analysis of the GCC Warning

The warning you're seeing is a __false positive__ from GCC 14's static analysis. Here's what's happening:

__Root Cause:__

- The warning `'void operator delete(void*, std::size_t)' called on unallocated object '<anonymous>'` is triggered in your `RadixTree::Mutate` function
- Specifically at lines 290 and 330-332 where you use `std::make_unique<Node>`
- GCC 14's aggressive inlining and static analysis is getting confused by the complex destructor chain involving `std::variant` with `std::unique_ptr` members

__Why it's a False Positive:__

- Your code is correctly using `std::make_unique` which properly allocates objects on the heap
- The warning is caused by GCC's static analyzer tracing through the deeply nested template instantiations of `std::variant` destructors and losing track of object lifetimes
- The complex inlining of variant operations makes the compiler think an object allocated in one context is being deleted in another

### Option 1: Suppress the Warning via Compiler Flags (Recommended)

Add the `-Wno-free-nonheap-object` flag to suppress this specific warning globally or for specific files.

You need to modify the `cmake/Modules/valkey_search.cmake` file in the `valkey_search_target_update_compile_flags` function Line 142:

__Add this line:__

```cmake
target_compile_options(${TARGET} PRIVATE -Wno-free-nonheap-object)
```


### Option 2: Use Pragma Directives for Local Suppression

Add `#pragma GCC diagnostic` directives around the problematic code sections to suppress the warning locally while keeping it enabled elsewhere.

__Around lines 290-291:__

```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wfree-nonheap-object"
                std::unique_ptr<Node> new_leaf = std::make_unique<Node>(
                    Node{0, std::nullopt, std::monostate{}});
#pragma GCC diagnostic pop
```

__Around lines 330-332:__

```cpp
#pragma GCC diagnostic push  
#pragma GCC diagnostic ignored "-Wfree-nonheap-object"
                std::unique_ptr<Node> new_node = std::make_unique<Node>(Node{
                    0, std::nullopt,
                    std::pair{path.substr(match), std::move(child.second)}});
#pragma GCC diagnostic pop
```

testing/radix_test.cc
BCathcart: Thanks for the fix!

Aksha1812:
> I don't think anyone has reviewed it closely. I was hoping to clean it up today but I'm not sure if I'll get to it.

sure nw. I don't think it will take too much time to clean it up, but if @allenss-amazon has some time to review, we can close two prs and have delete functionality ready.

src/indexes/text/radix_tree.h
```diff
@@ -339,16 +375,149 @@ void RadixTree<Target, reverse>::Mutate(
...
+    n->target = std::nullopt;
+    node_path.pop_back();
+
+    // Reconstruct the tree
+    std::visit(overloaded{
+                   [&](const std::monostate&) {
+                     // Leaf case - We need to trim the branch from the tree
```
Aksha1812: nit. could these cases be divided into smaller helper functions. some of the functions in this file are becoming very long. might be difficult for other people to make changes later.

src/indexes/text/radix_tree.h
Aksha1812, line 304: nit. do we need any input validation/ sanity check on word input here, i guess it should already have been validated by the tokenizer. but maybe a simple check like checking for null or if it doesn't exist in the trie. or maybe we can add in the tests below

Aksha1812:
> @BCathcart regarding the radix tree related build warning (vibe)
> 
> ## Analysis of the GCC Warning
> The warning you're seeing is a **false positive** from GCC 14's static analysis. Here's what's happening:
> 
> **Root Cause:**
> 
> * The warning `'void operator delete(void*, std::size_t)' called on unallocated object '<anonymous>'` is triggered in your `RadixTree::Mutate` function
> * Specifically at lines 290 and 330-332 where you use `std::make_unique<Node>`
> * GCC 14's aggressive inlining and static analysis is getting confused by the complex destructor chain involving `std::variant` with `std::unique_ptr` members
> 
> **Why it's a False Positive:**
> 
> * Your code is correctly using `std::make_unique` which properly allocates objects on the heap
> * The warning is caused by GCC's static analyzer tracing through the deeply nested template instantiations of `std::variant` destructors and losing track of object lifetimes
> * The complex inlining of variant operations makes the compiler think an object allocated in one context is being deleted in another
> 
> ### Option 1: Suppress the Warning via Compiler Flags (Recommended)
> Add the `-Wno-free-nonheap-object` flag to suppress this specific warning globally or for specific files.
> 
> You need to modify the `cmake/Modules/valkey_search.cmake` file in the `valkey_search_target_update_compile_flags` function Line 142:
> 
> **Add this line:**
> 
> ```cmake
> target_compile_options(${TARGET} PRIVATE -Wno-free-nonheap-object)
> ```
> 
> ### Option 2: Use Pragma Directives for Local Suppression
> Add `#pragma GCC diagnostic` directives around the problematic code sections to suppress the warning locally while keeping it enabled elsewhere.
> 
> **Around lines 290-291:**
> 
> ```c++
> #pragma GCC diagnostic push
> #pragma GCC diagnostic ignored "-Wfree-nonheap-object"
>                 std::unique_ptr<Node> new_leaf = std::make_unique<Node>(
>                     Node{0, std::nullopt, std::monostate{}});
> #pragma GCC diagnostic pop
> ```
> 
> **Around lines 330-332:**
> 
> ```c++
> #pragma GCC diagnostic push  
> #pragma GCC diagnostic ignored "-Wfree-nonheap-object"
>                 std::unique_ptr<Node> new_node = std::make_unique<Node>(Node{
>                     0, std::nullopt,
>                     std::pair{path.substr(match), std::move(child.second)}});
> #pragma GCC diagnostic pop
> ```

was any of this useful?

Aksha1812 (approved):

src/indexes/text/radix_tree.h
BCathcart: I have `CHECK(!word.empty()) << "Can't mutate the target at an empty word";` just below to check that it's a non-empty string. If it doesn't exist in the trie, then that means we're adding it for the first time which is expected.

src/indexes/text/radix_tree.h
BCathcart: I moved all of the tree cleanup code outside Mutate() which hopefully helps a bit

BCathcart:
> was any of this useful?

It was useful in reminding me to go back and look at the compiler warning 🙂. I had an unnecessary temporary stack allocation when creating the nodes. That AI insight is probably right that it was a false positive warning, but it still did indicate I had done something unusual.

Aksha1812 (approved):

PR #335, opened 2025-08-20
Fix ASAN test failure

allenss-amazon:
Apparently, memory accounting doesn't work with ASAN. no surprise.

### DIFF
--- a/integration/test_info.py
+++ b/integration/test_info.py
@@ -87,5 +87,5 @@ class TestVSSBasic(ValkeySearchTestCaseBase):
         for field in bytes_fields:
             assert field in info_data
             bytes_value = info_data[field]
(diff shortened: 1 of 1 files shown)


integration/test_info.py
```diff
@@ -91,4 +87,4 @@ def test_info_fields_present(self):
         for field in bytes_fields:
             assert field in info_data
             bytes_value = info_data[field]
-            assert isinstance(bytes_value, str) and bytes_value.endswith("iB")
+            assert (isinstance(bytes_value, str) and bytes_value.endswith("iB")) or bytes_value == 0
```
yairgott: I think that it would be clearer if the assert expression would `or` on the os environment variable "SAN_BUID" not equals to "no"

eifrah-aws (approved):

PR #336, opened 2025-08-20
UpdateRecord, RemoveRecord Implementation for Text

Aksha1812:
Draft for UpdateRecord, RemoveRecord Implementation for Text


src/indexes/text.cc
```diff
@@ -66,12 +66,51 @@ absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
 
 absl::StatusOr<bool> Text::RemoveRecord(const InternedStringPtr& key,
                                         DeletionType deletion_type) {
-  throw std::runtime_error("Text::RemoveRecord not implemented");
+  
+  // TODO: Key Tracking
+  
+  if (deletion_type == DeletionType::kRecord){
+    // Collect all words to check (to avoid modifying while iterating)
+    std::vector<std::string> words_to_check;
```
Aksha1812: Is there a way to get the words in the current key's value directly from here so that we don't have to check through every word if that key exists. I am adding implementation for `absl::flat_hash_map<Key, TextIndex> by_key_;` that would help deleting by key. but what would happen to the original main index where we would have to go to words see if key exists in the target(posting), that is little expensive. would we even delete the key there. or during search we would just check if the key is in untracked_keys list or maybe not present here `absl::flat_hash_map<Key, TextIndex> by_key_;` and we would not return results for it. @allenss-amazon

src/indexes/text.cc
```diff
@@ -66,12 +66,51 @@ absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
...
+    auto iter = text_index_schema_->text_index_->prefix_.GetWordIterator("");
+    while (!iter.Done()) {
+      words_to_check.emplace_back(iter.GetWord());
+      iter.Next();
+    }
```
allenss-amazon: Why copy this into a temporary data structure, why not just use this iterator int he loop below?

Aksha1812: updated

src/indexes/text.cc
```diff
@@ -66,12 +66,47 @@ absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
 
 absl::StatusOr<bool> Text::RemoveRecord(const InternedStringPtr& key,
                                         DeletionType deletion_type) {
-  throw std::runtime_error("Text::RemoveRecord not implemented");
+  
+  // TODO: Key Tracking
+  
+  if (deletion_type == DeletionType::kRecord){
+    auto iter = text_index_schema_->text_index_->prefix_.GetWordIterator("");
```
BCathcart, line 73: This is getting every word in the schema-level tree where we really just want every word in the key-specific tree. We need to implement the per-key tree which can be done as part of this PR if you have the bandwidth. Otherwise I'll probably get to it next week and we can wait to merge this PR on top of it.

PR #337, opened 2025-08-20
Adding retry to fanout base class

boda26:
Adding retry for ft.info, which introduces retry mechanism for fanout when encountering inconsistent state/network error within timeout limit (default 5s). This would provide a foundation for FT.CREATE/DROPINDEX consistency check. Provided several sets of integration tests for base class, primary and cluster mode operations.


src/commands/ft_info.cc
```diff
@@ -114,11 +114,11 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
 
   if (is_primary) {
     auto op = new query::primary_info_fanout::PrimaryInfoFanoutOperation(
-        std::string(index_schema_name), timeout_ms);
+        std::string(index_schema_name), timeout_ms, true);
```
allenss-amazon: Why is a parameter needed? For those use-cases where no retry is needed, just have the "RetryRequested" virtual function return a false for that case. wouldn't that simplify the base class?

src/commands/ft_info.cc
```diff
@@ -114,11 +114,11 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
     op->StartOperation(ctx);
   } else if (is_cluster) {
     auto op = new query::cluster_info_fanout::ClusterInfoFanoutOperation(
-        std::string(index_schema_name), timeout_ms);
+        std::string(index_schema_name), timeout_ms, true);
```
allenss-amazon: same

src/query/cluster_info_fanout_operation.cc
```diff
@@ -25,7 +27,7 @@ ClusterInfoFanoutOperation::ClusterInfoFanoutOperation(std::string index_name,
       backfill_in_progress_(false) {}
 
 unsigned ClusterInfoFanoutOperation::GetTimeoutMs() const {
-  return timeout_ms_.value_or(5000);
```
allenss-amazon, line 28: Ugly constant. What is the usage for this? Why not require a timeout value and avoid the conditional test.

src/query/cluster_info_fanout_operation.cc
```diff
@@ -104,10 +106,10 @@ void ClusterInfoFanoutOperation::OnResponse(
 
 coordinator::InfoIndexPartitionResponse
 ClusterInfoFanoutOperation::GetLocalResponse(
-    ValkeyModuleCtx* ctx, const coordinator::InfoIndexPartitionRequest& request,
+    int db_id, const coordinator::InfoIndexPartitionRequest& request,
```
allenss-amazon: Why add db_id ?

src/query/cluster_info_fanout_operation.cc
```diff
@@ -192,4 +194,18 @@ int ClusterInfoFanoutOperation::GenerateReply(ValkeyModuleCtx* ctx,
   return VALKEYMODULE_OK;
 }
 
+void ClusterInfoFanoutOperation::ResetForRetry() {
+  FanoutOperationBase::ResetForRetry();
```
allenss-amazon: Have the base class do this. that way derived classes don't have to -- and the programmer won't forget. Why should it be the responsibility of the derived class to tell the base class to do something in all cases?

src/query/fanout_operation_base.h
```diff
@@ -67,12 +65,82 @@ class FanoutOperationBase {
     delete static_cast<FanoutOperationBase*>(privdata);
   }
 
+  void StartFanoutRound(ValkeyModuleCtx* ctx) {
+    auto targets = GetTargets(ctx);
+    {
+      absl::MutexLock lock(&mutex_);
+      outstanding_ = targets.size();
+    }
```
allenss-amazon: Why is the lock needed here? What cases would you have another thread looking at this data structure?

src/query/fanout_operation_base.h
```diff
@@ -67,12 +65,82 @@ class FanoutOperationBase {
...
+          case coordinator::FanoutErrorType::INCONSISTENT_STATE_ERROR:
+            this->OnError(
+                grpc::Status(grpc::StatusCode::INTERNAL, ""),
+                coordinator::FanoutErrorType::INCONSISTENT_STATE_ERROR, target);
+            break;
+          case coordinator::FanoutErrorType::COMMUNICATION_ERROR:
+            this->OnError(grpc::Status(grpc::StatusCode::INTERNAL, ""),
+                          coordinator::FanoutErrorType::COMMUNICATION_ERROR,
+                          target);
+            break;
```
allenss-amazon: How can there be a communication error for a local callback?

src/query/fanout_operation_base.h
```diff
@@ -149,66 +225,31 @@ class FanoutOperationBase {
...
-          },
-          timeout_ms);
+  unsigned GetCurrentTimeoutMs() const {
+    using namespace std::chrono;
+    auto elapsed_ms =
+        duration_cast<milliseconds>(steady_clock::now() - start_tp_).count();
+    if (elapsed_ms >= GetTimeoutMs()) {
+      return 0;
     }
+    return static_cast<unsigned>(GetTimeoutMs() - elapsed_ms);
```
allenss-amazon: this appears misnamed. It looks like it's returning the remaining time before a timeout. Do you actually care about that value? Actually, you don't if you look at the usage all it really cares about is whether I'm timed-out or not. I'd recommend that this funciton be renamed and return a boolean (IsOperationTimedOut or something like that)

src/coordinator/server.cc
```diff
@@ -182,6 +182,59 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
   return reactor;
 }
 
+coordinator::InfoIndexPartitionResponse Service::GenerateInfoResponse(
+    const std::string& index_name) {
```
allenss-amazon: IMO, you should pass in the Request Message itself.

src/coordinator/server.cc
```diff
@@ -182,6 +182,59 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
+  coordinator::InfoIndexPartitionResponse response;
+  auto status_or_schema =
+      SchemaManager::Instance().GetIndexSchema(/*db=*/0, index_name);
```
allenss-amazon: db should come from the request message.

src/coordinator/server.cc
```diff
@@ -182,6 +182,59 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
+  if (global_metadata->type_namespace_map().contains(
+          kSchemaManagerMetadataTypeName)) {
+    const auto& entry_map = global_metadata->type_namespace_map().at(
+        kSchemaManagerMetadataTypeName);
+    if (entry_map.entries().contains(index_name)) {
+      const auto& entry = entry_map.entries().at(index_name);
+      fingerprint = entry.fingerprint();
+      version = entry.version();
+    }
+  }
```
allenss-amazon: This is concerning, when you can have a message that responds with index information yet, cannot have the fingerprint and version? Without those you don't know if the message is valid. If it's truely the case that sometimes these two fields are present, then I think it's an error case and should be responded accordingly. But if it shoudn't ever happen (which I believe is the case), then assert (CHECK) is the correct response.

src/coordinator/server.cc
```diff
@@ -190,60 +243,14 @@ grpc::ServerUnaryReactor* Service::InfoIndexPartition(
...
-    auto status_or_schema =
-        SchemaManager::Instance().GetIndexSchema(/*db=*/0, idx);
-    if (!status_or_schema.ok()) {
-      response->set_exists(false);
-      response->set_index_name(idx);
-      response->set_error(status_or_schema.status().ToString());
-      response->set_error_type(coordinator::FanoutErrorType::INDEX_NAME_ERROR);
-      reactor->Finish(ToGrpcStatus(status_or_schema.status()));
-      return;
+    *response = Service::GenerateInfoResponse(index_name);
```
allenss-amazon: I don't like that the base class "knows" that a message has an "Exists" method. IMO it would be better to return two values, an error code and a message. The base class shouldn't know anything about the contents of the message.

src/query/cluster_info_fanout_operation.cc
```diff
@@ -13,12 +13,14 @@
 namespace valkey_search::query::cluster_info_fanout {
 
 ClusterInfoFanoutOperation::ClusterInfoFanoutOperation(std::string index_name,
-                                                       unsigned timeout_ms)
+                                                       unsigned timeout_ms,
+                                                       uint32_t db_num)
```
allenss-amazon: let's reorder these so that db_num is next to index_name. Also, here you're passing in index_name as a string value. Conventionally, you'd pass in a reference and let the object decide if it wants to make a copy or not.

src/query/primary_info_fanout_operation.h
```diff
@@ -25,7 +25,8 @@ class PrimaryInfoFanoutOperation : public fanout::FanoutOperationBase<
                                        coordinator::InfoIndexPartitionResponse,
                                        fanout::FanoutTargetMode::kPrimary> {
  public:
-  PrimaryInfoFanoutOperation(std::string index_name, unsigned timeout_ms);
+  PrimaryInfoFanoutOperation(std::string index_name, unsigned timeout_ms,
+                             uint32_t db_num = 0);
```
allenss-amazon: Let's avoid default values, there are some cases that justify usage of default values. But this is the opposite, there's no meaningful default value for db_num, just like there's no meaningful default value for index_name. db_num is an extension of index_name, it makes no sense to have a default for one and not the other....

src/commands/ft_info.cc
```diff
@@ -111,14 +111,15 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
       PrefixACLPermissions(kInfoCmdPermissions, kInfoCommand);
   VMSDK_RETURN_IF_ERROR(
       AclPrefixCheck(ctx, permissions, index_schema->GetKeyPrefixes()));
-
+  
+  // operation(db_num, index_name, timeout)
   if (is_primary) {
     auto op = new query::primary_info_fanout::PrimaryInfoFanoutOperation(
-        std::string(index_schema_name), timeout_ms);
+        0, std::string(index_schema_name), timeout_ms);
```
allenss-amazon: This should be fetched from the ctx using the right module function.

src/commands/ft_info.cc
```diff
@@ -111,14 +111,15 @@ absl::Status FTInfoCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
...
     op->StartOperation(ctx);
   } else if (is_cluster) {
     auto op = new query::cluster_info_fanout::ClusterInfoFanoutOperation(
-        std::string(index_schema_name), timeout_ms);
+        0, std::string(index_schema_name), timeout_ms);
```
allenss-amazon: same

integration/test_info_cluster.py
```diff
@@ -9,31 +9,36 @@
...
-            return False
-        backfill_in_progress = int(info.get("backfill_in_progress", 1))
-        state = info.get("state", "")
-        return backfill_in_progress == 0 and state == "ready"
+        try:
+            raw = node.execute_command("FT.INFO", index_name, "CLUSTER")
+            info = _parse_info_kv_list(raw)
+            if not info:
+                return False
+            backfill_in_progress = int(info.get("backfill_in_progress", 1))
```
allenss-amazon: the "get" call means that you're not sure whether "backfill_in_progress" is present in the response or not. is that correct? shouldn't it be in every response?

allenss-amazon: Same for state.

integration/test_info_cluster.py
```diff
@@ -9,31 +9,36 @@
...
+            state = info.get("state", "")
+            return backfill_in_progress == 0 and state == "ready"
+        except Exception as e:
+            if "Communication error between nodes found" in str(e):
```
allenss-amazon: If this is expected, then there is pytest.raises which you can use to wrap the communication call. If this isn't expected you shouldn't just swallow it.

integration/test_info_cluster.py
```diff
@@ -9,31 +9,36 @@
...
             "PREFIX", "1", "doc:",
             "SCHEMA", "price", "NUMERIC"
         ) == b"OK"
-
-        N = 5
-        for i in range(N):
-            cluster.execute_command("HSET", f"doc:{i}", "price", str(10 + i))
         
-        waiters.wait_for_equal(lambda: self.is_indexing_complete(node0, index_name), True, timeout=5)
+        waiters.wait_for_equal(lambda: self.is_indexing_complete(node0, index_name), True, timeout=10)
```
allenss-amazon: Typically, we use the default timeout value for "wait_for_equal", unless it's a special case. If this is a special case there should be comment explaining why.

allenss-amazon: Also, I think there's a "wait_for_true".

integration/test_info_primary.py
```diff
@@ -25,13 +25,18 @@ def verify_error_response(client, cmd, expected_err_reply):
...
-        num_docs = int(info.get("num_docs", 0))
-        num_records = int(info.get("num_records", 0))
-        return num_docs >= N and num_records >= N
+        try:
+            raw = node.execute_command("FT.INFO", index_name, "PRIMARY")
+            info = _parse_info_kv_list(raw)
+            if not info:
+                return False
+            num_docs = int(info.get("num_docs", 0))
+            num_records = int(info.get("num_records", 0))
```
allenss-amazon: Same comments as above.

integration/test_info_primary.py
```diff
@@ -22,15 +22,32 @@ def verify_error_response(client, cmd, expected_err_reply):
...
     def is_indexing_complete(self, node, index_name, N):
         raw = node.execute_command("FT.INFO", index_name, "PRIMARY")
         info = _parse_info_kv_list(raw)
         if not info:
             return False
-        num_docs = int(info.get("num_docs", 0))
-        num_records = int(info.get("num_records", 0))
+        num_docs = int(info["num_docs"])
+        num_records = int(info["num_records"])
         return num_docs >= N and num_records >= N
```
allenss-amazon, line 44: >= that's odd, why wouldn't you know the exact correct value?

src/coordinator/client.cc
```diff
@@ -37,7 +37,7 @@ namespace valkey_search::coordinator {
 constexpr absl::string_view kRetryPolicy =
     "{\"methodConfig\" : [{"
     "   \"name\" : [{\"service\": \"valkey_search.coordinator.Coordinator\"}],"
-    "   \"waitForReady\": false,"
+    "   \"waitForReady\": true,"
```
allenss-amazon, line 42: Did you discover that this did something useful??

src/coordinator/client.cc
```diff
@@ -191,6 +192,7 @@ void ClientImpl::InfoIndexPartition(
   auto args = std::make_unique<InfoIndexPartitionArgs>();
   args->context.set_deadline(
       absl::ToChronoTime(absl::Now() + absl::Milliseconds(timeout_ms)));
+  args->context.set_wait_for_ready(true);
```
allenss-amazon: Is this redundant with the policy document above?

src/coordinator/coordinator.proto
```diff
@@ -139,8 +139,9 @@ message GetGlobalMetadataResponse {
 }
 
 message InfoIndexPartitionRequest {
-  string index_name = 1;
-  uint64 timeout_ms = 2;
+  uint32 db_num = 1;
+  string index_name = 2;
+  uint64 timeout_ms = 3;
```
allenss-amazon: Why is the timeout relevant to the recipients of this message?

src/query/fanout_operation_base.h
```diff
@@ -138,77 +195,43 @@ class FanoutOperationBase {
...
-            }
-            this->RpcDone();
-          },
-          timeout_ms);
+  unsigned IsOperationTimedOut() const {
+    using namespace std::chrono;
+    auto elapsed_ms =
+        duration_cast<milliseconds>(steady_clock::now() - start_tp_).count();
+    if (elapsed_ms >= GetTimeoutMs()) {
+      return 0;
```
allenss-amazon: Just as a style point, I usually find it clearer to compute a deadline time at the start of the operation rather than remembering the start time, then I can compare the current clock directly to the deadline time without doing a bunch of conversions, etc.

allenss-amazon (commented):
You should add some INFO field counters to provide visibility into the various retry and error paths. Testing these values will prove that you've exercised the right paths in the code for the future testing requirements.

With that in mind, what's needed to complete testing of the base class is:
1. Test at least one guaranteed retry.
2. Test at least one timeout due to a shard-down situation

For each of the derived classes, you want to ensure at least two tests:
1. That in the happy path, each returned result has a correct value. This may actually take several tests depending on the actual information being collected.
2. Ensure that at least one retry happens and you still get the right answers for each variable returned.

Now some of derived class tests can be dual purpose, i.e., they are also exercising the retry logic in the base class, so you might not need separate tests for that part of the base class's logic.

For example, for the backfill computation, we want a test where we know that the backfill is complete (which you can determine by waiting on each shard to be complete). But we also want tests where some (or all) of the nodes have NOT completed backfill. This requires the ability to suspend the backfill operation or you will have a test that's flaky due to races. There may already be configurable values to enable the "freezing" of backfill, but if not then you'll have to add that -- and I can help with that.

src/query/cluster_info_fanout_operation.cc
```diff
@@ -192,4 +146,18 @@ int ClusterInfoFanoutOperation::GenerateReply(ValkeyModuleCtx* ctx,
...
+  version_.reset();
+  backfill_complete_percent_max_ = 0.0f;
+  backfill_complete_percent_min_ = 0.0f;
+  backfill_in_progress_ = false;
+}
+
+bool ClusterInfoFanoutOperation::ShouldRetry() {
+  return !inconsistent_state_error_nodes.empty() ||
+         !communication_error_nodes.empty();
+}
```
KarthikSubbarao, line 163: Can we clarify both cases in a comment above? 1 is inconsistence state 2 is a network error

src/query/fanout_operation_base.h
```diff
@@ -9,17 +9,35 @@
...
+#include "vmsdk/src/debug.h"
 #include "vmsdk/src/log.h"
+#include "vmsdk/src/module_config.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
 namespace valkey_search::query::fanout {
 
+constexpr absl::string_view kFanoutForceRemoteFailConfig{
+    "fanout-force-remote-fail"};
+
```
KarthikSubbarao: Rather than a new module configuration just for testing, it would be better to include this testing capability in a command like FT._DEBUG which is intended for internal developer use. A new option in this command would be able to provide similar capabilities 

Configs are visible to users explicitly and would be confusing to find this

integration/test_fanout_base.py
```diff
@@ -0,0 +1,74 @@
...
+from valkey.cluster import ValkeyCluster
+from valkey.client import Valkey
+from valkeytestframework.conftest import resource_port_tracker
+from valkeytestframework.util import waiters
+from test_info_primary import is_index_on_all_nodes
+from valkey.exceptions import ResponseError
+import pytest
+import re
+
+class TestFanoutBase(ValkeySearchClusterTestCaseDebugMode):
```
KarthikSubbarao, line 12: Can we test a real scenario where a cluster is down? Right now, we mock "failure" scenarios using internal states using FT._DEBUG?

It will be good to have coverage on essential scenarios such as remote server down (we can shut down the remote server with a simple update to the test framework)

src/query/fanout_operation_base.h
allenss-amazon: Recently added is "CONTROLLED_BOOLEAN" which addresses this.

src/commands/ft_info.cc
```diff
@@ -26,9 +26,9 @@
 namespace valkey_search {
 
 constexpr absl::string_view kFTInfoTimeoutMsConfig{"ft-info-timeout-ms"};
-constexpr uint32_t kDefaultFTInfoTimeoutMs{5000};
+constexpr uint32_t kDefaultFTInfoTimeoutMs{1000};
 constexpr uint32_t kMinimumFTInfoTimeoutMs{100};
-constexpr uint32_t kMaximumFTInfoTimeoutMs{300000};  // 5 minutes max
+constexpr uint32_t kMaximumFTInfoTimeoutMs{10000};  // 10s max
```
KarthikSubbarao: Question: Why do we need a different timeout config for FT.INFO? Can we not use the existing timeout config?

src/query/fanout_operation_base.h
```diff
@@ -31,14 +50,11 @@ class FanoutOperationBase {
...
     blocked_client_->MeasureTimeStart();
-
-    auto targets = GetTargets(ctx);
-    outstanding_ = targets.size();
-    unsigned timeout_ms = GetTimeoutMs();
-    for (const auto& target : targets) {
-      auto req = GenerateRequest(target, timeout_ms);
-      IssueRpc(ctx, target, req, timeout_ms);
-    }
+    deadline_tp_ = std::chrono::steady_clock::now() +
```
KarthikSubbarao, line 44: Note: We might want to think about the deadline value used by the coordinator / gRPC.

If the deadline is equal to the timeout used for the overall operation, then we will never retry incase gRPC if the server is down / unresponsive as we will get the deadline exceeded error. If we want to retry and not wait for the entire command timeout duration, gRPC will need a smaller deadline duration

https://grpc.io/docs/guides/deadlines/

src/commands/ft_info.cc
allenss-amazon: Because of different expectations and usages. Timeout for search is used to truncate an inherently unpredictable process. For ft.info the underlying process isn't unpredictable but the timeout is used to address crashed or partitioned nodes.

allenss-amazon: Because of different expectations and usages. Timeout for search is used to truncate an inherently unpredictable process. For ft.info the underlying process isn't unpredictable but the timeout is used to address crashed or partitioned nodes.

KarthikSubbarao: I understand we might want to use a different value for the timeout config for FT.SEARCH vs FT.INFO due to the underlying operations and that in FT.SEARCH we are planning to support partial results after retries. But it felt excessive to have one timeout for each operation. Maybe we can have one for search/aggregate related operations and one timeout for other operations related?

>Timeout for search is used to truncate an inherently unpredictable process

Will FT.CREATE re-use the same timeout as FT.INFO then (because it is not used to truncate responses)?

integration/test_fanout_base.py
```diff
@@ -0,0 +1,102 @@
...
+                return False
+            except ConnectionError:
+                return True
+    
+        waiters.wait_for_true(lambda: is_node_down(node1), timeout=5)
+        
+        with pytest.raises(ResponseError) as excinfo:
+            node0.execute_command("FT.INFO", index_name, "CLUSTER")
+        
+        assert "Communication error between nodes found" or "Request timed out" in str(excinfo.value)
```
KarthikSubbarao: Can you add a new line at the end of the file?

integration/test_fanout_base.py
```diff
@@ -0,0 +1,102 @@
...
+        assert node0.execute_command(
+            "FT.CREATE", index_name,
+            "ON", "HASH",
+            "PREFIX", "1", "doc:",
+            "SCHEMA", "price", "NUMERIC"
+        ) == b"OK"
+
+        waiters.wait_for_true(lambda: is_index_on_all_nodes(self, index_name))
+
+        assert node0.execute_command("FT._DEBUG CONTROLLED_VARIABLE SET ForceTimeout yes") == b"OK"
```
KarthikSubbarao: Adding a new `ForceRetry` will better than trying to use the `ForceTimeout` to trigger retries.

src/commands/ft_info.cc
```diff
@@ -28,7 +28,7 @@ namespace valkey_search {
 constexpr absl::string_view kFTInfoTimeoutMsConfig{"ft-info-timeout-ms"};
 constexpr uint32_t kDefaultFTInfoTimeoutMs{5000};
 constexpr uint32_t kMinimumFTInfoTimeoutMs{100};
-constexpr uint32_t kMaximumFTInfoTimeoutMs{300000};  // 5 minutes max
+constexpr uint32_t kMaximumFTInfoTimeoutMs{10000};  // 10s max
```
KarthikSubbarao: We can leave this as it is

src/coordinator/client.cc
```diff
@@ -60,6 +62,8 @@ static constexpr int kCoordinatorQueryDefaultTimeout{120};
 static constexpr int kCoordinatorQueryMinTimeout{1};
 static constexpr int kCoordinatorQueryMaxTimeout{3600};
 
+static constexpr int kInfoRpcTimeout{2500};
```
KarthikSubbarao: Can you add a comment that we will consider adding a config here later on?

src/query/cluster_info_fanout_operation.cc
```diff
@@ -192,4 +146,18 @@ int ClusterInfoFanoutOperation::GenerateReply(ValkeyModuleCtx* ctx,
   return VALKEYMODULE_OK;
 }
 
+void ClusterInfoFanoutOperation::ResetForRetry() {
```
KarthikSubbarao, line 149: Question: Should `state_` be reset?

src/query/cluster_info_fanout_operation.h
```diff
@@ -51,12 +51,17 @@ class ClusterInfoFanoutOperation : public fanout::FanoutOperationBase<
   int GenerateReply(ValkeyModuleCtx* ctx, ValkeyModuleString** argv,
                     int argc) override;
 
+  void ResetForRetry() override;
+
+  bool ShouldRetry() override;
+
```
KarthikSubbarao, line 59: Let's add a short description for these new functions

src/query/fanout_operation_base.h
```diff
@@ -67,29 +73,89 @@ class FanoutOperationBase {
...
+        auto [status, resp] = this->GetLocalResponse(request, target);
+        if (status.ok()) {
+          this->OnResponse(resp, target);
+        } else {
+          this->OnError(status, resp.error_type(), target);
+        }
+        this->RpcDone();
+      });
+    } else {
+      // force the remote to fail 10 times for testing only
```
KarthikSubbarao:
```suggestion
      // Testing code path. This forces 10 retries for testing only when X is enabled
```

integration/test_fanout_base.py
```diff
@@ -0,0 +1,104 @@
...
+        assert node0.execute_command("FT._DEBUG CONTROLLED_VARIABLE SET ForceRetry yes") == b"OK"
+
+        node0.execute_command("FT.INFO", index_name, "PRIMARY")
+        info_search_str = str(node0.execute_command("INFO SEARCH"))
+
+        pattern = r'search_fanout_retry_count:(\d+)'
+        match = re.search(pattern, info_search_str)
+        if not match:
+            assert False, f"search_fanout_retry_count not found in INFO SEARCH results!"
+        retry_count = int(match.group(1))
```
allenss-amazon: This looks like code that's likely to get repeated in the future. We have an existing FT_INFO parser, we should look at enhancing it for future usage.

integration/test_fanout_base.py
```diff
@@ -0,0 +1,104 @@
...
+        assert node0.execute_command(
+            "FT.CREATE", index_name,
+            "ON", "HASH",
+            "PREFIX", "1", "doc:",
+            "SCHEMA", "price", "NUMERIC"
+        ) == b"OK"
+
+        waiters.wait_for_true(lambda: is_index_on_all_nodes(self, index_name))
+
+        assert node0.execute_command("FT._DEBUG CONTROLLED_VARIABLE SET ForceRetry yes") == b"OK"
```
allenss-amazon: What is the functionality of the "ForceRetry" controlled variable? On it's face, it doesn't seem like it's creating a remote failure. Perhaps it should be renamed or is this testing something different than the comment?

integration/test_fanout_base.py
```diff
@@ -0,0 +1,104 @@
...
+                return False
+            except ConnectionError:
+                return True
+    
+        waiters.wait_for_true(lambda: is_node_down(node1), timeout=5)
+        
+        with pytest.raises(ResponseError) as excinfo:
+            node0.execute_command("FT.INFO", index_name, "CLUSTER")
+        
+        assert "Request timed out" in str(excinfo.value)
```
allenss-amazon: I thought that this scenario should generate a "couldn't communicate with entire cluster" error, right?

integration/test_fanout_base.py
allenss-amazon: Maybe _parse_info_kv_list(raw) is all you need?

src/query/fanout_operation_base.h
```diff
@@ -9,17 +9,31 @@
...
 #include "vmsdk/src/blocked_client.h"
+#include "vmsdk/src/debug.h"
 #include "vmsdk/src/log.h"
+#include "vmsdk/src/module_config.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
+namespace vmsdk::debug {
+extern Controlled<bool> ForceRetry;
+}
+
```
allenss-amazon: This belongs in your namespace, not in the vmsdk::debug namespace.

vmsdk/src/debug.cc
```diff
@@ -21,6 +21,9 @@
 namespace vmsdk {
 namespace debug {
 
+// consider changing controlled boolean from static to inline type
+Controlled<bool> ForceRetry("ForceRetry", false);
+
```
allenss-amazon: This belongs elsewhere, not here.

src/query/fanout_operation_base.h
allenss-amazon: Actually, it should be a static declared in the .cc file that uses it. It doesn't need to be known anywhere else.

src/query/fanout_operation_base.h
```diff
@@ -67,29 +81,90 @@ class FanoutOperationBase {
...
+      if (vmsdk::debug::ForceRetry.GetValue() &&
+          Metrics::GetStats().fanout_retry_cnt < 10) {
+        std::thread([this, target]() {
+          this->OnError(grpc::Status(grpc::StatusCode::INTERNAL,
+                                     "Forced remote failure for testing"),
+                        coordinator::FanoutErrorType::COMMUNICATION_ERROR,
+                        target);
+          this->RpcDone();
+        }).detach();
+        return;
```
allenss-amazon: Why the "< 10" ?? This seems brittle and unexpected. You won't be able to use this for two tests without restarting the system....

Also, this doesn't "ForceRetry" --> This simulates a local gRPC error, like the GetClient failure below. Let's name it that way, i.e., something like "Fanout.Grpc.ForceLocalError" 

If you want to have the system limit the number of iterations for something like this. I'd recommend that you switch to an integer controlled variable and then decrement it once for each failure iteration. If the variable hits zero then it's back to being "off".

Alternatively you can stick with the boolean and just have it set itself "off" after it's been triggered once.

integration/test_fanout_base.py
allenss-amazon: <later> Yes, this doesn't match the test description, i.e., this particular test never leaves the local box. Which is definitely one test case. But doesn't cover the case where the target node is actually down. That's covered later in the shutdown case.

integration/test_fanout_base.py
allenss-amazon: This covers the case where the target node is completely down. But that might affect the cluster map and relies on the TCP stack behavior (i.e., connection timing out). It doesn't catch the case where the target system simply never responds (i.e., the request is delivered, but the target system is wedged and never responds). I think you can add another controlled variable to the gRPC receive path and have it just swallow the request without generated a response to simulate this case.

integration/test_info_primary.py
```diff
@@ -69,12 +88,52 @@ def test_ft_info_primary_counts(self):
...
+        raw = node0.execute_command("FT.INFO", index_name, "PRIMARY")
+
+        # check retry count
+        info_search_str = str(node0.execute_command("INFO SEARCH"))
+        pattern = r'search_fanout_retry_count:(\d+)'
+        match = re.search(pattern, info_search_str)
+        if not match:
+            assert False, f"search_fanout_retry_count not found in INFO SEARCH results!"
+        retry_count = int(match.group(1))
+        assert retry_count > 0, f"Expected retry_count to be greater than 0, got {retry_count}"
```
allenss-amazon: Isn't there an info parser?

integration/test_info_primary.py
```diff
@@ -69,12 +88,52 @@ def test_ft_info_primary_counts(self):
...
+        mode = info.get("mode")
+        index_name = info.get("index_name")
+        assert (mode in (b"primary", "primary"))
+        assert (index_name in (b"index1", "index1"))
+        num_docs = int(info["num_docs"])
+        num_records = int(info["num_records"])
+        hash_fail = int(info["hash_indexing_failures"])
+        assert num_docs == N
+        assert num_records == N
+        assert hash_fail == 0
```
allenss-amazon: Stylistically, this is wordy. I'd rather see something like:

```
assert N == int(info["num_records"])
```
This is clearer.

src/coordinator/client.cc
allenss-amazon: Let's just make it a config.

src/query/fanout_operation_base.h
```diff
@@ -29,16 +43,13 @@ class FanoutOperationBase {
...
-    outstanding_ = targets.size();
-    unsigned timeout_ms = GetTimeoutMs();
-    for (const auto& target : targets) {
-      auto req = GenerateRequest(target, timeout_ms);
-      IssueRpc(ctx, target, req, timeout_ms);
-    }
+    deadline_tp_ = std::chrono::steady_clock::now() +
+                   std::chrono::milliseconds(GetTimeoutMs());
+    targets_ = GetTargets(ctx);
+    Metrics::GetStats().fanout_retry_cnt.store(0);
```
allenss-amazon: I don't think we should ever be resetting this. It destroys the value of this in the operational environment.

src/query/fanout_operation_base.h
```diff
@@ -67,29 +81,90 @@ class FanoutOperationBase {
...
+    coordinator::ClientPool* client_pool_ =
+        ValkeySearch::Instance().GetCoordinatorClientPool();
+
+    if (target.type == FanoutSearchTarget::Type::kLocal) {
+      vmsdk::RunByMain([this, target, request]() {
+        auto [status, resp] = this->GetLocalResponse(request, target);
+        if (status.ok()) {
+          this->OnResponse(resp, target);
+        } else {
+          this->OnError(status, resp.error_type(), target);
```
allenss-amazon, line 108: We should have a counter for this error path.

src/query/fanout_operation_base.h
```diff
@@ -67,29 +81,90 @@ class FanoutOperationBase {
...
+          client.get(), request,
+          [this, target](grpc::Status status, Response& resp) {
+            if (status.ok()) {
+              this->OnResponse(resp, target);
+            } else {
+              VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+                  << "FANOUT_DEBUG: InvokeRemoteRpc error on target "
+                  << target.address << ", status code: " << status.error_code()
+                  << ", error message: " << status.error_message();
+              this->OnError(status, resp.error_type(), target);
```
allenss-amazon, line 136: let's have a counter for this error path.

src/metrics.h
```diff
@@ -124,6 +124,8 @@ class Metrics {
     std::atomic<uint64_t> time_slice_write_time{0};  // microseconds, cumulative
     std::atomic<uint64_t> time_slice_upserts{0};
     std::atomic<uint64_t> time_slice_deletes{0};
+
+    std::atomic<uint64_t> fanout_retry_cnt{0};
```
allenss-amazon: this name is a bit generic. You're using it strictly for the FT.INFO fanout, right? There are other "fanout" cases, like search.

allenss-amazon (approved):

PR #338, opened 2025-08-21
Fulltext query parsing + search logic base setup + Prefix Wildcard Search

KarthikSubbarao:
Fulltext query parsing, search logic base setup, and wildcard prefix search.

1. Creating / Defining all the required text predicates
2. Exact Term query parsing of Term, Prefix, Infix, Suffix, Fuzzy
3. Multi Term query parsing of Proximity (for ANDs and exact phrase)
4. Search layout changes
5. Wild card prefix search. supports searches on specific fields 
6. Updating the term/prefix search iterator handling for non existing words (which currently crashes)
7. Safely rejects any query with unsupported text predicates (everything other than Term & Prefix)


src/query/predicate.h
```diff
@@ -158,24 +258,59 @@ class TextPredicate : public Predicate {
...
+    return terms_.empty() ? "" : terms_[0]->GetIdentifier();
+  }
+
+  vmsdk::UniqueValkeyString GetRetainedIdentifier() const override {
+    return terms_.empty() ? vmsdk::UniqueValkeyString() : terms_[0]->GetRetainedIdentifier();
+  }
+
+  absl::string_view GetTextString() const override {
+    return terms_.empty() ? "" : terms_[0]->GetTextString();
+  }
```
KarthikSubbarao: ProximityPredicate should not support these operations.

However, it implements the TextPredicate, so it needs to implement the operations

src/indexes/text/term.cc
```diff
@@ -0,0 +1,47 @@
+#include "src/indexes/text/term.h"
```
Aksha1812, line 8: BSD license

src/indexes/text/term.h
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#ifndef _VALKEY_SEARCH_INDEXES_TEXT_PHRASE_H_
```
Aksha1812: VALKEY_SEARCH_INDEXES_TEXT_TERM_H_

src/indexes/text/wildcard.h
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#ifndef VALKEY_SEARCH_INDEXES_TREE_WILDCARD_ITERATOR_H_
```
Aksha1812: VALKEY_SEARCH_INDEXES_TEXT_WILDCARD_ITERATOR_H_

src/commands/filter_parser.cc
```diff
@@ -87,6 +87,105 @@ constexpr double kNegativeInf = -std::numeric_limits<double>::infinity();
...
+      break;
+    }
+    case query::PredicateType::kTag: {
+      const auto* tp = dynamic_cast<const query::TagPredicate*>(pred);
+      VMSDK_LOG(WARNING, nullptr) << prefix << "TAG(" << tp->GetTagString() << ")_"
+                                  << tp->GetIdentifier() << "\n";
+      break;
+    }
+    default:
+      valid |= false;
```
allenss-amazon: ? A fancy noop.

src/commands/filter_parser.cc
```diff
@@ -361,6 +433,214 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+  if (!tok.empty() && tok.back()  == '*') tok.remove_suffix(1);
+  return tok;
+}
+
+// Fuzzy: allow 1..3 '%' on both sides: %x%, %%x%%, %%%x%%%
+inline size_t GetFuzzyDistance(absl::string_view tok) {
+  if (tok.size() < 3) return 0;
+  auto count_leading = [](absl::string_view s) {
+    size_t n = 0; while (n < s.size() && s[n] == '%') ++n; return n;
+  };
```
allenss-amazon: Let's not hard-code the 3.

src/commands/filter_parser.cc
```diff
@@ -361,6 +433,214 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+    VMSDK_ASSIGN_OR_RETURN(auto t, BuildSingleTextPredicate(field_for_default, tok));
+    terms.push_back(std::move(t));
+    tok.clear();
+    return absl::OkStatus();
+  };
+  if (Match('"')) {
+    std::string curr;
+    while (!IsEnd()) {
+      char c = Peek();
+      if (c == '"') { ++pos_; break; }
```
allenss-amazon: Eventually this needs to handle escaped characters?

src/commands/filter_parser.cc
```diff
@@ -361,6 +433,214 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+                    VMSDK_ASSIGN_OR_RETURN(auto numeric, ParseNumericPredicate(current_field));
+                    extra_terms.push_back(std::move(numeric));
+                    continue;
+                } else if (next == '{') {
+                    ++pos_; // consume '{' for tag
+                    VMSDK_ASSIGN_OR_RETURN(auto tag, ParseTagPredicate(current_field));
+                    extra_terms.push_back(std::move(tag));
+                    continue;
+                }
+            }
```
allenss-amazon: Error if the IsEnd is reached?

src/query/predicate.h
allenss-amazon: Should you be asserting?

src/commands/filter_parser.cc
KarthikSubbarao: This (3) will actually be changed to be a modifiable module config. I planned to implement configs in a follow up PR, for now, I will update it to use a Static variable

src/commands/filter_parser.cc
KarthikSubbarao: Yes, escaped characters is a task I am tracking. When query string processing is implemented, we need to handle punctuation, stopword removal, handle escaped characters, as well as optional stemming

src/commands/filter_parser.cc
KarthikSubbarao: Do you mean the `valid` check? This is a temporary function for logging and for rejecting the currently unsupported text predicates. I plan to remove this function once we support all the remaining text predicates and also move the logging of the query syntax tree  into the FT.EXPLAINCLI command to help print out the query syntax tree for the query string

src/indexes/text/proximity.h
```diff
@@ -62,7 +62,7 @@ void process_one_key(KeyIterators[*]) {
   }
 }
 */
-class PhraseIterator : public indexes::EntriesFetcherIteratorBase {
+class ProximityIterator : public indexes::EntriesFetcherIteratorBase {
  public:
   PhraseIterator(const std::vector<WordIterator>& words,
```
Aksha1812: PromixityIterator* . now this wouldn't take Worditerator or field mask as input right?

src/indexes/text/term.cc
```diff
@@ -0,0 +1,54 @@
...
+  // On a Begin() call, we initialize the target_posting_ and key_iter_.
+  if (begin_) {
+    if (word_.Done()) {
+      nomatch_ = true;
+      return;
+    }
+    target_posting_ = word_.GetTarget();
+    key_iter_ = target_posting_->GetKeyIterator();
+    begin_ = false;  // Set to false after the first call to Next.    
+  } else {
```
Aksha1812, line 40: This doesn't seem right. if begin_ == false wouldn't we call NextKey() every single time before checking if it contains field mask.

pseudo code :

```
if begin 
  .
  .
  .
  while !Done() && !key_iter_.ContainsFields(field_mask_)
     NextKey()
  begin = false
else
  while !Done() && !key_iter_.ContainsFields(field_mask_)
     NextKey()
```

src/indexes/wildcard_iterator.cc
Aksha1812, line 1: I think this is added by mistake

src/indexes/text/term.cc
KarthikSubbarao: > wouldn't we call NextKey() every single time before checking if it contains field mask.

That is what a Term Match is supposed to do, right? We need to iterate through all the keys that match the word, so we stay on the same position in the word iterator. We need to advance to the next key every time Next is called, we hit the else code path and then check if it is contains the field. If not, we continue advancing till we find a field match

src/indexes/text/proximity.h
KarthikSubbarao: Yes, This is a place holder. It will need to be updated when implementing the PhraseIterator. Rough input will look like a vector of EntriesIterator base class, Slop, Inorder boolean.

src/indexes/text/term.cc
Aksha1812: it should not call NextKey() until ContainsFields is true.

KarthikSubbarao: if we ever get in the else path, we reach the while loop and then it stays there until we match the condition of `ContainsFields`

allenss-amazon: What is the purpose of begin_ ?

KarthikSubbarao: When the `EntriesFetcherIteratorBase` are being created in the Begin call after the `Search()` function, we need to initialize the Search Iterator as well as potentially have slightly different on the first call of Next.

`begin_`  is used on the first call of `Next` here for these two purposes

```
std::unique_ptr<EntriesFetcherIteratorBase> Text::EntriesFetcher::Begin() {
  if (auto term = dynamic_cast<const query::TermPredicate*>(predicate_)) {
    auto iter = text_index_->prefix_.GetWordIterator(term->GetTextString());
    auto itr = std::make_unique<text::TermIterator>(iter, term->GetTextString(), field_mask_, untracked_keys_);
    itr->Next();
    return itr;
  } else if (auto prefix = dynamic_cast<const query::PrefixPredicate*>(predicate_)) {
    auto iter = text_index_->prefix_.GetWordIterator(prefix->GetTextString());
    auto itr = std::make_unique<text::WildCardIterator>(iter, text::WildCardOperation::kPrefix, field_mask_, untracked_keys_);
    itr->Next();
    return itr;
  }
  CHECK(false) << "Unsupported TextPredicate operation";
  return nullptr;
}

```

src/commands/filter_parser.cc
```diff
@@ -361,6 +435,218 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+enum class WildcardType { kNone, kPrefix, kSuffix, kInfix };
+
+// detect wildcard kind based on '*'
+WildcardType DetectWildcard(absl::string_view tok) {
+  const bool starts = !tok.empty() && tok.front() == '*';
+  const bool ends   = !tok.empty() && tok.back()  == '*';
+  if (!starts && !ends) return WildcardType::kNone;
+  if (starts && ends)   return WildcardType::kInfix;
+  if (starts)           return WildcardType::kSuffix; // "*x"
+  return WildcardType::kPrefix;                       // "x*"
```
allenss-amazon: The returned classifications don't make sense to me. You can detect suffix and prefix  by checking each end, but you can't distinguish Infix from None. What was the purpose of this?

src/commands/filter_parser.cc
```diff
@@ -361,6 +435,218 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+}
+
+// strip one leading and/or trailing '*' for prefix/suffix/infix
+absl::string_view StripWildcardMarkers(absl::string_view tok) {
+  if (!tok.empty() && tok.front() == '*') tok.remove_prefix(1);
+  if (!tok.empty() && tok.back()  == '*') tok.remove_suffix(1);
+  return tok;
```
allenss-amazon: This could remove a wildcard and front AND end, is that what you want?

src/commands/filter_parser.cc
```diff
@@ -361,6 +435,218 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+  auto count_trailing = [](absl::string_view s) {
+    size_t n = 0; while (n < s.size() && s[s.size()-1-n] == '%') ++n; return n;
+  };
+  size_t lead = count_leading(tok);
+  size_t tail = count_trailing(tok);
+  if (lead != tail || lead == 0  || lead > FUZZY_MAX_DISTANCE) return 0;
+  if ((lead + tail) < tok.size()) {
+    return lead;
+  }
+  return 0;
```
allenss-amazon: There are lots of cases of mismatch counts of %, etc. that just return 0, rather than an error. Is there a place where those types of errors are detected, why wouldn't that combined here?

src/commands/filter_parser.cc
```diff
@@ -361,6 +435,218 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+}
+
+absl::string_view StripFuzzyMarkers(absl::string_view tok, size_t& distance) {
+  // remove up to the `distance` leading/trailing '%'
+  size_t i = 0; while (i < tok.size() && tok[i] == '%' && i < distance) ++i;
+  size_t j = 0; while (j < tok.size() && tok[tok.size()-1-j] == '%' && j < distance) ++j;
+  tok.remove_prefix(i);
+  tok.remove_suffix(j);
+  return tok;
+}
```
allenss-amazon: Why loop? If you have a validated distance, then you can just chop the view accordingly. If you don't then this function will really obscure what's going on. For example if you have mismatched count of '%' is this really what you want to do?

allenss-amazon: It feels to me like you'll want to rewrite the functions above. One pass over the token should be sufficient to detect all of the various wildcard and fuzzy information. It feels more complex now, but I think it will actually be simpler. Some counters a couple of state variables should be sufficient.

src/commands/filter_parser.cc
```diff
@@ -361,6 +435,218 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+      return absl::InvalidArgumentError("Empty fuzzy token");
+    }
+    return std::make_unique<query::FuzzyPredicate>(text_index, identifier, field_name, std::string(core),
+                                                   fuzzy_distance);
+  }
+  // Plain term
+  if (token.empty()) {
+    return absl::InvalidArgumentError("Empty text token");
+  }
+  return std::make_unique<query::TermPredicate>(text_index, identifier, field_name, std::string(token));
```
allenss-amazon: This is the block that feels like a rewrite is wanted. Breaking it up into these different pieces seems like there's a lot of potential cases that don't get well handled well.

src/commands/filter_parser.cc
```diff
@@ -361,6 +435,218 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+    if (!current_field.empty()) {
+      field_for_atom = current_field;
+    }
+    // Field override or numeric/tag
+    if (c == '@') {
+      VMSDK_ASSIGN_OR_RETURN(current_field, ParseFieldName());
+      field_for_atom = current_field;
+      SkipWhitespace();
+      if (!IsEnd()) {
+        if (Match('[')) {
```
allenss-amazon, line 613: This doesn't handle @(F1 | F2): I assume that's coming later?

src/indexes/text/term.cc
allenss-amazon: The exact relationship between:

1. Constructing the EntriesFetcher.
2. The semantics of Next
3. The semantics of Done

aren't clear. It would be good to have a large comment to explain those as it will help to under the code. Specifically here, there is ambiguity as to the state of the iterator after the construction. Do you need to do a Next to "prime the pump" or is the iterator valid after the construction?

Is the expected usage case 1:
```
Iterator x(<construction...>);
  while (!x.Done()) {
  ......
  x.Next()
}
```
or
```
iterator x(<construction>);
x.Next()
while (!x.Done()) {
  ....
  x.Next();
}
```

allenss-amazon: If it's the former, then I don't understand why you need the "begin_" variable?

src/indexes/text/wildcard.cc
```diff
@@ -0,0 +1,64 @@
...
+void WildCardIterator::Next() {
+  // Note: Currently only supports Prefix search.
+  if (begin_) {
+    if (word_.Done()) {
+      nomatch_ = true;
+      return;
+    }
+    target_posting_ = word_.GetTarget();
+    key_iter_ = target_posting_->GetKeyIterator();
+    begin_ = false;
```
allenss-amazon, line 38: Let's chat about this.

src/commands/filter_parser.cc
KarthikSubbarao: This is not considered valid syntax in redissearch query string semantics 
```
127.0.0.1:6380> ft.search products '@(desc | desc2):hello' dialect 2
(error) Syntax error at offset 15 near desc2
127.0.0.1:6380> ft.search products '@(desc):hello' dialect 2
(error) Syntax error at offset 7 near desc
127.0.0.1:6380> ft.search products '@desc:hello' dialect 2
1) (integer) 4
2) "product:2"
.
.
.
```

src/indexes/text/term.cc
KarthikSubbarao: In the Proximity Iterator PR, I will introduce the Hierarchy definition which we will follow for text search operations

allenss-amazon (approved):

PR #342, opened 2025-08-22
Implement lexical scanner

VoletiRam:
Implements a complete stateless lexical scanner for text field indexing with tokenization, case normalization, and stemmer integration. This establishes the foundation for full-text search by providing robust text processing capabilities within the existing storage architecture.

### __Lexical Tokenization Pipeline__

- __Punctuation-based word splitting__ with configurable separators
- __Case normalization__ (UPPERCASE → lowercase)
- __UTF-8 validation__ with proper error handling
- __Stemmer integration__ using existing Snowball stemmer
- __Configurable minimum stem size__ threshold

### __Stateless Architecture Design__

- __Zero configuration storage__ in Lexer class
- __Parameter-driven processing__ - all config passed from Text class
- __Shared configuration optimization__ via TextIndexSchema
- __Clean separation__ between storage (TextIndexSchema) and processing (Lexer)

### __Lexer Pipeline__

```
text → UTF-8 validation → tokenize → lowercase → stop word removal(TODO) → stem → tokens
```

Tested with debug logs in local

commands tested:
```

127.0.0.1:6379> FT.CREATE idx_case ON HASH SCHEMA title TEXT
OK
127.0.0.1:6379> HSET test:3 title "HELLO World MiXeD cAsE"
(integer) 1
127.0.0.1:6379> FT.CREATE idx_utf8 ON HASH SCHEMA title TEXT
OK
127.0.0.1:6379> HSET test:4 title "hello 世界 café résumé"
(integer) 1
127.0.0.1:6379> HSET test:5 title "Привет мир こんにちは"
(integer) 1
127.0.0.1:6379>
```

Server logs for the above commands
```
12433:M 22 Aug 2025 04:43:07.220 * Server initialized
12433:M 22 Aug 2025 04:43:07.220 * Ready to accept connections tcp
12433:M 22 Aug 2025 04:43:16.990 # <search> [notice], tid: 140703581000448, /src/index_schema.cc:70: Starting backfill for index schema in DB 0: idx_case (size: 0)
12433:M 22 Aug 2025 04:43:17.044 # <search> [notice], tid: 140703581000448, /src/index_schema.cc:682: Index schema idx_case finished backfill. Scanned 0 keys in 54.3789ms
12433:M 22 Aug 2025 04:43:18.628 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:61: [LEXER] 'HELLO World MiXeD cAsE' -> 4 tokens:
12433:M 22 Aug 2025 04:43:18.628 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:63: [LEXER]   [0]: 'hello'
12433:M 22 Aug 2025 04:43:18.628 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:63: [LEXER]   [1]: 'world'
12433:M 22 Aug 2025 04:43:18.628 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:63: [LEXER]   [2]: 'mix'
12433:M 22 Aug 2025 04:43:18.628 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:63: [LEXER]   [3]: 'case'
12433:M 22 Aug 2025 04:43:47.600 # <search> [notice], tid: 140703581000448, /src/index_schema.cc:70: Starting backfill for index schema in DB 0: idx_utf8 (size: 1)
12433:M 22 Aug 2025 04:43:47.642 # <search> [notice], tid: 140703581000448, /src/index_schema.cc:682: Index schema idx_utf8 finished backfill. Scanned 1 keys in 42.478161ms
12433:M 22 Aug 2025 04:43:47.643 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:61: [LEXER] 'HELLO World MiXeD cAsE' -> 4 tokens:
12433:M 22 Aug 2025 04:43:47.643 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:63: [LEXER]   [0]: 'hello'
12433:M 22 Aug 2025 04:43:47.643 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:63: [LEXER]   [1]: 'world'
12433:M 22 Aug 2025 04:43:47.643 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:63: [LEXER]   [2]: 'mix'
12433:M 22 Aug 2025 04:43:47.643 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:63: [LEXER]   [3]: 'case'
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:61: [LEXER] 'hello 世界 café résumé' -> 4 tokens:
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:61: [LEXER] 'hello 世界 café résumé' -> 4 tokens:
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:63: [LEXER]   [0]: 'hello'
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:63: [LEXER]   [0]: 'hello'
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:63: [LEXER]   [1]: '世界'
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:63: [LEXER]   [2]: 'café'
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:63: [LEXER]   [1]: '世界'
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140700779804416, /src/indexes/text/lexer.cc:63: [LEXER]   [3]: 'résumé'
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:63: [LEXER]   [2]: 'café'
12433:M 22 Aug 2025 04:43:53.572 # <module> [warning], tid: 140701331334912, /src/indexes/text/lexer.cc:63: [LEXER]   [3]: 'résumé'
12433:M 22 Aug 2025 04:44:14.786 # <module> [warning], tid: 140700635096832, /src/indexes/text/lexer.cc:61: [LEXER] 'Привет мир こんにちは' -> 3 tokens:
12433:M 22 Aug 2025 04:44:14.786 # <module> [warning], tid: 140700635096832, /src/indexes/text/lexer.cc:63: [LEXER]   [0]: 'Привет'
12433:M 22 Aug 2025 04:44:14.786 # <module> [warning], tid: 140700635096832, /src/indexes/text/lexer.cc:63: [LEXER]   [1]: 'мир'
12433:M 22 Aug 2025 04:44:14.786 # <module> [warning], tid: 140700737844992, /src/indexes/text/lexer.cc:61: [LEXER] 'Привет мир こんにちは' -> 3 tokens:
12433:M 22 Aug 2025 04:44:14.786 # <module> [warning], tid: 140700635096832, /src/indexes/text/lexer.cc:63: [LEXER]   [2]: 'こんにちは'
12433:M 22 Aug 2025 04:44:14.786 # <module> [warning], tid: 140700737844992, /src/indexes/text/lexer.cc:63: [LEXER]   [0]: 'Привет'
12433:M 22 Aug 2025 04:44:14.786 # <module> [warning], tid: 140700737844992, /src/indexes/text/lexer.cc:63: [LEXER]   [1]: 'мир'
12433:M 22 Aug 2025 04:44:14.786 # <module> [warning], tid: 140700737844992, /src/indexes/text/lexer.cc:63: [LEXER]   [2]: 'こんにちは'


```

### DIFF
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -75,6 +75,7 @@ IndexSchema::BackfillJob::BackfillJob(ValkeyModuleCtx *ctx,
 absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
     ValkeyModuleCtx *ctx, IndexSchema *index_schema,
     const data_model::Attribute &attribute,
+    const data_model::IndexSchema *index_schema_proto,
     std::optional<SupplementalContentChunkIter> iter) {
   const auto &index = attribute.index();
   switch (index.index_type_case()) {
@@ -85,10 +86,17 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
       return std::make_shared<indexes::Numeric>(index.numeric_index());
     }
     case data_model::Index::IndexTypeCase::kTextIndex: {
-      // Create or reuse shared TextIndexSchema
       if (!index_schema->GetTextIndexSchema()) {
-        // TODO : get the index_schema_proto the right way here to pass schema level properties
-        index_schema->CreateTextIndexSchema();
+        std::vector<std::string> stop_words(
+            index_schema_proto->stop_words().begin(),
+            index_schema_proto->stop_words().end()
+        );
+        index_schema->CreateTextIndexSchema(
+            index_schema_proto->language(),
+            index_schema_proto->punctuation(),
+            index_schema_proto->with_offsets(),
+            stop_words
+        );
       }
       return std::make_shared<indexes::Text>(index.text_index(),
                                              index_schema->GetTextIndexSchema());
@@ -181,7 +189,7 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::Create(
     for (const auto &attribute : index_schema_proto.attributes()) {
       VMSDK_ASSIGN_OR_RETURN(
           std::shared_ptr<indexes::IndexBase> index,
-          IndexFactory(ctx, res.get(), attribute, std::nullopt));
+          IndexFactory(ctx, res.get(), attribute, &index_schema_proto, std::nullopt));
       VMSDK_RETURN_IF_ERROR(
           res->AddIndex(attribute.alias(), attribute.identifier(), index));
     }
@@ -222,6 +230,7 @@ IndexSchema::IndexSchema(ValkeyModuleCtx *ctx,
       }
     }
   }
+
   // The protobuf has volatile fields that get save/restores in the RDB. here we
   // reconcile the source of the index_schema_proto (reload or not) and restore those fields
   if (reload) {
@@ -851,6 +860,14 @@ std::unique_ptr<data_model::IndexSchema> IndexSchema::ToProto() const {
   index_schema_proto->mutable_subscribed_key_prefixes()->Add(
       subscribed_key_prefixes_.begin(), subscribed_key_prefixes_.end());
   index_schema_proto->set_attribute_data_type(attribute_data_type_->ToProto());
+
+  // Get text configurations from TextIndexSchema for RDB serialization
+  if (this->text_index_schema_) {
+    index_schema_proto->set_language(this->text_index_schema_->GetLanguage());
+    index_schema_proto->set_punctuation(this->text_index_schema_->GetPunctuation());
+    index_schema_proto->set_with_offsets(this->text_index_schema_->GetWithOffsets());
+  }
+
   auto stats = index_schema_proto->mutable_stats();
   stats->set_documents_count(stats_.document_cnt);
   std::transform(
@@ -957,7 +974,7 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::LoadFromRDB(
           supplemental_content->index_content_header().attribute();
       VMSDK_ASSIGN_OR_RETURN(std::shared_ptr<indexes::IndexBase> index,
                              IndexFactory(ctx, index_schema.get(), attribute,
-                                          supplemental_iter.IterateChunks()));
+                                          index_schema_proto.get(), supplemental_iter.IterateChunks()));
       VMSDK_RETURN_IF_ERROR(index_schema->AddIndex(
           attribute.alias(), attribute.identifier(), index));
     } else if (ABSL_PREDICT_TRUE(!skip_loading_index_data) &&
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -87,16 +87,7 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
     }
     case data_model::Index::IndexTypeCase::kTextIndex: {
       if (!index_schema->GetTextIndexSchema()) {
-        std::vector<std::string> stop_words(
-            index_schema_proto->stop_words().begin(),
-            index_schema_proto->stop_words().end()
-        );
-        index_schema->CreateTextIndexSchema(
-            index_schema_proto->language(),
-            index_schema_proto->punctuation(),
-            index_schema_proto->with_offsets(),
-            stop_words
-        );
+        index_schema->CreateTextIndexSchema();
       }
       return std::make_shared<indexes::Text>(index.text_index(),
                                              index_schema->GetTextIndexSchema());
@@ -216,6 +207,11 @@ IndexSchema::IndexSchema(ValkeyModuleCtx *ctx,
       attribute_data_type_(std::move(attribute_data_type)),
       name_(std::string(index_schema_proto.name())),
       db_num_(index_schema_proto.db_num()),
+      language_(index_schema_proto.language()),
+      punctuation_(index_schema_proto.punctuation()),
+      with_offsets_(index_schema_proto.with_offsets()),
+      stop_words_(index_schema_proto.stop_words().begin(),
+                  index_schema_proto.stop_words().end()),
       mutations_thread_pool_(mutations_thread_pool),
       time_sliced_mutex_(CreateMrmwMutexOptions()) {
   ValkeyModule_SelectDb(detached_ctx_.get(), db_num_);
@@ -861,12 +857,11 @@ std::unique_ptr<data_model::IndexSchema> IndexSchema::ToProto() const {
       subscribed_key_prefixes_.begin(), subscribed_key_prefixes_.end());
   index_schema_proto->set_attribute_data_type(attribute_data_type_->ToProto());
 
-  // Get text configurations from TextIndexSchema for RDB serialization
-  if (this->text_index_schema_) {
-    index_schema_proto->set_language(this->text_index_schema_->GetLanguage());
-    index_schema_proto->set_punctuation(this->text_index_schema_->GetPunctuation());
-    index_schema_proto->set_with_offsets(this->text_index_schema_->GetWithOffsets());
-  }
+  // Always serialize text configurations from stored members
+  index_schema_proto->set_language(language_);
+  index_schema_proto->set_punctuation(punctuation_);
+  index_schema_proto->set_with_offsets(with_offsets_);
+  index_schema_proto->mutable_stop_words()->Assign(stop_words_.begin(), stop_words_.end());
 
   auto stats = index_schema_proto->mutable_stats();
   stats->set_documents_count(stats_.document_cnt);
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -46,8 +46,8 @@ absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
     return tokens.status();
   }
 
-  uint32_t position = 0;
-  for (const auto& token : *tokens) {
+  for (uint32_t position = 0; position < tokens->size(); ++position) {
+    const auto& token = (*tokens)[position];
     text_index_schema_->text_index_->prefix_.Mutate(
         token,
         [&](std::optional<std::shared_ptr<text::Postings>> existing)
@@ -65,7 +65,6 @@ absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
           postings->InsertPosting(key, text_field_number_, position);
           return postings;
         });
-    position++;
   }
 
   return true;
@@ -90,7 +89,8 @@ bool Text::IsTracked(const InternedStringPtr& key) const {
 }
 
 uint64_t Text::GetRecordCount() const {
-  throw std::runtime_error("Text::GetRecordCount not implemented");
+  // TODO: Implement proper record count tracking when key management is added
+  return 0;
 }
 
 std::unique_ptr<data_model::Index> Text::ToProto() const {
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -13,6 +13,7 @@
 #include "absl/strings/string_view.h"
 #include "src/index_schema.pb.h"
 #include "src/indexes/text/posting.h"
+#include "src/indexes/text/lexer.h"
 
 namespace valkey_search::indexes {
 
@@ -28,39 +29,45 @@ Text::Text(const data_model::TextIndex& text_index_proto,
 
 absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
                                      absl::string_view data) {
-  // TODO: Replace this tokenizing with the proper lexer functionality when it's
-  // implemented
-  int prev_pos = 0;
-  uint32_t position = 0;
-  
-  for (int i = 0; i <= data.size(); i++) {
-    if (i == data.size() || data[i] == ' ') {
-      if (i > prev_pos) {
-        absl::string_view word = data.substr(prev_pos, i - prev_pos);
-        text_index_schema_->text_index_->prefix_.Mutate(
-            word,
-            [&](std::optional<std::shared_ptr<text::Postings>> existing)
-                -> std::optional<std::shared_ptr<text::Postings>> {
-              std::shared_ptr<text::Postings> postings;
-              if (existing.has_value()) {
-                postings = existing.value();
-              } else {
-                // Create new Postings object with schema configuration
-                // TODO: Get save_positions from IndexSchema, for now assume true
-                bool save_positions = text_index_schema_->with_offsets_;
-                uint8_t num_text_fields = text_index_schema_->num_text_fields_;
-                postings = std::make_shared<text::Postings>(save_positions, num_text_fields);
-              }
-              
-              // Add the key and position to postings
-              postings->InsertPosting(key, text_field_number_, position);
-              return postings;
-            });
-        position++;
-      }
-      prev_pos = i + 1;
+  valkey_search::indexes::text::Lexer lexer;
+
+  auto tokens = lexer.Tokenize(
+      data,
+      text_index_schema_->GetPunctuationBitmap(),
+      text_index_schema_->GetStemmer(),
+      !no_stem_,
+      min_stem_size_
+  );
+
+  if (!tokens.ok()) {
+    if (tokens.status().code() == absl::StatusCode::kInvalidArgument) {
+      return false;  // UTF-8 errors → hash_indexing_failures
     }
+    return tokens.status();
   }
+
+  uint32_t position = 0;
+  for (const auto& token : *tokens) {
+    text_index_schema_->text_index_->prefix_.Mutate(
+        token,
+        [&](std::optional<std::shared_ptr<text::Postings>> existing)
+            -> std::optional<std::shared_ptr<text::Postings>> {
+          std::shared_ptr<text::Postings> postings;
+          if (existing.has_value()) {
+            postings = existing.value();
+          } else {
+            // Create new Postings object with schema configuration
+            bool save_positions = text_index_schema_->GetWithOffsets();
+            uint8_t num_text_fields = text_index_schema_->num_text_fields_;
+            postings = std::make_shared<text::Postings>(save_positions, num_text_fields);
+          }
+
+          postings->InsertPosting(key, text_field_number_, position);
+          return postings;
+        });
+    position++;
+  }
+
   return true;
 }
 
--- a/src/indexes/text.h
+++ b/src/indexes/text.h
@@ -8,10 +8,8 @@
 #ifndef VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 #define VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 
-// C++ standard library headers
 #include <memory>
 
-// Third-party library headers
 #include "absl/base/thread_annotations.h"
 #include "absl/functional/any_invocable.h"
 #include "absl/status/status.h"
@@ -19,7 +17,6 @@
 #include "absl/strings/string_view.h"
 #include "absl/synchronization/mutex.h"
 
-// Project headers
 #include "src/indexes/index_base.h"
 #include "src/indexes/text/text_index.h"
 #include "src/utils/string_interning.h"
--- a/src/indexes/text.h
+++ b/src/indexes/text.h
@@ -8,18 +8,20 @@
 #ifndef VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 #define VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 
+// C++ standard library headers
 #include <memory>
-#include <optional>
+
+// Third-party library headers
 #include "absl/base/thread_annotations.h"
-#include "absl/container/flat_hash_map.h"
 #include "absl/functional/any_invocable.h"
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/string_view.h"
 #include "absl/synchronization/mutex.h"
+
+// Project headers
 #include "src/indexes/index_base.h"
 #include "src/indexes/text/text_index.h"
-#include "src/index_schema.pb.h"
 #include "src/utils/string_interning.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 #include "src/indexes/text/term.h"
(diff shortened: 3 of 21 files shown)


src/index_schema.cc
```diff
@@ -85,10 +86,17 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexFactory(
...
     }
     case data_model::Index::IndexTypeCase::kTextIndex: {
-      // Create or reuse shared TextIndexSchema
       if (!index_schema->GetTextIndexSchema()) {
-        // TODO : get the index_schema_proto the right way here to pass schema level properties
-        index_schema->CreateTextIndexSchema();
+        std::vector<std::string> stop_words(
+            index_schema_proto->stop_words().begin(),
+            index_schema_proto->stop_words().end()
+        );
```
allenss-amazon: Should index_schema_proto should be a parameter to CreateTextIndexSchema? This loop and the following code would then move into that code, right?

src/index_schema.cc
```diff
@@ -851,6 +860,14 @@ std::unique_ptr<data_model::IndexSchema> IndexSchema::ToProto() const {
...
       subscribed_key_prefixes_.begin(), subscribed_key_prefixes_.end());
   index_schema_proto->set_attribute_data_type(attribute_data_type_->ToProto());
+
+  // Get text configurations from TextIndexSchema for RDB serialization
+  if (this->text_index_schema_) {
+    index_schema_proto->set_language(this->text_index_schema_->GetLanguage());
+    index_schema_proto->set_punctuation(this->text_index_schema_->GetPunctuation());
+    index_schema_proto->set_with_offsets(this->text_index_schema_->GetWithOffsets());
+  }
+
```
allenss-amazon: Why is this conditional? Seems like schema-level attributes (Language, Punctuation, WithOffset) should just be save/restored like any other. What should be different is that these should be marked "optional" in the protobuf and if they aren't present, then we should initialize them to a default value -- so that the remainder of the code doesn't treat it as a special case. Even if there aren't any text fields for this Schema, I don't care about burning a few extra bytes in the protobuf if it simplifes my life later.

We might think about making the protobuf have a struct named something like SchemaExtensionV2 that has the extra fields and then the entire struct would be marked optional. Seems a bit cleaner from a documentation perspective and avoids creating failure modes because one of these fields was specified but the others weren't.

src/indexes/text.cc
```diff
@@ -28,39 +29,45 @@ Text::Text(const data_model::TextIndex& text_index_proto,
...
+      !no_stem_,
+      min_stem_size_
+  );
+
+  if (!tokens.ok()) {
+    if (tokens.status().code() == absl::StatusCode::kInvalidArgument) {
+      return false;  // UTF-8 errors → hash_indexing_failures
     }
+    return tokens.status();
+  }
```
allenss-amazon: Why wouldn't this be just a VMSDK_ASSIGN_OR_RETURN ?

src/indexes/text.cc
```diff
@@ -28,39 +29,45 @@ Text::Text(const data_model::TextIndex& text_index_proto,
...
+
+  uint32_t position = 0;
+  for (const auto& token : *tokens) {
```
allenss-amazon: I think it would have been clearer to make the for loop iterate position from 0 to tokens.size() and then index into the array. Just a nit.

src/indexes/text/lexer.cc
```diff
@@ -8,48 +8,91 @@
...
+    size_t word_start = pos;
+    while (pos < text.size() &&
+           !punct_bitmap[static_cast<unsigned char>(text[pos])]) {
+      pos++;
+    }
+
+    if (pos > word_start) {
+      std::string word(text.data() + word_start, pos - word_start);
+
+      word = absl::AsciiStrToLower(word);
```
allenss-amazon: Let's make a string_view to pass into absl, avoid a meaningless malloc/free and copy of the data.

src/indexes/text/lexer.cc
```diff
@@ -8,48 +8,91 @@
...
-  CHECK(stemmer_) << "Stemmer not initialized";
+std::string Lexer::StemWord(
+    const std::string& word,
+    sb_stemmer* stemmer) const {
   
   if (word.empty()) {
     return word;
   }
   
+  CHECK(stemmer) << "Stemmer not initialized";
```
allenss-amazon, line 67: On it's face, it looks like this code might be susceptible to leaking the memory associated with the output of the stemmer. If that's NOT a problem, then a comment about the ownership of memory associated with the stemmer would be useful here.

src/indexes/text/text_index.cc
```diff
@@ -0,0 +1,27 @@
...
+
+TextIndexSchema::~TextIndexSchema() {
+  if (stemmer_) {
+    sb_stemmer_delete(stemmer_);
+  }
+}
+
+sb_stemmer* TextIndexSchema::GetStemmer() const {
+  if (!stemmer_) {
+    stemmer_ = sb_stemmer_new(GetLanguageString().c_str(), "UTF_8");
```
allenss-amazon, line 22: Has the language been verified at this point in time? If so, then this should never fail -- right? Please add a CHECK for that.

src/indexes/text/text_index.h
```diff
@@ -8,20 +8,31 @@
...
 #include "src/index_schema.pb.h"
 
+struct sb_stemmer;
+
 namespace valkey_search::indexes::text {
 
+namespace {
+
+bool IsWhitespace(char c) {
+  return std::isspace(static_cast<unsigned char>(c)) || std::iscntrl(static_cast<unsigned char>(c));
```
allenss-amazon: Are the casts really needed here?

src/indexes/text/text_index.h
```diff
@@ -70,16 +87,69 @@ struct TextIndexSchema {
...
+    // Add all whitespace characters as word separators (RFC requirement)
+    for (int i = 0; i < 256; ++i) {
+      if (IsWhitespace(static_cast<char>(i))) {
+        punct_bitmap_.set(i);
+      }
+    }
+
+    // Add user-specified punctuation characters
+    for (char c : punctuation_) {
+      if (static_cast<unsigned char>(c) < 256) {  // Only ASCII for now
```
allenss-amazon: I don't believe this test can fail.

src/utils/scanner.h
```diff
@@ -0,0 +1,235 @@
...
+      return std::nullopt;
+    }
+    double d = 0.0;
+#if 0
+    // CLANG doesn't support from_chars for floating point types.
+    auto [ptr, ec] = std::from_chars(&sv_[pos_], sv_.data() + sv_.size(), d);
+    if (ec == std::errc::invalid_argument) {
+      return std::nullopt;
+    }
+    RedisModule_Assert(ec == std::errc());
```
allenss-amazon: The dread pirate Redis.

testing/lexer_test.cc
```diff
@@ -0,0 +1,194 @@
...
+  ASSERT_TRUE(result.ok());
+  const auto& words = *result;
+  EXPECT_TRUE(words.empty());
+}
+
+TEST_F(LexerTest, DefaultPunctuationHandling) {
+  auto result = lexer_->Tokenize("hello,world!this-is_a.test", text_schema_->GetPunctuationBitmap(), text_schema_->GetStemmer(), stemming_enabled_, min_stem_size_);
+  ASSERT_TRUE(result.ok());
+  const auto& words = *result;
+  EXPECT_EQ(words.size(), 6);
```
allenss-amazon: can't you do EXPECT_EQ(words, { "hello", "world", ....}) ?

allenss-amazon: It feels like a lot of these tests could just be handled with a single test that has an array of values & responses. This will be similarly easy to maintain and long term, faster to build, etc.

testing/text_test.cc
```diff
@@ -0,0 +1,86 @@
...
+  std::string data = "";
+  
+  // Adding an empty record should still succeed
+  auto result = text_index_->AddRecord(key, data);
+  ASSERT_TRUE(result.ok()) << result.status();
+  EXPECT_TRUE(result.value());
+}
+
+TEST_F(TextTest, LargeDocument) {
+  auto key = valkey_search::StringInternStore::Intern("large_doc_key");
```
allenss-amazon: I'm not seeing much value in this set of tests. It just proves that AddRecord can return OK. If you wanted to add validation that the index structures were built correctly, that would be different....

src/indexes/text/text_index.cc
Aksha1812: this should be ensured at query level.but yes adding a CHECK here would be a good sanity check

src/indexes/text/text_index.h
```diff
@@ -70,16 +87,69 @@ struct TextIndexSchema {
...
+
+  bool GetWithOffsets() const {
+    return with_offsets_;
+  }
+
+  const PunctuationBitmap& GetPunctuationBitmap() const {
+    return punct_bitmap_;
+  }
+
+  std::string GetLanguageString() const {
```
Aksha1812, line 114: I guess we can have a proto or some other constants file in third_party/snowball folder where this would be automatically picked from rather than extending the switch statement for each langauge. Maybe not needed in this pr but definitely before adding support for other languages.

src/indexes/text/text_index.h
```diff
@@ -70,16 +87,69 @@ struct TextIndexSchema {
...
+    switch (language_) {
+      case data_model::LANGUAGE_ENGLISH:
+        return "english";
+      default:
+        return "english";
+    }
+  }
+
+ private:
+  void BuildOptimizedStructures() {
```
Aksha1812: It is not clear what the function does from the name of this function. Alter the name if you find a better alternative .

src/indexes/text.h
```diff
@@ -8,18 +8,20 @@
 #ifndef VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 #define VALKEYSEARCH_SRC_INDEXES_TEXT_H_
 
+// C++ standard library headers
```
Aksha1812: maybe these comments not needed

src/indexes/text/lexer.cc
Aksha1812: @VoletiRam you can use a concise version of below

```
// sb_stemmer_stem() returns a pointer to internal stemmer memory that is managed
// by the stemmer itself. The returned memory remains valid until the next call to
// sb_stemmer_stem() or until the stemmer is destroyed. We immediately copy this
// data into a std::string, so no manual memory management is required and there is
// no risk of memory leaks.
```

src/indexes/text.h
VoletiRam: I haven't touched the stemmer related code as it was from an approved PR. Can you own the stemmer related suggestions from this PR?

I also have other ideas over adding CHECKS in stemming related code. I would prefer index creation failures instead of crashing with checks. We need to think about multi-lingual support of stemmer too.

src/indexes/text/text_index.h
VoletiRam: This will just convert raw array of stings of punctuations, stop word etc to optimized versions of bitmap, hashsets.

I will add a comment and rename the function.

src/indexes/text/text_index.h
VoletiRam: This is related to stemming. Will leave this to you

src/index_schema.cc
VoletiRam: The text_index_schema_ object is only created when there's actually a TEXT field in the index schema. Though these are schema-level confifgs, we choose to store text related configs in text index schema. We do lazy initialization currently and store these configs in native members when there is TEXT field in ft.create. The current design ensure we save memory and avoid unecessary object creation/save/restore of text index schema if it is not in use.

To apply your suggestion, 
1. I need to move storage of language, punctuation, with_offsets to index_schema instead.
(or)
2. Create and maintain Text Index Schema object always. 

I am in favor of option 1

src/index_schema.cc
VoletiRam: Sure, will make this change.

src/indexes/text.h
Aksha1812: ok sure. these comments though are unrelated and are about including library headers. you can remove them

src/indexes/text/text_index.h
Aksha1812: Yes, I will add that when making any future changes to stemmer/ when we add multi language support.

src/indexes/text.h
VoletiRam: Yaa, I am not talking about this comment. Will address it.

For some reason, I am not getting reply option for other comments so added here

src/utils/scanner.h
VoletiRam: Haha. Ack, will rename. Thanks.

src/indexes/text.cc
VoletiRam: We can't use `VMSDK_ASSIGN_OR_RETURN` here because `return false` vs `return error_status` go to different counters:

- `return false` → `skipped_cnt` → `hash_indexing_failures` (data validation issues)
- `return error_status` → `failure_cnt` → system failures

I choose to return false because UTF validation is more of data issue than system failure

testing/text_test.cc
VoletiRam: Makes sense. will update.

src/indexes/text.h
allenss-amazon: You said: _I also have other ideas over adding CHECKS in stemming related code. I would prefer index creation failures instead of crashing with checks_ 

In general, programming errors should be detected with CHECK, not a failing return code. For 2 reasons:
1. You'll get more information about the failure, making it quicker/easier to fix.
2. It reduces upper layer complexity by not having to handle these error cases.

src/index_schema.cc
allenss-amazon: Option 1, yes.

src/indexes/text/lexer.cc
```diff
@@ -8,48 +8,91 @@
...
 
+// UTF-8 validation using Scanner
+bool Lexer::IsValidUtf8(absl::string_view text) const {
+  valkey_search::utils::Scanner scanner(text);
+
+  // Try to parse each UTF-8 character - Scanner counts invalid sequences
+  while (scanner.GetPosition() < text.size()) {
+    valkey_search::utils::Scanner::Char ch = scanner.NextUtf8();
+    if (ch == valkey_search::utils::Scanner::kEOF) {
+      break;
```
KarthikSubbarao, line 89: So, if EOF is reached before `scanner.GetPosition() < text.size()`, it can still be considered valid?

src/indexes/text.cc
allenss-amazon: Well, the upper level needs to distinguish the three potential results, right? Encoding this way is fine. It might make the upper layer a bit easier.

src/index_schema.cc
```diff
@@ -851,6 +860,14 @@ std::unique_ptr<data_model::IndexSchema> IndexSchema::ToProto() const {
   index_schema_proto->mutable_subscribed_key_prefixes()->Add(
       subscribed_key_prefixes_.begin(), subscribed_key_prefixes_.end());
   index_schema_proto->set_attribute_data_type(attribute_data_type_->ToProto());
+
+  // Get text configurations from TextIndexSchema for RDB serialization
+  if (this->text_index_schema_) {
+    index_schema_proto->set_language(this->text_index_schema_->GetLanguage());
+    index_schema_proto->set_punctuation(this->text_index_schema_->GetPunctuation());
+    index_schema_proto->set_with_offsets(this->text_index_schema_->GetWithOffsets());
```
KarthikSubbarao: Could we save stopwords now (I know we are not using them in the ingestion currently)? 

Or do we want to add it later for a particular reason?

src/indexes/text/lexer.cc
```diff
@@ -8,48 +8,91 @@
...
+           !punct_bitmap[static_cast<unsigned char>(text[pos])]) {
+      pos++;
+    }
+
+    if (pos > word_start) {
+      std::string word(text.data() + word_start, pos - word_start);
+
+      word = absl::AsciiStrToLower(word);
+
+      // TODO: Stop word removal
```
KarthikSubbarao, line 46: This would be helpful to add as a separate API. In the query path, we will not able to call the Tokenize function verbatim because there are nuanced query syntax around which we start and stop the tracking of a new word.

If there is an API to check if a word is a stopword, that will be great

src/indexes/text/lexer.cc
VoletiRam: Yes, if EOF is reached, the text can still be considered valid, because EOF simply indicates that all characters in the input have been successfully processed. The validity depends solely on scanner.GetInvalidUtf8Count() == 0.

Btw, The scenario you described cannot occur with the current implementation.

NextUtf8() only returns kEOF when pos_ >= sv_.size() (i.e., when we've reached or exceeded the end of the input string). The loop continues while scanner.GetPosition() < text.size(), where GetPosition() returns pos_ and text.size() equals sv_.size().  If NextUtf8() returns kEOF, it means pos_ >= sv_.size(), which means scanner.GetPosition() >= text.size(). This would make the loop condition scanner.GetPosition() < text.size() evaluate to false on the next iteration.

src/indexes/text/lexer.cc
```diff
@@ -8,48 +8,91 @@
...
+    while (pos < text.size() &&
+           punct_bitmap[static_cast<unsigned char>(text[pos])]) {
+      pos++;
+    }
+
+    size_t word_start = pos;
+    while (pos < text.size() &&
+           !punct_bitmap[static_cast<unsigned char>(text[pos])]) {
+      pos++;
+    }
```
KarthikSubbarao: This would be helpful to add as a separate API. In the query path, we will not able to call the Tokenize function verbatim because there are nuanced query syntax around which we start and stop the tracking of a new word.

If there is an API to check if a character should be considered as punctuation, that would be great

src/indexes/text/lexer.cc
```diff
@@ -8,48 +8,91 @@
...
+
+    if (pos > word_start) {
+      std::string word(text.data() + word_start, pos - word_start);
+
+      word = absl::AsciiStrToLower(word);
+
+      // TODO: Stop word removal
+
+      if (stemming_enabled && word.length() >= min_stem_size) {
```
KarthikSubbarao: Can we handle this logic (`stemming_enabled && word.length() >= min_stem_size`) inside the `StemWord` function? It will help the caller sites such as the query parsing logic

src/indexes/text/lexer.cc
KarthikSubbarao: +1

src/index_schema.cc
VoletiRam: There is no particular reason. I was planning to add in stop word removal PR where I can test it. Wanted to limit the changes scoped within the task.

src/indexes/text/lexer.cc
VoletiRam: Sure, will move.

src/indexes/text/text_index.cc
KarthikSubbarao: > this should be ensured at query level.but yes adding a CHECK here would be a good sanity check

The actually initialization of the stemmer can happen as a result of the FT.CREATE command. The language here can be specified (or we use the default english). Any specification other than english should be rejected.

However, we cannot assume that documents are all in english, If there are multi lingual documents, stemming will fail and we should not crash or expect that the snowball stemmer always succeeds

src/indexes/text/lexer.cc
VoletiRam: Okay. I can modularize into more specific component and will add API to check for punctuations and stop words.

src/indexes/text/lexer.cc
```diff
@@ -8,48 +8,91 @@
...
+  CHECK(stemmer) << "Stemmer not initialized";
+
   const sb_symbol* stemmed = sb_stemmer_stem(
-    stemmer_, 
+    stemmer,
     reinterpret_cast<const sb_symbol*>(word.c_str()),
     word.length()
   );
   
   DCHECK(stemmed) << "Stemming failed for word: " + word;
```
KarthikSubbarao, line 75: We cannot assume that documents are all in english, If there are multi lingual documents, stemming will fail and we should not crash or expect that the snowball stemmer always succeeds.

How does this handle non english words?

This is a note for the entire tokenization process (to safely work around non english documents)

src/indexes/text/lexer.cc
KarthikSubbarao: > Btw, The scenario you described cannot occur with the current implementation.

If the scenario should not occur, can the return condition be updated to include a validation that we are at the final position? 
```
 // If any invalid UTF-8 sequences were encountered, text is invalid
  return scanner.GetInvalidUtf8Count() == 0;
```

src/indexes/text/lexer.cc
VoletiRam: I am directing all the comments related to stemming to @Aksha1812

src/index_schema.cc
KarthikSubbarao: Ack, I was asking since FT.CREATE work is already done and we can provide stopwords during index creation currently. So, you could technically already test this out.

If FT.INFO work was done, you could already test out that after RDB save and load, the same stop words exist.

If we want to add it later, that is fine, but can we add a comment here for tracking?

src/indexes/text/lexer.cc
VoletiRam: Okay, will update this to:
```
return scanner.GetInvalidUtf8Count() == 0 && scanner.GetPosition() == text.size();
```

src/index_schema.cc
VoletiRam: Ack. I will add a comment.

src/indexes/text/text_index.h
VoletiRam: The casts are necessary but I can avoid intermediate casting from int to char to unsigned char. The lexer processes UTF-8 text from user input character-by-character. When users input international text like "café" or "naïve", the UTF-8 bytes for non-ASCII characters (128-255) will be negative values if char is signed (which is common).

KarthikSubbarao (approved):

PR #345, opened 2025-08-22
Implementation of FT.AGGREGATE Command

allenss-amazon:
Rebased version of aggregation, some changes post the implementation of non-vector commands.

Start of a test framework for comparing against Redisearch.


integration/compatibility/data_sets.py
```diff
@@ -0,0 +1,302 @@
...
+                    "t3": "all_the_same_value",
+                    "v1": array_encode(key_type, [i for _ in range(VECTOR_DIM)]),
+                    "e1" : 1,
+                    "e2" : "two",
+                },
+            )
+            for i in range(len(sortable_numbers))
+        ]
+        #
+        # Sortable numbers, designed so that sorted keys for this index don't have any duplications
```
Aksha1812, line 146: Don't we need to check for that case as well. Will you be adding more test cases for that?

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+  vmsdk::ArgsIterator itr{argv, argc};
+  std::string index_schema_name;
+  VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, index_schema_name));
+  VMSDK_ASSIGN_OR_RETURN(
+      auto index_schema,
+      SchemaManager::Instance().GetIndexSchema(ValkeyModule_GetSelectedDb(ctx),
+                                               index_schema_name));
+  RealIndexInterface index_interface(index_schema);
+  auto params = std::make_unique<AggregateParameters>(
+      options::GetDefaultTimeoutMs().GetValue(), &index_interface);
```
Aksha1812, line 117: Why do parameters need a timeout value?

src/commands/ft_aggregate_parser.cc
```diff
@@ -0,0 +1,337 @@
...
+        VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, cnt));
+        if (cnt == 0) {
+          return absl::OutOfRangeError("Groupby requires arguments");
+        }
+        for (auto i = 0; i < cnt; ++i) {
+          VMSDK_ASSIGN_OR_RETURN(auto group_string, itr.PopNext());
+          auto group_string_view = vmsdk::ToStringView(group_string);
+          if (group_string_view.empty() || group_string_view[0] != '@') {
+            return absl::InvalidArgumentError(
+                absl::StrCat("Group field reference must start with '@'"));
```
Aksha1812, line 204: just a nit. Maybe "Invalid Group field name reference" or "Group field reference empty" might be more appropriate here

src/commands/ft_aggregate_parser.cc
```diff
@@ -0,0 +1,337 @@
...
+              auto group, parameters.MakeReference(group_string_view, false));
+          groupby->groups_.emplace_back(std::unique_ptr<Attribute>(
+              dynamic_cast<Attribute *>(group.release())));
+        }
+        while (itr.PopIfNextIgnoreCase(kReduceParam)) {
+          GroupBy::Reducer r;
+          VMSDK_ASSIGN_OR_RETURN(auto name, itr.PopNext(),
+                                 _ << "Missing Reducer name");
+          auto uc_name =
+              expr::FuncUpper(expr::Value(vmsdk::ToStringView(name)));
```
Aksha1812, line 217: No case conversion required for rest of parsing?

src/commands/ft_aggregate_exec.cc
```diff
@@ -0,0 +1,325 @@
...
+  size_t record_field_count = 0;
+  while (!records.empty()) {
+    auto record = records.pop_front();
+    if (record_field_count == 0) {
+      record_field_count = record->fields_.size();
+    } else {
+      CHECK(record_field_count == record->fields_.size());
+    }
+    GroupKey k;
+    // todo: How do we handle keys that have a missing attribute in the key??
```
Aksha1812, line 169: What are we currently doing. Maybe we can check this on redis and add a compatibility test to confirm

src/commands/ft.aggregate.json
```diff
@@ -0,0 +1,26 @@
...
+        "type": "key"
+      },
+      {
+        "key_spec_index": 1,
+        "name": "query",
+        "type": "string"
+      }
+    ],
+    "arity": -3,
+    "complexity": "O(log N)",
```
Aksha1812, line 21: Is this average complexity for all operations? How did we get to this value.

testing/ft_aggregate_exec_test.cc
```diff
@@ -0,0 +1,283 @@
...
+  }
+}
+
+TEST_F(AggregateExecTest, ReducerTest) {
+  struct Testcase {
+    std::string text_;
+    size_t m;
+    std::vector<double> values_;
+  };
+  Testcase testcases[]{
```
Aksha1812, line 240: AVG test case missing. Also I see more reducers in redis. Are those scheduled to be implemented in later versions ?

integration/compatibility/generate.py
```diff
@@ -0,0 +1,401 @@
...
+                        "*",
+                        "load",
+                        "2",
+                        "__key",
+                        "t2",
+                        "apply",
+                        f"({lop}){op}({rop})",
+                        "as",
+                        "nn",
+                )
```
Aksha1812, line 401: Maybe we can add few more tests if they go here. there are multiple test files so not sure if these are added somewhere else. also not sure how holistic are above tests
- Filter tests
- queries with multiple stages
- also looks like none of the above tests test error conditions.
- params tests
- some reducer tests seem to be missing as well
- also do we support aggregation like groupby on vector fields, i see aggregation on only t1 which is tag field right?

src/commands/ft_search.h
```diff
@@ -32,6 +32,9 @@ struct Result {
 int Reply(ValkeyModuleCtx *ctx, [[maybe_unused]] ValkeyModuleString **argv,
           [[maybe_unused]] int argc);
 
+int Timeout(ValkeyModuleCtx *ctx, [[maybe_unused]] ValkeyModuleString **argv,
+            [[maybe_unused]] int argc);
```
Aksha1812, line 36: just curious what is this maybe_unused?

src/commands/ft_search_parser.cc
```diff
@@ -186,7 +182,7 @@ absl::StatusOr<FilterParseResults> ParsePreFilter(
 absl::Status ParseKNN(query::VectorSearchParameters &parameters,
                       absl::string_view filter_str) {
   if (filter_str.empty()) {
-    return absl::InvalidArgumentError("Vector query clause is missing");
+    return absl::InvalidArgumentError("Vector query clause is missing1");
```
Aksha1812, line 185: typo?

src/commands/ft_search_parser.cc
```diff
@@ -350,9 +346,24 @@ vmsdk::KeyValueParser<query::VectorSearchParameters> CreateSearchParser() {
...
     CreateSearchParser();
 
-absl::Status ParseQueryString(query::VectorSearchParameters &parameters) {
+}  // namespace
+
+//
+// We don't have values for the $ substitution yet. so we break the parsing into
+// two pieces
+//
+absl::Status PreParseQueryString(query::VectorSearchParameters &parameters) {
```
Aksha1812, line 355: Why is this just for VectorSearch parameters. shouldn't we validate query string regardless of what kind of search?

src/utils/scanner.h
```diff
@@ -0,0 +1,236 @@
+/*
```
Aksha1812, line 1: This is the same file, which Ram had added for Lexical PR right? or are there any changes. also adding it in this PR, is it intentional or a mistake? this will also change commit history for this file

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+      absl::string_view identifier) const override {
+    return schema_->GetAlias(identifier);
+  }
+  RealIndexInterface(std::shared_ptr<IndexSchema> schema) : schema_(schema) {}
+};
+
+absl::Status ManipulateReturnsClause(AggregateParameters &params) {
+  // Figure out what fields actually need to be returned by the aggregation
+  // operation. And modify the common search returns list accordingly
+  DBG << "Manipulating returns clause for: " << params.index_schema_name
```
yairgott, line 54: why not `VMSDK_LOG(DEBUG, nullptr)`?

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+        DBG << " *skipped*";
+        params.load_key = true;
+        continue;
+      }
+      if (load == vmsdk::ToStringView(params.score_as.get())) {
+        DBG << " *skipping score*";
+        continue;
+      }
+      VMSDK_ASSIGN_OR_RETURN(auto indexer, params.index_schema->GetIndex(load));
+      auto field_type = indexer->GetIndexerType();
```
yairgott, line 80: Following the existing naming, this should be `indexer_type`

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+  return std::move(params);
+}
+
+bool ReplyWithValue(ValkeyModuleCtx *ctx,
+                    data_model::AttributeDataType data_type,
+                    std::string_view name, indexes::IndexerType field_type,
+                    const expr::Value &value, int dialect) {
+  if (value.IsNil()) {
+    return false;
+  } else {
```
yairgott, line 163: Here and other places, avoid unnecessary `else` indentations when `if` returns.

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+  RealIndexInterface(std::shared_ptr<IndexSchema> schema) : schema_(schema) {}
+};
+
+absl::Status ManipulateReturnsClause(AggregateParameters &params) {
+  // Figure out what fields actually need to be returned by the aggregation
+  // operation. And modify the common search returns list accordingly
+  DBG << "Manipulating returns clause for: " << params.index_schema_name
+      << "\n";
+  CHECK(!params.no_content);
+  if (params.loadall_) {
```
yairgott, line 57: Here and other places: prefer early return and avoid unnecessary indentations.

src/commands/ft_aggregate.cc
yairgott: +1

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+            auto double_size = snprintf(double_storage, sizeof(double_storage),
+                                        "%.11g", *dble);
+            value_view = std::string_view(double_storage, double_size);
+            DBG << "JSON kNumeric:" << value_view << "\n";
+            break;
+          }
+          default:
+            DBG << "Unsupported field type for reply: " << int(field_type)
+                << "\n";
+            assert("Unsupported field type" == nullptr);
```
yairgott, line 199: Here and in other places where `assert` is used, prefer `CHECK`

src/expr/value.h
```diff
@@ -0,0 +1,186 @@
...
+#include <optional>
+#include <variant>
+
+#include "absl/container/inlined_vector.h"
+#include "absl/strings/string_view.h"
+
+namespace valkey_search {
+namespace expr {
+
+class Value {
```
yairgott, line 20: I wonder why can't we extend the existing type conversion [methods](https://github.com/valkey-io/valkey-search/blob/13fa8d4f8d1f17cb59c517ee3acf64c79a874c8b/vmsdk/src/type_conversions.h).

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+          case indexes::IndexerType::kTag:
+          case indexes::IndexerType::kNone: {
+            value_view = value.AsStringView();
+            DBG << "JSON kTag: " << value_view << "\n";
+            break;
+          }
+          case indexes::IndexerType::kNumeric: {
+            auto dble = value.AsDouble();
+            if (!dble) {
+              return false;
```
yairgott, line 188: should we return false here or return the value as string?

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+                << "\n";
+            assert("Unsupported field type" == nullptr);
+        }
+      }
+      ValkeyModule_ReplyWithSimpleString(ctx, name.data());
+      if (dialect == 2) {
+        ValkeyModule_ReplyWithStringBuffer(ctx, value_view.data(),
+                                           value_view.size());
+      } else {
+        std::string s;
```
yairgott, line 207: nit: better to use `absl::StrCat` in this case

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+      }
+    }
+  }
+  return true;
+}
+
+absl::Status SendReplyInner(ValkeyModuleCtx *ctx,
+                            std::deque<indexes::Neighbor> &neighbors,
+                            AggregateParameters &parameters) {
+  auto identifier =
```
yairgott, line 222: Better to verify the existing of the identifier in the query parsing phase.

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+                                              indexes::IndexerType::kNone);
+  }
+  if (parameters.IsVectorQuery()) {
+    auto score_sv = vmsdk::ToStringView(parameters.score_as.get());
+    scores_index = parameters.AddRecordAttribute(score_sv, score_sv,
+                                                 indexes::IndexerType::kNone);
+  }
+
+  //
+  //  1. Process the collected Neighbors into Aggregate Records.
```
yairgott, line 245: Each phase should have its own dedicated function.

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+  //
+  auto data_type = parameters.index_schema->GetAttributeDataType().ToProto();
+  RecordSet records(&parameters);
+  // Todo: fix this  for (auto &n : neighbors) {
```
yairgott, line 249: The comment seems redundant

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+  for (auto &n : neighbors) {
+    auto rec =
+        std::make_unique<Record>(parameters.record_indexes_by_alias_.size());
+    DBG << "Neighbor: " << n << " Empty Record:" << *rec << "\n";
+    if (parameters.load_key) {
+      rec->fields_.at(key_index) = expr::Value(n.external_id.get()->Str());
+    }
+    if (/* todo: parameters.addscores_ */ true) {
```
yairgott, line 257: Is the todo required?

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+      rec->fields_.at(scores_index) = expr::Value(n.distance);
+    }
+    // For the fields that were fetched, stash them into the RecordSet
+    if (n.attribute_contents.has_value() && !parameters.no_content) {
+      for (auto &[name, records_map_value] : *n.attribute_contents) {
+        auto value = vmsdk::ToStringView(records_map_value.value.get());
+        size_t record_index;
```
yairgott, line 264: It would be cleaner if `record_index` is `std::optional<size_t>`  and eliminate `found_index`

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+        bool found_index = false;
+        if (auto by_alias = parameters.record_indexes_by_alias_.find(name);
```
yairgott, line 266: As optimization to avoid very frequent hash map lookups, we can use an axillary vector which contains optionals of `record_index`. This vector will be initialized once outside of this loops.

src/commands/ft_aggregate.cc
```diff
@@ -0,0 +1,445 @@
...
+
+absl::Status FTAggregateCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
+                            int argc) {
+  auto status = [&]() {
+    auto &schema_manager = SchemaManager::Instance();
+    VMSDK_ASSIGN_OR_RETURN(
+        auto parameters,
+        aggregate::ParseCommand(ctx, argv + 1, argc - 1, schema_manager));
+    parameters->index_schema->ProcessMultiQueue();
+    bool inside_multi =
```
yairgott, line 399: Below code seems mostly copy & paste from [ft_search](https://github.com/valkey-io/valkey-search/blob/f683fffb3ff822a38967aca496fe1f4328369bd7/src/commands/ft_search.cc#L270-L307). Consider make adjustments to allow code reuse.

yairgott (commented):
First pass until ft_aggregate.cc

src/expr/value.h
allenss-amazon: It's not about converting types. It's a dynamically typed value. There will be additional future types, like ARRAY and VECTOR with more functions. Some of the Redisearch functionality requires ARRAY and I've already seen usecases where VECTOR functionality would be useful.

src/commands/ft_aggregate.cc
allenss-amazon: This isn't the value. This indicates that the value couldn't be converted. Which means that it won't be included in the output. In this case, there is an attribute declared as numeric, but the key contains a field that can't be converted to a number. Ergo, the field is not included in the output. I believe this is how Redisearch handles this case.

src/commands/ft_aggregate.cc
allenss-amazon: removed

src/commands/ft_aggregate.cc
allenss-amazon: Changed to IsVectorQuery. This will need updating when non-vector queries get scored.

src/commands/ft_aggregate.cc
allenss-amazon: Agreed. changed

src/commands/ft_aggregate.cc
allenss-amazon: No you can't. This is inherent in the RecordsMap data structure. The data structure is self defining and requires name lookups to be hash lookups for each instance of the RecordsMap. Aggregation uses a more complex, but more efficient data structure where indexes into the Records data structure are pre-computed at compile time and only require vector indexing after that. Here I'm converting from RecordsMap to a Record. I have two choices for loops. 

1. I can iterator over the RecordsMap fields and then do a hash lookup in the Records indexing structure. Or, 
2. I can iterate over the Records indexing structure, doing a hash lookup in RecordsMap. But method 2 must also handle the fields in RecordsMap that aren't in the Records indexing data structure. Which isn't easy to do efficiently.

I implemented method 1 because it's much easier to do.

PR #346, opened 2025-08-23
Implement Pause point

allenss-amazon:
A test facility to controllable pause the execution of background threads.

See description in ft_debug.cc

### DIFF
--- a/src/commands/ft_debug.cc
+++ b/src/commands/ft_debug.cc
@@ -70,7 +70,7 @@ absl::Status PausePointControlCmd(ValkeyModuleCtx *ctx, vmsdk::ArgsIterator& itr
 
 absl::Status FTDebugCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
                         int argc) {
-  if (false /* !vmsdk::config::IsDebugModeEnabled() */) {
+  if (!vmsdk::config::IsDebugModeEnabled()) {
     // Pretend like we don't exist
     std::ostringstream msg;
     msg << "ERR unknown command '" << vmsdk::ToStringView(argv[0]) << "', with args beginning with:";
--- a/src/commands/ft_debug.cc
+++ b/src/commands/ft_debug.cc
@@ -6,37 +6,98 @@
  */
 
 #include <absl/base/no_destructor.h>
+#include <absl/strings/ascii.h>
 
-#include "info.h"
+#include "module_config.h"
 #include "src/commands/commands.h"
+#include "vmsdk/src/debug.h"
 #include "vmsdk/src/command_parser.h"
+#include "vmsdk/src/info.h"
+#include "vmsdk/src/log.h"
 #include "vmsdk/src/status/status_macros.h"
 
 extern vmsdk::module::Options options;  // Declared in module_loader.cc
 namespace valkey_search {
 
-enum SubCommands {
-  kShowInfo,
-};
-
-const absl::flat_hash_map<absl::string_view, SubCommands> kDebugSubcommands({
-    {"SHOW_INFO", SubCommands::kShowInfo},
-});
+//
+// FT.DEBUG PAUSEPOINT [ SET | RESET | TEST | LIST] <pausepoint>
+//
+// Connects to the vmsdk::debug mechanism
+// A pausepoint is a mechanism to pause a thread at a specific location for testing purposes.
+//
+// Individual pausepoints are labelled by a unique string. No checking for the uniqueness is done.
+// A pause point in the code is enabled by calling vmsdk::debug::PausePoint("<string>");
+// If that pausepoint is not set, then this call does nothing. But if the pausepoint is set then
+// the calling thread "hangs" at that point.
+//
+// The TEST option can be used to determine if one or more threads are paused at a particular pausepoint,
+// the number of paused threads is returned by that command, with 0 indicating that no threads are paused.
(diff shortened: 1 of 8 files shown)


vmsdk/src/debug.cc
```diff
@@ -0,0 +1,77 @@
...
+#include "vmsdk/src/debug.h"
+
+namespace vmsdk {
+namespace debug {
+
+absl::Mutex pause_point_lock;
+absl::flat_hash_map<std::string, size_t> pause_point_waiters;
+
+void PausePoint(absl::string_view point) {
+  CHECK(!IsMainThread()) << "Pause point not allowed on main thread.";
```
eifrah-aws, line 42: Do we want to control the pause points based on configuration entry? something like:

```c++
if (!options::GetPausePointsEnabled()) {
  return;
}
```

vmsdk/src/debug.cc
```diff
@@ -0,0 +1,77 @@
...
+  {
+    absl::MutexLock lock(&pause_point_lock);
+    auto it = pause_point_waiters.find(point);
+    if (it == pause_point_waiters.end()) {
+      return;
+    }
+    it->second++; // Indicate that I'm waiting.
```
eifrah-aws: Maybe we can add some more context to the pause point? like some kind of string. Could be handy when we want to query a pause point, also, I would also record timestamp that a thread hit the pause point

vmsdk/src/debug.cc
```diff
@@ -0,0 +1,77 @@
...
+#include "utils.h"
+#include "vmsdk/src/log.h"
+
+#include "vmsdk/src/debug.h"
+
+namespace vmsdk {
+namespace debug {
+
+absl::Mutex pause_point_lock;
+absl::flat_hash_map<std::string, size_t> pause_point_waiters;
```
eifrah-aws: Can we use the following instead of `absl::flat_hash_map<std::string, size_t>` ? 

```c++
struct Waiter {
  std::string name;     // the thread name, e.g. "worker-read-1"
  int64_t timestamp{0}; // the time the pause point was hit for this thread
};
```

```c++
absl::flat_hash_map<std::string, std::vector<Waiter>> pause_point_waiters;
```

src/commands/ft_debug.cc
```diff
@@ -6,37 +6,74 @@
...
 
-enum SubCommands {
-  kShowInfo,
-};
-
-const absl::flat_hash_map<absl::string_view, SubCommands> kDebugSubcommands({
-    {"SHOW_INFO", SubCommands::kShowInfo},
-});
+//
+// FT.DEBUG PAUSEPOINT [ SET | RESET | TEST ] <pausepoint>
```
eifrah-aws: We could also add:

```c++
FT.DEBUG PAUSEPOINT LIST
```

this will generate list of pause points and who is waiting on them.

vmsdk/src/debug.cc
```diff
@@ -0,0 +1,77 @@
...
+    {
+      absl::MutexLock lock(&pause_point_lock);
+      auto it = pause_point_waiters.find(point);
+      if (it == pause_point_waiters.end()) {
+        VMSDK_LOG(WARNING, nullptr) << "End of waiting at pause point " << point;
+        return;
+      }
+    }
+    if (absl::Now() > message_time) {
+      VMSDK_IO_LOG_EVERY_N_SEC(WARNING, nullptr, 10) << "Waiting > 10 seconds at pause point " << point;
```
eifrah-aws: Maybe include the pause point name?

vmsdk/src/debug.cc
```diff
@@ -0,0 +1,77 @@
...
+    }
+    it->second++; // Indicate that I'm waiting.
+    VMSDK_LOG(WARNING, nullptr) << "Waiting at pause point " << point;
+  }
+  auto message_time = absl::Now() + absl::Seconds(10);
+  while (true) {
+    absl::SleepFor(absl::Milliseconds(50));
+    {
+      absl::MutexLock lock(&pause_point_lock);
+      auto it = pause_point_waiters.find(point);
```
eifrah-aws: nit:

```c++
if (!pause_point_waiters.contains(point)) {
  ...
}
```

vmsdk/src/debug.cc
allenss-amazon: I couldn't figure out how to set debug-mode in the integration environment, so I couldn't use it. Once we tackle that problem, I'll convert this.

vmsdk/src/debug.cc
allenss-amazon: added file/line & timestamp.

vmsdk/src/debug.cc
allenss-amazon: done

src/commands/ft_debug.cc
allenss-amazon: done

vmsdk/src/debug.cc
allenss-amazon: done.

vmsdk/src/debug.cc
allenss-amazon: fixed.

vmsdk/src/debug.cc
eifrah-aws: I wrote you in private :)

PR #349, opened 2025-08-25
Merge "main" into "fulltext" and apply clang formatting

BCathcart:
I merged main into the fulltext branch and then applied clang formatting to ensure any previously unformatted code specific to the fulltext branch is now formatted.


BCathcart:
I'm not sure why there are DCO failures for those commits coming in from main. It looks like they've all been signed off on.

PR #353, opened 2025-08-27
Implement stop words removal in ingestion pipeline

VoletiRam:
Implement stop words removal in ingestion pipeline. Add integration tests for ingestion pipeline.


integration/test_fulltext.py
```diff
@@ -200,3 +267,81 @@ def test_text_search_per_field(self):
...
+        """
+        Test comprehensive ingestion pipeline: FT.CREATE → HSET → FT.SEARCH with full tokenization
+        """
+        client: Valkey = self.server.get_new_client()
+        client.execute_command(ingestion_pipeline_index)
+        client.execute_command(*ingestion_pipeline_doc)
+        
+        # Punctuation tokenization (using prefix to handle stemming)
+        for query in ingestion_punctuation_queries:
+            assert client.execute_command(*query)[0] == ingestion_match_expected
```
allenss-amazon: This is only assuring that 1 match is found. What if it's the wrong match? It's tedious, but I think you should add the "right" answer to the test case.

src/indexes/text/lexer.h
```diff
@@ -19,13 +19,14 @@ and Text classes, then passed to lexer methods as parameters.
...
 2. Convert to lowercase 
-3. TODO: stop word removal (not implemented in this phase)
+3. Stop word removal (filter out common words)
 4. Apply stemming based on language and field settings
 
 */
 
 #include <bitset>
 #include <string>
+#include <unordered_set>
```
allenss-amazon: we should be using the abseil flat_hash_map, it's considerably better at the sacrifice of functionality that we don't care about.

src/indexes/text/lexer.h
```diff
@@ -41,14 +42,21 @@ struct Lexer {
...
 
   // Punctuation checking API
   static bool IsPunctuation(char c, const std::bitset<256>& punct_bitmap) {
     return punct_bitmap[static_cast<unsigned char>(c)];
   }
 
+  // Stop word checking API (expects lowercase input)
+  static bool IsStopWord(const std::string& lowercase_word, 
+                         const std::unordered_set<std::string>& stop_words_set) {
+    return stop_words_set.find(lowercase_word) != stop_words_set.end();
```
allenss-amazon: contains.

src/indexes/text/text_index.h
```diff
@@ -88,6 +89,9 @@ struct TextIndexSchema {
   // Optimized structures (shared across all text fields)
   PunctuationBitmap punct_bitmap_;
 
+  // Stop words set for filtering during tokenization
+  std::unordered_set<std::string> stop_words_set_;
```
allenss-amazon: absl::flat_hash_map

testing/lexer_test.cc
```diff
@@ -108,24 +111,87 @@ INSTANTIATE_TEST_SUITE_P(
...
+  EXPECT_FALSE(Lexer::IsStopWord("test", stop_words));
+  
+  // Test edge cases
+  EXPECT_FALSE(Lexer::IsStopWord("", stop_words));
+  EXPECT_FALSE(Lexer::IsStopWord("THE", stop_words)); // Should be lowercase input
+  EXPECT_FALSE(Lexer::IsStopWord("And", stop_words)); // Should be lowercase input
+  
+  // Test words that contain stop words but are not stop words
+  EXPECT_FALSE(Lexer::IsStopWord("theorist", stop_words));
+  EXPECT_FALSE(Lexer::IsStopWord("androgynous", stop_words));
```
allenss-amazon: I don't get to say this very often, BUT, This is a BIT overboard on testing. We're testing that we lower-cased it before we put it into a hash map?

testing/lexer_test.cc
```diff
@@ -108,24 +111,87 @@ INSTANTIATE_TEST_SUITE_P(
...
+      "the cat and dog",
+      no_stop_schema->GetPunctuationBitmap(),
+      no_stop_schema->GetStemmer(),
+      true,
+      3,
+      empty_set
+  );
+  
+  ASSERT_TRUE(result.ok());
+  EXPECT_EQ(*result, std::vector<std::string>({"the", "cat", "and", "dog"}));
```
allenss-amazon: This seems like a better test, having a few different phrases that get tokenized, including oddball stuff like punctuation, etc.

testing/text_index_schema_test.cc
```diff
@@ -51,21 +51,51 @@ TEST_F(TextIndexSchemaTest, FieldAllocationAcrossMultipleTexts) {
...
+  auto stemmer1 = custom_schema->GetStemmer();
+  EXPECT_NE(nullptr, stemmer1);
   
-  auto key = StringInternStore::Intern("test_key");
-  auto result = field.AddRecord(key, "hello world");
+  auto stemmer2 = custom_schema->GetStemmer();
+  EXPECT_EQ(stemmer1, stemmer2); // Same instance reused
   
-  EXPECT_TRUE(result.ok());
-  EXPECT_TRUE(*result);
```
allenss-amazon: again, there's not much value in testing that a few major structures got setup. Almost any testing at a slightly higher level will guarantee that.

allenss-amazon (changes requested):

integration/test_fulltext.py
VoletiRam: Sure, will add the assertion for matching the document. Thank you.

testing/lexer_test.cc
VoletiRam: Ack. This is overlooked from vibe coding. Will remove this test.

testing/text_index_schema_test.cc
VoletiRam: Okay, I understand now. I will emphasize more on business logic testing. Will avoid such shallow/obvious testing from next time. Thank you for the feedback.

integration/test_fulltext.py
```diff
@@ -73,6 +73,82 @@
...
+
+ingestion_stopword_queries = [
+    ["FT.SEARCH", "idx", '@content:"the"'],
+    ["FT.SEARCH", "idx", '@content:"are"'],
+    ["FT.SEARCH", "idx", '@content:"but"']
+]
+
+ingestion_case_queries = [
+    ["FT.SEARCH", "idx", '@content:"quick*"'],
+    ["FT.SEARCH", "idx", '@content:"effect*"']
```
allenss-amazon: why isn't this a duplicate of the punctuation queries? The label suggests that it's somehow testing the "case" but I don't see it....

integration/test_fulltext.py
```diff
@@ -200,3 +276,102 @@ def test_text_search_per_field(self):
...
+            result = client.execute_command(*query)
+            assert result[0] == ingestion_match_expected
+            assert result[1] == ingestion_expected_key
+            assert result[2] == ingestion_expected_fields
+        
+        # Non-existent terms
+        for query in ingestion_nomatch_queries:
+            result = client.execute_command(*query)
+            assert result[0] == ingestion_nomatch_expected
+
```
allenss-amazon: This structure is very difficult to follow. As far as I can tell, you have several lists of queries and you're simply testing whether the query finds the one and only key, right? How about making a list of those queries with a pass/fail and then just running those in a loop with a single pass/fail. The extra verbiage confuses rather than enlightends.

integration/test_fulltext.py
```diff
@@ -200,3 +276,102 @@ def test_text_search_per_field(self):
...
+        """
+        Test that per-index PUNCTUATION actually affects search results
+        """
+        client: Valkey = self.server.get_new_client()
+        client.execute_command(punctuation_test_index)
+        client.execute_command(*punctuation_test_doc)
+        result = client.execute_command(*punctuation_test_query)
+        assert result[0] == ingestion_match_expected  # Dot separator worked
+        assert result[1] == ingestion_expected_key
+        assert result[2] == punctuation_expected_fields
```
allenss-amazon: Again very difficult to follow, separating the data from the flow makes it idfficult. Why not just put the constants here?

VoletiRam: I was following existing pattern that we used in search tests. Wanted to be coherent with existing structure. Will move constants inside function instead.

integration/test_fulltext.py
VoletiRam: Sure, will do.

allenss-amazon (approved):

integration/test_fulltext.py
```diff
@@ -200,3 +200,102 @@ def test_text_search_per_field(self):
...
+        }
+
+        result = client.execute_command("FT.SEARCH", "idx", '@title:"run"')
+        actual_fields = dict(zip(result[2][::2], result[2][1::2]))
+        assert actual_fields == expected_value
+
+        result = client.execute_command("FT.SEARCH", "idx", '@content:"run"')
+        assert result[0] == 0  # Should not find (NOSTEM)
+
+    def test_custom_stopwords(self):
```
KarthikSubbarao, line 252: Can this test also validate the case where no stop words are provided?

integration/test_fulltext.py
```diff
@@ -200,3 +200,102 @@ def test_text_search_per_field(self):
...
+        
+        # non stop words should be findable
+        result = client.execute_command("FT.SEARCH", "idx", '@content:"are"')
+        assert result[0] == 1  # Regular word indexed
+        assert result[1] == b'doc:1'
+        assert result[2] == [b'content', b"the cat and dog are good"]
+
+    def test_nostem(self):
+        """
+        End-to-end test: FT.CREATE NOSTEM config actually affects stemming in search
```
KarthikSubbarao, line 273: > FT.CREATE NOSTEM config actually affects stemming in search

This only works right now because there is no query processing in the query string :D (even when stemming is enabled on the index side)

integration/test_fulltext.py
```diff
@@ -200,3 +200,102 @@ def test_text_search_per_field(self):
...
+        client.execute_command("FT.CREATE idx ON HASH NOSTEM SCHEMA content TEXT")
+        client.execute_command("HSET", "doc:1", "content", "running quickly")
+        
+        # With NOSTEM, exact forms should be findable
+        result = client.execute_command("FT.SEARCH", "idx", '@content:"running"')
+        assert result[0] == 1  # Exact form "running" found
+        assert result[1] == b'doc:1'
+        assert result[2] == [b'content', b"running quickly"]
+
+    def test_custom_punctuation(self):
```
KarthikSubbarao, line 285: We should be able to validate an RDB save and load now

Issue #139, opened 2025-05-22
FT.SEARCH fan-out to primaries on non-READONLY connections


ChristophKoerner:
Today, FT.SEARCH queries are routed to a random node of each shard in cluster-mode. In a 1 primary / 1 replica configuration, a search query sent to the primary will be routed to its replica 50% of the time. While this routing logic maximizes query throughput, it makes it difficult to reason about the query execution and consistency of results. In addition, it prevents us from improving read consistency in the future - for example to allow read-after-write consistency mention by the changes in https://github.com/valkey-io/valkey-search/issues/112.

Even though Search's command routing and fan-out mechanism is different from existing command routing, it would be useful to stay closer to Valkey's command routing semantics to provide better expectation about the execution of search commands. As discussed in yesterday's sync, we could fan-out search queries to only primaries by default, and use the `READONLY` connection flag to fan-out queries to also replica nodes. This improves predictability of query executions while maintaining ability to optimize for read throughput when needed via `READONLY` flag

This improvement would work well with https://github.com/valkey-io/valkey-search/issues/114, resulting in query locality for non-READONLY connections.

PR #356, opened 2025-08-29
Implement Fanout Test, including correct logic for READONLY in Valkey 9 (Fix for #139)

allenss-amazon:
In Valkey 8, a module cannot detect whether the client is READONLY or READWRITE. A fix was merged into Valkey 9 unstable that adds this bit to the VM_GetClientInfoById interface, making it available to modules.

With this PR the fanout logic for FT.SEARCH is modified as follows:

When paired to Valkey servers less than version 9 (i.e., 8.x), fanout is done randomly to all nodes in the cluster. This is the same behavior as with the Search 1.0 release (which was on Valkey 8).

For Valkey servers of version 9 and greater, fanout is done according to the clients readonly/readwrite status.

For readonly clients, fanout is done the same as in 1.0, i.e., randomly across all nodes in the cluster.

For readwrite clients, fanout is done ONLY to the primaries.

This fixes issue #139.

This PR has been manually tested with both Valkey 8.1.1 and the Valkey / Unstable branch as of this date.

### DIFF
--- a/integration/test_vss_basic.py
+++ b/integration/test_vss_basic.py
@@ -18,8 +18,15 @@ class TestVSSBasic(ValkeySearchTestCaseBase):
         module_list_count = len(module_list_data)
         # We expect JSON & Search to be loaded
         assert module_list_count == 2
-        assert b"search" in module_list_data
-        assert b"json" in module_list_data
+        module_loaded = False
+        json_loaded = False
+        for module in module_list_data:
+            if module[b"name"] == b"search":
+                module_loaded = True
+            elif module[b"name"] == b"json":
+                json_loaded = True
+        assert module_loaded
+
 
 class TestVSSClusterBasic(ValkeySearchClusterTestCase):
 
@@ -29,7 +36,5 @@ class TestVSSClusterBasic(ValkeySearchClusterTestCase):
         module_list_count = len(module_list_data)
         # We expect JSON & Search to be loaded
         assert module_list_count == 2
-        assert b"search" in module_list_data
-        assert b"json" in module_list_data
         cluster_client: ValkeyCluster = self.new_cluster_client()
         assert cluster_client.set("hello", "world") == True


scripts/common.rc
```diff
@@ -167,11 +167,12 @@ function setup_valkey_server() {
 
   # Clone and build it
   VALKEY_VERSION="8.1.1"
```
zvi-code, line 169: why are we using 8.1.1? Shouldn't we use unstable or 9.0 rc1?

src/commands/ft_search.cc
```diff
@@ -243,7 +243,26 @@ void Free([[maybe_unused]] ValkeyModuleCtx *ctx, void *privdata) {
...
 
-vmsdk::config::Boolean ForceReplicasOnly("test-force-replicas-only", false);
+vmsdk::config::Boolean ForceReplicasOnly("debug-force-replicas-only", false);
+
+auto ComputeQueryFanoutMode(ValkeyModuleCtx *ctx) {
+  if (ForceReplicasOnly.GetValue()) {
+    // Testing only
+    return query::fanout::FanoutTargetMode::kReplicasOnly;
+  }
+  if (ValkeyModule_GetServerVersion() < 0x90000) {
```
zvi-code, line 253: Why not take more specific approach, by just having the default behavior map to flags value of zero?

integration/test_fanout.py
```diff
@@ -0,0 +1,81 @@
...
+            r_result = replica.execute_command(*search_command(index.name))
+            assert(len(r_result) > 1)
+        readonly_primary_count = sum_of_remote_searches(self.get_primaries()) - readwrite_primary_count
+        readonly_replica_count = sum_of_remote_searches(self.get_replicas()) - readwrite_replica_count
+
+        version_string = primary.info("SERVER")['valkey_version']
+        major_version = int(version_string.split('.')[0])
+        if major_version >= 9:
+            # Valkey 9+ honors the readonly flag in the client context
+            assert(readwrite_primary_count != 0)
```
zvi-code, line 72: should be `readwrite_primary_count == 100` ?

integration/test_fanout.py
```diff
@@ -0,0 +1,81 @@
...
+            assert(readwrite_replica_count == 0)
+            assert(readonly_primary_count != 0)
+            assert(readonly_replica_count != 0)
```
zvi-code, line 75: should also add `readonly_primary_count+ readonly_replica_count = 100`

integration/test_fanout.py
```diff
@@ -0,0 +1,81 @@
...
+        #
+        for _ in range(100):
+            r_result = replica.execute_command(*search_command(index.name))
+            assert(len(r_result) > 1)
+        readonly_primary_count = sum_of_remote_searches(self.get_primaries()) - readwrite_primary_count
+        readonly_replica_count = sum_of_remote_searches(self.get_replicas()) - readwrite_replica_count
+
+        version_string = primary.info("SERVER")['valkey_version']
+        major_version = int(version_string.split('.')[0])
+        if major_version >= 9:
```
zvi-code, line 70: basically the functionality is not tested because we are not running with version 9

PR #358, opened 2025-08-30
Implement Controlled Variables and TEST_COUNTER macros

allenss-amazon:
More debugging infrastructure.

The CONTROLLED_VARIABLE facility is created. Controlled variables are similar to CONFIG variables except:

1. They are ephemeral, not preserved.
2. They are node-local, i.e., not replicated.
3. They are controlled through the FT._DEBUG command -- which is only available in --debug-mode.
4. They are not required to be stable across releases.
5. They won't appear in documentation.

The purpose is to alter the behavior of Search in controlled ways, i.e., to force non-standard behavior. This is a key issue in good end-to-end testing. A typical use-case can be seen in the cancellation logic, where a pair of controlled variables alter the behavior of the cancellation logic to allow tests to stimulate the timeout paths in ways that are independent of the timing, meaning that these tests are not timing dependent, eliminating a major source of flaky tests.

Accompanying the controlled variables is the TEST_COUNTER macro. This macro is short hand for declaring a counter in the code that is only used in testing. Counters are available in the INFO section, but again only in --debug-mode.

These two facilities will allow end-to-end complex multi-thread testing to be done in a timing independent way while automatically ensuring that these facilities are not visible to normal users.

### DIFF
--- a/testing/vector_test.cc
+++ b/testing/vector_test.cc
@@ -56,7 +56,7 @@ const absl::flat_hash_map<data_model::DistanceMetric, std::string>
 };
 
 static cancel::Token& CancelNever() {
-  static cancel::Token cancel_never = cancel::Make(100000, nullptr);
+  static cancel::Token cancel_never = cancel::Make(1000000, nullptr);
   return cancel_never;
 }
 


src/commands/ft_debug.cc
```diff
@@ -76,6 +85,70 @@ absl::Status PausePointControlCmd(ValkeyModuleCtx *ctx,
   return absl::OkStatus();
 }
 
+//
+// FT.DEBUG CONTROLLED_VARIABLE SET <test_control> <value>
```
eifrah-aws: nit:

I would preserve the `CONFIG` syntax, seems more persistent to me (besides, `CONTROLLED_VARIABLE` seems too long for me):

```
FT.DEBUG CONFIG SET <name> <value>
```

src/commands/ft_debug.cc
```diff
@@ -76,6 +85,70 @@ absl::Status PausePointControlCmd(ValkeyModuleCtx *ctx,
...
+    ValkeyModule_ReplyWithError(
+        ctx, absl::StrCat("Unknown keyword", keyword).data());
+  }
+  return absl::OkStatus();
+}
+
+absl::Status HelpCmd(ValkeyModuleCtx *ctx, vmsdk::ArgsIterator& itr) {
+  VMSDK_RETURN_IF_ERROR(CheckEndOfArgs(itr));
+  static std::vector<std::pair<std::string, std::string>> help_text {
+    {"FT._DEBUG SHOW_INFO", "Show Info Variable Information"},
```
eifrah-aws: nit: 

In your previous comment, you used `FT.DEBUG` and here `FT._DEBUG` - if `FT.DEBUG` is supported, I suggest using that in the help output.

PR #362, opened 2025-09-02
Add configurable thread pool fairness mechanism with comprehensive testing

yulazariy:
## Summary

This PR implements a configurable thread pool fairness mechanism that allows fine-grained control over the scheduling balance between high and low priority tasks, while maintaining strict precedence for maximum priority tasks.

## Motivation

Previously, the thread pool used strict priority ordering where high priority tasks would completely starve low priority tasks. This implementation adds a weighted fairness algorithm that allows configurable balance between task priorities, enabling better resource utilization and preventing starvation scenarios.

### Core Implementation
- **`vmsdk/src/thread_pool.h`**: Added fairness API and private members
  - `SetHighPriorityWeight(int weight)` - Configure high priority percentage (0-100)
  - `GetHighPriorityWeight()` - Get current weight setting
  - `TryGetNextTask()` - New fairness-aware task selection algorithm
  - Added atomic weight storage and random number generator

- **`vmsdk/src/thread_pool.cc`**: Implemented weighted random task selection
  - kMax priority always takes precedence (unchanged behavior)
  - High/Low priorities use weighted random selection based on configured percentages
  - Thread-safe weight updates during runtime

### Configuration Integration
- **`src/valkey_search_options.h/.cc`**: Added configuration options for the fairness weights
  - Integration with valkey-search module configuration system

### Testing & Fixes
- **`vmsdk/testing/thread_pool_test.cc`**: Comprehensive test suite (+354 lines)
  - `ThreadPoolFairnessTest` suite with 6 test cases
  - `ThreadPoolFairnessDistributionTest` with parameterized weight testing (0%, 25%, 50%, 75%, 100%)
  - Statistical distribution validation with 5% tolerance
  - Concurrent weight change testing
  - Starvation prevention validation
  - Performance regression testing
  - **Fix**: Changed `WithParamInterface<int>` to `TestWithParam<int>` to resolve compilation error
  - **Enhancement**: Added descriptive test names (`HighPriority0Percent`, `HighPriority25Percent`, etc.)

## Algorithm Details

```cpp
// kMax priority: Always executed first (unchanged)
if (!max_queue.empty()) return max_task;

// High/Low fairness: Weighted random selection
int high_weight = GetHighPriorityWeight(); // 0-100
int random_val = uniform_distribution(0, 99);
Priority selected = (random_val < high_weight) ? kHigh : kLow;
```

## Testing

All 35 tests pass, including:
- ✅ Basic fairness functionality (weight setter/getter, clamping)
- ✅ Statistical distribution validation across all weight percentages
- ✅ kMax priority preservation under all weight configurations
- ✅ Concurrent weight changes during task execution
- ✅ Starvation prevention (low priority tasks eventually execute)
- ✅ Performance regression (< 50% overhead vs strict priority)
- ✅ Edge cases (empty queues, rapid weight changes)

## Backward Compatibility

- **Default behavior unchanged**: Default weight is 100% (high priority), maintaining existing strict priority behavior
- **API additions only**: No breaking changes to existing ThreadPool interface
- **kMax priority unchanged**: Critical tasks still get absolute priority

## Usage Example

```cpp
ThreadPool pool("worker-pool", 4);
pool.SetHighPriorityWeight(75); // 75% high, 25% low priority execution
pool.Schedule(task, ThreadPool::Priority::kHigh);
pool.Schedule(task, ThreadPool::Priority::kLow);
```

### DIFF
--- a/vmsdk/src/thread_pool.cc
+++ b/vmsdk/src/thread_pool.cc
@@ -293,6 +293,23 @@ void ThreadPool::SetHighPriorityWeight(int weight) {
   // Clamp weight to valid range [0, 100]
   weight = std::max(0, std::min(100, weight));
   high_priority_weight_.store(weight, std::memory_order_relaxed);
+
+  // Pre-compute pattern for efficiency
+  if (weight == 0) {
+    pattern_length_.store(1, std::memory_order_relaxed);
+    high_ratio_.store(0, std::memory_order_relaxed);
+  } else if (weight == 100) {
+    pattern_length_.store(1, std::memory_order_relaxed);
+    high_ratio_.store(1, std::memory_order_relaxed);
+  } else {
+    int low_weight = 100 - weight;
+    int gcd_val = std::gcd(weight, low_weight);
+    int high_ratio = weight / gcd_val;
+    int pattern_len = high_ratio + (low_weight / gcd_val);
+
+    pattern_length_.store(pattern_len, std::memory_order_relaxed);
+    high_ratio_.store(high_ratio, std::memory_order_relaxed);
+  }
 }
 
 int ThreadPool::GetHighPriorityWeight() const {
@@ -334,11 +351,9 @@ std::optional<absl::AnyInvocable<void()>> ThreadPool::TryGetNextTask() {
       uint32_t counter_val =
           fairness_counter_.fetch_add(1, std::memory_order_relaxed);
 
-      // Calculate pattern for better latency distribution
-      int low_weight = 100 - high_weight;
-      int gcd_val = std::gcd(high_weight, low_weight);
-      int high_ratio = high_weight / gcd_val;
-      int pattern_length = high_ratio + (low_weight / gcd_val);
+      // Use pre-computed pattern values for better performance
+      int high_ratio = high_ratio_.load(std::memory_order_relaxed);
+      int pattern_length = pattern_length_.load(std::memory_order_relaxed);
 
       int position_in_pattern = counter_val % pattern_length;
       selected_priority =
--- a/vmsdk/src/thread_pool.h
+++ b/vmsdk/src/thread_pool.h
@@ -150,6 +150,10 @@ class ThreadPool {
   std::atomic<int> high_priority_weight_{100};
   std::atomic<uint32_t> fairness_counter_{0};
 
+  // Pattern-based weighted round robin for better latency distribution
+  std::atomic<int> pattern_length_{1};  // Length of the repeating pattern
+  std::atomic<int> high_ratio_{1};  // Number of high priority tasks in pattern
+
   FRIEND_TEST(ThreadPoolTest, DynamicSizing);
 };
 


src/valkey_search_options.cc
```diff
@@ -171,6 +171,25 @@ static auto log_level =
 constexpr absl::string_view kEnablePartialResults{"enable-partial-results"};
 static config::Boolean enable_partial_results(kEnablePartialResults, true);
 
+/// Configure the weight for high priority tasks in thread pools (0-100)
+/// Low priority weight = 100 - high_priority_weight
```
ohadshacham, line 175: I would add that max always have higher priority than both. 
Also, I would also add that this mechanism defines fairness between mutations waiting at the low/high queue.

vmsdk/src/thread_pool.cc
```diff
@@ -289,4 +289,53 @@ void ThreadPool::Resize(size_t count, bool wait_for_resize) {
...
+    selected_priority = Priority::kLow;
+  } else if (!low_has_tasks) {
+    selected_priority = Priority::kHigh;
+  } else {
+    // Both have tasks - use weighted random selection
+    int high_weight = high_priority_weight_.load();
+    std::uniform_int_distribution<int> dist(0, 99);
+    int random_val = dist(random_generator_);
+
+    selected_priority =
```
ohadshacham: nit: why not 0, 100 and <= below?

src/valkey_search_options.cc
```diff
@@ -171,6 +171,25 @@ static auto log_level =
...
+constexpr absl::string_view kHighPriorityWeight{"high-priority-weight"};
+static auto high_priority_weight =
+    config::NumberBuilder(kHighPriorityWeight, 100, 0,
+                          100)  // Default 100%, range 0-100
+        .WithModifyCallback([](auto new_value) {
+          // Update both reader and writer thread pools
+          auto* reader_pool = ValkeySearch::Instance().GetReaderThreadPool();
+          auto* writer_pool = ValkeySearch::Instance().GetWriterThreadPool();
```
eifrah-aws: No need for `*` just `auto`

src/valkey_search_options.cc
eifrah-aws: We have 3 priorities in the queue: Max, High & Low. we only handle High & Low?

vmsdk/src/thread_pool.cc
```diff
@@ -289,4 +289,53 @@ void ThreadPool::Resize(size_t count, bool wait_for_resize) {
...
 }
+
+void ThreadPool::SetHighPriorityWeight(int weight) {
+  // Clamp weight to valid range [0, 100]
+  weight = std::max(0, std::min(100, weight));
+  high_priority_weight_.store(weight);
+}
+
+int ThreadPool::GetHighPriorityWeight() const {
+  return high_priority_weight_.load();
```
eifrah-aws:
```c++
return high_priority_weight_.load(std::memory_order_relaxed);
```

vmsdk/src/thread_pool.cc
```diff
@@ -289,4 +289,53 @@ void ThreadPool::Resize(size_t count, bool wait_for_resize) {
...
+}
+
+std::optional<absl::AnyInvocable<void()>> ThreadPool::TryGetNextTask() {
+  // Check for kMax priority first - always takes precedence
```
eifrah-aws, line 320: Why do we ignore the `Max` priority? IMO, the solution should be generic

vmsdk/src/thread_pool.cc
```diff
@@ -289,4 +289,53 @@ void ThreadPool::Resize(size_t count, bool wait_for_resize) {
...
+  }
+
+  Priority selected_priority;
+  if (!high_has_tasks) {
+    selected_priority = Priority::kLow;
+  } else if (!low_has_tasks) {
+    selected_priority = Priority::kHigh;
+  } else {
+    // Both have tasks - use weighted random selection
+    int high_weight = high_priority_weight_.load();
```
eifrah-aws: Here and everywhere: use `std::memory_order_relaxed` if all you need is atomicity.

vmsdk/src/thread_pool.h
```diff
@@ -100,7 +102,18 @@ class ThreadPool {
   void WorkerThread(std::shared_ptr<Thread> thread)
       ABSL_LOCKS_EXCLUDED(queue_mutex_);
 
+  /// Set the weight for high priority tasks (0-100)
```
eifrah-aws: nit:

Use `[0-100)` to indicate that the value can be between 0 (inclusive) - 100 (exclusive)

vmsdk/src/thread_pool.cc
```diff
@@ -289,4 +289,53 @@ void ThreadPool::Resize(size_t count, bool wait_for_resize) {
...
+    std::uniform_int_distribution<int> dist(0, 99);
+    int random_val = dist(random_generator_);
```
zvi-code: Is the randomness really needed here? why not have simple counter based? something like:
```
class ThreadPool {
    alignas(64) std::atomic<uint32_t> counter_{0};  // Avoid false sharing
    
    bool ShouldUseHighPriority() {
        uint32_t val = counter_.fetch_add(1, std::memory_order_relaxed);
        int threshold = high_priority_weight_.load(std::memory_order_relaxed);
        return (val % 100) < threshold;
    }
};

**Comment by zvi-code** [comment]
A general question, if the motivation is to avoid starvation, why not use time-base fairness? it's very likely that backfill, for example, have much less work that actual inserts, so task count will not be comparable, right?

**Comment by yairgott** [comment]
Can you please share more info about the motivation to provide such fairness control? Pointing out that high priority tasks are aimed to execute tasks for which a client is waiting to be completed while low priority tasks are background tasks, like backfill, for which no client is waiting.

File: src/valkey_search_options.cc (hunk 1 above)
Comment (yulazariy) [reply src/valkey_search_options.cc:175]: will add

File: vmsdk/src/thread_pool.cc (hunk 2 above)
Comment (yulazariy) [reply vmsdk/src/thread_pool.cc]: you mean 1, 100 and <= ? (if we use 0 it means low must win)

File: vmsdk/src/thread_pool.cc (hunk 5 above)
Comment (yulazariy) [reply vmsdk/src/thread_pool.cc:320]: Max needs to be done over all others no matter what

File: vmsdk/src/thread_pool.cc (hunk 6 above)
Comment (yulazariy) [reply vmsdk/src/thread_pool.cc]: will fix

File: vmsdk/src/thread_pool.h (hunk 7 above)
Comment (yulazariy) [reply vmsdk/src/thread_pool.h]: it actually [0, 100] as the edge means I favor only one of the priorities

File: src/valkey_search_options.cc (hunk 3 above)
Comment (yulazariy) [reply src/valkey_search_options.cc]: fixed

File: vmsdk/src/thread_pool.cc (hunk 4 above)
Comment (yulazariy) [reply vmsdk/src/thread_pool.cc]: fixed

File: vmsdk/src/thread_pool.cc (hunk 8 above)
Comment (yulazariy) [reply vmsdk/src/thread_pool.cc]: will use the counter solution as it is faster (if no high contention)

**Review by zvi-code — APPROVED** [review]
I think this is good for first step. We can enhance\extend the mechanism when we identify the need.

**Comment by zvi-code** [comment]
@yairgott , what do you think? do you have concern merging this?

File: vmsdk/src/thread_pool.cc [hunk 9]
```diff
@@ -289,4 +288,55 @@ void ThreadPool::Resize(size_t count, bool wait_for_resize) {
...
+  } else if (!low_has_tasks) {
+    selected_priority = Priority::kHigh;
+  } else {
+    // Both have tasks - use counter-based fairness
+    int high_weight = high_priority_weight_.load(std::memory_order_relaxed);
+    uint32_t counter_val =
+        fairness_counter_.fetch_add(1, std::memory_order_relaxed);
+
+    selected_priority =
+        ((counter_val % 100) < static_cast<uint32_t>(high_weight))
```
yairgott: Say high priority is set to 75%, the scheduling would lead to 75 high priority tasks followed by 25 low, followed by 75 high,  etc. The 76th high priority task would experience high latency due to waiting for the first 25 low tasks. In this case, IMO handling 3 high priority followed by 1 low, would be preferred as it would decrease the latency variability.

yulazariy: You are correct and I will fix it to use the ratio of the weights

zvi-code (approved):

PR #368, opened 2025-09-04
Update ft info

Nivesh-01:
This changes makes the FT.INFO more faithful with attributes and statistics in valkey-search module. We are doing this by removing any statics that were place holders made just to be compatible with redis search. We have also added additional statistics, added attributes and chaged the design of FT.INFO which we are going to show to the users. This is in continuation to the PR in the main branch where only attributes and statistics not related to fulltext are added.

### DIFF
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -4,6 +4,7 @@ from valkey.client import Valkey
 from valkey_search_test_case import ValkeySearchTestCaseBase
 from valkeytestframework.conftest import resource_port_tracker
 from ft_info_parser import FTInfoParser
+from valkeytestframework.util import waiters
 
 """
 This file contains tests for full text search.
@@ -204,19 +205,25 @@ class TestFullText(ValkeySearchTestCaseBase):
         tag_attr = parser.get_attribute_by_name("category")
         assert tag_attr["type"] == "TAG"
         assert tag_attr.get("SEPARATOR") == "|"
-        
-        # Add validation checks for specific fields
-        assert parser.num_docs >= 0, f"num_docs should be non-negative, got: {parser.num_docs}"
-        assert parser.num_records >= 0, f"num_records should be non-negative, got: {parser.num_records}"
-        assert parser.hash_indexing_failures >= 0, f"hash_indexing_failures should be non-negative, got: {parser.hash_indexing_failures}"
+
+        def check_is_backfill_complete(idxname):
+            """
+            Helper function to check if backfill is complete.
+            """
+            info = client.execute_command("FT.INFO", idxname)
+            parser = FTInfoParser(info)
+            return parser.is_backfill_complete()
         
         # Validate backfill fields
-        assert isinstance(parser.backfill_in_progress, bool), f"backfill_in_progress should be boolean, got: {type(parser.backfill_in_progress)}"
-        assert isinstance(parser.backfill_complete_percent, (int, float)), f"backfill_complete_percent should be numeric, got: {type(parser.backfill_complete_percent)}"
-        assert 0.0 <= parser.backfill_complete_percent <= 1.0, f"backfill_complete_percent should be between 0.0 and 1.0, got: {parser.backfill_complete_percent}"
+        waiters.wait_for_equal(lambda: check_is_backfill_complete("idx4"), True, timeout=5)
+        
+        # Add validation checks for specific fields
+        assert parser.num_docs == 0, f"num_docs should be zero"
+        assert parser.num_records == 0, f"num_records should be zero"
+        assert parser.hash_indexing_failures == 0, f"hash_indexing_failures should be zero"
         
         # Validate queue and delay fields
-        assert parser.mutation_queue_size >= 0, f"mutation_queue_size should be non-negative, got: {parser.mutation_queue_size}"
+        assert parser.mutation_queue_size == 0, f"mutation_queue_size should be non-negative, got: {parser.mutation_queue_size}"
         assert isinstance(parser.recent_mutations_queue_delay, str), f"recent_mutations_queue_delay should be string, got: {type(parser.recent_mutations_queue_delay)}"
         
         # Validate state field
@@ -234,7 +241,7 @@ class TestFullText(ValkeySearchTestCaseBase):
         
         # Validate with_offsets setting
         with_offsets = parser.parsed_data.get("with_offsets")
-        assert with_offsets in [0, 1, "0", "1", True, False], f"with_offsets should be boolean-like, got: {with_offsets}"
+        assert with_offsets == 1, f"with_offsets is set to true any other value is wrong"
         
         # Validate language setting
         language = parser.parsed_data.get("language", "")
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -3,6 +3,7 @@ from valkey import ResponseError
 from valkey.client import Valkey
 from valkey_search_test_case import ValkeySearchTestCaseBase
 from valkeytestframework.conftest import resource_port_tracker
+from ft_info_parser import FTInfoParser
 
 """
 This file contains tests for full text search.
@@ -107,7 +108,7 @@ class TestFullText(ValkeySearchTestCaseBase):
         assert result[0] == 2  # Number of documents found. Both docs below start with Grea* => Great and Greased
         assert result[1] == b"product:1" and result[3] == b"product:5" or result[1] == b"product:5" and result[3] == b"product:1"
 
-    def test_ft_create(self):
+    def test_ft_create_and_info(self):
         """
         Test basic text search for FT.CREATE with multiple cases.
         Validates that the command parsing works correctly even though TEXT indexing is not yet implemented.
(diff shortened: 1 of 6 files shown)


integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+        assert any(state in str_repr for state in ["ready", "backfill_in_progress"])
+        
+        # Test dictionary conversion
+        data_dict = parser.to_dict()
+        assert isinstance(data_dict, dict)
+        assert data_dict["index_name"] == "parser_test_idx"
+
+    def test_ft_info_comprehensive_validation(self):
+        """
+        Comprehensive test validating all major FT.INFO response fields and structure
```
BCathcart: If this is truly comprehensive, then is there any need for the other tests?

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
         # @ NOT configured as separator - should not be able with split words
         result = client.execute_command("FT.SEARCH", "idx", '@content:"test"')
         assert result[0] == 0
+
+    def test_ft_info_basic_text_index(self):
```
BCathcart: We should think about how we structure our amztests at some point. I don't think isolating all our testing in fulltext specific files will always be the way to go. For example, we may want a test_ft_info.py file that contains tests for all ft.info fields. It makes more sense to me to group based on feature (the FT.INFO command) rather than index type in this case. Although if we can reduce the number of tests we're adding here, I think it's fine to leave just one or two in this file for now.

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+        parser = FTInfoParser(info_response)
+        
+        # Validate basic index properties
+        assert parser.index_name == "text_idx"
+        # Accept both ready and backfill_in_progress states as valid
+        assert parser.state in ["ready", "backfill_in_progress"]
+        assert parser.num_docs == 0
+        # During backfill, terms and records may be 0
+        assert parser.num_terms >= 0
+        assert parser.num_records >= 0
```
BCathcart: We haven't populated any records yet have we?

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+        assert not parser.has_indexing_failures()
+        # Index may be in backfill state initially
+        assert parser.state in ["ready", "backfill_in_progress"]
```
BCathcart: This is already checked above

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+        
+        # Validate index definition
+        index_def = parser.index_definition
+        assert index_def["key_type"] == "HASH"
+        assert index_def["prefixes"] == ["doc:"]
+        assert index_def["default_score"] == 1  # Returned as integer, not string
```
BCathcart: "# Returned as integer, not string"

Let's try to avoid comments that are evident from the code

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+        
+        # Get updated info
+        info_response = client.execute_command("FT.INFO", "text_idx")
+        parser = FTInfoParser(info_response)
+        
+        # Verify document count increased
+        assert parser.num_docs == 2
+        # During backfill, terms and records may remain 0 until indexing completes
+        assert parser.num_terms >= 0  # May be 0 during backfill
+        assert parser.num_records >= 0  # May be 0 during backfill
```
BCathcart: Just for my understanding, when are the number of docs updated compared to the number of terms and records in the indexing flow?

I'd also wait for the backfill to complete so that we can check the number of terms and records are updated to what we eventually expect. You can wait for the parser.state to be "ready" before checking for example.

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+            # Attributes may not be fully parsed during backfill
+            if title_attr is not None:
+                assert title_attr.get("type") == "TEXT" or isinstance(title_attr, list)
+            if desc_attr is not None:
+                assert desc_attr.get("type") == "TEXT" or isinstance(desc_attr, list)
+        except (AttributeError, TypeError):
+            # Parser may fail during backfill - this is acceptable
+            pass
+
+    def test_ft_info_mixed_field_types(self):
```
BCathcart: The tests starting at this point seem better suited for a general FT.INFO test file

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+        # Create an index for scope testing
+        client.execute_command("FT.CREATE", "test_idx", "ON", "HASH", "SCHEMA", "field", "TEXT")
+        
+        # Test invalid scope parameter
+        with pytest.raises(ResponseError, match="Invalid scope parameter"):
+            client.execute_command("FT.INFO", "test_idx", "INVALID_SCOPE")
+
+    def test_ft_info_local_scope(self):
+        """
+        Test FT.INFO with explicit LOCAL scope (default behavior)
```
BCathcart: Are we creating multiple shards in this test? If not, I'd expect the GLOBAL scope to return the same thing anyways

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+        # Add multiple documents
+        for i in range(10):
+            client.execute_command("HSET", f"doc:{i}", "content", f"document number {i} with unique content")
+        
+        # Get updated state
+        info_response = client.execute_command("FT.INFO", "activity_idx")
+        updated_parser = FTInfoParser(info_response)
+        
+        # Verify statistics updated
+        assert updated_parser.num_docs >= 10  # May include docs from other tests
```
BCathcart: Do we not clean up state between tests?

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+        # During backfill, terms and records may not be updated immediately
+        assert updated_parser.num_terms >= initial_parser.num_terms
+        assert updated_parser.num_records >= initial_parser.num_records
+        assert updated_parser.state in ["ready", "backfill_in_progress"]
```
BCathcart: Again I think we should wait for the indexing to finish. These checks aren't telling us much right now

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+
+    def test_ft_info_parser_functionality(self):
+        """
+        Test the FTInfoParser utility functions and properties
```
BCathcart: Testing test code feels odd. Aren't we inherently validating it works by using it in the other tests already?

integration/test_fulltext.py
BCathcart: I'd consider if we can get away with one comprehensive happy case test and the error case test

integration/test_fulltext.py
```diff
@@ -299,3 +300,370 @@ def test_custom_punctuation(self):
...
+            if price_field and isinstance(price_field, dict):
+                assert price_field.get("type") == "NUMERIC"
+                
+            category_field = parser.get_attribute_by_name("category")
+            if category_field and isinstance(category_field, dict):
+                assert category_field.get("type") == "TAG"
+                
+        except (AttributeError, TypeError):
+            # Parser may fail during backfill - this is acceptable
+            # The important thing is that the index was created successfully
```
BCathcart: Why does the parser fail during backfill? This doesn't seem like the important thing given the test is for FT.INFO validation and not index creation

testing/ft_info_test.cc
```diff
@@ -464,6 +469,141 @@ INSTANTIATE_TEST_SUITE_P(
                     },
                 },
         },
+        {
+            .test_name = "happy_path_text_defaults",
```
BCathcart, line 473: Would these already be covered by `happy_path_text_all_options` below?

integration/test_fulltext.py
Nivesh-01: done

integration/test_fulltext.py
Nivesh-01: I agree, I added it just in case. Actually I tend to add extra tests eventhough I know that they are probably not needed just in case. This one for example is totally removable and I can do it if it is better.

integration/test_fulltext.py
Nivesh-01: agreed, will remove it.

testing/ft_info_test.cc
Nivesh-01: yes they will. I added this test out of habbit of being completely thorough but yeah this is covered in the next step and can be removed

integration/test_fulltext.py
Nivesh-01: This is partially done. So for vector, numeric and tag it is implemented but for text it is not properly implemented yet.

integration/test_fulltext.py
Nivesh-01: It actually is comprehensive, it also have vector fields in addition to text fields, so tehnically we could get away with one test.

integration/test_fulltext.py
Nivesh-01: done

integration/test_fulltext.py
Nivesh-01: For now I am thinking of having 2 tests only related to text field in test_fulltext and then moving the other more comprehensive ones to a new test_fulltext_ft_info file. What do you think?

integration/test_fulltext.py
Nivesh-01: Same as above - For now I am thinking of having 2 tests only related to text field in test_fulltext and then moving the other more comprehensive ones to a new test_fulltext_ft_info file. What do you think?

integration/test_fulltext.py
Nivesh-01: Removing it from these tests

integration/test_fulltext.py
Nivesh-01: I was going to but stopped since this is going to be my next task anyways. Let it be here just in case we revisit and want more rigorous testing. But I think you are right does not make sense to have additional placeholder kind tests so removing it.

integration/test_fulltext.py
Nivesh-01: num_terms is still a place holder, num docs counts every document and is immediatetly updated when addRecords() function is called for any untracked key, for num_records for each addRecord in different index fields
Relatively num_docs happens fastest and then num_records.

I could do that.

integration/test_fulltext.py
Nivesh-01: I have not done that yet

Nivesh-01: right now in all these test I am more focused on wether everything that we want is being shown for FT.info instead of how correctly the metrics have been added, since that is probably not somehting ft.info is concerned about. I could add those in as well

testing/ft_info_test.cc
BCathcart: Thorough is good, redundant is not 🙂

integration/test_fulltext.py
BCathcart: I'd personally have all ft.info focused tests together in a `test_ft_info.py` file

integration/test_fulltext.py
BCathcart: I mean in this test we haven't added any keys/documents so they would have to be 0 right?

integration/test_fulltext.py
BCathcart: I have used "documents" and "records" interchangeably in the past but if I remember correctly what @Aksha1812 told me last week, "records" is actually the total number of words indexed from documents and "terms" is the total number of unique words (if we are to copy Redis search). Since you mentioned they aren't implemented yet, maybe we shouldn't be verifying this functionality with a test at this point and instead introduce a test along with the code that implements it?

integration/test_fulltext.py
BCathcart: > May include docs from other tests

I'm just commenting on the fact that I don't think this is possible. I'd hope our test infra here cleans up state between tests

integration/test_fulltext.py
BCathcart: In general it's better to over-test than under-test, but every test should still serve some clear purpose. There's developer overhead to maintaining tests as well as compute and time costs to running them.

integration/test_fulltext.py
Nivesh-01: What I am trying to do here is see if they are at least showing up when FT.INFO, so I could remove it but keeping it there also makes sense I believe, what are your thoughts on this.

BCathcart: Your unit test already validates that they exist in FT.INFO output so I don't think it's necessary. The test also reads like it's validating metric updates after docs are added.

src/index_schema.cc
```diff
@@ -733,8 +733,26 @@ uint64_t IndexSchema::CountRecords() const {
   return record_cnt;
 }
 
+bool IndexSchema::HasTextFields() const {
+  for (const auto &attribute : attributes_) {
```
BCathcart: A shortcut here would be to check if `text_index_schema_` is not null

src/index_schema.cc
```diff
@@ -844,6 +862,36 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
...
+    ValkeyModule_ReplyWithArray(ctx, stop_words_.size());
+    for (const auto &stop_word : stop_words_) {
+      ValkeyModule_ReplyWithSimpleString(ctx, stop_word.c_str());
+    }
+
+    ValkeyModule_ReplyWithSimpleString(ctx, "with_offsets");
+    ValkeyModule_ReplyWithSimpleString(ctx, with_offsets_ ? "1" : "0");
+  }
+
+  if (language_ != data_model::LANGUAGE_UNSPECIFIED) {
```
BCathcart: Do we need this check since it seems like it always defaults to english? We can probably remove the data_model::LANGUAGE_UNSPECIFIED variant entirely

src/indexes/text.cc
```diff
@@ -78,7 +78,22 @@ absl::StatusOr<bool> Text::ModifyRecord(const InternedStringPtr& key,
...
+  // Show only one: if no_stem is specified, show no_stem, otherwise show
+  // min_stem_size
+  if (no_stem_) {
+    ValkeyModule_ReplyWithSimpleString(ctx, "NO_STEM");
+    ValkeyModule_ReplyWithSimpleString(ctx, "1");
+  } else {
+    ValkeyModule_ReplyWithSimpleString(ctx, "MIN_STEM_SIZE");
+    ValkeyModule_ReplyWithLongLong(ctx, min_stem_size_);
+  }
+  // Text fields do not include a size field (unlike numeric/tag/vector fields)
```
BCathcart: Is this something we'll need to add in the future?

integration/test_fulltext.py
```diff
@@ -154,15 +155,90 @@ def test_ft_create(self):
...
+        ]
+        
+        for field in required_top_level_fields:
+            assert field in parser.parsed_data, f"Missing required field: {field}"
+        
+        # Validate index definition structure
+        index_def = parser.index_definition
+        assert "key_type" in index_def
+        assert "prefixes" in index_def
+        assert "default_score" in index_def
```
BCathcart, line 196: I think we might as well assert the values are exactly what we expect throughout this test

Nivesh-01: done

src/index_schema.cc

src/index_schema.cc

src/indexes/text.cc
Nivesh-01: done

BCathcart (approved):

PR #370, opened 2025-09-05
added and removed stats and attributes from ft.info

Nivesh-01:
This changes makes the FT.INFO more faithful with attributes and statistics in valkey-search module. We are doing this by removing any statics that were place holders made just to be compatible with redis search. We have also added additional statistics that we are going to show to the users. This will be matched by an update in the fulltext branch as well where only attributes and statistics related to fulltext will be added.


src/index_schema.cc
```diff
@@ -754,55 +752,17 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
...
   ValkeyModule_ReplyWithLongLong(ctx, 0);
-  ValkeyModule_ReplyWithSimpleString(ctx, "global_total");
+  ValkeyModule_ReplyWithSimpleString(ctx, "index_memory");
   ValkeyModule_ReplyWithLongLong(ctx, 0);
-  ValkeyModule_ReplyWithSimpleString(ctx, "index_capacity");
+  ValkeyModule_ReplyWithSimpleString(ctx, "total_inverted_index_blocks");
   ValkeyModule_ReplyWithLongLong(ctx, 0);
-  ValkeyModule_ReplyWithSimpleString(ctx, "index_total");
+  ValkeyModule_ReplyWithSimpleString(ctx, "total_terms");
   ValkeyModule_ReplyWithLongLong(ctx, 0);
```
KarthikSubbarao: Question about these additional metrics and `memory_per_document` below. If these metrics are fulltext specific, would it make sense to add those metric into the fulltext branch? Since releases (e.g. 1.1) will be created from main, it would be good to not include fulltext metrics until we actually support them (e.g. in the 2.0 release with fulltext support).

I agree with the other changes

Aksha1812: @Nivesh-01 Yes, you can remove these from main for now. We are not yet sure if we are going to use the same fields as well or not. And valkey and redis might different interpretations of below fields, so better to leave it out for now.

KarthikSubbarao (approved):

PR #377, opened 2025-09-08
Memory allocation tracking for Fulltext

Aksha1812:
Use memory tracking using the VMSDK framework to accurately measure memory usage for text index components used in fulltext search. Used IsolatedMemoryScope for tracking memory in Posting and Radix Trie.

Added below info fields to ft.info to track memory for text index

```
"num_unique_terms":,                  // Total number of unique terms in the text index
"num_total_terms":,                   // Total frequency of all terms across all documents  
"num_records":,                       // Total number of records across all indexes
"posting_sz_mb":,                     // Memory used by posting lists (inverted index data) in MB
"position_sz_mb":,                    // Memory used by position information for phrase queries in MB
"total_postings":,                    // Total number of posting lists (equals unique terms)
"radix_sz_mb":,                       // Memory used by the radix tree (term dictionary) in MB
"total_text_index_sz_mb":,            // Total memory used by all text index components in MB
"total_terms_per_doc_avg":,           // Average number of terms per document
```

Pending tests

### DIFF
--- a/src/indexes/text/posting.cc
+++ b/src/indexes/text/posting.cc
@@ -17,7 +17,6 @@
 #include "absl/log/check.h"
 #include "src/index_schema.h"
 
-
 namespace valkey_search::indexes::text {
 
 // Internal FieldMask classes - not part of external interface


src/indexes/text.cc
```diff
@@ -138,4 +141,117 @@ std::unique_ptr<EntriesFetcherIteratorBase> Text::EntriesFetcher::Begin() {
   return nullptr;
 }
 
+uint64_t Text::GetTotalPositions() const {
```
BCathcart: We'll need to keep track of these metrics during ingestion. We can't iterate through the entire index multiple times every time INFO is called.

src/index_schema.cc
```diff
@@ -764,11 +765,53 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
...
-  ValkeyModule_ReplyWithSimpleString(ctx, "num_terms");
-  ValkeyModule_ReplyWithLongLong(ctx, 0);
+  
+  // Get text index for info fields
+  std::shared_ptr<indexes::Text> text_index = nullptr;
+  for (const auto& attribute : attributes_) {
+    auto index = attribute.second.GetIndex();
+    if (index->GetIndexerType() == indexes::IndexerType::kText) {
+      text_index = std::dynamic_pointer_cast<indexes::Text>(index);
+      break;
```
BCathcart: I'm a bit confused by the code flow. We get a single text attribute and then call a function on it (e.g. GetNumTerms()), which then comes back to the schema level and iterates over the whole index. Should we track these metrics in `TextIndexSchema` and call into it directly? Or do we intend to keep track of metrics for each text attribute separately in some way?

src/index_schema.h
```diff
@@ -181,6 +181,14 @@ class IndexSchema : public KeyspaceEventSubscription,
   uint64_t GetBackfillDbSize() const;
   InfoIndexPartitionData GetInfoIndexPartitionData() const;
 
+  // Text index statistics helper methods
+  uint64_t GetTextIndexMemoryUsage() const;
+  uint64_t GetRadixTreeMemoryUsage() const;
+  uint64_t GetPostingsMemoryUsage() const;
+  uint64_t GetTotalPositions() const;
+  uint64_t GetNumTerms() const;
+  uint64_t GetTotalTermFrequency() const;
```
BCathcart: Are these used?

src/indexes/text.cc
```diff
@@ -78,7 +78,10 @@ absl::StatusOr<bool> Text::ModifyRecord(const InternedStringPtr& key,
 }
 
 int Text::RespondWithInfo(ValkeyModuleCtx* ctx) const {
-  throw std::runtime_error("Text::RespondWithInfo not implemented");
+  auto num_replies = 2;
+  ValkeyModule_ReplyWithSimpleString(ctx, "type");
+  ValkeyModule_ReplyWithSimpleString(ctx, "TEXT");
+  return num_replies;
```
BCathcart: Do you know if we need any metrics on a per-text attribute level to be compatible with redis labs or is everything tracked at the schema level?

src/indexes/text.h
```diff
@@ -110,6 +110,23 @@ class Text : public IndexBase {
       const query::TextPredicate& predicate,
       bool negate) const ABSL_NO_THREAD_SAFETY_ANALYSIS;
 
+  // Statistics methods for FT.INFO
+  uint64_t GetTotalPositions() const;
```
BCathcart: I see this isn't reported in FT.INFO. Should it be?

src/indexes/text.cc
Aksha1812: ok makes sense. will update

src/index_schema.cc
Aksha1812: If i call any text attribute it would lead to same text index or text indexschema.

src/index_schema.h
Aksha1812: No i had moved them from here to text.cc . only definitions accidently left out. will remove it.

src/indexes/text.cc
Aksha1812: per text attribute there is only one more field which is "WEIGHT", which is used for boosting relevance for a particular field during search scoring. We are not doing that currently.

src/indexes/text.h
Aksha1812: I use it to calculate positions memory size , where i just multiplty num_of_positions * sizeof(uint32_t). so indirectly reporting it. didn't think it would be useful to show both number of positions and position,memory size.

src/indexes/text/posting.cc
```diff
@@ -179,18 +186,23 @@ class Postings::Impl {
...
 Postings::Postings(bool save_positions, size_t num_text_fields)
     : impl_(std::make_unique<Impl>(save_positions, num_text_fields)) {
+  IsolatedMemoryScope scope{memory_pool_};
   CHECK(impl_ != nullptr) << "Failed to create Postings implementation";
 }
 
 // Automatic cleanup via unique_ptr
-Postings::~Postings() = default;
+Postings::~Postings() {
+  IsolatedMemoryScope scope{memory_pool_};
```
allenss-amazon: This is probably a noop.

src/indexes/text/posting.cc
```diff
@@ -205,6 +217,8 @@ void Postings::InsertPosting(const Key& key, size_t field_index,
     effective_position = 0;
   }
 
+  bool is_new_key =
+      impl_->key_to_positions_.find(key) == impl_->key_to_positions_.end();
```
allenss-amazon: ? this appears unused.

src/indexes/text/posting.cc
```diff
@@ -223,14 +237,19 @@ void Postings::InsertPosting(const Key& key, size_t field_index,
 
 // Remove a document key and all its positions
 void Postings::RemoveKey(const Key& key) {
-  impl_->key_to_positions_.erase(key);
+  IsolatedMemoryScope scope{memory_pool_};
+
+  auto it = impl_->key_to_positions_.find(key);
+  if (it != impl_->key_to_positions_.end()) {
+    impl_->key_to_positions_.erase(it);
+  }
```
allenss-amazon, line 264: what's wrong with erase?

src/indexes/text.cc
allenss-amazon: The original spec assumed that the position counts were bubbled up through the PrefixTree, i.e., that the counts in the prefix tree. You should be able to just query the root node.

src/index_schema.cc
allenss-amazon: Why aren't we just querying the per-index-schema TextIndex ?

src/indexes/text.cc
Aksha1812: Not sure how it would be bubbled up through the prefix tree. Are you saying i need to add that functionality or is it already there? would we update this total at the root node each time when we add a position? Also this calculation is will change with Position Optimization, but yeah we would still need to go over each Posting for that. I understood what Brennan said that we could keep track of this during ingestion and deletion. @allenss-amazon are you saying the same thing or is there an easier way?

src/index_schema.cc
Aksha1812: Ok yeah that makes more sense. I'll move all the info related functions to text_index/-schema and use the thext_index_schema variable which is already there in index_schema.

src/indexes/text/posting.cc
Aksha1812: Oh. I thought it would track the memory when a Posting is deleted and when we deallocate its memory using the deconstructor. how should i ensure I track the memory which is deallocated too?

integration/ft_info_parser.py
```diff
@@ -298,6 +298,46 @@ def with_offsets(self) -> Optional[bool]:
...
+
+    @property
+    def total_postings(self) -> Optional[int]:
+        """Get the total number of posting lists (equals unique terms)."""
+        return self.parsed_data.get("total_postings")
+
+    @property
+    def radix_sz_mb(self) -> Optional[float]:
+        """Get the memory used by the radix tree (term dictionary) in MB."""
+        return self.parsed_data.get("radix_sz_mb")
```
allenss-amazon: We should be returning bytes, not MB.

integration/test_fulltext.py
```diff
@@ -639,3 +642,86 @@ def delete_documents(client_id):
...
+        print(f"  Unique Terms: {info_data['num_unique_terms']}")
+        print(f"  Total Terms: {info_data['num_total_terms']}")
+        print(f"  Total Postings: {info_data['total_postings']}")
+        print(f"  Avg Terms/Doc: {info_data['total_terms_per_doc_avg']:.2f}")
+        print(f"  Total Memory: {info_data['total_text_index_sz_mb']:.4f} MB")
+        print(f"  Posting Memory: {info_data['posting_sz_mb']:.4f} MB")
+        print(f"  Position Memory: {info_data['position_sz_mb']:.4f} MB")
+        print(f"  Radix Memory: {info_data['radix_sz_mb']:.4f} MB")
+        
+        # TODO : Add verification of cleanup after keys deleted.
```
allenss-amazon: I think this test should delete the docs and then restest the FT info information. Then it should delete the index and retest.

src/indexes/text/posting.cc
allenss-amazon: You need to cause the destructors for Postings to run while the IsolatedMemoryScope is still in-scope. You can do that for the nested objects by explicitly invoking the actual destructor here WITHOUT causing infinite recursion.

Here's how to do that.

1) Create a thread-local boolean which we'll call "InPostingsDestructor", assume it's initialied as False (all thread-locals are initialized to zero at thread startup time.
```

~Postings() {
   if (!inPostingsDestructor) {
      inPostingsDestructor = true;
      this->~Postings() ; // call myself recursively
      inPostingsDestructor = false;
   }
}

```

src/indexes/text/text_index.cc
```diff
@@ -37,9 +52,13 @@ std::optional<std::shared_ptr<text::Postings>> RemoveKeyFromPostings(
   CHECK(existing.has_value()) << "Per-key tree became unaligned";
   auto postings = existing.value();
   postings->RemoveKey(key);
+
   if (!postings->IsEmpty()) {
     return postings;
   } else {
+    auto& metadata = GetTextIndexMetadata();
+    std::lock_guard<std::mutex> lock(metadata.mtx);
+    metadata.num_unique_terms--;
```
allenss-amazon: should num_unique_terms be atomic?

src/indexes/text/text_index.cc
```diff
@@ -154,4 +173,76 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr& key) {
   }
 }
 
+uint64_t TextIndexSchema::GetTotalPositions() const {
+  auto& metadata = GetTextIndexMetadata();
+  std::lock_guard<std::mutex> lock(metadata.mtx);
+  return metadata.total_positions;
```
allenss-amazon: The lock here is a no-op. It doesn't guarantee that metadata remains in existence and it doesn't guarantee that metadata.total_positions remains constant. (It could change between the return and the usage of it....

I suspect these counters should be atomics and then all of this false locking just goes away.

src/indexes/text/text_index.cc
```diff
@@ -154,4 +173,76 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr& key) {
...
+  return static_cast<double>(GetPositionMemoryUsage()) / num_total_terms;
+}
+
+double TextIndexSchema::GetTotalTextIndexSizePerTermAvg() const {
+  uint64_t num_terms = GetNumUniqueTerms();
+  if (num_terms == 0) {
+    return 0.0;
+  }
+  return static_cast<double>(GetTotalTextIndexMemoryUsage()) / num_terms;
+}
```
allenss-amazon, line 221: If we use these internally, fine. But if this is done solely for INFO reporting, then doesn't the application have the numerator and denominator available and can do it's own division? (let's avoid bloating the INFO calls)

src/indexes/text/text_index.h
```diff
@@ -27,6 +27,16 @@ struct sb_stemmer;
 
 namespace valkey_search::indexes::text {
 
+// FT.INFO counters for text info fields
+struct TextIndexMetadata {
+  uint64_t total_positions = 0;
+  uint64_t num_unique_terms = 0;
+  uint64_t total_term_frequency = 0;
```
allenss-amazon: I think these should be atomics, and then just drop the lock.

src/indexes/text/text_index.cc
Aksha1812: Redis had a bunch of aggregate info fields like this, but sure i can remove it for now, and later if we absolutely need a particular field we can add it.

src/indexes/text/text_index.h
Aksha1812: done

src/indexes/text/text_index.cc
Aksha1812: done

src/indexes/text/text_index.cc
Aksha1812: done

src/indexes/text/text_index.cc
Aksha1812: done

integration/test_fulltext.py
Aksha1812: done. but not everything will zero out. as we delete the the entire key from the per key index and in the per key index we are having duplicate Postings currently, that will get resaolved soon when we just have one Posting object in the schema level index and Radix and Total memory would also zero down when Brennan handles radix tree memory cleanup.

integration/ft_info_parser.py
Aksha1812: done

integration/test_fulltext.py
```diff
@@ -300,13 +301,15 @@ def test_ft_create_and_info(self):
         assert isinstance(stop_words, list), f"stop_words should be list, got: {type(stop_words)}"
         assert set(stop_words) == {"the", "and"}, f"Expected stop_words ['the', 'and'], got: {stop_words}"
         
-        # Validate with_offsets setting
+        # Validate with_offsets setting (only present when text_index_schema exists)
```
BCathcart: Why does this test need to be changed? Aren't we validating `idx4` here which does have text fields so we should probable expect that these fields exist?

integration/ft_info_parser.py
```diff
@@ -298,6 +298,41 @@ def with_offsets(self) -> Optional[bool]:
...
         return bool(offsets) if offsets is not None else None
 
+    @property
+    def num_unique_terms(self) -> Optional[int]:
+        """Get the number of unique terms in the text index."""
+        return self.parsed_data.get("num_unique_terms")
+
+    @property
+    def num_total_terms(self) -> Optional[int]:
+        """Get the total frequency of all terms across all documents."""
```
BCathcart, line 308: Have we compared the behaviour of this metric against Redis search? For example the value will be much different if offsets are enabled or disabled

integration/ft_info_parser.py
```diff
@@ -298,6 +298,41 @@ def with_offsets(self) -> Optional[bool]:
...
+        return self.parsed_data.get("num_total_terms")
+
+    @property
+    def posting_sz_bytes(self) -> Optional[int]:
+        """Get the memory used by posting lists (inverted index data) in bytes."""
+        return self.parsed_data.get("posting_sz_bytes")
+
+    @property
+    def position_sz_bytes(self) -> Optional[int]:
+        """Get the memory used by position information for phrase queries in bytes."""
```
BCathcart, line 318: Might be good to mention that it's a subset of the postings object memory which I assume is also already included in `posting_sz_bytes`. It's not immediately clear from these comments how the memory is divided or double-counted across the different pieces here

src/indexes/text/text_index.cc
```diff
@@ -11,23 +11,43 @@
...
 
 namespace valkey_search::indexes::text {
 
+// Static memory pool for TextIndexSchema
+MemoryPool TextIndexSchema::memory_pool_{0};
+
+int64_t TextIndexSchema::GetMemoryUsage() { return memory_pool_.GetUsage(); }
+
+TextIndexMetadata& GetTextIndexMetadata() {
+  static TextIndexMetadata metadata;
```
BCathcart: Don't we need to track the metadata for each text schema independently? `FT.INFO` is specific to an index and this looks to be accumulating metadata across all text schemas. I guess the same applies to the memory pools which are also shared

src/indexes/text/text_index.cc
```diff
@@ -154,4 +180,43 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr& key) {
...
+  }
+  return Postings::GetMemoryUsage();
+}
+
+uint64_t TextIndexSchema::GetRadixTreeMemoryUsage() const {
+  if (!text_index_) {
+    return 0;
+  }
+  return RadixTree<std::shared_ptr<Postings>, false>::GetMemoryUsage() -
+         GetPostingsMemoryUsage();
```
BCathcart: Just curious why did you switch from isolated memory scope to using the nested scope?

src/indexes/text/posting.h
```diff
@@ -98,8 +99,8 @@ struct Postings {
   // Total number of keys
   size_t GetKeyCount() const;
 
-  // Total number of postings for all keys
-  size_t GetPostingCount() const;
+  // Total number of positions for all keys
+  size_t GetPositionCount() const;
 
   // Total frequency of the term across all keys and positions
   size_t GetTotalTermFrequency() const;
```
BCathcart, line 105: Should we remove this? Looks like it's unused and term frequency is tracked elsewhere

Aksha1812: I am using it for unit testing.

src/indexes/text/text_index.cc
Aksha1812: As per our design TextIndexSchema -> Radix Tree -> Postings. With the isolated memory scope the memory was not bubbling up to total text index. Also the main issue was when suppose we delete a key in the key index, this would trigger cascading deallocation from the key_index to Postings. It is very difficult to put an Isolatedmemoryscope around that to accurately track the individual deallocations. and the numbers would not make that much sense. We can discuss on it later offline

src/indexes/text/text_index.cc
Aksha1812: the pool is defined outside but the scope is still inside the schema, and around places where we need to track the memory.

integration/ft_info_parser.py
Aksha1812: ok added note at TextIndexSchema::GetPositionMemoryUsage() .

integration/ft_info_parser.py
Aksha1812: they just had one info field that is `num_terms` representing number of distinct terms, which looked vague to us earlier and hence decide to keep num_total_terms and num_unique terms separate. which makes more sense as of now. Currently we are not worrying about if it would be compatible or not, just including the fields which we actually need and might be relevant for us now

integration/test_fulltext.py
Aksha1812: It was throwing error earlier, will, try to run wihtout these changes again

src/indexes/text/text_index.cc
Aksha1812: understand it now, do you have any suggestions on how to handle

integration/ft_info_parser.py
BCathcart: Ok eventually we should have a broader discussion to finalize what info we want to expose through FT.INFO

src/indexes/text/posting.h
BCathcart: Ok it's not high priority but typically unit tests should focus on the actual API that's used. Let's consider if we can get rid of it in the future

src/indexes/text/text_index.cc
```diff
@@ -55,6 +76,9 @@ TextIndexSchema::TextIndexSchema(data_model::Language language,
 absl::StatusOr<bool> TextIndexSchema::IndexAttributeData(
     const InternedStringPtr& key, absl::string_view data,
     size_t text_field_number, bool stem, size_t min_stem_size, bool suffix) {
+  current_schema_ = this;
```
BCathcart, line 79: This feels like an unusual pattern but I think it works for now. Essentially any time an ingestion thread calls a modifying function on the TextIndexSchema, it is given a reference to the relevant metadata which can be used by the sub-structure code (postings object and radix tree). Typically I would just say to pass the metadata object between functions but I guess the memory tracking stuff complicates things because it's used in destructors that aren't explicitly called. Am I following the right train of thought? I'm a little concerned about maintainability of this pattern but I think that's something we can consider in a follow-up

Since we only really care about the metadata, can we store a pointer to it directly? I don't think there is any need to expose the whole schema in a global variable

allenss-amazon: I think yes. The ease-of-use of the memory tracking hasn't met expectations and I have some thoughts about how to make it less brittle. But later.

src/index_schema.cc
```diff
@@ -795,6 +799,39 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
...
+      text_index_schema_ ? text_index_schema_->GetTotalTermFrequency() : 0);
+  ValkeyModule_ReplyWithSimpleString(ctx, "num_unique_terms");
+  ValkeyModule_ReplyWithLongLong(
+      ctx, text_index_schema_ ? text_index_schema_->GetNumUniqueTerms() : 0);
+  ValkeyModule_ReplyWithSimpleString(ctx, "total_postings");
+  ValkeyModule_ReplyWithLongLong(
+      ctx, text_index_schema_ ? text_index_schema_->GetNumUniqueTerms() : 0);
+
+  // Memory statistics are only shown when debug mode is enabled
+  if (vmsdk::config::IsDebugModeEnabled()) {
```
BCathcart, line 815: `ft_info_test.cc` also needs to be updated to change expectations based on the debug mode

PR #382, opened 2025-09-11
Early return mutation handling on record not found and not tracked

yairgott:


src/index_schema.cc
```diff
@@ -1066,6 +1072,20 @@ vmsdk::BlockedClientCategory IndexSchema::GetBlockedCategoryFromProto() const {
       return vmsdk::BlockedClientCategory::kOther;
   }
 }
+
+bool IndexSchema::InTrackedMutationRecords(
+    const InternedStringPtr &key, const std::string &identifier) const {
+  absl::MutexLock lock(&mutated_records_mutex_);
```
eifrah-aws, line 1078: Can we use a reader lock here?

yulazariy (approved):

src/index_schema.cc
yairgott: It's possible but it wouldn't have much value as this is called only by the main thread while the worker threads require write lock.

PR #386, opened 2025-09-15
FT.CREATE consistency check

boda26:
1. Index creation consistency check after FT.CREATE command (changing FT.CREATE from a non-blocking command to a blocking command, only in cluster mode & coordinator is on)
2. relocate ft.info timeout config to valkey_search_options

### DIFF
--- a/src/coordinator/metadata_manager.cc
+++ b/src/coordinator/metadata_manager.cc
@@ -146,7 +146,7 @@ absl::StatusOr<google::protobuf::Any> MetadataManager::GetEntry(
   return metadata.type_namespace_map().at(type_name).entries().at(id).content();
 }
 
-absl::StatusOr<NewEntryFingerprintVersion> MetadataManager::CreateEntry(
+absl::StatusOr<IndexFingerprintVersion> MetadataManager::CreateEntry(
     absl::string_view type_name, absl::string_view id,
     std::unique_ptr<google::protobuf::Any> contents) {
   auto &registered_types = registered_types_.Get();
@@ -188,7 +188,10 @@ absl::StatusOr<NewEntryFingerprintVersion> MetadataManager::CreateEntry(
   metadata.mutable_version_header()->set_top_level_fingerprint(
       ComputeTopLevelFingerprint(metadata.type_namespace_map()));
   BroadcastMetadata(detached_ctx_.get(), metadata.version_header());
-  return NewEntryFingerprintVersion{fingerprint, version};
+  IndexFingerprintVersion index_fingerprint_version;
+  index_fingerprint_version.set_fingerprint(fingerprint);
+  index_fingerprint_version.set_version(version);
+  return index_fingerprint_version;
 }
 
 absl::Status MetadataManager::DeleteEntry(absl::string_view type_name,
--- a/src/coordinator/metadata_manager.h
+++ b/src/coordinator/metadata_manager.h
@@ -30,11 +30,6 @@
 
 namespace valkey_search::coordinator {
 
-struct NewEntryFingerprintVersion {
-  uint64_t fingerprint;
-  uint32_t version;
-};
-
 using FingerprintCallback = absl::AnyInvocable<absl::StatusOr<uint64_t>(
     const google::protobuf::Any &metadata)>;
 using MetadataUpdateCallback = absl::AnyInvocable<absl::Status(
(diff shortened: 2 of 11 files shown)


src/query/create_consistency_check_fanout_operation.cc
```diff
@@ -0,0 +1,125 @@
...
+    return;
+  }
+  if (!resp.exists()) {
+    grpc::Status status =
+        grpc::Status(grpc::StatusCode::INTERNAL, "Index does not exists");
+    OnError(status, coordinator::FanoutErrorType::INDEX_NAME_ERROR, target);
+    return;
+  }
+  if (!schema_fingerprint_.has_value()) {
+    schema_fingerprint_ = resp.schema_fingerprint();
```
KarthikSubbarao: With this PR, How do we handle the failure mode where the fanout based consistency check fails?

src/coordinator/metadata_manager.h
```diff
@@ -30,6 +30,11 @@
 
 namespace valkey_search::coordinator {
 
+struct NewEntryFingerprintVersion {
```
allenss-amazon: Why "New" ?

src/query/cluster_info_fanout_operation.cc
```diff
@@ -53,34 +52,47 @@ void ClusterInfoFanoutOperation::OnResponse(
...
-    return;
+
+  // Determine if we need to call OnError, do it outside the lock
+  // prevent double locking issue in OnError
+  bool should_call_error = false;
+  grpc::Status error_status(grpc::StatusCode::OK, "");
+  coordinator::FanoutErrorType error_type;
+
+  {
+    absl::MutexLock lock(&mutex_);
```
allenss-amazon, line 63: Doesn't this contradict the comment?

src/query/cluster_info_fanout_operation.cc
```diff
@@ -53,34 +52,47 @@ void ClusterInfoFanoutOperation::OnResponse(
...
+    }
+    if (!version_.has_value()) {
+      version_ = resp.version();
+    } else if (version_.value() != resp.version()) {
+      should_call_error = true;
+      error_status =
+          grpc::Status(grpc::StatusCode::INTERNAL,
+                       "Cluster not in a consistent state, please retry.");
+      error_type = coordinator::FanoutErrorType::INCONSISTENT_STATE_ERROR;
+    }
```
allenss-amazon: This isn't treating the fingerprint+version as a pair. You shouldn't have one and not the other. I think you can fix the message to enforce this.

src/query/primary_info_fanout_operation.cc
```diff
@@ -53,34 +52,47 @@ void PrimaryInfoFanoutOperation::OnResponse(
...
+                       "Cluster not in a consistent state, please retry.");
+      error_type = coordinator::FanoutErrorType::INCONSISTENT_STATE_ERROR;
+    }
+    if (!version_.has_value()) {
+      version_ = resp.version();
+    } else if (version_.value() != resp.version()) {
+      should_call_error = true;
+      error_status =
+          grpc::Status(grpc::StatusCode::INTERNAL,
+                       "Cluster not in a consistent state, please retry.");
```
allenss-amazon: same comment about fingerprint/version as a pair.

src/query/cluster_info_fanout_operation.cc
boda26: The lock is applied inside the bracket scope. This scope is just for error detecting. If error is detected, OnError is called outside this scope, where the lock does not apply.

src/coordinator/metadata_manager.cc
```diff
@@ -188,7 +188,10 @@ absl::StatusOr<NewEntryFingerprintVersion> MetadataManager::CreateEntry(
   metadata.mutable_version_header()->set_top_level_fingerprint(
       ComputeTopLevelFingerprint(metadata.type_namespace_map()));
   BroadcastMetadata(detached_ctx_.get(), metadata.version_header());
-  return NewEntryFingerprintVersion{fingerprint, version};
+  IndexFingerprintVersion index_fingerprint_version;
```
allenss-amazon, line 191: I think you could also write this as:
return IndexFingerprintVersion {
   .fingerprint = fingerprint,
   .version = version
};

allenss-amazon (approved):

src/coordinator/metadata_manager.cc
boda26: I defined IndexFingerprintVersion to be a message in coordintaor.proto. Since it is a grpc message, it can only be initialized using setters, and this initialization style does not work.

PR #389, opened 2025-09-17
Fix eviction flakey tests due to valkey not eviction fast enough

yulazariy:
Add waits and retries to overcome Valkey async evictions

### DIFF
--- a/integration/test_eviction.py
+++ b/integration/test_eviction.py
@@ -69,14 +69,18 @@ class TestEviction(ValkeySearchTestCaseBase):
         # Set a relatively small memory limit to trigger eviction
         # This should be enough for initial data but will trigger eviction when we add more
         current_used_memory = client.info("memory")["used_memory"]
-        # new_maxmemory = (
-        #     current_used_memory
-        #     if policy == "noeviction"
-        #     else str(int(int(current_used_memory) * 1.1))
-        # )
-
-        client.config_set("maxmemory", current_used_memory)
+        
+        # For noeviction, set tight limit; for others, allow some buffer
+        if policy == "noeviction":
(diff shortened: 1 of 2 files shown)


integration/test_eviction.py
```diff
@@ -101,12 +96,29 @@ def _test_eviction_behavior(
...
+                    index.load_data(client, next_batch_end, start_index=current_docs)
+                current_docs = next_batch_end
+                
+                # Small delay to allow eviction to process
+                time.sleep(0.01)
+                
+            except OutOfMemoryError:
+                # If we hit OOM, wait a bit for eviction and retry with smaller batch
+                time.sleep(0.1)
+                batch_size = max(10, batch_size // 2)
```
zvi-code: We should fail the test after sufficient number of retries

yulazariy: fixed

integration/test_eviction.py
```diff
@@ -101,12 +96,36 @@ def _test_eviction_behavior(
...
+        # Load data gradually to allow eviction to keep up
+        batch_size = 100
+        current_docs = initial_docs
+        max_oom_retries = 5
+        
+        while current_docs < total_docs:
+            next_batch_end = min(current_docs + batch_size, total_docs)
+            oom_retries = 0
+            
+            while oom_retries < max_oom_retries:
```
zvi-code, line 109: if ` oom_retries >= max_oom_retries` it will loop forever because `current_docs < total_docs` will not change

yulazariy: correct need to raise here

zvi-code:
LGTM

PR #391, opened 2025-09-17
Exact Phrase / Proximity AND base support 

KarthikSubbarao:
Exact Phrase / Proximity AND base support


src/indexes/text/proximity.cc
```diff
@@ -0,0 +1,206 @@
...
+bool ProximityIterator::NextKey() {
+  // On the second call onwards, advance any text iterators that are still
+  // sitting on the old key.
+  auto advance = [&]() -> void {
+    for (auto& iter : iters_) {
+      if (!iter->DoneKeys() && iter->CurrentKey() == current_key_) {
+        iter->NextKey();
+      }
+    }
+  };
```
allenss-amazon, line 50: As we discussed, moving to a heap-based scheme rather than the linear scans will improve the performance here.

src/indexes/text/proximity.cc
```diff
@@ -0,0 +1,206 @@
...
+}
+
+bool ProximityIterator::SeekForwardKey(const Key& target_key) {
+  // If current key is already >= target_key, no need to seek
+  if (current_key_ && current_key_->Str() >= target_key->Str()) {
+    return true;
+  }
+  // Skip all keys less than target_key for all iterators
+  for (auto& iter : iters_) {
+    if (!iter->DoneKeys() && iter->CurrentKey()->Str() < target_key->Str()) {
```
allenss-amazon: Once one iterator hits Done, doesn't that imply that this iterator is Done ?

src/indexes/text/proximity.cc
```diff
@@ -0,0 +1,206 @@
...
+  // 3) Advance all iterators that are strictly behind the current max_key
+  for (auto& iter : iters_) {
+    iter->SeekForwardKey(max_key);
+  }
+  return false;
+}
+
+bool ProximityIterator::SeekForwardKey(const Key& target_key) {
+  // If current key is already >= target_key, no need to seek
+  if (current_key_ && current_key_->Str() >= target_key->Str()) {
```
allenss-amazon: This is comparing the contents of the keys. I thought for lexical ordering that it would be sufficient to just look at the address of the keys... why isn't that true?

src/indexes/text/proximity.h
```diff
@@ -8,85 +8,78 @@
...
-  const InternedStringSet* untracked_keys_;
-  InternedStringPtr current_key_;
   FieldMaskPredicate field_mask_;
-};
+  // Current key/position
+  Key current_key_;
+  std::optional<PositionRange> current_position_;
+  // Vectors used for positional checks
+  mutable std::vector<PositionRange> positions_;
+  mutable std::vector<std::pair<Position, size_t>> pos_with_idx_;
```
allenss-amazon: why mutable?

src/indexes/text/term.cc
```diff
@@ -9,45 +9,126 @@
...
 }
 
-const InternedStringPtr& TermIterator::operator*() const {
-  // Return the current key from the key iterator of the posting object.
-  return key_iter_.GetKey();
+bool TermIterator::NextKey() {
+  if (current_key_) {
+    for (auto& key_iter : key_iterators_) {
+      if (key_iter.IsValid() &&
+          key_iter.GetKey()->Str() == current_key_->Str()) {
```
allenss-amazon: Here, clearly you should only have to compare the key pointers. They are both interned, so therefore they are unique.

src/indexes/text/term.h
```diff
@@ -8,47 +8,63 @@
...
+returns true. Through this process, it "merges" multiple posting iterators.
+- Position iteration happens across all the position iterators, allowing us to
+search for positions in asc order across all the required words within the same
+key and same field. Once no more positions are found, DonePositions returns
+true. Thus, position iteration is a union of all position iterators obtained
+from all the posting iterators that are on the current key and field mask.
 
 */
-class TermIterator : public indexes::EntriesFetcherIteratorBase {
+class TermIterator : public TextIterator {
```
allenss-amazon, line 39: TermIterator seems a bit ambiguous, I'd like to think there's a vector name. Maybe TextIteratorUnion or TextIteratorIntersection, something more descriptive?

src/indexes/text/text_iterator.h
```diff
@@ -0,0 +1,93 @@
...
+  PositionRange() = default;
+  PositionRange(Position s, Position e) : start(s), end(e) {
+    CHECK(start <= end);
+  }
+};
+
+/* Base Class for all Text Search Iterators.
+ * This contract holds for both keys and positions.
+ * // Initializes the TextIterator and primes to the first match of
+ * keys/positions. TextIterator::Init();
```
allenss-amazon: do you mean the constructor? I don't see an ::Init function definition

src/indexes/text/text_iterator.h
```diff
@@ -0,0 +1,93 @@
...
+ * // Tells the caller site if there was no initial match upon init.
+ * // Post init, it tells us whether there are keys remaining that can be
+ * searched for.
+ *. If (!DoneKeys) {
```
allenss-amazon: while (!DoneKeys()) ?

src/indexes/text/text_iterator.h
```diff
@@ -0,0 +1,93 @@
...
+
+  // Key-level iteration
+  // Returns true if there is a match (i.e. `CurrentKey()` is valid) provided
+  // the TextIterator is used as described above. Use `CurrentKey()` to access
+  // the matching document. Otherwise, returns false. Returns false if we have
+  // exhausted all keys, and there are no more search results. In this case no
+  // more calls should be made to `NextKey()`.
+  virtual bool DoneKeys() const = 0;
+  // Returns the current key.
+  // PRECONDITION: !DoneKeys()
```
allenss-amazon: Let's be more specific. PRECONDITION -> ASSERT ?

src/indexes/text/text_iterator.h
```diff
@@ -0,0 +1,93 @@
...
+  virtual const Key& CurrentKey() const = 0;
+  // Advances the key iteration until there is a match OR until we have
+  // exhausted all keys. Returns true when there is a match wrt constraints
+  // (e.g. field, position, inorder, slop, etc). Returns false otherwise. When
+  // false is returned, `CurrentKey()` should no longer be accessed.
+  // This resets the Positions.
+  // PRECONDITION: !DoneKeys()
+  virtual bool NextKey() = 0;
+  // Seeks forward to the first key >= target_key that matches all constraints.
+  // Returns true if such a key is found, false if no more matching keys exist.
```
allenss-amazon, line 68: In case of false, doesn't this imply DoneKeys and DonePositions also? Might be good to add to the description.

src/indexes/text/text_iterator.h
```diff
@@ -0,0 +1,93 @@
...
+  // `CurrentPosition()` to access the matching document. Otherwise, returns
+  // false. Returns false if we have exhausted all keys, and there are no more
+  // search results. In this case no more calls should be made to
+  // `NextPosition()`.
+  virtual bool DonePositions() const = 0;
+  // Returns the current position as a closed interval within a document.
+  // Represents start and end. start == end in all TextIterators except for the
+  // OR Proximity since it can contain a nested proximity block.
+  // PRECONDITION: !DonePositions()
+  virtual PositionRange CurrentPosition() const = 0;
```
allenss-amazon: Curious, CurrentKey returns a reference, but this is a value. Why the asymettry?

src/indexes/text.cc
```diff
@@ -129,31 +127,86 @@ std::unique_ptr<Text::EntriesFetcher> Text::Search(
...
-    auto iter = text_index_->prefix_.GetWordIterator(prefix->GetTextString());
-    auto itr = std::make_unique<text::WildCardIterator>(
-        iter, text::WildCardOperation::kPrefix, field_mask_, untracked_keys_);
-    itr->Next();
-    return itr;
+// Recursive function to return the appropriate text iterator based on the
+// predicate.
+std::unique_ptr<text::TextIterator> Text::EntriesFetcher::BuildTextIterator(
+    const query::TextPredicate* predicate) {
+  if (auto term = dynamic_cast<const query::TermPredicate*>(predicate)) {
```
allenss-amazon: isn't the dynamic_cast redundant? Why any cast at all?

src/indexes/text.cc
```diff
@@ -129,31 +127,86 @@ std::unique_ptr<Text::EntriesFetcher> Text::Search(
...
+        text_index_->prefix_.GetWordIterator(prefix->GetTextString());
+    std::vector<text::Postings::KeyIterator> key_iterators;
+    while (!word_iter.Done()) {
+      key_iterators.emplace_back(word_iter.GetTarget()->GetKeyIterator());
+      word_iter.Next();
+    }
+    return std::make_unique<text::TermIterator>(std::move(key_iterators),
+                                                field_mask_, untracked_keys_);
+  }
+  if (auto suffix = dynamic_cast<const query::SuffixPredicate*>(predicate)) {
```
allenss-amazon: why the dynamic cast? dynamic_cast isn't free.

allenss-amazon (commented):
The comparisons of the keys is consistently using the text of the strings. I don't think this is required. I think it's sufficient to order things by the addresses, i.e., the contents of the InternedStringPtr.

What I'd recommend is that we wrap these comparisons in some functions, i.e., KeyOrder_EQ, KeyOrder_GT, KeyOrder_GE, etc. That way we can centralize this and prove/disprove my assertions that the key ordering is ordering and thus can use the pointers instead of having to actually look at the key text itself (which is going to get fairly expensive).

src/indexes/text/proximity.cc
KarthikSubbarao: Yes, you mean we don't have to check for `iter->DoneKeys()` before calling this function?

That is correct in terms of the semantics

src/indexes/text/proximity.cc
KarthikSubbarao: Yes, this is effort that is planned for subsequent optimizations in the term and proximity iterators

allenss-amazon (approved):

src/indexes/text/term.h
KarthikSubbarao: For ProximityOR, we will have come up with a plan on handling mixed predicate (text / non text).

One option is making all the indexes implement what is called now as the "TextIterator" interface. I need to spend time on looking into this option more to see how it will fit for non text operations. 

But yes, these are core interfaces which we will rename based on how they are used

src/indexes/text/posting.cc
```diff
@@ -171,7 +171,7 @@ class Postings::Impl {
  public:
   bool save_positions_;
   size_t num_text_fields_;
-  std::map<Key, PositionMap, InternedStringPtrLess> key_to_positions_;
+  std::map<Key, PositionMap> key_to_positions_;
```
Aksha1812, line 174: @KarthikSubbarao why was this deleted?

PR #392, opened 2025-09-17
Lexical Ordering for Keys in Postings KeyIterator

Aksha1812:
Implement custom comparator for Postings::KeyIterator, which compares the actual key strings itself and not the addresses to keep it sorted. Necessary for search results to be displayed in lexical order of key strings.


KarthikSubbarao:
Can we fix the formatting?

KarthikSubbarao (approved):

PR #396, opened 2025-09-23
Cluster metadata Validation

Nivesh-01:
We validate wether the metadata has been propogated accross the cluster properly. We check the consistency accross different nodes in a cluster to see if everything is properly migrated.
This PR also has some changes related to https://github.com/valkey-io/valkey-search/pull/395 -> this is going to be merged into main and then in to fulltext but would be a blocking change hence added those same changes in this PR so that the testing can be comprehensive.

### DIFF
--- a/integration/indexes.py
+++ b/integration/indexes.py
@@ -7,7 +7,6 @@ from typing import Tuple, Union
 # from pyparsing import abstractmethod
 import valkey
 from ft_info_parser import FTInfoParser
-from utils import IndexingTestHelper
 import logging, json
 import struct
 
@@ -182,10 +181,9 @@ class Index:
         return d
 
     def info(self, client: valkey.client) -> FTInfoParser:
-        parser = IndexingTestHelper.get_ft_info(client, self.name)
-        if parser is None:
-            raise Exception(f"Failed to get FT.INFO for index {self.name}")
-        return parser
+        res = client.execute_command("FT.INFO", self.name)
+        return FTInfoParser(res)
     
     def backfill_complete(self, client: valkey.client) -> bool:
-        return IndexingTestHelper.is_backfill_complete_on_node(client, self.name)
+        res = self.info(client)
+        return res.backfill_in_progress == 0
--- a/integration/test_ft_metadata_cluster_validation.py
+++ b/integration/test_ft_metadata_cluster_validation.py
@@ -18,21 +18,11 @@ class TestFTMetadataClusterValidation(ValkeySearchClusterTestCase):
     return consistent results across all cluster nodes.
     """
 
-    def wait_for_indexing_complete_on_all_nodes(self, index_name: str, timeout: int = 10):
+    def wait_for_indexing_complete_on_all_nodes(self, index_name: str):
         """Wait for indexing to complete on all cluster nodes."""
         nodes = [self.new_client_for_primary(i) for i in range(self.CLUSTER_SIZE)]
-        success = IndexingTestHelper.wait_for_indexing_complete_on_all_nodes(nodes, index_name, timeout)
-        if not success:
-            raise TimeoutError(f"Indexing did not complete on all nodes within {timeout} seconds for index '{index_name}'")
+        IndexingTestHelper.wait_for_indexing_complete_on_all_nodes(nodes, index_name)
 
-    def get_ft_list_from_all_nodes(self) -> List[List[bytes]]:
-        """Get FT._LIST results from all cluster nodes."""
-        results = []
-        for i in range(self.CLUSTER_SIZE):
-            node = self.new_client_for_primary(i)
-            result = node.execute_command("FT._LIST")
-            results.append(result)
-        return results
 
     def get_ft_info_from_all_nodes(self, index_name: str) -> List[FTInfoParser]:
         """Get FT.INFO results from all cluster nodes."""
@@ -45,18 +35,12 @@ class TestFTMetadataClusterValidation(ValkeySearchClusterTestCase):
 
     def validate_ft_list_consistency(self, expected_indexes: List[str]):
         """Validate that FT._LIST returns consistent results across all nodes."""
-        ft_list_results = self.get_ft_list_from_all_nodes()
-        
-        # Convert all results to sets of strings for comparison
+        # Get normalized results from all nodes using the updated get_ft_list function
         normalized_results = []
-        for result in ft_list_results:
-            normalized = set()
-            for item in result:
-                if isinstance(item, bytes):
-                    normalized.add(item.decode('utf-8'))
-                else:
-                    normalized.add(str(item))
-            normalized_results.append(normalized)
+        for i in range(self.CLUSTER_SIZE):
+            node = self.new_client_for_primary(i)
+            result = IndexingTestHelper.get_ft_list(node)
+            normalized_results.append(result)
         
         # All nodes should have the same set of indexes
         first_result = normalized_results[0]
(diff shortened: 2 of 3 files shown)


BCathcart:
I'm going to push back a bit on the amount of tests and whether they all belong in the same file again.

First off, I'm not sure `test_basic_text_index_metadata_validation` is needed. I think just having `test_complex_text_index_metadata_validation` would be fine. I see that you're adding other field types in the complex test which I think is good because we don't seem to have a FT.CREATE integration test that already does this for the existing fields. Since we're testing the effects of FT.CREATE, I would consider moving it to the existing `test_ft_create.py`.

Next it looks like `test_index_with_data_metadata_validation` has good intentions to test that we can actually use the index cluster-wide. I'm not sure it's actually testing anything right now though. It just asserts that the indexes are in the ready state. I'd approach this more as an end-to-end sanity test where we create the index, ingest docs, and then do queries on them.

`test_error_handling_metadata_validation` tests invalid requests for FT.INFO and FT.CREATE. I'd check if these aren't already covered by any existing testing. From a quick glance it looks like there might already be coverage for FT.INFO in our integration tests and FT.CREATE in unit tests. If you see gaps in the invalid FT.CREATE command coverage, I'd add additional unit tests to `ft_create_test.cc`

integration/test_ft_metadata_cluster_validation.py
```diff
@@ -0,0 +1,283 @@
...
+        ft_info_results = self.get_ft_info_from_all_nodes(index_name)
+        
+        # All nodes should have the index
+        for i, parser in enumerate(ft_info_results):
+            assert parser is not None, f"Index '{index_name}' not found on node {i}"
+        
+        # Compare key metadata fields across all nodes
+        first_parser = ft_info_results[0]
+        
+        for i, parser in enumerate(ft_info_results[1:], 1):
```
BCathcart, line 95: I think it would be good to simplify the logic below. What causes `parser.to_dict() == first_parser.to_dict()` to fail?

integration/test_ft_metadata_cluster_validation.py
```diff
@@ -0,0 +1,283 @@
...
+    def is_indexing_complete_on_node(self, node: Valkey, index_name: str) -> bool:
+        """Check if indexing is complete on a specific node."""
+        try:
+            info_response = node.execute_command("FT.INFO", index_name)
+            parser = FTInfoParser(info_response)
+            return parser.is_backfill_complete() and parser.is_ready()
+        except ResponseError:
+            return False
+
+    def wait_for_indexing_complete_on_all_nodes(self, index_name: str, timeout: int = 10):
```
BCathcart: I'd consider moving this to `utils.py`. This will be a common pattern.

integration/test_ft_metadata_cluster_validation.py
```diff
@@ -0,0 +1,283 @@
...
+            "subcategory": {
+                "type": "TAG",
+                "identifier": "subcategory",
+                "SEPARATOR": ",",
+                "CASESENSITIVE": 1
+            }
+        }
+        self.validate_ft_info_consistency(index_name, expected_attributes)
+        
+        # Additional validation for global index settings
```
BCathcart: This looks like more FT.INFO validation. Can it not also be included in `validate_ft_info_consistency`?

src/indexes/tag.cc
```diff
@@ -167,16 +167,14 @@ absl::StatusOr<bool> Tag::RemoveRecord(const InternedStringPtr& key,
 }
 
 int Tag::RespondWithInfo(ValkeyModuleCtx* ctx) const {
-  auto num_replies = 6;
+  auto num_replies = 8;
```
BCathcart, line 170: These source code changes were merged a while ago. I'd merge them in just to make the diff more clear

integration/test_ft_metadata_cluster_validation.py
Nivesh-01: I do not think there is any function as above, but anywhere there is a comparison between parser and first_parser it is to signify that there is inconsistency between the data in two nodes in a cluster and that the metadata is not being propagated correctly.

BCathcart: The parser parses ft.info output into a dictionary. I'm wondering why we can't just compare the dictionaries directly

integration/test_ft_metadata_cluster_validation.py
Nivesh-01: Done

integration/test_ft_metadata_cluster_validation.py
Nivesh-01: done

integration/test_fulltext.py
```diff
@@ -240,8 +241,8 @@ def test_ft_create_and_info(self):
         assert client.execute_command(*command_args) == b"OK"
         assert b"idx4" in client.execute_command("FT._LIST")
         
-        info_response = client.execute_command("FT.INFO", "idx4")
-        parser = FTInfoParser(info_response)
+        parser = IndexingTestHelper.get_ft_info(client, "idx4")
+        assert parser is not None
```
Nivesh-01, line 245: Should we add the error handling in the utils file itself

src/indexes/tag.cc
Nivesh-01: done

integration/test_fulltext.py
BCathcart: I don't think there is any need to catch exceptions in the utils functions. Just let any exceptions propagate and crash the test unless the caller explicitly catches them

integration/indexes.py
```diff
@@ -181,9 +182,10 @@ def make_data(self, row: int) -> dict[str, Union[str, bytes]]:
...
-        return FTInfoParser(res)
+        parser = IndexingTestHelper.get_ft_info(client, self.name)
+        if parser is None:
+            raise Exception(f"Failed to get FT.INFO for index {self.name}")
+        return parser
     
     def backfill_complete(self, client: valkey.client) -> bool:
-        res = self.info(client)
-        return res.backfill_in_progress == 0
+        return IndexingTestHelper.is_backfill_complete_on_node(client, self.name)
```
BCathcart: These functions don't serve any purpose anymore. Let's just replace the invocations with calls to the `IndexingTestHelper` functions directly

integration/test_ft_metadata_cluster_validation.py
```diff
@@ -0,0 +1,206 @@
...
+            raise TimeoutError(f"Indexing did not complete on all nodes within {timeout} seconds for index '{index_name}'")
+
+    def get_ft_list_from_all_nodes(self) -> List[List[bytes]]:
+        """Get FT._LIST results from all cluster nodes."""
+        results = []
+        for i in range(self.CLUSTER_SIZE):
+            node = self.new_client_for_primary(i)
+            result = node.execute_command("FT._LIST")
+            results.append(result)
+        return results
```
BCathcart: Let's add `get_ft_list()` to `IndexingTestHelper` and  then `get_ft_list_from_all_nodes` and `get_ft_info_from_all_nodes` aren't really needed

integration/test_ft_metadata_cluster_validation.py
```diff
@@ -0,0 +1,206 @@
...
+                        # Try to sort by 'identifier' if present, otherwise keep original order
+                        if all('identifier' in item for item in normalized_items):
+                            normalized[key] = sorted(normalized_items, key=lambda x: x.get('identifier', ''))
+                        else:
+                            normalized[key] = normalized_items
+                    else:
+                        normalized[key] = value
+                except (TypeError, AttributeError):
+                    # If sorting fails, keep original order
+                    normalized[key] = value
```
BCathcart, line 77: We should raise an exception if the function fails to do what it's intended to do. No need to catch it and return the original array

integration/test_ft_metadata_cluster_validation.py
```diff
@@ -0,0 +1,206 @@
...
+        expected_set = set(expected_indexes)
+        assert expected_set.issubset(first_result), f"Expected indexes {expected_set} not found in {first_result}"
+
+    def _normalize_dict_for_comparison(self, data: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Normalize dictionary for comparison by sorting arrays and recursively normalizing nested structures.
+        This ensures consistent comparison regardless of array ordering.
+        """
+        if not isinstance(data, dict):
+            return data
```
BCathcart: We should fail in this case

integration/test_ft_metadata_cluster_validation.py
```diff
@@ -0,0 +1,206 @@
...
+        for key, value in data.items():
+            if isinstance(value, list):
+                # Sort lists that contain comparable elements (strings, numbers)
+                try:
+                    if value and all(isinstance(item, (str, int, float)) for item in value):
+                        normalized[key] = sorted(value)
+                    elif value and all(isinstance(item, dict) for item in value):
+                        # For lists of dicts, normalize each dict and sort by a stable key if possible
+                        normalized_items = [self._normalize_dict_for_comparison(item) for item in value]
+                        # Try to sort by 'identifier' if present, otherwise keep original order
```
BCathcart: Why are we looking for 'identifier' explicitly here?

integration/test_ft_metadata_cluster_validation.py
BCathcart: We can probably get rid of this function now or get rid of `IndexingTestHelper.wait_for_indexing_complete_on_all_nodes` and let the test loop call `is_indexing_complete_on_node` for each node itself

integration/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    def is_indexing_complete_on_node(node: Valkey, index_name: str) -> bool:
+        """
+        Check if indexing is complete on a specific node.
+        
+        This is the most comprehensive check that verifies both backfill completion
+        and that the index is in ready state.
+        
+        """
+        parser = IndexingTestHelper.get_ft_info(node, index_name)
+        return parser.is_backfill_complete() and parser.is_ready() if parser else False
```
BCathcart: Let's get rid of the error handling in these functions. For example if False is returned here, the caller doesn't know if backfilling is in progress or there is a problem fetching FT.INFO.

integration/utils.py
```diff
@@ -0,0 +1,71 @@
...
+        except ResponseError:
+            return False
+    
+    @staticmethod
+    def wait_for_indexing_complete_on_all_nodes(nodes: list, index_name: str, timeout: int = 10) -> bool:
+        """Wait for indexing to complete on all provided nodes."""
+        import time
+        start_time = time.time()
+        
+        while time.time() - start_time < timeout:
```
BCathcart: There are common utility functions for waiting for a condition. I think you've used them in a past PR for reference.

integration/test_fulltext.py
Nivesh-01: ok

integration/indexes.py
Nivesh-01: leaving as it was before made be allen ---> might need a followup later

integration/test_ft_metadata_cluster_validation.py
Nivesh-01: done

PR #399, opened 2025-09-30
Fix integration framework for JSON inex creation

yanamolo:
Fixed integration tests framework for JSON index creation, specifically to work with non-vector indexes. Also added a test for non-vector indexes creation for JSON, including search operation.


yanamolo:
Will rename to KeyDataType

allenss-amazon (approved):

PR #401, opened 2025-10-01
Revise Save/Restore for true pit snapshot.

### LINKED ISSUE #41: Avoid Backfill After RDB is Loaded
Currently, after an RDB load, the backfill process is triggered to ensure consistency between indexes and the keyspace. Although measurements show that the backfill completes relatively quickly, even for large keyspaces, achieving a consistent state faster remains valuable. To eliminate the need for the backfill process after RDB is loaded, the following improvements are required:

1. **Serialize Non-Vector Indexes:** Currently, only vector indexes are serialized. Non-vector indexes must also be included to avoid discrepancies.
2. **Serialize the Mutation Tracker:** This ensures that pending mutations are preserved and applied consistently after the load.
3. **Serialize the Multi-Exec Queue:** Capturing the multi-exec queue state will prevent loss of transactional operations during recovery.

allenss-amazon:
Revises the save/restore logic to support saving of all index data and pending mutation queue as discussed earlier.

This new logic is fully backward and forward compatible. RDB files written using the previous binaries (aka V1) will still be loaded and subject to the same stale data correction and forced backfill logic as existed before this PR. In other words, for RDB files written with the old code, the new code handles them exactly the same.

New format files (aka V2) contain additional data that will be ignored by old code. In other words, V2 RDB files can still be loaded into prior code which will ignore the extra data, i.e., they are treated like V1 files and be subjected to stale data correction and forced backfill.

V2 files which are loaded into the new code are not subjected to stale data correction and forced backfill. Once a V2 file is loaded it can "open for business" immediately. 

The V2 format contains an additional section for the contents of non-vector indexes (Tags, Numeric). These sections contain the key and attribute-value for each entry in the index. This external format differs from the internal format, allowing the internal data structures to be freely optimized in the future.

The V2 format also has a section for the mutation queue and index backfill status. This additional section records the keys that were in the mutation queue at the time of the save as well as the index backfill status (complete or incomplete). This allows the recreation of the mutation queue and backfill status (backfill needed or not) to match the state at the time of the save. (One small quirk is that the saved mutation queue doesn't bother to record whether the saved keys are in the mutation queue due to a backfill or not. This behavior should not affect the client-visible semantics, but would affect internal counters for the ingestion pipeline as they would be mislead by the source of the mutation.

Two controls are provided to override the normal behavior of this code.

The configurable: ```search.rdb_write_v2``` (default "yes") controls whether a generated RDB file is in the V1 or V2 format. 

The configurable: ```search.rdb_read_v2``` (default "yes") controls whether the V2 data is used. Setting this to "no" will force the code to treat a V2 file as a V1 file, i.e., to ignore the extra V1 data.

FIxes #41

### DIFF
--- a/integration/test_saverestore.py
+++ b/integration/test_saverestore.py
@@ -239,11 +239,11 @@ class TestMutationQueue(ValkeySearchTestCaseDebugMode):
         assert reads == [len(records)]
 
     def get_pausepoint(self, p):
+        result = self.client.execute_command(f"ft._debug pausepoint test {p}")
         try:
-            result = self.client.execute_command(f"ft._debug pausepoint test {p}")
             return int(result)
-        except ResponseError:
-            assert "not found" in result
+        except ValueError:
+            assert b"not found" in result
             return 0
     
     def test_multi_exec_queue(self):


src/rdb_serialization.h
```diff
@@ -336,6 +371,15 @@ class RDBChunkOutputStream : public hnswlib::OutputStream {
     }
   }
   absl::Status SaveChunk(const char *data, size_t len) override;
+  absl::Status SaveString(const std::string_view s) {
+    return SaveChunk(s.data(), s.size());
+  }
+  template <typename T, std::enable_if_t<std::is_trivial<T>::value &&
```
Aksha1812, line 364: Do we need enable_if_t ? I saw it limits this function only to certain types. What types does it limit the template to. In future are we expected to serialize any complex information ?

testing/common.h
```diff
@@ -323,10 +326,10 @@ class MockThreadPool : public vmsdk::ThreadPool {
   MockThreadPool(const std::string& name, size_t num_threads)
       : vmsdk::ThreadPool(name, num_threads) {
     ON_CALL(*this, Schedule(testing::_, testing::_))
-        .WillByDefault(testing::Invoke(
+        .WillByDefault(
```
Aksha1812, line 330: I had seen this was causing build warnings , glad this is being fixed, what was the actual issue?

src/index_schema.cc
```diff
@@ -824,49 +915,94 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+
+  return absl::OkStatus();
+}
+
+absl::Status IndexSchema::SaveMutationQueue(RDBChunkOutputStream out) const {
+  VMSDK_LOG(NOTICE, nullptr) << "Writing Mutation Queue, records = "
+                             << tracked_mutated_records_.size();
+  VMSDK_RETURN_IF_ERROR(out.SaveObject(tracked_mutated_records_.size()));
+  for (const auto &[key, value] : tracked_mutated_records_) {
+    VMSDK_RETURN_IF_ERROR(out.SaveString(key->Str()));
```
Aksha1812, line 1100: We don't save the value of records in the mutation queue?

src/indexes/numeric.cc
```diff
@@ -259,4 +272,78 @@ uint64_t Numeric::GetRecordCount() const {
...
+                             << " Untracked:" << untracked_key_count;
+
+  for (const auto& key_ptr : untracked_keys_) {
+    VMSDK_RETURN_IF_ERROR(output.SaveString(key_ptr->Str()));
+    //
+    // Ensure no duplicate keys (which would cause a load error)
+    //
+    if (tracked_keys_.contains(key_ptr)) {
+      DCHECK(false);
+      return absl::InternalError("Numeric field save duplicate key");
```
Aksha1812: what happens when tracked/ untracked keys overlap between loading different indexes. I assume it would be handled automatically while backfilling?

integration/test_saverestore.py
```diff
@@ -0,0 +1,212 @@
...
+from valkey_search_test_case import ValkeySearchTestCaseBase, ValkeySearchTestCaseDebugMode
+from valkeytestframework.conftest import resource_port_tracker
+from indexes import *
+import pytest
+import logging
+from util import waiters
+import threading
+from ft_info_parser import FTInfoParser
+
+index = Index("index", [Vector("v", 3, type="HNSW", m=2, efc=1), Numeric("n"), Tag("t")])
```
Aksha1812, line 20: maybe we could add a flat vector field as well  just to make testing more extensive. also maybe adding some deletes, updates and then saving restoring would also remove any doubts

src/index_schema.cc
```diff
@@ -889,58 +1025,94 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::LoadFromRDB(
...
+                SkipSupplementalContent(supplemental_iter, "mutation queue"));
+          } else {
+            if (index_schema) {
+              VMSDK_RETURN_IF_ERROR(index_schema->LoadMutationQueue(
+                  ctx, RDBChunkInputStream(supplemental_iter.IterateChunks())));
+              bool backfilling =
+                  supplemental_content->mutation_queue_header().backfilling();
+              if (!backfilling) {
+                VMSDK_LOG(DEBUG, ctx) << "Backfill suppressed.";
+                index_schema->backfill_job_.Get() = std::nullopt;
```
Aksha1812, line 1251: I imagine some consistency check here, but i guess that is a separate task

src/rdb_serialization.h
allenss-amazon: This implementation won't handle complex information. If the need for a more sophisticated serialization framework comes up we'll tackle it then. The enable_if_t is seat belts for developers.

testing/common.h
allenss-amazon: Not sure why this wasn't caught earlier. I believe that testing::Invoke predates lambda functions and isn't needed for them.

src/index_schema.cc
allenss-amazon: The values are not needed because they are "by definition" duplicates of what is in the main database.

src/indexes/numeric.cc
allenss-amazon: In theory tracked/untracked are disjoint sets. In the save/restore world it's better to crash the save than the restore. In other words, let's not write an invalid RDB file.

integration/test_saverestore.py
allenss-amazon: The PR doesn't change the save/restore of a vector index. It does suppress the post load scan for stale keys. But this logic isn't affected by HNSW vs FLAT.

src/index_schema.cc
allenss-amazon: Yes, in a perfect world we'd have additional checking code that could be enabled/disabled.

src/rdb_serialization.h
```diff
@@ -306,6 +308,24 @@ class RDBChunkInputStream : public hnswlib::InputStream {
...
+    return *str.release();
+  }
+
+  template <typename T, std::enable_if_t<std::is_trivial<T>::value &&
+                                             std::is_standard_layout<T>::value,
+                                         bool> = true>
+  absl::StatusOr<T> LoadObject() {
+    VMSDK_ASSIGN_OR_RETURN(auto buffer, LoadChunk());
+    if (buffer->size() != sizeof(T)) {
+      return absl::InternalError("Mismatched size protocol error");
```
yairgott, line 324: Consider asserting using CHECK or DCHECK

src/index_schema.cc
```diff
@@ -837,49 +903,186 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+      oracle_index->GetTrackedKeyCount() + oracle_index->GetUnTrackedKeyCount();
+  //
+  // Now, make sure all the other indexes have the same key count, except for
+  // vector indexes which may have less keys
+  //
+  for (const auto &[name, attr] : attributes_) {
+    auto idx = attr.GetIndex();
+    size_t cnt = idx->GetTrackedKeyCount() + idx->GetUnTrackedKeyCount();
+    if (cnt != oracle_key_count) {
+      if (IsVectorIndex(idx) && cnt < oracle_key_count) {
```
yairgott: shouldn't this be:
```
if ((IsVectorIndex(idx) && cnt <= oracle_key_count) || (!IsVectorIndex(idx) && cnt == oracle_key_count)) {
  continue;
```

src/indexes/numeric.h
```diff
@@ -93,18 +93,21 @@ class Numeric : public IndexBase {
...
-  inline void ForEachTrackedKey(
-      absl::AnyInvocable<void(const InternedStringPtr&)> fn) const override {
-    absl::MutexLock lock(&index_mutex_);
-    for (const auto& [key, _] : tracked_keys_) {
-      fn(key);
-    }
-  }
-  uint64_t GetRecordCount() const override;
+
+  size_t GetTrackedKeyCount() const override;
```
yairgott: Here and other places, including other index classes, use ABSL_LOCKS_EXCLUDED(index_mutex_)

src/index_schema.cc
```diff
@@ -901,59 +1104,76 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::LoadFromRDB(
...
+            SUPPLEMENTAL_CONTENT_INDEX_EXTENSION: {
+          VMSDK_LOG(NOTICE, nullptr) << "Loading Mutation Queue";
+          if (!RDBReadV2()) {
+            VMSDK_RETURN_IF_ERROR(
+                SkipSupplementalContent(supplemental_iter, "mutation queue"));
+          } else {
+            if (index_schema) {
+              VMSDK_RETURN_IF_ERROR(index_schema->LoadIndexExtension(
+                  ctx, RDBChunkInputStream(supplemental_iter.IterateChunks())));
+              bool backfilling =
```
yairgott: nit: there is no point in defining a variable if used only once.

src/index_schema.cc
```diff
@@ -837,49 +903,186 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+  auto oracle_index = attributes_.begin()->second.GetIndex();
+  auto oracle_name = attributes_.begin()->first;
+  for (const auto &attribute : attributes_) {
+    if (!IsVectorIndex(attribute.second.GetIndex())) {
+      oracle_index = attribute.second.GetIndex();
+      oracle_name = attribute.first;
+      break;
+    }
+  }
+  size_t oracle_key_count =
```
yairgott, line 990: shouldn't we break if non-vector index if found?

src/index_schema.cc
```diff
@@ -837,49 +903,186 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+          return absl::OkStatus();
+        }
+      };
+      auto status1 = larger_index->ForEachTrackedKey(key_check);
+      if (!status1.ok()) {
+        status = status1;
+      }
+      auto status2 = larger_index->ForEachUnTrackedKey(key_check);
+      if (!status2.ok()) {
+        status = status1;
```
yairgott: should this be:
`status = status2;` ?

src/index_schema.cc
```diff
@@ -837,49 +903,186 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+}
+
+absl::Status IndexSchema::SaveIndexExtension(RDBChunkOutputStream out) const {
+  if (RDBValidateOnWrite()) {
+    VMSDK_RETURN_IF_ERROR(ValidateIndex());
+  }
+  //
+  // Need to find an attribute index that has the right tracked/untracked
+  // keys. Any non-vector index will do. But it there are only vector
+  // indexes we will use that.
```
yairgott: what is the value of doing so if there are no non-vector indexes?

src/rdb_serialization.h
allenss-amazon: I don't think crashing is the correct response to an invalid RDB file.

src/index_schema.cc
allenss-amazon: Well, they are equivalent. How about this?

```
    if (IsVectorIndex(idx) ? cnt <= oracle_key_count
                           : cnt == oracle_key_count) {
      continue;
    }
```

src/index_schema.cc
allenss-amazon: Which variable? backfilling is used in the next line.

src/index_schema.cc
allenss-amazon: Isn't that exactly what this does?

src/index_schema.cc
allenss-amazon: good catch.

src/index_schema.cc
allenss-amazon: Hmm, good point. Worse, the list will be wrong if there are multiple vector indexes across disjoint sets of keys. I think you're correct that it can be skipped.

allenss-amazon:
> I might had missed it, but I haven't seen that the queued multi/exec are being serialized as well.

Good catch.

src/index_schema.cc
yairgott: `oracle_index` would point to the first vector index.

src/index_schema.cc
yairgott: Right but my point is that it is used just by the next line. I'm suggesting to change the next line to:
```
if (!supplemental_content->mutation_queue_header().backfilling()) {
```

src/rdb_serialization.h
yairgott: I suggest to use DCHECK such that we'll get a clear signal for such in the lab.

src/index_schema.cc
yairgott: SG!

src/index_schema.cc
allenss-amazon: Fixed in latest update. No validation when there aren't any non-vector fields. tests were extended to include vector-only, non-vector-only and mixed indexes.

src/index_schema.cc
allenss-amazon: I suspect I was printing it earlier. But I changed it as suggested.

src/rdb_serialization.h
allenss-amazon: Ok, I can support that. I thought you were complaining about the enable_if ugliness. Done.

src/index_schema.cc
```diff
@@ -837,49 +905,237 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+
+absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
+                                             RDBChunkInputStream input) {
+  VMSDK_ASSIGN_OR_RETURN(size_t key_count, input.LoadObject<size_t>());
+  rdb_load_keys.Increment(key_count);
+  VMSDK_LOG(DEBUG, ctx) << "Loading Index Extension, keys = " << key_count;
+  for (size_t i = 0; i < key_count; ++i) {
+    VMSDK_ASSIGN_OR_RETURN(auto keyname_str, input.LoadString());
+    auto keyname = vmsdk::MakeUniqueValkeyString(keyname_str);
+    ProcessKeyspaceNotification(ctx, keyname.get(), false);
```
yairgott, line 1130: Here and other calls to `ProcessKeyspaceNotification`, It's important to preserve the `is backfill original entry value (third parameter) to avoid situations that the backfill mutations causing huge delay for client driven mutations.

src/rdb_serialization.h
yairgott: Kindly reminder

vmsdk/src/utils.cc
```diff
@@ -315,4 +315,17 @@ std::optional<std::string> JsonUnquote(absl::string_view sv) {
   return result;
 }
 
+std::string StringToHex(std::string_view s) {
```
yairgott, line 318: This seems not in-used

src/index_schema.cc
yairgott: Kindly reminder

vmsdk/src/utils.cc
allenss-amazon: True, I used it during debugging and it's not used now. But, it's useful going forward for debugging. so no reason to eliminate it.

src/index_schema.cc
allenss-amazon: Fixed, both from_backill and from_multi are saved, from_backfill is used during a restore. Currently from_multi is unused.

src/index_schema.cc
```diff
@@ -837,48 +916,243 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+      return out.SaveString(key->Str());
+    };
+    VMSDK_RETURN_IF_ERROR(index->ForEachTrackedKey(write_a_key));
+    VMSDK_RETURN_IF_ERROR(index->ForEachUnTrackedKey(write_a_key));
+    CHECK(key_count == 0) << "Key count mismatch for index " << GetName();
+  }
+  //
+  // Write out the mutation queue entries
+  //
+  VMSDK_LOG(NOTICE, nullptr) << "Writing Mutation Queue, records = "
```
yairgott: shouldn't below be gated by a check of `RDBWriteV2()`?

src/index_schema.cc
```diff
@@ -837,48 +916,243 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+  for (const auto &key : multi_mutations_.Get().keys) {
+    CHECK(tracked_mutated_records_.find(key) != tracked_mutated_records_.end());
+    VMSDK_RETURN_IF_ERROR(out.SaveString(key->Str()));
+  }
+  return absl::OkStatus();
+}
+
+absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
+                                             RDBChunkInputStream input) {
+  VMSDK_ASSIGN_OR_RETURN(size_t key_count, input.LoadObject<size_t>());
```
yairgott, line 1124: Consider adding assert to validate `RDBReadV2()` for safety

src/index_schema.cc
```diff
@@ -983,8 +1273,33 @@ void IndexSchema::OnSwapDB(ValkeyModuleSwapDbInfo *swap_db_info) {
 }
 
 void IndexSchema::OnLoadingEnded(ValkeyModuleCtx *ctx) {
-  // Clean up any potentially stale index entries that can arise from pending
-  // record deletions being lost during RDB save.
+  if (loaded_v2_) {
```
yairgott, line 1297: 1. `loaded_v2_` is needed to reset here or at the beginning of loading.
2.  The indentation should be avoided. Suggested code:
```
auto cleanup = absl::Cleanup([] {
        loaded_v2_ = false;
});
if (!loaded_v2_) {
    return;
} 
... 
```

src/index_schema.cc
```diff
@@ -983,8 +1273,33 @@ void IndexSchema::OnSwapDB(ValkeyModuleSwapDbInfo *swap_db_info) {
...
+    VMSDK_LOG(NOTICE, ctx) << "RDB load completed, "
+                           << " Mutation Queue contains "
+                           << tracked_mutated_records_.size() << " entries."
+                           << (backfill_job_.Get().has_value()
+                                   ? " Backfill still required."
+                                   : " Backfill not needed.");
+    while (DrainMutationQueue() &&
+           std::any_of(tracked_mutated_records_.begin(),
```
yairgott: wouldn't this be very in-efficient if the queue contains a lot of backfill mutations?

src/index_schema.cc
```diff
@@ -837,48 +916,243 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+  return absl::OkStatus();
+}
+
+absl::Status IndexSchema::ValidateIndex() const {
+  absl::Status status = absl::OkStatus();
+  //
+  // Find a non-vector index as the oracle
+  // If all indexes are vector indexes, no validation is needed
+  //
+  std::shared_ptr<indexes::IndexBase> oracle_index = nullptr;
```
yairgott: nit: no need to initialize a shared_ptr with `nullptr`

src/index_schema.cc
allenss-amazon: The entire function is predicated by RDBWriteV2.

src/index_schema.cc
allenss-amazon: sure, why not.

src/index_schema.cc
allenss-amazon: I'll reset it here. FWIW, I believe the reset is not needed as it's set in the constructor and any future reload will reconstruct the IndexSchema.

src/index_schema.cc
allenss-amazon: As discussed, will revise to not save/restore backfills when index is in backfilling state.

src/index_schema.cc
allenss-amazon: ack

src/index_schema.cc
allenss-amazon: Updated in current code. New code skips backfilling records UNLESS the index is marked as no longer backfilling. This is validated in new test. Also, new test to validate that a write of a pending backfill is correctly converted to a non-backfilling write as properly gets saved.

src/index_schema.cc
```diff
@@ -837,49 +918,261 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
...
+                                           [](const auto &entry) {
+                                             return !entry.second.from_backfill;
+                                           });
+  VMSDK_LOG(NOTICE, nullptr)
+      << "Writing mutation queue records = " << count
+      << " Total queue:" << tracked_mutated_records_.size();
+  VMSDK_RETURN_IF_ERROR(out.SaveObject(count));
+  rdb_save_mutation_entries.Increment(count);
+  for (const auto &[key, value] : tracked_mutated_records_) {
+    if (IsBackfillInProgress() && value.from_backfill) {
```
yairgott, line 1097: I think the `IsBackfillInProgress()` should be omitted. Consider that a backfill could be done but the `tracked_mutated_records_` may still have a very high volume of backfill entries.

allenss-amazon: When I wrote this, I was guarding against one specific scenario. It wasn't clear to me from reading the code if this is possible, but the cost of guarding against it isn't really that high. Yes, there may be thousands of keys in the queue, but the scan of the queue only happens once and it's done by a background thread (post fork). 

The scenario I was considering was right at the end of a backfill. It wasn't clear to me whether IsBackfillInProgress flips to false when the last key is scanned (and the backfills have been submitted) or when the last backfill completes ingestion. So, defensively, I assumed the former, i.e., that there is a small window of time where the index is marked as not backfilling, but there are keys marked as backfilling still in the queue. In this case, the backfilling keys are preserved and thus will get updated on a reload (just as the !Backfilling intends). Of course, if it's the later, i.e., IsBackfillInProgress correlated to the completion of the mutation queue, then this scan is not needed.

PR #403, opened 2025-10-02
Add more integration tests to fanout

boda26:
Add more integration tests to fanout to cover index name not found error failure cases and detect retry on that situation

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,7 @@
     "clangd.path": "/usr/bin/clangd",
     "C_Cpp.codeAnalysis.runAutomatically": true,
     "editor.formatOnSave": false,
-    "C_Cpp.intelliSenseEngine": "default",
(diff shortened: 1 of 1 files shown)


integration/test_ft_dropindex_consistency.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+        thread0.start()
+        thread1.start()
+        # Wait for both threads to complete
+        thread0.join()
+        thread1.join()
+
+        res1, res2 = results.values()
+        assert "OK" in str(res1)
+        assert "OK" in str(res2)
```
allenss-amazon: the barrier serves to synchronize the clients activities. But doesn't provide any synchronization at the server side. How do you know that this test is exercising the "simultaneous" case that we want to test? Seems to me that this test would report success with almost any sequencing of events on the server side.

PR #406, opened 2025-10-06
Stemming - Ingestion and Search handling

Aksha1812:
## Ingestion 

```
happy and happiness stem to -> happi
personal and person stem to -> person 

→ payload: {doc:1 happy, 
            doc:2 happiness, 
            doc:3 happi,
            doc:4 personal,
            doc:5 person
            }
            
Prefix Trie path:                  ()
                                /       \
                           happ      person   
                          /     \          \
                      iness      y          al  
                  
Posting objects to be stored :

 happy : Posting {doc:1},
 happi : Posting {doc:3} ,
 happiness : Posting {doc:2}, 
 personal: Posting {doc:4}, 
 person: Posting {doc:5}, 

Stem Trie path:                    ()
                                /       \
                           happi      person   

Stem Targets to be stored :

happi -> {happy, happiness} 
person -> {personal}

(even 'happi' stems to 'happi' but mapping a word to itslef doesnt make sense)
```

## Search

```
A non exact term search should return both original word and all stem parents of the stemmed root of original word

For instance for query FT.SEARCH testindex '@desc:happi'

Results should include the original word + All stem targets of word happi + Exact Stem root if exists



```

Other things added with this PR

- Addressed comments onGetLanguageString() function

### DIFF
--- a/src/indexes/text/term.cc
+++ b/src/indexes/text/term.cc
@@ -6,47 +6,127 @@
  */
 
 #include "src/indexes/text/term.h"
+#include "src/indexes/text/lexer.h"
+#include "src/indexes/text/text_index.h"
 
 namespace valkey_search::indexes::text {
 
 TermIterator::TermIterator(const WordIterator& word,
                            const absl::string_view data,
                            const FieldMaskPredicate field_mask,
-                           const InternedStringSet* untracked_keys)
+                           const InternedStringSet* untracked_keys,
+                           bool stemming_enabled,
+                           std::shared_ptr<TextIndexSchema> text_index_schema,
+                           uint32_t min_stem_size,
+                           std::shared_ptr<TextIndex> text_index)
     : word_(word),
       data_(data),
       field_mask_(field_mask),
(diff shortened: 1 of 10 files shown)


src/indexes/text/term.cc
```diff
@@ -6,47 +6,127 @@
...
-  } else {
+    begin_ = false;
+  } else if (key_iter_.IsValid()) {
     key_iter_.NextKey();
   }
-  // Advance until we find a valid key or reach the end
-  while (!Done() && !key_iter_.ContainsFields(field_mask_)) {
+  
+  if (stemming_enabled_) {
+    while (stem_target_ && stem_word_iter_ != stem_target_->end()) {
```
KarthikSubbarao, line 86: For stemming, it would be good to consider a separate word iterator. This code should be re-used through the GetStemmedWordIterator and not embedded in each search operation


Also, as discussed, we only will need to use this StemmedWordIterator during the initialization of the search operations once the open PR is merged in

PR #410, opened 2025-10-09
Full support for DB numbers other than zero in cluster mode.

### LINKED ISSUE #204: Support multi-db in CME
Seems like current code is assuming only single db in CME.

some pointers:
https://github.com/valkey-io/valkey-search/blob/b91e54993e9468d1ebce9735eb91013710e1140b/src/schema_manager.cc#L297

https://github.com/valkey-io/valkey-search/blob/b91e54993e9468d1ebce9735eb91013710e1140b/src/schema_manager.cc#L172
https://github.com/valkey-io/valkey-search/blob/b91e54993e9468d1ebce9735eb91013710e1140b/src/schema_manager.cc#L225

allenss-amazon:
fixes #204

### DIFF
--- a/src/coordinator/metadata_manager.cc
+++ b/src/coordinator/metadata_manager.cc
@@ -299,10 +299,10 @@ std::unique_ptr<GlobalMetadata> MetadataManager::GetGlobalMetadata() {
 }
 
 void MetadataManager::RegisterType(absl::string_view type_name,
-                                   vmsdk::ValkeyVersion encoding_version,
                                    FingerprintCallback fingerprint_callback,
                                    MetadataUpdateCallback callback,
-                                   MinVersionCallback min_version_callback) {
+                                   MinVersionCallback min_version_callback,
+                                   vmsdk::ValkeyVersion encoding_version) {
   auto insert_result =
       registered_types_.Get().insert(std::pair<std::string, RegisteredType>{
           type_name,
@@ -312,7 +312,7 @@ void MetadataManager::RegisterType(absl::string_view type_name,
               .update_callback = std::move(callback),
               .min_version_callback = std::move(min_version_callback)}});
   VMSDK_LOG(DEBUG, nullptr) << "Registering type: " << type_name;
-  CHECK(insert_result.second);
+  CHECK(insert_result.second) << "Type already registered: " << type_name;
 }
 
 void MetadataManager::BroadcastMetadata(ValkeyModuleCtx *ctx) {
--- a/src/coordinator/metadata_manager.h
+++ b/src/coordinator/metadata_manager.h
@@ -24,6 +24,7 @@
 #include "src/coordinator/client_pool.h"
 #include "src/coordinator/coordinator.pb.h"
 #include "src/rdb_serialization.h"
+#include "version.h"
 #include "vmsdk/src/command_parser.h"
 #include "vmsdk/src/managed_pointers.h"
 #include "vmsdk/src/utils.h"
@@ -96,17 +97,16 @@ class MetadataManager {
   // accept updates to that type both locally and over the cluster bus.
   //
   // * type_name should be a unique string identifying the type.
-  // * encoding_version should be bumped any time the underlying metadata format
-  // is changed.
   // * fingerprint_callback should be a function for computing the fingerprint
   // of the metadata for the given encoding version. This function can only
   // change when the encoding version is bumped.
   // * update_callback will be called whenever the metadata is updated.
+  // * encoding_version should only be set in unit tests.
   void RegisterType(absl::string_view type_name,
-                    vmsdk::ValkeyVersion encoding_version,
                     FingerprintCallback fingerprint_callback,
                     MetadataUpdateCallback callback,
-                    MinVersionCallback min_version_callback);
+                    MinVersionCallback min_version_callback,
+                    vmsdk::ValkeyVersion encoding_version = kModuleVersion);
 
   void BroadcastMetadata(ValkeyModuleCtx *ctx);
 
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -130,7 +130,7 @@ SchemaManager::SchemaManager(
               }});
   if (coordinator_enabled) {
     coordinator::MetadataManager::Instance().RegisterType(
-        kSchemaManagerMetadataTypeName, kModuleVersion, ComputeFingerprint,
+        kSchemaManagerMetadataTypeName, ComputeFingerprint,
         [this](const coordinator::ObjName &obj_name,
                const google::protobuf::Any *metadata, uint64_t fingerprint,
                uint32_t version) -> absl::Status {
--- a/testing/coordinator/metadata_manager_test.cc
+++ b/testing/coordinator/metadata_manager_test.cc
@@ -111,7 +111,7 @@ TEST_P(EntryOperationTest, TestEntryOperations) {
   std::vector<CallbackResult> callbacks_tracker;
   for (auto& type_to_register : test_case.types_to_register) {
     test_metadata_manager_->RegisterType(
-        type_to_register.type_name, type_to_register.encoding_version,
+        type_to_register.type_name,
         [&](const google::protobuf::Any& metadata) -> absl::StatusOr<uint64_t> {
           return type_to_register.fingerprint_to_return;
         },
@@ -126,7 +126,7 @@ TEST_P(EntryOperationTest, TestEntryOperations) {
           callbacks_tracker.push_back(std::move(callback_result));
           return type_to_register.status_to_return;
         },
-        [](auto) { return 1; });
+        [](auto) { return 1; }, type_to_register.encoding_version);
   }
   if (test_case.expect_num_broadcasts > 0) {
     EXPECT_CALL(*kMockValkeyModule,
@@ -545,7 +545,7 @@ TEST_P(MetadataManagerReconciliationTest, TestReconciliation) {
   std::vector<CallbackResult> callbacks_tracker;
   for (const auto& type_to_register : test_case.types_to_register) {
     test_metadata_manager_->RegisterType(
-        type_to_register.type_name, type_to_register.encoding_version,
+        type_to_register.type_name,
         [&](const google::protobuf::Any& metadata) -> absl::StatusOr<uint64_t> {
           return type_to_register.fingerprint_to_return;
         },
@@ -559,7 +559,7 @@ TEST_P(MetadataManagerReconciliationTest, TestReconciliation) {
           });
           return type_to_register.status_to_return;
         },
-        [](auto) { return kModuleVersion; });
+        [](auto) { return kModuleVersion; }, type_to_register.encoding_version);
   }
 
   if (test_case.expect_broadcast) {
@@ -1941,14 +1941,14 @@ TEST_F(MetadataManagerTimestampTest,
 
   // Register a type with a failing callback
   test_metadata_manager_->RegisterType(
-      "my_type", {0, 0, 1},
+      "my_type",
       [](const google::protobuf::Any& metadata) -> absl::StatusOr<uint64_t> {
         return 1234;
       },
       [](const ObjName& obj_name, const google::protobuf::Any* metadata,
          uint64_t fingerprint,
          uint32_t version) { return absl::InternalError("Callback failed"); },
-      [](auto) { return kModuleVersion; });
+      [](auto) { return kModuleVersion; }, {0, 0, 1});
 
   // Reconciliation should fail due to callback failure
   auto status =


src/schema_manager.h
```diff
@@ -36,6 +36,45 @@ namespace valkey_search {
 constexpr absl::string_view kSchemaManagerMetadataTypeName{"vs_index_schema"};
 constexpr uint32_t kMetadataEncodingVersion = 1;
 
+/*
+
+The SchemaManager uses the MetadataManager to manage a single name space of
+indexes, i.e., map<string, GlobalMetadataEntry>.
```
murphyjacob4: Appreciate the thought going into this. The interoperability is a headache

One question: when someone creates an index in db_num > 0 on Valkey 9, how would a shard running on Valkey 8 behave? I am guessing it would crash, right? I guess that behavior is fine, but we should document this

Another option I guess is that we have namespaces in the metadata manager. Right now we use "vs_index_schema". We could move to "vs_index_schema2"? Nodes on the old versions would just not get the index definitions for those created with vs_index_schema2, which should be fine, and we can register for callbacks of "vs_index_schema" in the new node for backwards compat. Probably a bit more work, but it would prevent the Valkey 8.0 nodes from crashing in such cases

src/coordinator/metadata_manager.cc
```diff
@@ -23,22 +23,44 @@
...
+With Valkey 9, the introduction of DB num into CME creates a desire for a
+three-level hierarchy: <type-name, db-num, object-name>. This 3-level
+internal namespace is mapped into an external two-level namespace to provide
+backward and some degree of forward compatibility.
+
+The mapping is done by manipulating the object name. It is known that
+pre-Valkey 9 object names cannot contain a valid hash-tag. This provides the
+necessary information to distinguish between an object name and an encoded
+<db_num, object_name> pair. Objects with a db_num of 0 are stored the same
+way as pre-Valkey 9 objects, providing backward compatibility.
```
murphyjacob4: Now that we have the encoding version added - could we just add a new field to the IndexSchema proto (e.g. `clustered_db_num`)? Old versions would silently drop it, but newer versions would support it.

This way, db_num can be removed from the metadata manager entirely (not encoded in the name and not added as a middle layer of the table).

allenss-amazon: That would disallow two indexes in different db's to have the same name.

allenss-amazon: The underlying problem is that MetadataManager has a two-level naming scheme and now we need a three-level naming scheme. Plus, this stuff is preserved in the RDB file, so there's a backward compatibility issue. We can certainly move to a three-level scheme with conversion from old code. But that's a lot of changes to MetadataManager.

allenss-amazon: Another scheme would be to reuse the top-level naming for each DB. That's a change in the philosophy of MetadataManager as it would require dynamic registration of the type level types.

murphyjacob4: I see, yeah that makes sense, was missing that part.

If I were redesigning it - the metadata entry would be indexed by a UUID instead of just the index name. The UUID could be computed by the original creator.

Is there some way we could move to a UUID based scheme now? I see two ways:

1. Leave the existing metadata type namespace, and just change the schema name to be UUID indexed:

If we assume the new UUIDs are unique (they should be, or at least have an astronomically low chance of collision) then they should be able to merge cleanly with the existing IDs. We would just make the "ID" for metadata manager work like: `DB!=0 -> UUID`, `DB==0 -> index name`

Reading the code, the only place the ID is used in the Schema Manager is we use this to delete the existing index before processing the newer version of the index. So it shouldn't cause a crash. If you do create an index on DB!=0 when the cluster has nodes on the older version of Valkey-Search, it should work okay actually. If you later went to delete it, it would skip the delete since it would not find the index by the ID (which is now a UUID), which I guess is okay since it isn't even accessible on that node (it is just wasted resources). If you were to recreate it - it would log an error and fail to process - which I also think is okay since it is inaccessible on that node. Once you upgrade to the latest version, it should be able to handle the metadata (maybe we need to add some MetadataManager <-> SchemaManager reconciliation on load)

2. Add a new metadata type namespace for UUID-indexed:

This would separate the two indexing schemes into two high level types, the old one, which is used for DB=0, and the new one, which is used for DB!=0.

Nodes on the old versions would just not apply the metadata for unknown types. So this works quite cleanly. Once they upgrade, ideally they would then apply the metadata

----

What do you think? 1 is less clean on the older version nodes, but this already feels like an edge case. As long as we can get it to not crash and it would eventually recover when fully upgraded, I am okay with some odd behaviors on the old nodes. But 2 should be quite clean, I just don't know if we want to support this two-type setup long term (it is complexity that we later need to clean up)

allenss-amazon: Today, the MetadataManager takes care of resolving consistency issues. In order to support UUID naming, it would have to have some way to determine if a newly received UUID conflicts with some existing UUID. That would require that registered types have some mechanism that would assist this. In other words a client would be queried with a new protobuf and would have to return some UUID that conflicted with it. That seems worse to me, the value of MetadataManager is that he understands the objects and performs collision resolution.

Here, the problem is fundamentally naming. If we confine the discussion to one registered type, then the current system has a flat namespace. I think you either redesign MetadataManager to have a more hierarchical namespace OR you encode the hierarchy into the existing flat namespace. I've chosen the second path -- because I believe you were on the right path before, but the addition of the dbnum, was just unexpected. (BTW, we'll face exactly the same issues when we implement aliasing -- which I plan to do soon).

murphyjacob4: It is fundamentally about deriving some "primary key" from the index metadata that we can use to uniquely identify an index. It seems like a fundamental inflexibility that the primary key cannot be updated. That's fine as long as we are sure it will never really update. If dbnum is some once-in-a-project problem, then I am okay with what you have proposed.

But - if we need to add another field to the primary key - it seems like we won't be able to take the same path out again, right?

How about this: could we use a proto for the primary key?

```
message IndexMetadataKey {
     string name = 1;
     int db_num = 2;
     ...
}
```

So we would then serialize this and use this as the key. If we ever add a new field to the key, it would be cross version compatible as long as the default value for that field is handled. It is a similar option to UUID, but it guarantees two semantically identically indexes collide and are reconciled by the metadata manager

src/coordinator/metadata_manager.cc
```diff
@@ -33,13 +34,33 @@
...
+
+*/
```
murphyjacob4: Is this comment still needed?

src/coordinator/metadata_manager.cc
murphyjacob4: Closing

src/coordinator/metadata_manager.cc
```diff
@@ -134,39 +181,55 @@ uint64_t MetadataManager::ComputeTopLevelFingerprint(
...
         entry.fingerprint(), entry.version());
   }
   VMSDK_LOG_EVERY_N_SEC(WARNING, detached_ctx_.get(), 10)
       << "No type registered for: " << type_name << ", skipping callback";
   return absl::OkStatus();
 }
 
-absl::StatusOr<google::protobuf::Any> MetadataManager::GetEntry(
-    absl::string_view type_name, absl::string_view id) {
+absl::StatusOr<const GlobalMetadataEntry *> MetadataManager::GetExistingEntry(
```
murphyjacob4: The naming is a little confusing across the three functions. WDYT of: GetEntry, GetEntryContents, and GetEntryFingerprint?

```suggestion
absl::StatusOr<const GlobalMetadataEntry *> MetadataManager::GetEntry(
```

src/coordinator/metadata_manager.cc
```diff
@@ -134,39 +181,55 @@ uint64_t MetadataManager::ComputeTopLevelFingerprint(
...
+        id_itr->second.has_content()) {
+      return &id_itr->second;
+    }
   }
-  return metadata.type_namespace_map().at(type_name).entries().at(id).content();
+  return absl::NotFoundError(
+      absl::StrCat("Entry not found: ", type_name, " ", obj_name));
+}
+
+absl::StatusOr<google::protobuf::Any> MetadataManager::GetEntry(
```
murphyjacob4:
```suggestion
absl::StatusOr<google::protobuf::Any> MetadataManager::GetEntryContent(
```

src/coordinator/metadata_manager.cc
```diff
@@ -134,39 +181,55 @@ uint64_t MetadataManager::ComputeTopLevelFingerprint(
...
+    absl::string_view type_name, const ObjName &obj_name) {
+  VMSDK_ASSIGN_OR_RETURN(auto entry, GetExistingEntry(type_name, obj_name));
+  return entry->content();
+}
+
+absl::StatusOr<IndexFingerprintVersion> MetadataManager::GetEntryInfo(
```
murphyjacob4:
```suggestion
absl::StatusOr<IndexFingerprintVersion> MetadataManager::GetEntryFingerprint(
```

src/coordinator/metadata_manager.cc
```diff
@@ -257,18 +328,20 @@ std::unique_ptr<GlobalMetadata> MetadataManager::GetGlobalMetadata() {
...
+                                   MetadataUpdateCallback callback,
+                                   MinVersionCallback min_version_callback) {
   auto insert_result =
       registered_types_.Get().insert(std::pair<std::string, RegisteredType>{
-          type_name, RegisteredType{.encoding_version = encoding_version,
-                                    .fingerprint_callback =
-                                        std::move(fingerprint_callback),
-                                    .update_callback = std::move(callback)}});
-  VMSDK_LOG_EVERY_N_SEC(WARNING, detached_ctx_.get(), 10)
-      << "Type already registered for: " << type_name;
```
murphyjacob4, line 270: Hmmm so were we just always logging this? 🤦

src/coordinator/metadata_manager.cc
```diff
@@ -257,18 +328,20 @@ std::unique_ptr<GlobalMetadata> MetadataManager::GetGlobalMetadata() {
...
-  DCHECK(insert_result.second);
+          type_name,
+          RegisteredType{
+              .encoding_version = encoding_version,
+              .fingerprint_callback = std::move(fingerprint_callback),
+              .update_callback = std::move(callback),
+              .min_version_callback = std::move(min_version_callback)}});
+  VMSDK_LOG(DEBUG, nullptr) << "Registering type: " << type_name;
+  CHECK(insert_result.second);
```
murphyjacob4:
```suggestion
  CHECK(insert_result.second) << "Type already registered for: " << type_name;
```

src/coordinator/metadata_manager.cc
```diff
@@ -257,18 +328,20 @@ std::unique_ptr<GlobalMetadata> MetadataManager::GetGlobalMetadata() {
...
-          type_name, RegisteredType{.encoding_version = encoding_version,
-                                    .fingerprint_callback =
-                                        std::move(fingerprint_callback),
-                                    .update_callback = std::move(callback)}});
-  VMSDK_LOG_EVERY_N_SEC(WARNING, detached_ctx_.get(), 10)
-      << "Type already registered for: " << type_name;
-  DCHECK(insert_result.second);
+          type_name,
+          RegisteredType{
+              .encoding_version = encoding_version,
```
murphyjacob4, line 310: Not required for this PR, but just thinking, should we just match encoding version to the current module version? Having each registered type manage it may be overkill

murphyjacob4: I guess the only registered type we have (SchemaManager) just provides the current version anyways. But just thinking - do we need RegisteredType to declare this or can we just assume all RegisteredTypes will use current version as encoding version (I can't think of a reason they wouldn't)

src/coordinator/metadata_manager.cc
allenss-amazon: Moved and combined with the comment on ObjName at the end of this file.

src/coordinator/metadata_manager.cc
allenss-amazon: Done.

src/coordinator/metadata_manager.cc
allenss-amazon: Done.

src/coordinator/metadata_manager.cc
allenss-amazon: done.

src/coordinator/metadata_manager.cc
allenss-amazon: Yes. Personally, I'd just crash here.

src/coordinator/metadata_manager.cc
allenss-amazon: done.

src/coordinator/metadata_manager.cc
allenss-amazon: I contemplated removing it, but that breaks a bunch of unit tests. But, I made a better solution. I moved the encoding_version parameter to the end of the RegisterType and defaulted it to the declared global constant. I updated the comment to indicate that it should only get set in a unit test.

PR #412, opened 2025-10-11
Text update and delete ingestion

BCathcart:
This PR populates a separate TextIndex for every key to support deletes and, later on, inline filtering. When deleting a key from the index, the key-specific TextIndex is traversed to gather all the words indexed from it. This tells us which paths in the schema-level TextIndex will lead to a postings object that the key needs to be removed from.

Any time an indexed key is modified, it's first removed from the text index (if one is present), and then re-indexed. The inefficiency here is if a key is only partially updated (e.g. some non-indexed metadata is changed), we may unnecessarily delete and re-index the same text. The hope is that text indexing will be fast enough we can live with this. Otherwise, we can consider a solution where Valkey provides insight into which fields of the Hash or JSON document were actually mutated.

I've also added:
- Code to populate the suffix tree, although I'm leaving this untested until the search side is hooked up.
- Rudimentary locking so at least it shouldn't crash on concurrency issues.
- Some small refactoring (e.g. the stemmer is now owned by the lexer).

### DIFF
--- a/src/indexes/text/radix_tree.h
+++ b/src/indexes/text/radix_tree.h
@@ -88,11 +88,12 @@ struct RadixTree {
   RadixTree() = default;
 
   //
-  // Adds a target to the for the given word. If the target already exists,
-  // it will be replaced by the new target. An empty target will cause the
-  // word to be deleted from the tree.
+  // Adds the target for the given word, replacing the existing target
+  // if there is one. Providing an empty target will cause the word to be
+  // deleted from the tree. Only use this API when you don't care about any
+  // existing target.
   //
-  // This function is explicitly multi-thread safe and is
+  // (TODO) This function is explicitly multi-thread safe and is
   // designed to allow other mutations to be performed on other words and
   // targets simultaneously, with minimal collisions.
   //
@@ -104,16 +105,17 @@ struct RadixTree {
   void SetTarget(absl::string_view word, std::optional<Target> new_target);
 
   //
-  // Applies the mutation function to the target of the given word to generate
+  // Applies the mutation function to the current target of the word to generate
   // a new target. If the word doesn't already exist, a path for it will be
-  // first added to the tree. The new target is also returned to the caller.
+  // first added to the tree and the target will be std::nullopt. The new target
+  // is returned to the caller.
   //
   // The input parameter to the mutate function will be nullopt if there is no
   // entry for this word. Otherwise it will contain the value for this word. The
   // return value of the mutate function is the new value for this word. if the
   // return value is nullopt then this word is deleted from the RadixTree.
   //
-  // This function is explicitly multi-thread safe and is
+  // (TODO) This function is explicitly multi-thread safe and is
   // designed to allow other mutations to be performed on other words and
   // targets simultaneously, with minimal collisions.
   //
--- a/src/indexes/text/text_index.cc
+++ b/src/indexes/text/text_index.cc
@@ -114,6 +114,10 @@ absl::StatusOr<bool> TextIndexSchema::IndexAttributeData(
     key_index = &per_key_text_indexes_[key];
   }
 
+  // TODO: Once we optimize the postings object for space efficiency, it won't
+  // be cheap to incrementally update. We likely want to build the position map
+  // structure up front for each word in the key and then merge them into the
+  // trees' posting objects at the end of the key ingestion.
   for (uint32_t position = 0; position < tokens->size(); ++position) {
     const auto& token = (*tokens)[position];
     const std::optional<std::string> reverse_token =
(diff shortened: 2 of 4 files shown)


src/indexes/text/lexer.h
```diff
@@ -55,8 +58,13 @@ struct Lexer {
   }
 
  private:
-  std::string StemWord(const std::string& word, sb_stemmer* stemmer,
-                       bool stemming_enabled, uint32_t min_stem_size) const;
+  mutable sb_stemmer* stemmer_ = nullptr;
+
+  // TODO: Create a little pool of stemmers protected by this mutex
+  mutable std::mutex stemmer_mutex_;
```
allenss-amazon, line 64: Hmm, I thought that Lexer was per-thread. Why does it need a lock then?

src/indexes/text/radix_tree.h
```diff
@@ -88,23 +88,41 @@ struct RadixTree {
...
   // designed to allow other mutations to be performed on other words and
   // targets simultaneously, with minimal collisions.
   //
-  // In all cases, the mutate function is invoked once under the locking
-  // provided by the RadixTree itself, so if the target objects are disjoint
-  // (which is normal) then no locking is required within the mutate function
-  // itself.
+  // It's expected that the caller will know whether or not the word
+  // exists. Passing in a word that doesn't exist along with a
+  // nullopt new_target will cause the word to be added and
```
allenss-amazon, line 102: It's not clear how this is different from MutateTarget.

allenss-amazon: After reading the code, it's clear that this is an optimization for when the caller doesn't care what was at the word before, right? If so, perhaps the comment could be improved to indicate that.

src/indexes/text/text_index.cc
```diff
@@ -71,4 +95,103 @@ const char* TextIndexSchema::GetLanguageString() const {
   }
 }
 
+absl::StatusOr<bool> TextIndexSchema::IndexAttributeData(
+    const InternedStringPtr& key, absl::string_view data,
+    size_t text_field_number, bool stem, size_t min_stem_size, bool suffix) {
+  auto tokens = lexer_.Tokenize(data, GetPunctuationBitmap(), stem,
+                                min_stem_size, GetStopWordsSet());
+
```
allenss-amazon, line 103: This implementation assumes that it's easy to incrementally update the postings list, i.e., to add/remove key and position entries. While that's true for keys, it won't be true for positions for fields. When Aksha rebuilds the position map to be space efficient it won't be cheap to incrementally edit it.

I believe we'll want to move to structure where per-field/per-position information coming out of the lexer is retained in some working storage object. Then at the end of the key ingestion that working storage object is "flushed" out. This allows the accumulation of words/positions/fields for the keys and they can then be ingested in an efficient fashion.

I don't think this is hard to do. In the ingestion engine, there's a per-attribute loop that ends up calling AddRecord/ModifyRecord for each attribute(field) of the index. It should be easy to add this new object as an extra parameter to those functions and have the higher level loop create and then "flush" that object.

A key advantage of this scheme is that Aksha never has to incrementally edit a positions list as the data for it can be stored in the temporary object in a fashion that's native to the Position map.

src/indexes/text/lexer.h
BCathcart: Previously we did have a Lexer object created per thread, but really it was just a stateless set of functions. We created one stemmer per index schema and would pass that down into the lexer functions. @Aksha1812 reported that the stemmer wasn't threadsafe. I switched things around here so that there is one lexer instance per index schema that holds the stemmer and manages its concurrency.

We could create a lexer/stemmer instance per thread, but the stemmer instantiates some internal buffers every time so that's probably going to hurt performance. I was planning to play around with a small pool of stemmers.

src/indexes/text/radix_tree.h
BCathcart: I'll update the comment. It's likely these APIs change when we get to improving the locking. Right now the new target generated during the prefix tree update is applied to the suffix tree update, which is fine given the entire thing is wrapped in a mutex. Once the locking gets moved into the trees, we'll need some sort of coordination to make sure the suffix tree updates don't get applied out of order.

src/indexes/text/text_index.cc
BCathcart: Good point. I'll leave a TODO and track this task

src/indexes/text/lexer.h
allenss-amazon: This feels like it's going to be a major ingestion performance issue. I've seen on the web that other people have tackled the single-thread-ishness. We'll need to look into that.

src/indexes/text/radix_tree.h
allenss-amazon: We might want to think about an external locking structure, instead of locks within the tree itself -- especially for the leaf nodes (postings objects)

BCathcart: Yea I was thinking through a couple different schemes yesterday including moving locking outside the tree. I'll get to writing a doc soon-ish

Issue #372, opened 2025-09-05
[BUG] Valkey-server 8.1.3 crashes when varlkey-search is loaded twice


dmitrypol:
```
docker pull valkey/valkey-bundle
docker run -p 6379:6379 valkey/valkey-bundle
docker exec -it valkey-bundle bash
# start another valkey process with custom config
valkey-server -p 7379 --enable-module-command yes
valkey-cli -p 7379 
MODULE LOAD /usr/lib/valkey/libsearch.so
MODULE LOAD /usr/lib/valkey/libsearch.so
```

Log output:

131:M 05 Sep 2025 10:35:09.202 * Legacy Redis Module /usr/lib/valkey/libsearch.so found
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
F0000 00:00:1757068509.202214     131 utils.cc:76] Check failed: !set_main_thread
*** Check failure stack trace: ***

=== VALKEY BUG REPORT START: Cut & paste starting from here ===
131:M 05 Sep 2025 10:35:09.202 # valkey 8.1.3 crashed by signal: 6, si_code: -6
131:M 05 Sep 2025 10:35:09.202 # Crashed running the instruction at: 0x7ff448902eec

------ STACK TRACE ------
EIP:
/lib/x86_64-linux-gnu/libc.so.6(+0x8aeec)[0x7ff448902eec]

132 bio_close_file
/lib/x86_64-linux-gnu/libc.so.6(+0x85f16)[0x7ff4488fdf16]
/lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x1e8)[0x7ff4489005d8]
valkey-server *:7379(bioProcessBackgroundJobs+0x197)[0x5605bc2b8a77]
/lib/x86_64-linux-gnu/libc.so.6(+0x891f5)[0x7ff4489011f5]
/lib/x86_64-linux-gnu/libc.so.6(__clone+0x40)[0x7ff448980b00]

138 read-worker-2
/lib/x86_64-linux-gnu/libc.so.6(syscall+0x19)[0x7ff4489797d9]
/usr/lib/valkey/libsearch.so(+0x4695da)[0x7ff44530b5da]
/usr/lib/valkey/libsearch.so(+0x469653)[0x7ff44530b653]
/usr/lib/valkey/libsearch.so(+0x469501)[0x7ff44530b501]
/usr/lib/valkey/libsearch.so(+0xcc2de)[0x7ff444f6e2de]
[...]

PR #415, opened 2025-10-14
prevent Valkey crash when module is loaded twice

dmitrypol:
addresses https://github.com/valkey-io/valkey-search/issues/372
I tested this on Valkey 8 and 9
@murphyjacob4 - I implemented the change per your recommendation.

### DIFF
--- a/vmsdk/src/utils.cc
+++ b/vmsdk/src/utils.cc
@@ -54,10 +54,10 @@ int StopTimerFromBackgroundThread(
 }
 
 bool verifyLoadedOnlyOnce() {
-    static bool prev_loaded = false;
-    if (prev_loaded) return false;
-    prev_loaded = true;
-    return true;
+  static bool prev_loaded = false;
+  if (prev_loaded) return false;
+  prev_loaded = true;
+  return true;
 }
 
 void TrackCurrentAsMainThread() {


murphyjacob4:
For context - @dmitrypol and I debugged - the problem is when you load two instances of the module, they will actually read from the same symbols and cause initialization problems. It surfaces as a problem here because the static main thread variable must be unset (but it set by the other shared object).

vmsdk/src/module.h
```diff
@@ -23,11 +23,11 @@
   extern "C" {                                                              \
   int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,  \
                           int argc) {                                       \
-    vmsdk::TrackCurrentAsMainThread();                                      \
     if (auto status = vmsdk::module::OnLoad(ctx, argv, argc, options);      \
         status != VALKEYMODULE_OK) {                                        \
       return status;                                                        \
     }                                                                       \
+    vmsdk::TrackCurrentAsMainThread();                                      \
```
murphyjacob4: Maybe it's worth having a "symbol collision" condition that we check prior to calling OnLoad? I guess it is nice if we can keep the main thread tracking before the OnLoad call, so that OnLoad will have it set.

So something like:

```
absl::Status verifyLoadedOnlyOnce() {
    static bool prev_loaded = false;
    if (prev_loaded) return absl::FailedPrecondition("The module is already loaded");
    prev_loaded = true;
    return absl::OkStatus();
}
...
  int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,  \
                          int argc) {                                       \
    VMSDK_RETURN_IF_ERROR(verifyLoadedOnlyOnce()); \
    ... <existing code>
```

allenss-amazon: well, don't use absl::Status as the return type, use a boolean. Don't use VMSDK_RETURN_IF_ERROR. 
Put a message in the log directly using a static string (no memory allocations ,please). Then return VALKEYMODULE_ERROR if it fails.

murphyjacob4: Sorry, coding in the comment section without an IDE is hard. I thought status was absl::Status here for some reason. But yeah you get the idea

yairgott: +1 on verifying that the module is not already loaded first

dmitrypol:
all tests passed `./build.sh --run-integration-tests --run-tests`

allenss-amazon:
It's failing the regression test. The formatting can be automated "./build.sh --format" will solve that for you. I'm not sure why the other tests are failing. all of the failures seem intermittent -- not what we normally see, an not something that would seem to be related to this delta.

I added you as a contributor (once you accept the invite), it should allow you to submit PRs that run the regression tests automatically.

Issue #407, opened 2025-10-07
[BUG] Enforce min Valkey version for loading module otherwise it will crash Valkey 7


dmitrypol:
**Describe the bug**
Valkey 7 process crashes after loading valkey-search module

**To Reproduce**
Compile Valkey 7.2.11
src/valkey-server --loadmodule .../valkey-search/.build-release/libsearch.so

```
=== REDIS BUG REPORT START: Cut & paste starting from here ===
213051:M 07 Oct 2025 13:52:48.539 # valkey 7.2.11 crashed by signal: 11, si_code: 1
213051:M 07 Oct 2025 13:52:48.539 # Accessing address: (nil)

------ STACK TRACE ------

Backtrace:
/lib64/libpthread.so.0(+0x129f0)[0x7f02f56489f0]
src/valkey-server *:6379(invalidFunctionWasCalled+0x0)[0x4ddf60]
/home/dpolyako/github/valkey-io/valkey-search/.build-release/libsearch.so(_ZN5vmsdk6module16AddACLCategoriesEP15ValkeyModuleCtxRKNSt7__cxx114listISt17basic_string_viewIcSt11char_traitsIcEESaIS8_EEE+0x45)[0x7f02e6f2dc0e]
/home/dpolyako/github/valkey-io/valkey-search/.build-release/libsearch.so(_ZN5vmsdk6module6OnLoadEP15ValkeyModuleCtxPP18ValkeyModuleStringiRKNS0_7OptionsE+0x1790)[0x7f02e6f2f7b0]
/home/dpolyako/github/valkey-io/valkey-search/.build-release/libsearch.so(ValkeyModule_OnLoad+0x32)[0x7f02e6ef683a]
src/valkey-server *:6379(moduleLoad+0xf5)[0x492e15]
src/valkey-server *:6379(moduleLoadFromQueue+0x39)[0x495309]
src/valkey-server *:6379(main+0x720)[0x450250]
/lib64/libc.so.6(__libc_start_main+0xe5)[0x7f02f52998a5]
src/valkey-server *:6379(_start+0x2e)[0x45060e]
```

**Expected behavior**
Simply verify min redis_version or valkey_version to prevent loading module.  Similar crash will occur on Redis 7.

allenss-amazon:
Fixed in #408

PR #427, opened 2025-10-19
Prevent crash on Valkey 7.2 when loading valkey-search

dmitrypol:
@allenss-amazon - per your guidance I moved minimum_valkey_version check to vmsdk::module.cc to prevent crashing on AddACLCategories which is not present in Valkey 7.2

Still keeping vmsdk generic so a specific version can be passed in

`./build.sh --run-tests --run-integration-tests` passed locally

@murphyjacob4 FYI

https://github.com/valkey-io/valkey-search/issues/407

### DIFF
--- a/.github/workflows/clang_tidy_format.yml
+++ b/.github/workflows/clang_tidy_format.yml
@@ -60,7 +60,7 @@ jobs:
         id: build-comp-db
         run: |
           echo "Building CompDB"
-          docker run --rm  -v "$(pwd):/workspace" --user "ubuntu:ubuntu" presubmit-image bash -c "sudo su ubuntu && export USERNAME='ubuntu' && sudo chown -R ubuntu:ubuntu /workspace && export HOME='/home/ubuntu'  && export USERNAME='ubuntu' && ci/refresh_comp_db.sh "
+          docker run --rm --privileged -v "$(pwd):/workspace" --user "ubuntu:ubuntu" presubmit-image bash -c "sudo su ubuntu && export USERNAME='ubuntu' && sudo chown -R ubuntu:ubuntu /workspace && export HOME='/home/ubuntu'  && export USERNAME='ubuntu' && ci/refresh_comp_db.sh "
 
       - name: Find and process modified files
         run: |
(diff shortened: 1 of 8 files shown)


src/module_loader.cc
```diff
@@ -46,6 +45,7 @@ vmsdk::module::Options options = {
         valkey_search::kSearchCategory,
```
allenss-amazon, line 47: Stylistically, I'm not a fan of changing the #define to a member variable. People are more conditioned to fiddling a #define for versioning. Plus it's in all capital letters, which makes it even more obvious.... But it's a nit.

vmsdk/src/module.cc
```diff
@@ -82,6 +82,15 @@ int OnLoad(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc,
                      "Failed to init logging, %s", status.message().data());
     return VALKEYMODULE_ERR;
   }
+  auto server_version = ValkeyModule_GetServerVersion();
```
allenss-amazon, line 90: I would condition this "if" with a check that the variable "ValkeyModule_GetServerVersion" is not null. Really old versions don't have this interface and you'll end up here with it being null.

src/module_loader.cc
dmitrypol: no problem, updated the code

vmsdk/src/module.cc
dmitrypol: added if check

allenss-amazon:
If you can resolve the formatting errors, we can merge this.

vmsdk/src/module.cc
murphyjacob4: I think we need a small logical update:

1. If ValkeyModule_GetServerVersion doesn't exist but no options.minimum_valkey_version is set, we should load successfully
2. If ValkeyModule_GetServerVersion doesn't exist and options.minimum_valkey_version is set, we should fail to load
3. If ValkeyModule_GetServerVersion does exist, proceed with the comparison

I think "unset" needs to be treated specially

allenss-amazon: I don't see the value in supporting vmsdk where minimum_valkey_version is optional. Doesn't make any sense to me.

dmitrypol: I also like the idea of making minimum_valkey_version required.  We've had lots of issues with module / Valkey compatibility and requiring this check up front would have prevented them.

PR #428, opened 2025-10-20
Implement queue wait time aware local node preference in fanout operations

yulazariy:
## Summary
This PR introduces an intelligent fanout optimization that prefers local node execution when the system is under low CPU utilization, reducing network overhead and improving query latency.

## Changes
- **New Configuration**: Added `--low-utilization-threshold` option (default 50%) to control when local preference is applied
- **Smart Target Selection**: Fanout operations now monitor CPU utilization from both reader and writer thread pools
- **Fallback Strategy**: Maintains random selection when local preference is unavailable or system is under high load
- **Performance Optimization**: Refactored target creation to build targets upfront for efficient selection
- **Test Coverage**: Extended integration tests to validate CPU-aware fanout behavior

## Benefits
- Reduces network calls when local node has spare capacity
- Improves query latency by keeping operations local during low utilization periods
- Maintains existing behavior under high load conditions
- Configurable threshold allows fine-tuning for different deployment scenarios

## Testing
- Added comprehensive integration tests for fanout functionality
- Validated CPU utilization monitoring and threshold-based decisions
- Confirmed fallback behavior works correctly

### DIFF
--- a/integration/run.sh
+++ b/integration/run.sh
@@ -144,7 +144,7 @@ mkdir -p ${LOGS_DIR}
 function run_pytest() {
   zap valkey-server
   LOG_INFO "Running: ${PYTHON_PATH} -m pytest ${FILTER_ARGS} --capture=no --cache-clear -v ${ROOT_DIR}/integration/"
-  run_with_timeout 3600 ${PYTHON_PATH} -m pytest ${FILTER_ARGS} --capture=no --cache-clear -v ${ROOT_DIR}/integration/
+  timeout 60m ${PYTHON_PATH} -m pytest ${FILTER_ARGS} --capture=no --cache-clear -v ${ROOT_DIR}/integration/
   RUN_SUCCESS=$?
 }
 
--- a/scripts/common.rc
+++ b/scripts/common.rc
@@ -276,6 +276,7 @@ function run_with_timeout() {
     fi
     
     # Start the command in background
+    echo "Starting test in run_with_timeout"
     "$@" &
     local pid=$!
     
@@ -299,7 +300,7 @@ function run_with_timeout() {
     # Wait for the command to complete
     wait "$pid" 2>/dev/null
     local exit_code=$?
-    
+    echo "Finish test in run_with_timeout ${exit_code}"
     # Kill the timer if command finished
     kill "$timer_pid" 2>/dev/null
     wait "$timer_pid" 2>/dev/null


src/query/fanout_template.h
```diff
@@ -51,6 +55,45 @@ struct FanoutSearchTarget {
...
+
+    if (reader_pool) {
+      auto cpu_result = reader_pool->GetAvgCPUPercentage();
+      if (cpu_result.ok()) {
+        avg_cpu += cpu_result.value();
+        pool_count++;
+      }
+    }
+
+    if (writer_pool) {
```
yairgott: Say that there are no writes and heavy reads, IIUC, this logic may false positive identify as low utilization, right?

yulazariy: I guess you are correct in that case thanks!
I think that since the pools share the same resources i.e. cpus we need to look at the max between the averages right?

yairgott: Considering the maximum value may not be ideal in scenarios with a high frequency of writes (e.g., during a backfill process) and a low frequency of reads: While the time slice mutex would allow queries to execute, the read thread pool may be underutilized.

Alternatively, we could base the logic solely on read thread pool utilization, measuring the percentage of time the read thread pool was fully utilized during the read time slice quota. Another way to look at this is that underutilization occurs when processing an additional query during the read time slice would not increase the ingestion latency or the query time for queries ahead in the queue. For completeness, we define a state of no queries as underutilization. 

Noting that we need to make sure that the overhead of underutilization detection remains low. 

wdyt?

yulazariy: I think that those are valid concerns with the implementation but need to keep in mind that this is a minor optimization to prefer run query on the local node in the shard if the system thinks it is underutilize. So if it fails to detect low utilization it can still run locally since we use the random mechanism we have today.

Also
1. If we are under backfill, isn't it valid scenario where we may want to run remotely? (also for the recall). And in cases where we have low frequency of reads, the system will detect it is underutilize and prefer to run locally i.e. effectively using the read time slice the system has.
2. I think that looking only for reader pool is also not ideal as the resources are shared between the pools but again I don't think in real life scenarios we will have many write/modify operations on existing index (except for backfilling)
3. If we want to consider the MRMW mutex, we will need in each decision to consider all the locks we have (as the number of the indexes) and that will be expensive. But that can also be a good optimization that we can add to the lock mechanism itself and make it more adaptive with the time slices it has

yairgott: > If we are under backfill, isn't it valid scenario where we may want to run remotely? (also for the recall). And in cases where we have low frequency of reads, the system will detect it is underutilize and prefer to run locally i.e. effectively using the read time slice the system has.

This is a little tricky as in most cases, it's very likely that the replica is also performing backfilling.

> I think that looking only for reader pool is also not ideal as the resources are shared between the pools but again I don't think in real life scenarios we will have many write/modify operations on existing index (except for backfilling)

Right but my suggestion is that we consider underutilization state as if we were to process an additional query during the read time slice, the ingestion latency is not increased and neither the query time of queries ahead in the queue. For completeness, we define a state of no queries as underutilization.

> If we want to consider the MRMW mutex, we will need in each decision to consider all the locks we have (as the number of the indexes) and that will be expensive. But that can also be a good optimization that we can add to the lock mechanism itself and make it more adaptive with the time slices it has

There are a couple of approaches that we may take. My high level thought is something along the lines of: 
1. Measure the avg query processing time.
2. During the time slice, measure how many queries were actually processed, by all the reader threads.
3. Based on the avg query processing time, calculate the query capacity, across all the reader threads.

Underutilization is considered as if the amount of processed queries during a time slice is below the overall query capacity. We may want to incorporate some moving avg calculation and consider factoring up the avg query processing time.

WDYT?

allenss-amazon: > I think that looking only for reader pool is also not ideal as the resources are shared between the pools but again I don't think in real life scenarios we will have many write/modify operations on existing index (except for backfilling)

Semantic caching is a use-case that's being actively pushed for this technology. In that use-case we should expect some level of simultaneous read/write activity.

yulazariy: I think that it sound like a good approach and we can use a configuration to define the threshold of the spare capacity we allow to run another query locally

The only thing that is not clear to me is that each index has it's own time slice mutex so we will have to consider them all as part of the check right?

yairgott: That is a good point and the algorithm I suggested wouldn't work. Say for example say we have 10 indexes and a thread pool size of 10, the thread pool could be fully utilized, each thread is used for a different index. 

Accurately determining under utilization seems to me a complex problem but maybe a close enough detection is good enough. Examples of such:
1. Comparing the captured CPU utilization to a minimum configuration threshold value. Say 10 CPU cores, the config may be set to 500%.
2.  Comparing the queue wait time to a minimum configuration threshold value.

wdyt?

yulazariy: definitely determining under utilization is an interesting (and hard) problem!

Regarding the approximation, I think that option 1 is simpler and intuitive and it is also reduce the coupling with the worker types since we only consider the CPU resource.

vmsdk/src/thread_pool.cc
```diff
@@ -316,13 +324,61 @@ int ThreadPool::GetHighPriorityWeight() const {
...
+                          : 0.0;
+
+  wait_time_samples_[sample_index_] = wait_time_ms;
+
+  if (current_sample_count_ < sample_queue_size_) current_sample_count_++;
+
+  double current_avg = recent_avg_wait_time_.load();
+  double new_avg =
+      current_avg + (wait_time_ms - old_sample) /
+                        std::min(current_sample_count_, sample_queue_size_);
```
yairgott: it seems that there is a risk of dividing by zero?

src/query/fanout_template.h
```diff
@@ -51,6 +55,38 @@ struct FanoutSearchTarget {
...
+      if (!queue_wait_result.ok()) {
+        // If we can't get queue wait time, assume high utilization for safety
+        return false;
+      }
+
+      double queue_wait_time = queue_wait_result.value();
+
+      // System is under low utilization if queue wait time is below threshold
+      return queue_wait_time < threshold;
+    } catch (...) {
```
yairgott: valkey-search's code is exception free. The only exception is hnswlib code base. Please remove the `try catch` if not required, otherwise catch the specific exception and log, with rate limiting, its message.

src/query/fanout_template.h
```diff
@@ -176,35 +212,46 @@ class FanoutTemplate {
...
         }
 
-        // Create target object only for the selected node
-        if (flags & VALKEYMODULE_NODE_MYSELF) {
-          selected_targets.push_back(create_local_target());
-        } else {
-          selected_targets.push_back(create_remote_target(
-              absl::StrCat(ip, ":", coordinator::GetCoordinatorPort(port))));
+        // Prefer local target when system utilization is low
+        if (prefer_local && shard_targets.size() > 1) {
```
yairgott: can you please clarify why `shard_targets.size() > 1` ?

yulazariy: shard_targets are all the available targets in a given shard so we only attempt local preference if there are multiple targets to choose from (if size == 1, we'll use that single target regardless of locality) 

I also added it as comment in code

src/query/fanout_template.h
yulazariy: you are correct I removed it and did a proper error handling

vmsdk/src/thread_pool.cc
yulazariy: Looking at the code, `current_sample_count_` is incremented first (`current_sample_count_++`), so it should never be `0` at the division point. However, let's use the cleaner max approach such as 

```std::max(current_sample_count_, 1)```

src/query/fanout_template.h
yairgott: What's confusing me are lines 253/254. IIUC, shouldn't those lines be conditioned on not adding a target in the loop to `selected_targets`?

yulazariy: so previously, the code uniformly choose one of the indexes of the potential target within a shard (old code line 186-187) then fetch the data on it from valkey (old code lines 190-197) and lastly add it to the `selected_targets` (old code lines 202-206).

With the new change it is now 

1. optimize to save the fetch data from valkey so we only do it once.
2. before we uniformly choose one of the nodes (new code lines 253-254), it first try to force choose the local node iff a) the system is under utilize i.e. the `prefer_local` is `true` and we have more the 1 potential target (otherwise it will be chosen no matter what)

yairgott: ahh, ok, it's difficult to follow the curly braces with the github's diff tool :) It looked to me as if the  random selection is done outside of the `shard_id_to_targets` for loop. Thanks for clarifying this!

LGTM!

yairgott (approved):

PR #429, opened 2025-10-20
Cluster Map Interface

boda26:
Initial design of cluster map interface. This interface enables instant retrieval of cluster slot status, shard info and fingerprints, and primary/replica/random node targets.

### DIFF
--- a/vmsdk/src/cluster_map.h
+++ b/vmsdk/src/cluster_map.h
@@ -77,6 +77,8 @@ class ClusterMap {
   const std::vector<NodeInfo>& GetReplicaTargets() const;
   const std::vector<NodeInfo>& GetAllTargets() const;
 
+  bool GetIsClusterMapFull() const;
+
   // generate a random targets vector from cluster bus
   std::vector<NodeInfo> GetRandomTargets(ValkeyModuleCtx* ctx);
 
@@ -98,6 +100,8 @@ class ClusterMap {
 
   // Cluster-level fingerprint (hash of all shard fingerprints)
   uint64_t cluster_slots_fingerprint_;
+
+  bool is_cluster_map_full_;
 
   // Pre-computed target lists
   std::vector<NodeInfo> primary_targets_;
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,7 @@
     "clangd.path": "/usr/bin/clangd",
     "C_Cpp.codeAnalysis.runAutomatically": true,
     "editor.formatOnSave": true,
-    "C_Cpp.intelliSenseEngine": "disabled",
+    "C_Cpp.intelliSenseEngine": "default",
     "C_Cpp.codeAnalysis.clangTidy.enabled": true,
     "C_Cpp.codeAnalysis.clangTidy.useBuildPath": true,
     "clangd.arguments": [
@@ -17,5 +17,8 @@
         "--query-driver=**"
     ],
     "clangd.inactiveRegions.useBackgroundHighlight": true,
-    "clangd.enableCodeCompletion": true
+    "clangd.enableCodeCompletion": true,
+    "files.associations": {
+        "string_view": "cpp"
+    }
 }
--- a/src/commands/ft_dropindex.cc
+++ b/src/commands/ft_dropindex.cc
@@ -25,7 +25,7 @@ class DropConsistencyCheckFanoutOperation
     : public query::fanout::FanoutOperationBase<
           coordinator::InfoIndexPartitionRequest,
           coordinator::InfoIndexPartitionResponse,
-          vmsdk::cluster_map::FanoutTargetMode::kAll> {
+          fanout::FanoutTargetMode::kAll> {
  public:
   DropConsistencyCheckFanoutOperation(uint32_t db_num,
                                       const std::string& index_name,
@@ -33,14 +33,10 @@ class DropConsistencyCheckFanoutOperation
       : query::fanout::FanoutOperationBase<
             coordinator::InfoIndexPartitionRequest,
             coordinator::InfoIndexPartitionResponse,
-            vmsdk::cluster_map::FanoutTargetMode::kAll>(),
+            fanout::FanoutTargetMode::kAll>(),
         db_num_(db_num),
         index_name_(index_name),
         timeout_ms_(timeout_ms){};
-
-  std::vector<vmsdk::cluster_map::NodeInfo> GetTargets() const {
-    return ValkeySearch::Instance().GetClusterMap()->GetAllTargets();
-  }
 
   unsigned GetTimeoutMs() const override { return timeout_ms_; }
 
--- a/src/query/cluster_info_fanout_operation.cc
+++ b/src/query/cluster_info_fanout_operation.cc
@@ -25,11 +25,6 @@ ClusterInfoFanoutOperation::ClusterInfoFanoutOperation(
       backfill_complete_percent_min_(0.0f),
       backfill_in_progress_(false) {}
 
-std::vector<vmsdk::cluster_map::NodeInfo>
-ClusterInfoFanoutOperation::GetTargets() const {
-  return ValkeySearch::Instance().GetClusterMap()->GetAllTargets();
-}
-
 unsigned ClusterInfoFanoutOperation::GetTimeoutMs() const {
   return timeout_ms_;
 }
--- a/src/query/cluster_info_fanout_operation.h
+++ b/src/query/cluster_info_fanout_operation.h
@@ -25,12 +25,10 @@ class ClusterInfoFanoutOperation
     : public fanout::FanoutOperationBase<
           coordinator::InfoIndexPartitionRequest,
           coordinator::InfoIndexPartitionResponse,
-          vmsdk::cluster_map::FanoutTargetMode::kAll> {
+          fanout::FanoutTargetMode::kAll> {
  public:
   ClusterInfoFanoutOperation(uint32_t db_num, const std::string& index_name,
                              unsigned timeout_ms);
-
-  std::vector<vmsdk::cluster_map::NodeInfo> GetTargets() const override;
 
   unsigned GetTimeoutMs() const override;
 
--- a/src/query/fanout_operation_base.h
+++ b/src/query/fanout_operation_base.h
@@ -31,8 +31,7 @@ namespace valkey_search::query::fanout {
 
 constexpr unsigned kNoValkeyTimeout = 86400000;
 
-template <typename Request, typename Response,
-          vmsdk::cluster_map::FanoutTargetMode kTargetMode>
+template <typename Request, typename Response, FanoutTargetMode kTargetMode>
 class FanoutOperationBase {
  public:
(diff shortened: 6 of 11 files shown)


vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,73 @@
...
+
+#include <bitset>
+#include <string>
+#include <vector>
+
+#include "src/query/fanout_template.h"
+#include "src/valkeymodule.h"
+
+namespace vmsdk {
+
```
allenss-amazon, line 57: Let's put this in it's own namespace.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,73 @@
...
+ public:
+  // create a new cluster map in the background
+  static std::shared_ptr<ClusterMap> CreateNewClusterMap(ValkeyModuleCtx* ctx);
+
+  // slot ownership checks
+  bool IsSlotOwned(uint16_t slot) const;
+
+  // shard lookups
+  const ShardInfo* GetShardById(const std::string& shard_id) const;
+  const std::string& GetShardIdBySlot(uint16_t slot) const;
```
allenss-amazon: I doubt the shard name is all that useful, I think returning the ShardInfo object would be more useful.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,73 @@
...
+  // flexible to add other getter methods
+ public:
+  // create a new cluster map in the background
+  static std::shared_ptr<ClusterMap> CreateNewClusterMap(ValkeyModuleCtx* ctx);
+
+  // slot ownership checks
+  bool IsSlotOwned(uint16_t slot) const;
+
+  // shard lookups
+  const ShardInfo* GetShardById(const std::string& shard_id) const;
```
allenss-amazon: How is shard-not-found signalled? Also, use string_view to pass in string parameters.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,73 @@
...
+  const std::string& GetShardIdBySlot(uint16_t slot) const;
+  const absl::flat_hash_map<std::string, ShardInfo>& GetAllShards() const;
+
+  // get cluster level slot fingerprint
+  uint64_t GetClusterSlotsFingerprint() const;
+
+  // get fingerprint for a specific shard
+  uint64_t GetShardSlotsFingerprint(const std::string& shard_id) const;
```
allenss-amazon: How is this different from GetShardById(...)->fingerprint  ??

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,73 @@
...
+
+ private:
+  // 1: slot is owned by this cluster, 0: slot is not owned by this cluster
+  std::bitset<16384> owned_slots_;
```
allenss-amazon: NUM_SLOTS constant.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,73 @@
...
+  // slot ownership checks
+  bool IsSlotOwned(uint16_t slot) const;
+
+  // shard lookups
+  const ShardInfo* GetShardById(const std::string& shard_id) const;
+  const std::string& GetShardIdBySlot(uint16_t slot) const;
+  const absl::flat_hash_map<std::string, ShardInfo>& GetAllShards() const;
+
+  // get cluster level slot fingerprint
+  uint64_t GetClusterSlotsFingerprint() const;
```
allenss-amazon, line 93: Do you mean for the whole cluster?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,73 @@
...
+
+  // Cluster-level fingerprint (hash of all shard fingerprints)
+  uint64_t cluster_slots_fingerprint_;
+
+  // Pre-computed target lists
+  std::vector<valkey_search::query::fanout::FanoutSearchTarget>
+      primary_targets_;
+  std::vector<valkey_search::query::fanout::FanoutSearchTarget>
+      replica_targets_;
+  std::vector<valkey_search::query::fanout::FanoutSearchTarget> random_targets_;
```
allenss-amazon: does random_targets make sense here? How do we randomize from one call to the next call?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,73 @@
...
+  absl::flat_hash_map<std::string, ShardInfo> shards_;
+
+  // Cluster-level fingerprint (hash of all shard fingerprints)
+  uint64_t cluster_slots_fingerprint_;
+
+  // Pre-computed target lists
+  std::vector<valkey_search::query::fanout::FanoutSearchTarget>
+      primary_targets_;
+  std::vector<valkey_search::query::fanout::FanoutSearchTarget>
+      replica_targets_;
```
allenss-amazon: I don't see any interfaces that get any of these. Also, in looking at FanoutSearchTarget, I'm wondering how that's different from some per-Node information structure which belongs to this class. I'd recommend moving that struct into this class and naming it something like NodeInfo (parallel to ShardInfo).

vmsdk/src/cluster_map.h
boda26: Yes, I think we can hash the slot fingerprints from all shards to get a single cluster-wide fingerprint, which makes it faster for cluster-wide slot consistency check.

vmsdk/src/cluster_map.h
boda26: For here I think kRandom would just return the same set of nodes in a certain period. When the CreateNewClusterMap being called and refresh the cluster map, another set of random nodes are get selected.

vmsdk/src/cluster_map.h
boda26: I think we can create a NodeInfo Struct in this class. Also all the current FanoutSearchTarget code can be moved entirely into the cluster map object, including the selection of nodes (kPrimary/kReplica/kRandom/kAll)

vmsdk/src/cluster_map.h
boda26: If shard not found then this function would return nullptr. Another choice is to return std::optional

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,112 @@
...
+  kAll            // Select all nodes (both primary and replica)
+};
+
+struct NodeInfo {
+  enum Type {
+    kLocal,
+    kRemote,
+  };
+  std::string node_id;
+  Type type;
```
allenss-amazon: The primary/replica flags should be located here, right?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,112 @@
...
+    os << "NodeInfo{type: " << target.type << ", address: " << target.address
+       << "}";
+    return os;
+  }
+};
+
+struct ShardInfo {
+  // shard_id is the primary node id
+  std::string shard_id;
+  std::string primary_address;
```
allenss-amazon: Shouldn't this be either a pointer to a NodeInfo or a NodeInfo itself? The address is inside of the NodeInfo. Note, it's possible to have shards that don't have a current primary.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,112 @@
...
+  std::vector<std::string> replica_addresses;
```
allenss-amazon: Same here. Nodeinfo, not strings....

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,112 @@
...
+  std::vector<uint16_t> owned_slots;
```
allenss-amazon: I believe an ordered set would be more useful than a vector. When you fingerprint, it will eliminate any order dependency.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,112 @@
...
+
+  // shard lookups
+  const ShardInfo* GetShardById(std::string_view shard_id) const;
+  const absl::flat_hash_map<std::string, ShardInfo>& GetAllShards() const;
+
+  // get cluster level slot fingerprint
+  uint64_t GetClusterSlotsFingerprint() const;
+
+  // get fingerprint for a specific shard
+  uint64_t GetShardSlotsFingerprint(std::string_view shard_id) const;
```
allenss-amazon: Not sure this is useful, since you're exposing ShardInfo above....

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,112 @@
...
+
+ private:
+  // 1: slot is owned by this cluster, 0: slot is not owned by this cluster
+  std::bitset<NUM_SLOTS> owned_slots_;
+
+  // slot-to-shard lookup
+  std::array<std::string, NUM_SLOTS> slot_to_shard_id_;
```
allenss-amazon: Hmm, is this needed? It's a bit expensive (1.5MB by my estimate). Which API do you expect to need this?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,112 @@
...
+
+  absl::flat_hash_map<std::string, ShardInfo> shards_;
+
+  // Cluster-level fingerprint (hash of all shard fingerprints)
+  uint64_t cluster_slots_fingerprint_;
+
+  // Pre-computed target lists
+  std::vector<NodeInfo> primary_targets_;
+  std::vector<NodeInfo> replica_targets_;
+  std::vector<NodeInfo> random_targets_;
```
allenss-amazon: You're proposing to tie the update frequency of the random_targets_ to a particular time-rate (the rate at which the overall ClusterMap is regenerated). This suggests a relationship between the time-rate and the duration of the search operations. What time-rate were you targetting? What is the expected duration of the search operations?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,112 @@
...
+  std::vector<NodeInfo> all_targets_;
+
+  // private helper function to refresh targets in CreateNewClusterMap
+  std::vector<NodeInfo> GetTargets(ValkeyModuleCtx* ctx,
```
allenss-amazon: I think this should return a const reference.

vmsdk/src/cluster_map.h
boda26: Should it be an unordered_set or a set here, is the order matter here?

vmsdk/src/cluster_map.h
boda26: This is actually not needed for now

vmsdk/src/cluster_map.h
boda26: I am moving the current GetTargets function from fanout_template to cluster_map object. This function will be called only in CreateNewClusterMap use the existing logic from fanout_template to build the target vectors. The getters will be called for efficient retrieval and they will return const references.

vmsdk/src/cluster_map.h
boda26: Let's discuss this in meeting

PR #431, opened 2025-10-23
Thread-local stemmers

BCathcart:
Since a stemmer instance is not thread-safe, stemmers will be owned by threads and shared amongst the Lexer instances. The type of stemmer is determined by the language, so every ingestion worker thread creates a stemmer for each language it tokenizes at least once.

I also continued some refactoring that attempts to clean up the text_index and lexer APIs. Basically just encapsulated more data that is only relevant to tokenization in the Lexer class.

### DIFF
--- a/src/indexes/text/lexer.cc
+++ b/src/indexes/text/lexer.cc
@@ -6,6 +6,7 @@
 
 #include "src/indexes/text/lexer.h"
 
+#include "absl/container/flat_hash_map.h"
 #include "absl/log/check.h"
 #include "absl/status/status.h"
 #include "absl/strings/ascii.h"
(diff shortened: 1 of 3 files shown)


src/indexes/text/lexer.cc
```diff
@@ -14,30 +14,96 @@
...
+  }
+  return stop_words_set;
+}
+
+const char* GetLanguageString(data_model::Language language) {
+  switch (language) {
+    case data_model::LANGUAGE_ENGLISH:
+      return "english";
+    default:
+      return "english";
```
allenss-amazon: should the default be an assert? Wouldn't it be a huge internal bug if we got there?

src/indexes/text/lexer.h
```diff
@@ -26,45 +26,56 @@ Tokenization Pipeline:
...
+  std::bitset<256> punct_bitmap_;
+  absl::flat_hash_set<std::string> stop_words_set_;
+
+  // Thread-local stemmer cache. Since a stemmer instance is not thread-safe,
+  // stemmers will be owned by threads and shared amongst the Lexer instances.
+  // Each ingestion worker thread gets a stemmer for each language it tokenizes
+  // at least once.
+  class StemmerCache {
+   private:
+    std::unordered_map<data_model::Language, sb_stemmer*> cache_;
```
allenss-amazon: A nit, for the clearnup of the stemmer, you could have used a unique_ptr here with a custom deleter class.

src/indexes/text/lexer.cc
```diff
@@ -14,30 +14,96 @@
...
+  }
+}
+
+}  // namespace
+
+thread_local Lexer::StemmerCache Lexer::stemmer_cache_;
+
+// Clean up thread-local stemmers when the thread exits
+Lexer::StemmerCache::~StemmerCache() {
```
Aksha1812: Would there be any stemmer objects in the main thread? I am thinking there has to be atleast one for stemming query parsed tokens during search. thast would mean that this map would also exist till all stemmers are empty. should we destruct this cache_ map too here if it is empty.

src/indexes/text/lexer.h
BCathcart: Done

src/indexes/text/lexer.cc
BCathcart: Every thread that uses a Lexer instance to tokenize will have its own copy of the stemmer_cache_. When a thread dies, it's stemmer_cache_ instance will go out of scope and the destructor will automatically be called.

Does that clarify things?

PR #432, opened 2025-10-23
Track the list of inflight keys

baswanth09:
### Track In-Flight  Keys

When we schedule a mutation task for a key on the threadpool, we track `mutation_queue_size_` but not which keys are actually in the queue. We'll need to know the in-flight keys so that when we evaluate predicates on the main thread at the end of the FT.SEARCH flow, we need to know if all relevant keys have been indexed or not.  These checks will be added in followup PRs.

### DIFF
--- a/src/index_schema.h
+++ b/src/index_schema.h
@@ -25,6 +25,7 @@
 #include "absl/synchronization/blocking_counter.h"
 #include "absl/synchronization/mutex.h"
 #include "absl/time/time.h"
+#include "command_parser.h"
 #include "gtest/gtest_prod.h"
 #include "src/attribute.h"
 #include "src/attribute_data_type.h"
@@ -193,6 +194,9 @@ class IndexSchema : public KeyspaceEventSubscription,
   uint64_t GetBackfillDbSize() const;
   InfoIndexPartitionData GetInfoIndexPartitionData() const;
 
+  static absl::Status TextInfoCmd(ValkeyModuleCtx *ctx,
+                                  vmsdk::ArgsIterator &itr);
+
  protected:
   IndexSchema(ValkeyModuleCtx *ctx,
               const data_model::IndexSchema &index_schema_proto,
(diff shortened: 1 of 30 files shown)


baswanth09:
Yes. `InternedStringMap<DocumentMutation> tracked_mutated_records_` can be used as source of truth to check if the key is in-flight.
In `IndexSchema::TrackMutatedRecord` if the key is not new we simply update with the latest values.
```
for (auto &mutated_attribute : mutated_attributes) {
    itr->second.attributes.value()[mutated_attribute.first] =
        std::move(mutated_attribute.second);
  }
 ```
 For multi-exec as well, keys are added to `tracked_mutated_records_` 
 ```
HSET inside MULTI/EXEC
  → ProcessMutation (inside_multi_exec=true)
    → EnqueueMultiMutation
      → multi_mutations.keys.push(key)
      → IF queue.size >= thread_pool_size:
          → vmsdk::RunByMain → ProcessMultiQueue (early flush)
    → TrackMutatedRecord (adds to tracked_mutated_records_)
    → return
 ```
 
 So, I have just added a utility function to check if a key is in-flight or not using  `tracked_mutated_records_`

File: src/index_schema.h [hunk 1]
```diff
@@ -155,6 +155,10 @@ class IndexSchema : public KeyspaceEventSubscription,
   virtual void OnLoadingEnded(ValkeyModuleCtx *ctx);
 
   inline const Stats &GetStats() const { return stats_; }
+  bool IsKeyInflight(const InternedStringPtr &key) const {
```
allenss-amazon: Please add ABSL_LOCK_EXCLUDED(mutated_records_mutex_). Otherwise I'm g2g

allenss-amazon (approved):

baswanth09:
Added the annotation `ABSL_LOCKS_EXCLUDED(mutated_records_mutex_)`.

PR #433, opened 2025-10-24
FT.SEARCH basic params added

Nivesh-01:
Added the basic search params and their parsing for FT.SEARCH- 
  inorder{false};
  slop{0}; (Optional)
  verbatim{false};

### DIFF
--- a/src/commands/ft_search_parser.cc
+++ b/src/commands/ft_search_parser.cc
@@ -74,13 +74,7 @@ constexpr absl::string_view kLocalOnly{"LOCALONLY"};
 constexpr absl::string_view kVectorFilterDelimiter = "=>";
 constexpr absl::string_view kSlop{"SLOP"};
 constexpr absl::string_view kInorder{"INORDER"};
-constexpr absl::string_view kLanguage{"LANGUAGE"};
 constexpr absl::string_view kVerbatim{"VERBATIM"};
-
-// Language lookup map for enum parsing (English only, case-insensitive)
-const absl::NoDestructor<
-    absl::flat_hash_map<absl::string_view, data_model::Language>>
-    kLanguageByStr({{"ENGLISH", data_model::LANGUAGE_ENGLISH}});
 
 absl::StatusOr<absl::string_view> SubstituteParam(
     query::SearchParameters &parameters, absl::string_view source) {
@@ -353,9 +347,6 @@ vmsdk::KeyValueParser<query::SearchParameters> CreateSearchParser() {
       kVerbatim, GENERATE_FLAG_PARSER(query::SearchParameters, verbatim));
   parser.AddParamParser(kSlop,
                         GENERATE_VALUE_PARSER(query::SearchParameters, slop));
-  parser.AddParamParser(
-      kLanguage,
-      GENERATE_ENUM_PARSER(query::SearchParameters, language, *kLanguageByStr));
   return parser;
 }
 
--- a/src/query/search.h
+++ b/src/query/search.h
@@ -84,9 +84,8 @@ struct SearchParameters {
   FilterParseResults filter_parse_results;
   std::vector<ReturnAttribute> return_attributes;
   bool inorder{false};
-  unsigned int slop{0};
+  std::optional<uint32_t> slop;
   bool verbatim{false};
-  data_model::Language language{data_model::LANGUAGE_ENGLISH};
   struct ParseTimeVariables {
     // Members of this struct are only valid during the parsing of
     // VectorSearchParameters on the mainthread. They get cleared
--- a/testing/ft_search_parser_test.cc
+++ b/testing/ft_search_parser_test.cc
@@ -71,8 +71,7 @@ struct FTSearchParserTestCase {
   bool vector_query{true};
   bool verbatim{false};
   bool inorder{false};
-  int slop{0};
-  data_model::Language language{data_model::LANGUAGE_ENGLISH};
+  std::optional<unsigned> slop{};
 };
 
 class FTSearchParserTest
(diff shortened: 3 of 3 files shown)


KarthikSubbarao:
Can we fix the PR to only include the changes your commit? It has 6000 lines now

src/query/search.h
```diff
@@ -82,6 +82,11 @@ struct SearchParameters {
   bool no_content{false};
   FilterParseResults filter_parse_results;
   std::vector<ReturnAttribute> return_attributes;
+  bool no_stop_words{false};
+  bool inorder{false};
+  int slop{0};
+  bool verbatim{false};
+  std::string language;
```
KarthikSubbarao: We have an enum for Language in protobuf. Right now, as a string this is slightly hard to compare against and use later on

src/commands/ft_search_parser.cc
```diff
@@ -72,6 +72,11 @@ constexpr absl::string_view kTimeoutParam{"TIMEOUT"};
 constexpr absl::string_view kAsParam{"AS"};
 constexpr absl::string_view kLocalOnly{"LOCALONLY"};
 constexpr absl::string_view kVectorFilterDelimiter = "=>";
+constexpr absl::string_view kNoStopWords{"NOSTOPWORDS"};
```
KarthikSubbarao: worth noting that NOSTOPWORDS is deprecated in redis version 8.0 based on public documentation

src/commands/ft_search_parser.cc
```diff
@@ -338,6 +343,17 @@ vmsdk::KeyValueParser<query::SearchParameters> CreateSearchParser() {
...
+      kNoStopWords,
+      GENERATE_FLAG_PARSER(query::SearchParameters, no_stop_words));
+  parser.AddParamParser(kInorder,
+                        GENERATE_FLAG_PARSER(query::SearchParameters, inorder));
+  parser.AddParamParser(
+      kVerbatim, GENERATE_FLAG_PARSER(query::SearchParameters, verbatim));
+  parser.AddParamParser(kSlop,
+                        GENERATE_VALUE_PARSER(query::SearchParameters, slop));
+  parser.AddParamParser(
+      kLanguage, GENERATE_VALUE_PARSER(query::SearchParameters, language));
```
KarthikSubbarao: How do we validate the input here?

src/query/search.h
```diff
@@ -82,6 +82,11 @@ struct SearchParameters {
   bool no_content{false};
   FilterParseResults filter_parse_results;
   std::vector<ReturnAttribute> return_attributes;
+  bool no_stop_words{false};
+  bool inorder{false};
+  int slop{0};
```
KarthikSubbarao: slop cannot be negative

Nivesh-01: check added

src/commands/ft_search_parser.cc
Nivesh-01: we do all validation and checks in the absl::Status Verify(query::SearchParameters &parameters) function. That is also the place where slop cannot be negative is added

src/commands/ft_search_parser.cc
Nivesh-01: Removed stopwords

src/query/search.h
Nivesh-01: done

src/query/search.h
KarthikSubbarao: A check (< 0) will not be needed if we make this unsigned :D

src/commands/ft_search_parser.cc
```diff
@@ -72,6 +72,16 @@ constexpr absl::string_view kTimeoutParam{"TIMEOUT"};
...
+constexpr absl::string_view kSlop{"SLOP"};
+constexpr absl::string_view kInorder{"INORDER"};
+constexpr absl::string_view kLanguage{"LANGUAGE"};
+constexpr absl::string_view kVerbatim{"VERBATIM"};
+
+// Language lookup map for enum parsing (English only, case-insensitive)
+const absl::NoDestructor<
+    absl::flat_hash_map<absl::string_view, data_model::Language>>
+    kLanguageByStr({{"ENGLISH", data_model::LANGUAGE_ENGLISH},
+                    {"english", data_model::LANGUAGE_ENGLISH}});
```
KarthikSubbarao: Do we need both cases stored here? Instead, we can just store the lower case, and the caller site can convert to lower case and just check against that. That will also handle the mixed case

Nivesh-01: I did this to avoid wierd strings like EnGLisH. After researching I found that Redis accepts all of them so I will change it up and do the needful

Nivesh-01: Done

src/query/search.h
Nivesh-01: done

src/query/search.h
```diff
@@ -82,6 +83,10 @@ struct SearchParameters {
   bool no_content{false};
   FilterParseResults filter_parse_results;
   std::vector<ReturnAttribute> return_attributes;
+  bool inorder{false};
+  unsigned int slop{0};
```
KarthikSubbarao: nit: uint32_t

testing/ft_search_parser_test.cc
```diff
@@ -69,6 +69,10 @@ struct FTSearchParserTestCase {
   std::string search_parameters_str;
   uint64_t timeout_ms{query::kTimeoutMS};
   bool vector_query{true};
+  bool verbatim{false};
+  bool inorder{false};
+  int slop{0};
```
KarthikSubbarao: uint32_t

Nivesh-01: done

src/query/search.h
Nivesh-01: done

PR #435, opened 2025-10-27
Cluster Map Implementation

boda26:
1. Cluster Map Object Implementation
2. Refresh ClusterMap at the start of each fanout operation; fanout operations use targets from ClusterMap object

### DIFF
--- a/vmsdk/testing/cluster_map_test.cc
+++ b/vmsdk/testing/cluster_map_test.cc
@@ -113,6 +113,9 @@ class ClusterMapTest : public vmsdk::ValkeyTest {
     if (config.primary.has_value()) {
       range.push_back(
           CreateValkeyModuleCallReply(CreateNodeArray(config.primary.value())));
+    } else {
+      range.push_back(CreateValkeyModuleCallReply(
+          CreateNodeArray(CreateInvalidNodeConfig())));
     }
 
     for (const auto& replica : config.replicas) {
@@ -246,6 +249,9 @@ class ClusterMapTest : public vmsdk::ValkeyTest {
     if (config.primary.has_value()) {
       range.push_back(CreateValkeyModuleCallReply(
           CreateNodeArrayWithMap(config.primary.value())));
+    } else {
+      range.push_back(CreateValkeyModuleCallReply(
+          CreateNodeArrayWithMap(CreateInvalidNodeConfigWithMap())));
     }
 
     for (const auto& replica : config.replicas) {
@@ -293,6 +299,15 @@ class ClusterMapTest : public vmsdk::ValkeyTest {
     MockGetMyClusterID(local_node_id);
     MockClusterSlotsCallWithMap(ranges);
     return ClusterMap::CreateNewClusterMap(&fake_ctx);
+  }
+
+  // Helper: Create an invalid node config
+  NodeConfig CreateInvalidNodeConfig() {
+    return NodeConfig{"?", 30001, primary_ids.at(0), {}};
+  }
+
+  NodeConfigWithMap CreateInvalidNodeConfigWithMap() {
+    return NodeConfigWithMap{"?", 30001, primary_ids.at(0), {}};
   }
 
   // Helper: Create standard 3-shard configuration


vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,123 @@
...
+  };
+  std::string node_id;
+  NodeRole role;
+  NodeLocation location;
+  // Empty string if location is kLocal.
+  std::string address;
+
+  bool operator==(const NodeInfo& other) const {
+    return role == other.role && location == other.location &&
+           address == other.address;
```
allenss-amazon: Is the address sufficient?

Why not use the spaceship operator?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,123 @@
...
+  enum NodeRole { kPrimary, kReplica };
+  enum NodeLocation {
+    kLocal,
+    kRemote,
+  };
+  std::string node_id;
+  NodeRole role;
+  NodeLocation location;
+  // Empty string if location is kLocal.
+  std::string address;
```
allenss-amazon: Why doesn't a Node have a reference to a ShardInfo? Even if there's a transient situation where we don't know the Shard for a node. Seems like this would be more efficient to locate the fingerprint for that shard....

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,123 @@
...
+  std::set<uint16_t> owned_slots;
+  // Hash of owned_slots vector
+  uint64_t slots_fingerprint;
+};
+
+class ClusterMap {
+ public:
+  // Create a new cluster map by querying current cluster state
+  // This builds the map in the background and can be called from any thread
+  static std::shared_ptr<ClusterMap> CreateNewClusterMap(ValkeyModuleCtx* ctx);
```
allenss-amazon, line 96: This description is confusing. I'm creating a new one in the background, am I blocked until the background compeltes? If not, what's returned then?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,123 @@
...
+
+  // return pre-generated target vectors
+  const std::vector<NodeInfo>& GetPrimaryTargets() const;
+  const std::vector<NodeInfo>& GetReplicaTargets() const;
+  const std::vector<NodeInfo>& GetAllTargets() const;
+
+  bool GetIsClusterMapFull() const;
```
allenss-amazon: Not sure what you mean by "Full", perhaps a comment is required.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,123 @@
...
+
+  // generate a random targets vector from cluster bus
+  std::vector<NodeInfo> GetRandomTargets(ValkeyModuleCtx* ctx);
+
+  // slot ownership checks
+  bool IsSlotOwned(uint16_t slot) const;
```
allenss-amazon: IsSlotOwned by anybody? Or just me? How about "IOwnSlot(....)"

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,123 @@
...
+
+  // shard lookups, will return nullptr if shard does not exist
+  const ShardInfo* GetShardById(std::string_view shard_id) const;
+  const absl::flat_hash_map<std::string, ShardInfo>& GetAllShards() const;
+
+  // get cluster level slot fingerprint
```
allenss-amazon, line 129: What is this?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+
+namespace vmsdk {
+namespace cluster_map {
+
+namespace coordinator {
+// Coordinator port offset - same as in src/coordinator/util.h
+// This offset results in 26673 for Valkey default port 6379 - which is COORD
+// on a telephone keypad.
+static constexpr int kCoordinatorPortOffset = 20294;
```
allenss-amazon: Why aren't we just using the value in src/coordinator/util.h ?? Why is needed to be duplicated?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+
+inline int GetCoordinatorPort(int valkey_port) {
+  // TODO Make handling of TLS more robust
+  if (valkey_port == 6378) {
+    return valkey_port + kCoordinatorPortOffset + 1;
+  }
```
allenss-amazon: Same here. There should be a common routine for this.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+  return valkey_port + kCoordinatorPortOffset;
+}
+}  // namespace coordinator
+
+bool ClusterMap::GetIsClusterMapFull() const { return is_cluster_map_full_; }
```
allenss-amazon: These one-liner access functions are best done in the .h file.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,123 @@
...
+
+  // return pre-generated target vectors
+  const std::vector<NodeInfo>& GetPrimaryTargets() const;
+  const std::vector<NodeInfo>& GetReplicaTargets() const;
+  const std::vector<NodeInfo>& GetAllTargets() const;
+
+  bool GetIsClusterMapFull() const;
+
+  // generate a random targets vector from cluster bus
+  std::vector<NodeInfo> GetRandomTargets(ValkeyModuleCtx* ctx);
```
allenss-amazon: Why does this function need a ctx parameter?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+
+// return pre-generated all targets
+const std::vector<NodeInfo>& ClusterMap::GetAllTargets() const {
+  return all_targets_;
+}
+
+// generate a random targets vector from cluster bus (not pre-generated)
+std::vector<NodeInfo> ClusterMap::GetRandomTargets(ValkeyModuleCtx* ctx) {
+  return GetTargets(ctx, FanoutTargetMode::kRandom);
+}
```
allenss-amazon: This should be generated from the existing lists of shards and you don't need ctx for that.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+
+// slot ownership checks
+bool ClusterMap::IsSlotOwned(uint16_t slot) const { return owned_slots_[slot]; }
```
allenss-amazon: Should probably move this to the .h file.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+    return &it->second;
+  }
+  return nullptr;
+}
+
+// return all shards as a map
+const absl::flat_hash_map<std::string, ShardInfo>& ClusterMap::GetAllShards()
+    const {
+  return shards_;
+}
```
allenss-amazon: .h file

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+
+// get cluster level slot fingerprint
+uint64_t ClusterMap::GetClusterSlotsFingerprint() const {
+  return cluster_slots_fingerprint_;
+}
```
allenss-amazon: .h file

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+      selected_targets.push_back(std::move(node_info));
+    }
+  }
+  return selected_targets;
+}
+
+// create a new cluster map in the background
+std::shared_ptr<ClusterMap> ClusterMap::CreateNewClusterMap(
+    ValkeyModuleCtx* ctx) {
+  auto new_map = std::shared_ptr<ClusterMap>(new ClusterMap());
```
allenss-amazon, line 417: Where is the code that reuses old cluster maps that aren't too far out of date?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+  return shards_;
+}
+
+// get cluster level slot fingerprint
+uint64_t ClusterMap::GetClusterSlotsFingerprint() const {
+  return cluster_slots_fingerprint_;
+}
+
+// private helper function to refresh targets in CreateNewClusterMap
+std::vector<NodeInfo> ClusterMap::GetTargets(ValkeyModuleCtx* ctx,
```
allenss-amazon: This code is overly complicated and hard to understand. I think it should be replaced with a block of code that creates the ShardInfo database (which is pretty much the "All" case) and then the other cases (primaries, replicas, all) can be trivially generated by walking the shardinfo database. This will be much much clearer and therefore less prone to mistakes.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+  new_map->all_targets_ = GetTargets(ctx, FanoutTargetMode::kAll);
+
+  // call CLUSTER_SLOTS from Valkey Module API
+  auto reply = vmsdk::UniquePtrValkeyCallReply(
+      ValkeyModule_Call(ctx, "CLUSTER", "c", "SLOTS"));
+  if (reply == nullptr) {
+    // if Valkey Module API returns nullptr, return an empty map
+    VMSDK_IO_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+        << "CLUSTER_MAP_ERROR: CLUSTER SLOTS call returns nullptr";
+    new_map->is_cluster_map_full_ = false;
```
allenss-amazon: The is_cluster_map_full should be constructed as False. And only set to true later on.....

Also, I think this isn't a real failure case, I'd just CHECK(reply) << "ValkeyModule_Call failed". If it ever fails, it's because you've done something wrong (bad command, etc.).

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+
+  auto reply_type = ValkeyModule_CallReplyType(reply.get());
+  if (reply_type != VALKEYMODULE_REPLY_ARRAY) {
+    // if Valkey Module API returns incorrect type, return an empty map
+    VMSDK_IO_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+        << "CLUSTER_MAP_ERROR: CLUSTER SLOTS call returns incorrect type, "
+           "expect VALKEYMODULE_REPLY_ARRAY but got "
+        << reply_type;
+    new_map->is_cluster_map_full_ = false;
+    return new_map;
```
allenss-amazon: Again, CHECK. If you get a non-array reply CLUSTER SLOTS is broken.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+
+  // reply is an array of arrays
+  // each array element should contain at least 3 elements
+  // (1) start slot range (2) end slot range (3) master info (4)... replica info
+  for (size_t i = 0; i < len; ++i) {
+    ValkeyModuleCallReply* slot_range =
+        ValkeyModule_CallReplyArrayElement(reply.get(), i);
+    if (!slot_range ||
+        ValkeyModule_CallReplyType(slot_range) != VALKEYMODULE_REPLY_ARRAY) {
+      continue;
```
allenss-amazon: continue? Isn't this another CHECK situation? You're checking if the length of the reply was a lie OR if it's not an array or arrays. Having a check is fine. But realistically if it fails here something is seriously broken -- we're not going to recover from it. Better to crash at the point of failure to make it obvious what went wrong. Realistically, if you can't get a cluster map, then what recovery action do you expect the clients to do???

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+    }
+    size_t slot_len = ValkeyModule_CallReplyLength(slot_range);
+    // each array element should have at least 3 elements
+    if (slot_len < 3) {
```
allenss-amazon: Same here.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+      continue;
+    }
+    // Get start and end slots
+    ValkeyModuleCallReply* start_slot =
+        ValkeyModule_CallReplyArrayElement(slot_range, 0);
+    ValkeyModuleCallReply* end_slot =
+        ValkeyModule_CallReplyArrayElement(slot_range, 1);
+    long long start = ValkeyModule_CallReplyInteger(start_slot);
```
allenss-amazon: Check(ValkeyModule_CallReplyType(start_slot) == VALKEYMODULE_REPLY_INTEGER);

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+      }
+    }
+    // node info is an array
+    // (1) address (2) port number (3) node_id (4) hostname (optional)
+    // Get master node ID
+    ValkeyModuleCallReply* master_node =
+        ValkeyModule_CallReplyArrayElement(slot_range, 2);
+    if (!master_node || ValkeyModule_CallReplyLength(master_node) < 3) {
+      continue;
+    }
```
allenss-amazon: CHECK.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+    size_t slot_len = ValkeyModule_CallReplyLength(slot_range);
+    // each array element should have at least 3 elements
+    if (slot_len < 3) {
+      continue;
+    }
+    // Get start and end slots
+    ValkeyModuleCallReply* start_slot =
+        ValkeyModule_CallReplyArrayElement(slot_range, 0);
```
allenss-amazon: Indexes into the array should be named, not magic constants here.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+    ValkeyModuleCallReply* end_slot =
+        ValkeyModule_CallReplyArrayElement(slot_range, 1);
+    long long start = ValkeyModule_CallReplyInteger(start_slot);
+    long long end = ValkeyModule_CallReplyInteger(end_slot);
+    // Mark slots as assigned in cluster
+    std::set<uint16_t> slot_set;
+    for (long long slot = start; slot <= end; slot++) {
+      if (slot >= 0 && slot < k_num_slots) {
```
allenss-amazon: CHECK(). If cluster slots is returning bad answers, just crash. Also, move it out of the loop....

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+    // (1) address (2) port number (3) node_id (4) hostname (optional)
+    // Get master node ID
+    ValkeyModuleCallReply* master_node =
+        ValkeyModule_CallReplyArrayElement(slot_range, 2);
+    if (!master_node || ValkeyModule_CallReplyLength(master_node) < 3) {
+      continue;
+    }
+    size_t master_id_len;
+    const char* master_id_str = ValkeyModule_CallReplyStringPtr(
+        ValkeyModule_CallReplyArrayElement(master_node, 2), &master_id_len);
```
allenss-amazon: Here's you're very trusting of CallReplyArrayElement.... both that it's not null AND it's type == STRING. I'm fine with that, it's just inconsistent with above.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+    std::string master_id(master_id_str, master_id_len);
+
+    // Check if any node in this shard is local
+    bool is_local_shard = false;
+    for (size_t j = 2; j < slot_len; j++) {
+      ValkeyModuleCallReply* node =
+          ValkeyModule_CallReplyArrayElement(slot_range, j);
+      if (!node || ValkeyModule_CallReplyLength(node) < 3) continue;
```
allenss-amazon: check

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+          ValkeyModule_CallReplyArrayElement(node, 2), &node_id_len);
+      std::string node_id(node_id_str, node_id_len);
+
+      char ip[INET6_ADDRSTRLEN];
+      char master_buf[VALKEYMODULE_NODE_ID_LEN];
+      int port, flags;
+      if (ValkeyModule_GetClusterNodeInfo(ctx, node_id.c_str(), ip, master_buf,
+                                          &port, &flags) == VALKEYMODULE_OK &&
+          (flags & VALKEYMODULE_NODE_MYSELF)) {
+        is_local_shard = true;
```
allenss-amazon: You already know this, right? Isn't this in ShardInfo ?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+    // Update shards map
+    auto it = new_map->shards_.find(master_id);
+    if (it == new_map->shards_.end()) {
+      ShardInfo shard;
+      shard.shard_id = master_id;
+      shard.owned_slots = slot_set;
+      shard.slots_fingerprint = 0;
+      new_map->shards_[master_id] = std::move(shard);
+    } else {
+      it->second.owned_slots.insert(slot_set.begin(), slot_set.end());
```
allenss-amazon: What if this list of nodes doesn't match the previous set?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+    }
+  }
+  new_map->is_cluster_map_full_ = assigned_slots.all();
```
allenss-amazon: Where are the fingerprints for each shard being computed?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,445 @@
...
+          range_end = i;
+        } else {
+          if (!owned_ranges.empty()) owned_ranges += ", ";
+          owned_ranges +=
+              std::to_string(range_start) + "-" + std::to_string(range_end);
+          range_start = i;
+          range_end = i;
+        }
+      }
+    }
```
allenss-amazon, line 528: Why not save the slot ranges as they came in originally?

boda26:
Todo: create a SocketAddress struct for ip and port

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,144 @@
...
+    kLocal,
+    kRemote,
+  };
+  std::string node_id;
+  NodeRole role;
+  NodeLocation location;
+  // Empty string if location is kLocal.
+  std::string ip;
+  int port;
+  std::string address;
```
allenss-amazon: I'd still like to see a SocketAddress object. You could put it in vmsdk/src/utils.h

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,144 @@
...
+  // Pointer to the shard this node belongs to
+  const ShardInfo* shard = nullptr;
+
+  auto operator<=>(const NodeInfo&) const = default;
+
+  friend std::ostream& operator<<(std::ostream& os, const NodeInfo& target) {
```
allenss-amazon, line 64: If it's freestanding, you need to add inline. If you make it a member function then you don't need that.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,144 @@
...
+  }
+};
+
+struct ShardInfo {
+  // shard_id is the primary node id
+  std::string shard_id;
+  // primary node can be empty
+  std::optional<NodeInfo> primary;
+  std::vector<NodeInfo> replicas;
+  std::set<uint16_t> owned_slots;
```
allenss-amazon: This should be a set of slot ranges. map<uint16_t, uint16_t> It's a subset of the slot_to_shard_map_ in the ClusterMap object. It should be updateable at the same time.

vmsdk/src/cluster_map.cc
allenss-amazon: This doesn't belong in the coordinator namespace. That's an application thing, for vmsdk, I'd just put this into it's own namespace: clustermap remove the nested namespace, i.e., just put it into vmsdk.

vmsdk/src/cluster_map.cc
allenss-amazon: As we discussed before, the coordinator port number stuff belongs elsewhere.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+    return valkey_port + kCoordinatorPortOffset + 1;
+  }
+  return valkey_port + kCoordinatorPortOffset;
+}
+}  // namespace coordinator
+
+// configurable variable for cluster map expiration time
+static auto cluster_map_expiration_ms =
+    vmsdk::config::Number("cluster-map-expiration-ms",
+                          5000,      // default: 5 seconds
```
allenss-amazon: 5 seconds is too long for a default.  This number needs to balance CPU consumption with convergence time. 5 seconds is too long for the convergence time. I think this could be something like 1/4 second.

allenss-amazon: Min should be 0, which effectively disables caching.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+
+// return pre-generated replica targets
+const std::vector<NodeInfo>& ClusterMap::GetReplicaTargets() const {
+  return replica_targets_;
+}
+
+// return pre-generated all targets
+const std::vector<NodeInfo>& ClusterMap::GetAllTargets() const {
+  return all_targets_;
+}
```
allenss-amazon: Simple one-line getters should be moved to the .h file.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+
+// shard lookups, will return nullptr if shard does not exist
+const ShardInfo* ClusterMap::GetShardById(std::string_view shard_id) const {
+  auto it = shards_.find(std::string(shard_id));
+  if (it != shards_.end()) {
+    return &it->second;
+  }
+  return nullptr;
```
allenss-amazon: Personal style. I'd use 
```
return (it != shards_.end()) ? &it-.second : nullptr;
```

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+}
+
+// For shard fingerprint - just sum all slots
+uint64_t ClusterMap::compute_shard_fingerprint(
+    const std::set<uint16_t>& slots) {
+  uint64_t fingerprint = 0;
+  for (uint16_t slot : slots) {
+    fingerprint += slot;
+  }
+  return fingerprint;
```
allenss-amazon: I think we need a stronger fingerprint function than arithmetic sum. I think the fingerprinting for metadata uses HighwayHash, no reason not to duplicate that here.

Because we're caching this, the CPU time for the hashing isn't critical. Better to spend some additional cycles on a better hash.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+
+  for (const auto& [shard_id, shard] : shards) {
+    // Simple string hash that's deterministic
+    uint64_t shard_id_hash = 0;
+    for (char c : shard_id) {
+      shard_id_hash = shard_id_hash * 31 + c;
+    }
+
+    fingerprint += shard.slots_fingerprint + shard_id_hash;
+  }
```
allenss-amazon: Same here.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+  // reply is an array of arrays
+  // each array element should contain at least 3 elements
+  // (1) start slot range (2) end slot range (3) master info (4)... replica info
+  size_t len = ValkeyModule_CallReplyLength(reply.get());
+  for (size_t i = 0; i < len; ++i) {
+    ValkeyModuleCallReply* slot_range =
+        ValkeyModule_CallReplyArrayElement(reply.get(), i);
+    CHECK(slot_range)
+        << "CLUSTER_MAP_ERROR: ValkeyModule_Call CLUSTER SLOTS result returned "
+           "invalid slot range sub-array";
```
allenss-amazon: This extra verbiage detracts from reading the code. If this CHECK fails, you'll see the text "slot_range" and you'll get the source file-name and line number in the output. That's enough to see what went wrong. By adding the extra output, it just makes the code harder to read -- you don't really get any additional information.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+    auto slot_range_type = ValkeyModule_CallReplyType(slot_range);
+    CHECK(slot_range_type == VALKEYMODULE_REPLY_ARRAY)
```
allenss-amazon: Here, I'd put 
```
CHECK(ValkeyModule...callReplyType(slot_range) == VALKEY_MODULE_REPLY_ARRAY);
```
 Again, having all of the extra verbiage detracts from reading the code. Here if you wanted, you could add something like:

```
CHECK(ValkeyModule...callReplyType(slot_range) == VALKEY_MODULE_REPLY_ARRAY); << "Got:" << ValkeyModule_CallReplyType(slot_range);
```
Your code puts the type into a variable. Now the reader has to ask "What use will that variable be put to in the future?" -- it adds cognitive load, when you really aren't going to use that variable again. Eliminating that variable reduces the load.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+           "incorrect slot range sub-array type, expect "
+           "VALKEYMODULE_REPLY_ARRAY but got "
+        << slot_range_type;
+
+    size_t slot_len = ValkeyModule_CallReplyLength(slot_range);
+    // each array element should have at least 3 elements
+    CHECK(slot_len >= 3)
+        << "CLUSTER_MAP_ERROR: ValkeyModule_Call CLUSTER "
+           "SLOTS result returned a slot range sub-array with invalid length, "
+           "should be at least 3";
```
allenss-amazon: this is the worst of both worlds, redundant message -- but you didn't dump the actual offending value.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+        << "CLUSTER_MAP_ERROR: end slot expect INTEGER type, got"
+        << ValkeyModule_CallReplyType(end_slot);
+    long long start = ValkeyModule_CallReplyInteger(start_slot);
+    long long end = ValkeyModule_CallReplyInteger(end_slot);
+    // Mark slots as assigned in cluster
+    std::set<uint16_t> slot_set;
+    for (long long slot = start; slot <= end; slot++) {
+      if (slot >= 0 && slot < k_num_slots) {
+        slot_set.insert(static_cast<uint16_t>(slot));
+      }
```
allenss-amazon: Moving to a range map will eliminate this loop.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,531 @@
...
+    // get primary IP and port
+    size_t primary_ip_len;
+    const char* primary_ip_char = ValkeyModule_CallReplyStringPtr(
+        ValkeyModule_CallReplyArrayElement(primary_node_arr, 0),
+        &primary_ip_len);
+    CHECK(primary_ip_char)
+        << "CLUSTER_MAP_ERROR: primary node ip returned nullptr";
+    std::string primary_ip(primary_ip_char, primary_ip_len);
+
+    long long primary_port = ValkeyModule_CallReplyInteger(
```
allenss-amazon: All of this code is crying out for some additional subroutines to handle the checkes. Realistically, you're just getting integer or string our of some position in an array. Having some functions to do that will simplify things.  I'd look at absl::StatusOr<...> and the VMSDK_ASSIGN_OR_RETURN and VMSDK_RETURN_IF_ERROR macros.

boda26:
Added unit tests for ClusterMap

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,176 @@
...
+  kAll            // Select all nodes (both primary and replica)
+};
+
+const size_t k_num_slots = 16384;
+
+// forward declaration to solve circular dependency
+struct ShardInfo;
+
+struct NodeInfo {
+  enum NodeRole { kPrimary, kReplica };
```
allenss-amazon: A two-valued enum tends to be a bit baroque in the code. If we believe that this would have more values in the future, then an enum makes sense. But I don't think that's true here. So perhaps a boolean value with IsPrimary and IsReplica member functions makes the client code more readable.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,176 @@
...
+  enum NodeLocation {
+    kLocal,
+    kRemote,
+  };
+  std::string node_id;
+  NodeRole role;
+  NodeLocation location;
+  // Empty string if location is kLocal.
+  std::string ip;
+  // 0 if location is kLocal
```
allenss-amazon: Port is 0 if it's local? Hmm, that's a todo, there should be a way to get the local port.

You should have some interface functions, i.e., IsLocal(), IsRemote(), IsPrimary(), IsReplica()

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,176 @@
...
+  int port;
```
allenss-amazon: uint16_t for port.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,176 @@
...
+                           const char* my_node_id);
+
+  static void ProcessSlotRange(
+      ValkeyModuleCallReply* slot_range, const char* my_node_id,
+      absl::flat_hash_map<std::string, ShardInfo>& shards,
+      std::bitset<k_num_slots>& owned_slots,
+      std::vector<NodeInfo>& primary_targets,
+      std::vector<NodeInfo>& replica_targets,
+      std::vector<NodeInfo>& all_targets,
+      std::vector<SlotRangeInfo>& slot_ranges);
```
allenss-amazon: Why so many parameters?? Why isn't this a member function???

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,176 @@
...
+
+  static void BuildSlotToShardMap(
```
allenss-amazon: Same, why isn't this a member function?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,483 @@
...
+}
+
+// Helper function to parse node info from CLUSTER SLOTS reply
+NodeInfo ClusterMap::ParseNodeInfo(ValkeyModuleCallReply* node_arr,
+                                   bool is_local_shard,
+                                   NodeInfo::NodeRole role) {
+  CHECK(node_arr) << "CLUSTER_MAP_ERROR: node subarray returned nullptr";
+  CHECK(ValkeyModule_CallReplyLength(node_arr) >= 3)
+      << "CLUSTER_MAP_ERROR: node subarray returned invalid length, expected "
+         ">= 3";
```
allenss-amazon: If you're going to bother with a message, include the offending value. The failure message will have "3" in it, what it won't have is the output of CallReplyLength....

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,483 @@
...
+    const char* node_id_char = ValkeyModule_CallReplyStringPtr(
+        ValkeyModule_CallReplyArrayElement(node_arr, 2), &node_id_len);
+
+    if (node_id_char && node_id_len == VALKEYMODULE_NODE_ID_LEN &&
+        memcmp(node_id_char, my_node_id, VALKEYMODULE_NODE_ID_LEN) == 0) {
+      return true;
+    }
+  }
+  return false;
+}
```
allenss-amazon, line 249: Seems like you could have just scanned the ShardInfo at the end, right?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,483 @@
...
+
+// Helper function to parse slot range and create ShardInfo
+void ClusterMap::ProcessSlotRange(
+    ValkeyModuleCallReply* slot_range, const char* my_node_id,
+    absl::flat_hash_map<std::string, ShardInfo>& shards,
+    std::bitset<k_num_slots>& owned_slots,
+    std::vector<NodeInfo>& primary_targets,
+    std::vector<NodeInfo>& replica_targets, std::vector<NodeInfo>& all_targets,
+    std::vector<SlotRangeInfo>& slot_ranges) {
+  CHECK(slot_range)
```
allenss-amazon: Why so many parameters. Why isn't this a member function?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,483 @@
...
+  uint16_t expected_next = 0;
+  for (const auto& [start_slot, range_and_shard] : slot_map) {
+    if (start_slot != expected_next) {
+      return false;
+    }
+    expected_next = range_and_shard.first + 1;
+  }
+
+  return expected_next == k_num_slots;
+}
```
allenss-amazon: one thing that you can insulate yourself against is having a bad set of slot ranges. Here, you're testing that the slot ranges which are present covers 0..k_num_slots. Which is good. But it doesn't guarantee that there aren't other slot ranges which are duplicates. Here, if you counted up the number of ranges you visited and then compared it against the size of the map, if they weren't equal then you have a bad map. I think a CHECK for a bad map is sufficient.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,483 @@
...
+
+// Main function - now much cleaner
+std::shared_ptr<ClusterMap> ClusterMap::CreateNewClusterMap(
+    ValkeyModuleCtx* ctx) {
+  auto new_map = std::shared_ptr<ClusterMap>(new ClusterMap());
+  new_map->is_cluster_map_full_ = false;
```
allenss-amazon: Isn't this a duplicate of what the constructor should have done?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,483 @@
...
+  std::vector<SlotRangeInfo> slot_ranges;
+  size_t len = ValkeyModule_CallReplyLength(reply.get());
+  for (size_t i = 0; i < len; ++i) {
+    ValkeyModuleCallReply* slot_range =
+        ValkeyModule_CallReplyArrayElement(reply.get(), i);
+    ProcessSlotRange(slot_range, my_node_id, new_map->shards_,
+                     new_map->owned_slots_, new_map->primary_targets_,
+                     new_map->replica_targets_, new_map->all_targets_,
+                     slot_ranges);
+  }
```
allenss-amazon: This should be a member function....

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,483 @@
...
+  BuildSlotToShardMap(new_map->slot_to_shard_map_, new_map->shards_,
+                      slot_ranges);
+
+  // Check if cluster map is full
+  new_map->is_cluster_map_full_ =
+      CheckClusterMapFull(new_map->slot_to_shard_map_);
+
+  // Compute cluster-level fingerprint
+  new_map->cluster_slots_fingerprint_ =
+      compute_cluster_fingerprint(new_map->shards_);
```
allenss-amazon: same, these should all be member functions....

vmsdk/src/cluster_map.h
boda26: they are used in building new cluster map, not quite able to be member functions

vmsdk/src/cluster_map.cc
boda26: The current logic needs is_local information to build both NodeInfo and ShardInfo

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,192 @@
...
+struct ShardInfo;
+
+struct NodeInfo {
+  std::string node_id;
+  bool is_primary;
+  bool is_local;
+  SocketAddress socket_address;
+  // a map containing all additional network metadata(the fourth entry of
+  // CLUSTER SLOTS response); can be empty
+  std::unordered_map<std::string, std::string> additional_network_metadata;
```
allenss-amazon: let's use absl::flat_hash_map, it's more consistent with the rest of the codebase. I don't think you need stable iterators, which is the major functional difference.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,192 @@
...
+};
+
+struct ShardInfo {
+  // shard_id is the primary node id
+  std::string shard_id;
+  // primary node can be empty
+  std::optional<NodeInfo> primary;
+  std::vector<NodeInfo> replicas;
+  // map start slot to end slot
+  std::map<uint16_t, uint16_t> owned_slots;
```
allenss-amazon: absl::btree_map.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,192 @@
...
+
+  std::chrono::steady_clock::time_point GetExpirationTime() const {
+    return expiration_tp_;
+  }
+
+  // are all the slots assigned to some shard
+  bool IsConsistent() const { return is_consistent_; }
+
+  // generate a random targets vector from cluster bus
+  std::vector<NodeInfo> GetRandomTargets();
```
allenss-amazon: const?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,192 @@
...
+  // get cluster level slot fingerprint
+  uint64_t GetClusterSlotsFingerprint() const {
+    return cluster_slots_fingerprint_;
+  }
+
+ private:
+  std::chrono::steady_clock::time_point expiration_tp_;
+
+  // a map used for check duplicate socket addresses, node_id -> SocketAddress
+  std::unordered_map<std::string, SocketAddress> node_to_socket_map_;
```
allenss-amazon: absl::flat_hash_map.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,192 @@
...
+
+  // 1: slot is owned by this cluster, 0: slot is not owned by this cluster
+  std::bitset<k_num_slots> owned_slots_;
+
+  absl::flat_hash_map<std::string, ShardInfo> shards_;
+
+  // An ordered map, key is start slot, value is end slot and ShardInfo
```
allenss-amazon, line 145: Unordered?

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,192 @@
...
+  std::map<uint16_t, std::pair<uint16_t, const ShardInfo*>> slot_to_shard_map_;
```
allenss-amazon: absl::btree_map.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+  if (slot >= start_slot && slot <= end_slot) {
+    // Return the shard pointer
+    return it->second.second;
+  }
+  // Slot is in a gap, return nullptr
+  return nullptr;
+}
+
+// generate a random targets vector with one node from each shard
+std::vector<NodeInfo> ClusterMap::GetRandomTargets() {
```
allenss-amazon: const?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+    // Collect all nodes in this shard (primary + replicas)
+    std::vector<NodeInfo> shard_nodes;
+    if (shard_info.primary.has_value()) {
+      shard_nodes.push_back(shard_info.primary.value());
+    }
+    for (const auto& replica : shard_info.replicas) {
+      shard_nodes.push_back(replica);
+    }
+    // Randomly select one node from this shard
+    if (!shard_nodes.empty()) {
```
allenss-amazon: Just as a question, how can this shard be empty? How would it come into existence in the map? If it can't happen (which I think is correct), then maybe CHECK(!shard_nodes.empty());

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+      random_targets.push_back(shard_nodes[index]);
+    }
+  }
+  return random_targets;
+}
+
+// For shard fingerprint - hash the slot ranges
+uint64_t ClusterMap::ComputeShardFingerprint(
+    const std::map<uint16_t, uint16_t>& slot_ranges) {
+  if (slot_ranges.empty()) {
```
allenss-amazon: CHECK(!slot_ranges.empty()). How could it be empty?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+    return 0;
+  }
+
+  // Convert map to vector for contiguous memory
```
allenss-amazon: I don't think this is needed. You can compute hashes incrementally, See HighwayHashCatT

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+
+    shard_entries.push_back({.shard_id_hash = shard_id_hash,
+                             .slots_fingerprint = shard.slots_fingerprint});
+  }
+
+  // Sort for deterministic ordering
+  std::sort(shard_entries.begin(), shard_entries.end(),
+            [](const ShardEntry& a, const ShardEntry& b) {
+              return a.shard_id_hash < b.shard_id_hash;
+            });
```
allenss-amazon: Why do you need to sort again? Isn't the shards_ map already in shard-id order?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+  // Get IP
+  size_t node_ip_len;
+  const char* node_ip_char = ValkeyModule_CallReplyStringPtr(
+      ValkeyModule_CallReplyArrayElement(node_arr, 0), &node_ip_len);
+  CHECK(node_ip_char) << VALKEY_MODULE_CALL_ERROR_MSG;
+
+  // Get port
+  long long node_port = ValkeyModule_CallReplyInteger(
+      ValkeyModule_CallReplyArrayElement(node_arr, 1));
+  CHECK(node_port) << VALKEY_MODULE_CALL_ERROR_MSG;
```
allenss-amazon: If you're not going to use named_constants for the indexes, then at least do them in numerical order.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+
+  // Get additional network metadata
+  // Depending on the client RESP protocol version, the additional network
+  // metadata might be a map(RESP3), or a flattened array(RESP2)
+  std::unordered_map<std::string, std::string> additional_network_metadata;
+  auto reply_metadata = ValkeyModule_CallReplyArrayElement(node_arr, 3);
+  CHECK(reply_metadata) << VALKEY_MODULE_CALL_ERROR_MSG;
+  CHECK(ValkeyModule_CallReplyType(reply_metadata) == VALKEYMODULE_REPLY_MAP ||
```
allenss-amazon: It feels like this would be clearer with a switch.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+    size_t array_len = ValkeyModule_CallReplyLength(reply_metadata);
+    for (size_t i = 0; i + 1 < array_len; i += 2) {
+      auto arr_key = ValkeyModule_CallReplyArrayElement(reply_metadata, i);
+      auto arr_val = ValkeyModule_CallReplyArrayElement(reply_metadata, i + 1);
+      size_t key_len;
+      const char* key_str = ValkeyModule_CallReplyStringPtr(arr_key, &key_len);
+      size_t val_len = 0;
+      const char* val_str = ValkeyModule_CallReplyStringPtr(arr_val, &val_len);
+      additional_network_metadata[std::string(key_str, key_len)] =
+          std::string(val_str, val_len);
```
allenss-amazon: Looks like duplication of code.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+    }
+  }
+
+  std::string node_id_str = std::string(node_id_char, node_id_len);
+  SocketAddress addr{
+      .ip = is_local_shard ? "" : std::string(node_ip_char, node_ip_len),
```
allenss-amazon: Is it certain that he IP and Port for a local address aren't valid? If so, local IP should be 127.0.0.1, and you get the right port from the CLUSTER SLOTS command. Are you seeing zero???

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+      .port = is_local_shard ? 0 : static_cast<uint16_t>(node_port)};
+
+  // Check for duplicate socket addresses across different nodes
+  // Skip check for local nodes (they all have empty ip and port 0)
+  if (!is_local_shard) {
```
allenss-amazon: Why is local shard exempt from this? If you had the right ip/socket then you could avoid the special case, right?

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+    auto it = node_to_socket_map_.find(node_id_str);
```
allenss-amazon: Why do I care about the node_id. If you're looking for duplicate SocketAddresses, then make something that's indexed by SocketAddress. You could have a set<SocketAddress> (well absl::flat_hash_set<SocketAddress>) OR  if you want to know which nodes collided (not sure I care), you could have a map<SocketAddress, std::string>. In other case you simplify this code and avoid a loop.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+      // New node - check if this socket address is already used by another node
+      for (const auto& [existing_node_id, existing_addr] :
+           node_to_socket_map_) {
+        if (existing_addr == addr) {
+          VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+              << "Duplicate socket address " << addr.ip << ":" << addr.port
+              << " found for different nodes: " << existing_node_id << " and "
+              << node_id_str;
+          this->is_consistent_ = false;
+          break;
```
allenss-amazon: With the set you don't need this...

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+  size_t node_id_len;
+  const char* node_id_char = ValkeyModule_CallReplyStringPtr(
+      ValkeyModule_CallReplyArrayElement(node_arr, 2), &node_id_len);
+  CHECK(node_id_char) << VALKEY_MODULE_CALL_ERROR_MSG;
+
+  // Get IP
+  size_t node_ip_len;
+  const char* node_ip_char = ValkeyModule_CallReplyStringPtr(
+      ValkeyModule_CallReplyArrayElement(node_arr, 0), &node_ip_len);
+  CHECK(node_ip_char) << VALKEY_MODULE_CALL_ERROR_MSG;
```
allenss-amazon: You should read the description of this in the Redis documentation, there are many more options for the IP address than are handled here. see https://redis.io/docs/latest/commands/cluster-slots/

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+    ShardInfo shard;
+    shard.shard_id = shard_id;
+    shard.owned_slots[start] = end;
+    // compute shard fingerprint later
+    shard.slots_fingerprint = 0;
+    shard.primary = primary_node;
+    shard.replicas = std::move(replicas);
+
+    auto [inserted_it, success] = shards_.insert({shard_id, std::move(shard)});
+    shard_it = inserted_it;
```
allenss-amazon: Stylistically:

ShardInfo& shard = shards_[shard_id];
shard..... =
shard.... =
...

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+  } else {
+    // Existing shard
+    // check if the shard info is consistent
+    if (!IsExistingShardConsistent(shard_it->second, primary_node, replicas)) {
+      VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+          << "Inconsistency shard info found on existing slot ranges!";
+      is_consistent_ = false;
+    }
+    shard_it->second.owned_slots[start] = end;
```
allenss-amazon: Shouldn't the consistency check come after the new slot range is added? Or maybe in a batch at the end.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+
+  // Store slot range info for later
+  slot_ranges.push_back(
+      {static_cast<uint16_t>(start), static_cast<uint16_t>(end), shard_id});
+
+  // Fix shard pointers and populate target lists
+  if (is_new_shard) {
+    if (shard_it->second.primary.has_value()) {
+      shard_it->second.primary->shard = &(shard_it->second);
+      primary_targets_.push_back(shard_it->second.primary.value());
```
allenss-amazon: I'd recommend that primary_targets and all_targets get computed in a loop at the end. It simplifies this logic.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+
+// Helper function to check if cluster map is full
+bool ClusterMap::CheckClusterMapFull() {
+  if (slot_to_shard_map_.empty() || slot_to_shard_map_.begin()->first != 0) {
+    return false;
+  }
+
+  uint16_t expected_next = 0;
+  for (const auto& [start_slot, range_and_shard] : slot_to_shard_map_) {
+    CHECK(start_slot >= expected_next) << "Duplicate slo ranges found";
```
allenss-amazon: slot. Also, here inserting a constant message isn't useful, the crash will have the text of the CHECK statement and the file/line number which eliminates any use for a constant additional message. However, dumping out the two values that aren't comparing properly is a totally different story.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+    if (start_slot != expected_next) {
+      return false;
```
allenss-amazon, line 399: A log message here, please.

vmsdk/src/cluster_map.cc
```diff
@@ -0,0 +1,609 @@
...
+    }
+    expected_next = range_and_shard.first + 1;
+  }
+
```
allenss-amazon: This code ensures that each slot is covered. But it doesn't ensure that there aren't extra entries in the slot table that would by definition result in a duplicate coverage for some slots. By counting the number of times you go around the loop and then comparing it to the map size, you can rule out this additional failure mode.

vmsdk/src/cluster_map.cc
boda26: The shards_ map is a flat_hash_map which is unordered. If switching to use a btree_map, the btree_map does not keep pointer stability, which breaks the shard pointer in NodeInfo. I think either keep the sort here, or use a std::map for shards_. A sort here i think is affordable?

vmsdk/src/cluster_map.cc
boda26: The node_id is useful when two slot ranges belong to the same shard. In this case a node is populated twice, so we need to use a map(addr to node_id) to verify this address belongs to the same node).

vmsdk/src/cluster_map.cc
boda26: fixed-moved pointer linking after all insertions into btree_map is finished

vmsdk/src/cluster_map.cc
```diff
@@ -57,47 +57,39 @@ const ShardInfo* ClusterMap::GetShardBySlot(uint16_t slot) const {
...
-    if (!shard_nodes.empty()) {
-      size_t index = absl::Uniform(gen, 0u, shard_nodes.size());
-      random_targets.push_back(shard_nodes[index]);
+    CHECK(node_count > 0);
+    size_t index = absl::Uniform(gen, 0u, node_count);
+    if (shard_info.primary.has_value() && index == 0) {
+      random_targets.push_back(shard_info.primary.value());
+    } else {
+      size_t replica_index = shard_info.primary.has_value() ? index - 1 : index;
+      random_targets.push_back(shard_info.replicas[replica_index]);
```
allenss-amazon: Let's break out the code to randomly select a node from a shard into a separate API (here you can just call that). I've recently uncovered a case that wants this capability.

src/CMakeLists.txt
```diff
@@ -80,6 +80,7 @@ target_link_libraries(valkey_search PUBLIC utils)
 target_link_libraries(valkey_search PUBLIC status_macros)
 target_link_libraries(valkey_search PUBLIC valkey_module)
 target_link_libraries(valkey_search PUBLIC acl)
+target_link_libraries(valkey_search PUBLIC cluster_map)
```
allenss-amazon: Let's not create a new lib, please just add it another existing lib. I think we're going to get rid of most of these intermediate lib files.

src/valkey_search.h
```diff
@@ -97,6 +98,28 @@ class ValkeySearch {
...
+  std::shared_ptr<vmsdk::cluster_map::ClusterMap> GetClusterMap() const {
+    return std::atomic_load(&cluster_map_);
+  }
+
+  // Update cluster map with a new one (thread-safe atomic swap)
+  void UpdateClusterMap(
+      std::shared_ptr<vmsdk::cluster_map::ClusterMap> new_map) {
+    std::atomic_store(&cluster_map_, new_map);
+  }
+
```
allenss-amazon: Is there a reason this isn't private or just done inline? Who would do that?

src/valkey_search.h
```diff
@@ -97,6 +98,28 @@ class ValkeySearch {
...
+  // Refresh cluster map by creating a new one from current cluster state
+  void RefreshClusterMap(ValkeyModuleCtx *ctx) {
+    VMSDK_LOG(NOTICE, ctx) << "RefreshClusterMap called";
```
allenss-amazon: rate limit, please.

allenss-amazon: Also, it looks like a new clustermap is created on every call. Where is the code to reuse a recently created cluster map?

vmsdk/testing/cluster_map_test.cc
```diff
@@ -0,0 +1,813 @@
...
+    CallReplyArray range;
+    range.push_back(
+        CreateValkeyModuleCallReply(CallReplyInteger(config.start_slot)));
+    range.push_back(
+        CreateValkeyModuleCallReply(CallReplyInteger(config.end_slot)));
+
+    if (config.master.has_value()) {
+      range.push_back(
+          CreateValkeyModuleCallReply(CreateNodeArray(config.master.value())));
+    }
```
allenss-amazon: Wouldn't this be null if there wasn't a master defined?

vmsdk/testing/cluster_map_test.cc
```diff
@@ -0,0 +1,813 @@
...
+  int port;
+  std::string node_id;  // 40-character hex string
+  std::unordered_map<std::string, std::string> additional_network_metadata;
+};
+
+// Helper structure to define a slot range with its nodes
+struct SlotRangeConfig {
+  int start_slot;
+  int end_slot;
+  std::optional<NodeConfig> master;
```
allenss-amazon: Corporate policy, we don't use "master", we use "primary".

vmsdk/testing/cluster_map_test.cc
boda26: currently I am doing CHECK(master_node), the program would crash if master_node is null.

allenss-amazon: Doesn't the documentation say that the primary node can be NULL ? Here, the code is preparing to have an optional primary node -- which is good, we want to test that case. But if the primary node is missing, then you need to add a NULL value in to the result array, here you're skipping that entry entirely, which isn't the behavior of the cluster map.

vmsdk/src/utils.h
```diff
@@ -134,5 +134,25 @@ struct JsonQuotedStringView {
...
 }  // namespace vmsdk
+
+// Hash specialization for SocketAddress
+namespace std {
+template <>
+struct hash<vmsdk::SocketAddress> {
+  size_t operator()(const vmsdk::SocketAddress &addr) const {
+    size_t h1 = std::hash<std::string>{}(addr.primary_endpoint);
+    size_t h2 = std::hash<uint16_t>{}(addr.port);
+    return h1 ^ (h2 << 1);
```
allenss-amazon, line 153: Just xor the two, why the shift?

src/query/fanout_operation_base.h
```diff
@@ -124,14 +129,14 @@ class FanoutOperationBase {
...
-          [this, target](grpc::Status status, Response& resp) {
+          [this, target, client_address](grpc::Status status, Response& resp) {
             if (status.ok()) {
               this->OnResponse(resp, target);
             } else {
               ++Metrics::GetStats().info_fanout_fail_cnt;
               VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
                   << "FANOUT_DEBUG: InvokeRemoteRpc error on target "
-                  << target.address << ", status code: " << status.error_code()
+                  << client_address << ", status code: " << status.error_code()
```
allenss-amazon, line 139: same, here you're passing duplicate information. Imagine if there was a programming error that caused the client_address to not match target->GetAddress(), then the log message would actually hide the problem.

src/query/fanout_operation_base.h
```diff
@@ -110,12 +112,15 @@ class FanoutOperationBase {
...
+      std::string client_address = absl::StrCat(
+          target.socket_address.primary_endpoint, ":",
+          coordinator::GetCoordinatorPort(target.socket_address.port));
+      auto client = client_pool_->GetClient(client_address);
       if (!client) {
         ++Metrics::GetStats().info_fanout_fail_cnt;
         VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
             << "FANOUT_DEBUG: Found invalid client on target "
-            << target.address;
+            << client_address;
```
allenss-amazon, line 123: client->GetAddress() ?? If it doesn't have this ability, let's add it.

src/valkey_search.h
```diff
@@ -97,6 +98,15 @@ class ValkeySearch {
   // of the program.
   ValkeyModuleCtx *GetBackgroundCtx() const { return ctx_; }
 
+  // Get or create a new cluster map
+  std::shared_ptr<vmsdk::cluster_map::ClusterMap> GetOrRefreshClusterMap(
```
allenss-amazon, line 102: Just call it GetClusterMap, we can overload the name.

vmsdk/src/cluster_map.h
```diff
@@ -0,0 +1,198 @@
...
+#include "highwayhash/highwayhash.h"
+#include "vmsdk/src/utils.h"
+#include "vmsdk/src/valkey_module_api/valkey_module.h"
+
+namespace vmsdk {
+
+namespace cluster_map {
+
+static constexpr highwayhash::HHKey kHashKey{
+    0x9736bad976c904ea, 0x08f963a1a52eece9, 0x1ea3f3f773f3b510,
```
allenss-amazon, line 35: Does this need to be here? Can't it be in the .cc file?

allenss-amazon (approved):
A couple of nits, but those can be addressed in a followup. Once the tests complete let's merge this.

yairgott:
I'm concerned that refresh ClusterMap at the start of each fanout operation is very inefficient.

@boda26 ,  are you interested to drive an enhancement into the engine which allows a module to subscribe to a cluster map modification event notification?

PR #438, opened 2025-10-29
Query Processing Improvements

KarthikSubbarao:
* Escape Char Handling of fulltext content that enables users to escape query syntax characters
* Query syntax compat improvements
* Default text field (no field specification) handling
* Index based Punctuation
* Stemming, Stopwords
* Updated TextPredicate structures
* Applying the text properties based on the FT.SEARCH command arguments 

Added unit and integ testing

### DIFF
--- a/src/commands/filter_parser.cc
+++ b/src/commands/filter_parser.cc
@@ -450,7 +450,12 @@ std::unique_ptr<query::Predicate> WrapPredicate(
 
 static const uint32_t FUZZY_MAX_DISTANCE = 3;
 
-// Handle backslashes inside text content.
+// Handles backslash escaping for both quoted and unquoted text
+// Escape Syntax:
+// \\ -> \
+// \<punctuation> -> <punctuation>
+// \<non-punctuation> -> (break to new token)<non-punctuation>...
+// \<EOL> -> Return error
 absl::StatusOr<bool> FilterParser::HandleBackslashEscape(
     const indexes::text::Lexer& lexer, std::string& processed_content) {
   if (!Match('\\', false)) {
@@ -478,6 +483,12 @@ absl::StatusOr<bool> FilterParser::HandleBackslashEscape(
   }
 }
 
+// Returns a token within an exact phrase parsing it until reaching the
+// token boundary while handling escape chars.
+// Quoted Text Syntax:
+// word1 word2" word3 -> word1
+// word2" word3 -> word2
+// Token boundaries (separated by space): " <punctuation> \<non-punctuation>
 absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseQuotedTextToken(
     std::shared_ptr<indexes::text::TextIndexSchema> text_index_schema,
     FieldMaskPredicate field_mask, std::optional<uint32_t> min_stem_size) {
@@ -506,6 +517,18 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseQuotedTextToken(
       false};
 }
 
+// Returns a token after parsing it until the token boundary while handling
+// escape chars.
+// Unquoted Text Syntax:
+//   Term:    word
+//   Prefix:  word*
+//   Suffix:  *word
+//   Infix:   *word*
+//   Fuzzy:   %word% | %%word%% | %%%word%%%
+// Token boundaries:
+//   <punctuation> ( ) | @ " - { } [ ] : ; $
+// Reserved chars:
+//   { } [ ] : ; $ -> error
 absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseUnquotedTextToken(
     std::shared_ptr<indexes::text::TextIndexSchema> text_index_schema,
     FieldMaskPredicate field_mask, std::optional<uint32_t> min_stem_size) {
@@ -668,12 +691,14 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
 }
 
 // This function is called when the characters detected are potentially those of
-// a text predicate. It can parse an exact phrase, or simply multiple text
-// tokens (without field specifiers) and will return the grouped result of those
-// predicates. Currently, this is Proximity and will be changed to the
-// ComposedAND.
-// When non text query syntax is detected (not escaped), it breaks out and
-// returns back to the caller site with the parsed predicate.
+// a text predicate.
+// Text Parsing Syntax:
+//   Quoted: "word1 word2" -> ProximityPredicate(exact, slop=0, inorder=true)
+//   Unquoted: word1 word2 -> TermPredicate(word1) - stops at first token
+// Token boundaries for unquoted text: <punctuation> ( ) | @ " - { } [ ] : ; $
+// Quoted phrases (Exact Phrase) parse all tokens within quotes, unquoted
+// parsing stops after first token.
+// TODO: Update ProximityPredicate to ComposedAND.
 absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
     const std::optional<std::string>& field_or_default) {
   auto text_index_schema = index_schema_.GetTextIndexSchema();
@@ -709,6 +734,10 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
                                      min_stem_size));
     if (result.predicate) {
       terms.push_back(std::move(result.predicate));
+      // TODO: Uncomment this once we have ComposedAND evaluation functional for
+      // handling proximity checks. Until the, we handle unquoted text tokens
+      // by building a proximity predicate containing them.
+      // if (!exact_phrase) break;
     }
     if (result.break_on_query_syntax) {
       break;


src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+          break;
+        } else {
+          should_escape = true;
+        }
+      }
+      backslash_count = 0;
+      if (should_escape) {
+        processed_content.push_back(ch);
+        ++current_pos;
+        should_escape = false;
```
allenss-amazon: Won't this code treat \\ and \\\\ and \\\\\\ and \\\\\\\\, etc. as the same? I don't think that's correct.

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+    // Handle fuzzy token boundary detection
+    if (!in_quotes && ch == '%') {
+      if (current_pos == pos_) {
+        // Leading percent
+        while (current_pos < expression_.size() &&
+               expression_[current_pos] == '%') {
+          leading_percent_count++;
+          current_pos++;
+          if (leading_percent_count > FUZZY_MAX_DISTANCE) break;
+        }
```
allenss-amazon, line 576: This seems like it should be split into two functions, one for in-quotes and one for !in-quotes. The constant testing of "in-quotes" makes the code complex and I'm skeptical that the shared portions justify this complexity.

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+// When non text query syntax is detected (not escaped), it breaks out and
+// returns back to the caller site with the parsed predicate.
+absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
+    const std::optional<std::string>& field_for_default) {
+  auto text_index_schema = index_schema_.GetTextIndexSchema();
+  if (!text_index_schema) {
+    return absl::InvalidArgumentError("Index does not have any text field");
+  }
+  std::vector<std::unique_ptr<query::TextPredicate>> terms;
+  uint64_t field_mask;
```
allenss-amazon: FieldMask ?

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+    field_mask = 1ULL << text_index->GetTextFieldNumber();
+    if (text_index->IsStemmingEnabled()) {
+      min_stem_size = text_index->GetMinStemSize();
+    }
+  } else {
+    // Set identifiers to include all text fields in the index schema.
+    auto text_identifiers = index_schema_.GetAllTextIdentifiers();
+    for (const auto& identifier : text_identifiers) {
+      filter_identifiers_.insert(identifier);
+    }
```
allenss-amazon, line 682: Hmm, this list is a duplicate of the information that's in the field_mask. If the IndexSchema had a reverse mapping of field_mask bit position to TextIndex, then you wouldn't need to carry both around.

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
-          continue;
-        } else if (Match('{')) {
-          VMSDK_ASSIGN_OR_RETURN(auto tag, ParseTagPredicate(current_field));
-          extra_terms.push_back(std::move(tag));
-          continue;
-        }
-      } else {
-        return absl::InvalidArgumentError("Invalid query string");
+    if (c == '"') {
+      in_quotes = !in_quotes;
```
allenss-amazon, line 720: as stated above, I believe you should be breaking out to a in-quote specific block of code here. No reason it can't return a list of predicates which is what this loop is expecting....

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+        continue;
       }
+      break;
+    }
+    size_t token_start = pos_;
+    VMSDK_ASSIGN_OR_RETURN(
+        auto result, ParseTokenAndBuildPredicate(in_quotes, text_index_schema,
+                                                 field_mask, min_stem_size));
+    if (result.predicate) {
+      terms.push_back(std::move(result.predicate));
```
allenss-amazon, line 736: What kinds of Tokens wouldn't generate a predicate ?

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
-        std::move(all_terms), /*slop=*/0, /*inorder=*/true);
+    pos_ = result.end_pos;
+  }
+  std::unique_ptr<query::Predicate> pred;
+  if (terms.size() > 1) {
+    // TODO: Set these based on the FT.SEARCH command parameters.
+    uint32_t slop = 0;
+    bool inorder = false;
+    if (exact) {
+      slop = 0;
```
allenss-amazon, line 756: redundant

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
-  } else if (!all_terms.empty()) {
-    prox = std::make_unique<query::ProximityPredicate>(
-        std::move(all_terms), /*slop=*/0, /*inorder=*/true);
+    pos_ = result.end_pos;
+  }
+  std::unique_ptr<query::Predicate> pred;
+  if (terms.size() > 1) {
+    // TODO: Set these based on the FT.SEARCH command parameters.
+    uint32_t slop = 0;
+    bool inorder = false;
```
allenss-amazon: bool inorder = exact ?

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+      break;
+    }
+    size_t token_start = pos_;
+    VMSDK_ASSIGN_OR_RETURN(
+        auto result, ParseTokenAndBuildPredicate(in_quotes, text_index_schema,
+                                                 field_mask, min_stem_size));
+    if (result.predicate) {
+      terms.push_back(std::move(result.predicate));
+    }
+    if (result.break_on_query_syntax) {
```
allenss-amazon, line 742: This piece of state seems strange, is it related to the merging of in-quote and not-in-quote scanning?

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+  std::string processed_content;
+  // State tracking for predicate detection
+  bool starts_with_star = false;
+  bool ends_with_star = false;
+  size_t leading_percent_count = 0;
+  size_t trailing_percent_count = 0;
+  bool break_on_query_syntax = false;
+  while (current_pos < expression_.size()) {
+    char ch = expression_[current_pos];
+    // Handle backslashes
```
allenss-amazon: Why isn't this code using the scanner Peek & Match machinery?

src/commands/filter_parser.cc
```diff
@@ -449,218 +448,287 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
+  std::optional<uint32_t> min_stem_size = std::nullopt;
+  // Handle default / every field (no field specifier) and specific
+  // field query cases.
+  if (field_for_default.has_value()) {
+    auto index = index_schema_.GetIndex(field_for_default.value());
+    if (!index.ok() ||
+        index.value()->GetIndexerType() != indexes::IndexerType::kText) {
+      return absl::InvalidArgumentError("Index does not have any text field");
+    }
+    auto* text_index = dynamic_cast<const indexes::Text*>(index.value().get());
```
allenss-amazon: this feels like an validation that needs to happen at every term, it should probably get broken out into a separate function.

src/commands/filter_parser.cc
KarthikSubbarao: It is inorder either if the token is an exact phrase OR if the INORDER property was used in the FT.SEARCH command

src/commands/filter_parser.cc
KarthikSubbarao: Yes, it is redundant now, but I will merge another PR that adds the FT.SEARCH options and then it wouldn't be redundant since slop has some value from the command that is overridden in exact phrases

src/commands/filter_parser.cc
KarthikSubbarao: It is more related to breaking out of text specific parsing and start parsing of other operations (parenthesis, negate, a field specifier, etc). We need to know when to break out of this loop

src/commands/filter_parser.cc
KarthikSubbarao: Do you mean the default vs specific field handling / setup should be a separate function? Sure, I can move it into one

src/commands/filter_parser.cc
KarthikSubbarao: If the query string only has punctuation and then we run into query syntax characters (`( - @`, etc), the token will have a null predicate. The other case is it simply has punctuation characters and then we reach the end of the query string without being able to parse an actual text predicate

src/commands/filter_parser.cc
KarthikSubbarao: No, it will not treat them the same.

The escape character behavior is different based on:
(1) odd/even number of backslashes
(2) whether the character after the final backslash is a punctuation OR a normal character
(3) whether we are in quotes vs not in quotes
It also causes a "break" into two tokens or escape behavior based on the above

src/commands/filter_parser.cc
allenss-amazon: Yes a separate function.

src/commands/filter_parser.cc
allenss-amazon: Well the code counts up sequences of backslashes. As near as I can tell, the code then triggers on the next non-backslash character, this trigger is only sensitive to whether backslash_count is zero, even/odd or > 1. Nowhere is a there a loop based on backslash_count, ergo, long sequences of backslashes get treated like they were short sequences of backslashes, i.e., 0, 1 or 2. So you're correct that \, \\, and \\\ aren't treated the same. But any even number of backslashes (!= 0) are the same and any odd number of backslashes are treated the same. Is this really the way that Redisearch behaves?

KarthikSubbarao: Discussed offline that this was the compatible search behavior, but we have chosen to implement the escape char handling in a standard approach

integration/test_fulltext.py
```diff
@@ -690,3 +690,140 @@ def test_suffix_search(self):
...
+        ]
+        
+        # Validate that all text index fields are present and have reasonable values
+        for field in text_index_fields:
+            assert field in info_data, f"Missing text index field: {field}"
+            value = info_data[field]
+            
+            if field == "num_unique_terms":
+                assert isinstance(value, (int, float)) and value > 0, f"{field} should be positive number, got {value}"
+                assert value >= 10, f"Expected at least 10 unique terms, got {value}"
```
allenss-amazon, line 729: why isn't this test exact?

integration/test_fulltext.py
```diff
@@ -690,3 +690,140 @@ def test_suffix_search(self):
...
+                
+            elif field == "num_total_terms":
+                assert isinstance(value, (int, float)) and value > 0, f"{field} should be positive number, got {value}"
+                assert value >= info_data["num_unique_terms"], f"Total terms {value} should be >= unique terms {info_data['num_unique_terms']}"
```
allenss-amazon, line 733: same.

integration/test_fulltext.py
```diff
@@ -690,3 +690,140 @@ def test_suffix_search(self):
...
+        
+        # Note: Postings are currently stored in per-key indexes. When we implement shared
+        # Posting objects across schema and key indexes, these metrics will reach zero on deletion.
+        assert info_after_delete['num_unique_terms'] < initial_unique_terms, \
+            f"Unique terms should decrease after deletion: {info_after_delete['num_unique_terms']} >= {initial_unique_terms}"
+        assert info_after_delete['num_total_terms'] < initial_total_terms, \
+            f"Total terms should decrease after deletion: {info_after_delete['num_total_terms']} >= {initial_total_terms}"
+        assert info_after_delete['position_sz_bytes'] < initial_position_memory, \
+            f"Position memory should decrease after deletion: {info_after_delete['position_sz_bytes']} >= {initial_position_memory}"
+        assert info_after_delete['total_text_index_sz_bytes'] < initial_total_memory, \
```
allenss-amazon, line 800: Shouldn't these also be zero now?

src/commands/filter_parser.cc
```diff
@@ -448,216 +450,202 @@ std::unique_ptr<query::Predicate> WrapPredicate(
...
     // - characters in the middle of text tokens are not negate. If they are in
     // the beginning, break.
-    if (!in_quotes && ch == '-' && processed_content.empty()) {
+    if (ch == '-' && processed_content.empty()) {
       break_on_query_syntax = true;
       break;
     }
     // Break to complete an exact phrase or start a new exact phrase.
     if (ch == '"') break;
-    // Break on all punctuation characters, except text query syntax chars such
```
allenss-amazon: would this be clearer with a switch statement?

src/commands/filter_parser.cc
```diff
@@ -676,23 +664,49 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
...
+    return absl::InvalidArgumentError("Index does not have any text field");
+  }
+  std::vector<std::unique_ptr<query::TextPredicate>> terms;
+  // Handle default / every field (no field specifier) and specific
+  // field query cases.
+  FieldMaskPredicate field_mask;
+  std::optional<uint32_t> min_stem_size = std::nullopt;
+  VMSDK_RETURN_IF_ERROR(
+      SetupTextFieldConfiguration(field_mask, min_stem_size, field_or_default));
   bool in_quotes = false;
```
allenss-amazon, line 715: The logic of this escapes me. I understand the specialized code for double quotes. But assuming it's not in double quotes, isn't this just a vector of elements that get anded together (w or w/o proximity -- tbd later). Where an element could be anything that matches keys, i.e., prefix/suffix/numeric/tag/parenthesized expression, etc. If I understand this logic for a case like ```word1 @f:[1 2] word2``` you're going to parse the word1 and then break out of it for the numeric and then enter it again for word2? And then stitch these together? I don't understand that.

src/commands/filter_parser.cc
```diff
@@ -676,23 +664,49 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
...
-  bool exact = false;
+  bool exact_phrase = false;
   while (!IsEnd()) {
     char c = Peek();
     if (c == '"') {
       in_quotes = !in_quotes;
       ++pos_;
       if (in_quotes && terms.empty()) {
```
allenss-amazon, line 722: Can a double quote be escaped in an exact phrase?

src/commands/filter_parser.cc
allenss-amazon: Q: If a query has only punctuation, does it match 0 keys or all keys?

integration/test_fulltext.py
KarthikSubbarao: This test was not added in my PR. If you are reviewing the diff, this test was brought it after a merge with the latest fulltext branch to resolve merge conflicts

integration/test_fulltext.py
KarthikSubbarao: This test was not added in my PR. If you are reviewing the diff, this test was brought it after a merge with the latest fulltext branch to resolve merge conflicts

integration/test_fulltext.py
KarthikSubbarao: This test was not added in my PR. If you are reviewing the diff, this test was brought it after a merge with the latest fulltext branch to resolve merge conflicts

src/commands/filter_parser.cc
KarthikSubbarao: Yes

src/commands/filter_parser.cc
KarthikSubbarao: I currently prefer separate If statements since there are different behaviors and it seems more readable this way

src/commands/filter_parser.cc
KarthikSubbarao: It will be rejected

daddaman-amz (approved):

src/indexes/text.cc
```diff
@@ -127,6 +115,14 @@ std::unique_ptr<EntriesFetcherIteratorBase> Text::EntriesFetcher::Begin() {
 // Implement the TextPredicate BuildTextIterator virtual method
 namespace valkey_search::query {
 
+void* TextPredicate::Search(bool negate) const {
+  // TODO: Add logic to calculate the size based on number of keys estimated.
+  auto fetcher = std::make_unique<indexes::Text::EntriesFetcher>(
```
daddaman-amz, line 120: Do you remember why the size was hard-coded as zero?

PR #442, opened 2025-10-31
Predicate evaluator changes for text predicates

daddaman-amz:
## Summary

This PR refactors predicate evaluation for text fields to utilize the text index directly. In future we will have PR to wait if we have any inflight keys during the evaluation, which will ensure we evaluate from the latest version of Index.

### Key Changes

- **Modified predicate evaluation signature**: All predicate `Evaluate` methods now return an `EvaluationResult` containing:
  - A boolean flag indicating whether the predicate matches
  - An optional filter iterator for result traversal
  - The filter iterator would be nullptr for Numeric/Tags whereas it will be the respective text iterator for the Text predicate
  
  
- **Text index integration**: Text predicates  now evaluate against the `TextIndex` data structure unlike numeric / tags who read from keyspace.

- Added predicate proximity logic as part of composed AND

### DIFF
--- a/src/query/predicate.cc
+++ b/src/query/predicate.cc
@@ -44,6 +44,7 @@ EvaluationResult TermPredicate::Evaluate(Evaluator& evaluator) const {
   return evaluator.EvaluateText(*this);
 }
 
+// TermPredicate: Exact term match in the text index.
 EvaluationResult TermPredicate::Evaluate(
     const valkey_search::indexes::text::TextIndex& text_index,
     const std::shared_ptr<valkey_search::InternedString>& target_key) const {
@@ -88,6 +89,7 @@ EvaluationResult PrefixPredicate::Evaluate(Evaluator& evaluator) const {
   return evaluator.EvaluateText(*this);
 }
 
+// PrefixPredicate: Matches all terms that start with the given prefix.
 EvaluationResult PrefixPredicate::Evaluate(
     const valkey_search::indexes::text::TextIndex& text_index,
     const std::shared_ptr<valkey_search::InternedString>& target_key) const {
@@ -138,6 +140,7 @@ EvaluationResult SuffixPredicate::Evaluate(Evaluator& evaluator) const {
   return evaluator.EvaluateText(*this);
 }
 
+// SuffixPredicate: Matches terms that end with the given suffix
 EvaluationResult SuffixPredicate::Evaluate(
     const valkey_search::indexes::text::TextIndex& text_index,
     const std::shared_ptr<valkey_search::InternedString>& target_key) const {
@@ -327,6 +330,10 @@ ComposedPredicate::ComposedPredicate(std::unique_ptr<Predicate> lhs_predicate,
       slop_(slop),
       inorder_(inorder) {}
 
+// ComposedPredicate: Combines two predicates with AND/OR logic.
+// For text predicates with proximity constraints (slop/inorder), creates
+// ProximityIterator to validate term positions meet distance and order
+// requirements.
 EvaluationResult ComposedPredicate::Evaluate(Evaluator& evaluator) const {
   if (GetType() == PredicateType::kComposedAnd) {
     EvaluationResult lhs = lhs_predicate_->Evaluate(evaluator);
--- a/src/query/response_generator.cc
+++ b/src/query/response_generator.cc
@@ -130,20 +130,15 @@ class PredicateEvaluator : public query::Evaluator {
 
   EvaluationResult EvaluateText(
       const query::TextPredicate &predicate) override {
-    // If we have per-key indexes, use them for text evaluation
-    if (per_key_indexes_ && target_key_) {
-      // Find the per_key_text_index for the target key.
-      auto it = per_key_indexes_->find(target_key_);
-      if (it == per_key_indexes_->end()) {
-        VMSDK_LOG(WARNING, nullptr)
-            << "Target key not found in index for predicate evaluation";
-        return EvaluationResult(false);
-      }
-      // Just pass the text_index, predicate computes its own field_mask
-      return predicate.Evaluate(it->second, target_key_);
+    // Find the per_key_text_index for the target key.
+    auto it = per_key_indexes_->find(target_key_);
+    if (it == per_key_indexes_->end()) {
+      VMSDK_LOG(WARNING, nullptr)
+          << "Target key not found in index for predicate evaluation";
+      return EvaluationResult(false);
     }
-    // Fallback: no per-key index available
-    return EvaluationResult(false);
+    // Just pass the text_index, predicate computes its own field_mask
+    return predicate.Evaluate(it->second, target_key_);
   }
 
  private:


src/query/predicate.cc
```diff
@@ -34,14 +40,35 @@ TermPredicate::TermPredicate(const indexes::Text* index,
...
   // call dynamic dispatch on the evaluator
   return evaluator.EvaluateText(*this);
 }
 
-bool TermPredicate::Evaluate(const std::string_view& text) const {
-  if (text.empty()) return false;
-  return text == term_;  // exact match
+EvaluationResult TermPredicate::Evaluate(const valkey_search::indexes::text::TextIndex& text_index) const {
+  size_t field_number = index_->GetTextFieldNumber();
+  uint64_t field_mask = 1ULL << field_number;
```
KarthikSubbarao: This field_mask information will be accessible from the predicate directly with the PR here: https://github.com/valkey-io/valkey-search/pull/438

src/query/predicate.cc
```diff
@@ -34,14 +40,35 @@ TermPredicate::TermPredicate(const indexes::Text* index,
...
+  
+  auto key_iter = postings->GetKeyIterator();
+  if (!(key_iter.IsValid() && key_iter.ContainsFields(field_mask))) {
+    return EvaluationResult(false);
+  }
+
+  std::vector<indexes::text::Postings::KeyIterator> key_iterators;
+  key_iterators.emplace_back(std::move(key_iter));
+  auto iterator = std::make_unique<indexes::text::TermIterator>(
+      std::move(key_iterators), field_mask, nullptr);
```
KarthikSubbarao, line 72: We need to return false if there are no positions

src/query/predicate.cc
KarthikSubbarao: Same comment for all other predicate evaluations

src/query/predicate.cc
KarthikSubbarao: Same comment for all other predicate evaluations

src/query/predicate.cc
```diff
@@ -198,41 +253,83 @@ bool TagPredicate::Evaluate(
...
+    EvaluationResult rhs = rhs_predicate_->Evaluate(evaluator);
+    VMSDK_LOG(DEBUG, nullptr) << "Inline evaluate AND predicate rhs: " << rhs.matches;
+    // Short-circuit for AND
+    if (!rhs.matches) {
+      return EvaluationResult(false);
+    }
+
+    // Proximity check: Only if slop/inorder set and both sides have iterators.
+    // This ensures we only check proximity for text predicates, not numeric/tag.
+    if (slop_.has_value() && inorder_.has_value() &&
```
KarthikSubbarao: In the current PR, slop and inorder are always none (nullopt). Have we tested when they are set based on the query? filter_parser has the logic for this

daddaman-amz: I found that inorder=false crashes in general for existing proximity logic. will try to add fix in this CR along with style issues.

src/query/predicate.cc
```diff
@@ -53,13 +85,44 @@ PrefixPredicate::PrefixPredicate(const indexes::Text* index,
...
+    }
+    word_iter.Next();
+  }
+
+  if (key_iterators.empty()) {
+    return EvaluationResult(false);
+  }
+
+  auto iterator = std::make_unique<indexes::text::TermIterator>(
+      std::move(key_iterators), field_mask, nullptr);
```
KarthikSubbarao: Is this duplicate code between the Evaluator and the search logic in text.cc?


it will be even more duplicated once we have the Prefilter (InlineVector) Evaluator implemented. Is it possible to have one function used across these 3 call sites?


Same comment for other text predicate evaluate methods

src/query/predicate.cc
```diff
@@ -131,9 +193,15 @@ ProximityPredicate::ProximityPredicate(
...
       slop_(slop) {}
 
-bool ProximityPredicate::Evaluate(Evaluator& evaluator) const {
+EvaluationResult ProximityPredicate::Evaluate(Evaluator& evaluator) const {
   return evaluator.EvaluateText(*this);
 }
+      
+EvaluationResult ProximityPredicate::Evaluate(
+    const valkey_search::indexes::text::TextIndex& text_index) const {
+  // Temporarily make this pass through for tests.
```
KarthikSubbarao: This is fine, you can actually delete this once the Query processing improvement PR is merged. But can we update the field handling of the ProximityIterator to include validation there?

src/query/predicate.cc
daddaman-amz: done

src/query/predicate.cc
daddaman-amz: I will address this refactor when I do prefilter evalutor in a new pr

integration/test_fulltext.py
daddaman-amz, line 1: Add wait for backfill in the test

src/commands/filter_parser.cc
```diff
@@ -765,17 +804,35 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
   }
   std::unique_ptr<query::Predicate> pred;
   if (terms.size() > 1) {
+    VMSDK_LOG(WARNING, nullptr) << "terms.size() > 1";
     uint32_t slop = options_.slop.value_or(0);
     bool inorder = options_.inorder;
+    // std::optional<uint32_t> slop = std::nullopt;
+    // std::optional<bool> inorder = std::nullopt;
     if (exact_phrase) {
```
KarthikSubbarao: Since we have uncommented the above, if there are multiple terms in this function at this line of code, it will always be an exact phrase. Let us always set slop 0 and inorder here

src/commands/filter_parser.cc
```diff
@@ -765,17 +804,35 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
...
       slop = 0;
       inorder = true;
+      VMSDK_LOG(WARNING, nullptr)
+          << "EXACT PHRASE: slop=" << slop << " inorder=" << inorder;
     }
     // TODO: Swap ProximityPredicate with ComposedANDPredicate once it is
     // flattened. Once that happens, we need to add slop and inorder properties
     // to ComposedANDPredicate.
```
KarthikSubbarao: Let's remove the comment

src/commands/filter_parser.cc
```diff
@@ -752,7 +791,7 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
       // TODO: Uncomment this once we have ComposedAND evaluation functional for
       // handling proximity checks. Until the, we handle unquoted text tokens
       // by building a proximity predicate containing them.
```
KarthikSubbarao: Remove the TODO

src/commands/filter_parser.cc
```diff
@@ -436,16 +449,25 @@ inline std::unique_ptr<query::Predicate> MayNegatePredicate(
...
-std::unique_ptr<query::Predicate> WrapPredicate(
+std::unique_ptr<query::Predicate> FilterParser::WrapPredicate(
     std::unique_ptr<query::Predicate> prev_predicate,
     std::unique_ptr<query::Predicate> predicate, bool& negate,
     query::LogicalOperator logical_operator) {
   if (!prev_predicate) {
     return MayNegatePredicate(std::move(predicate), negate);
   }
+  std::optional<uint32_t> slop = options_.slop;
+  std::optional<bool> inorder = options_.inorder;
```
KarthikSubbarao: This does not have to optional since it is boolean

src/query/predicate.cc
```diff
@@ -193,41 +308,145 @@ bool TagPredicate::Evaluate(
...
+    //     << "Begin RHS iterator type: " << typeid(*rhs.filter_iterator).name()
+    //     << " CurrentKey: " << rhs.filter_iterator->CurrentKey()->Str();
+
+    // Proximity check: Only if slop/inorder set and both sides have
+    // iterators. This ensures we only check proximity for text predicates,
+    // not numeric/tag.
+    VMSDK_LOG(WARNING, nullptr)
+        << "In predicate composed AND predicate slop: " << slop_.has_value()
+        << " inorder_ value:" << inorder_.has_value();
+    if (slop_.has_value() && inorder_.has_value() && lhs.filter_iterator &&
```
KarthikSubbarao: If either one of slop or inorder are true, we need enter proximity

src/query/predicate.cc
```diff
@@ -193,41 +308,145 @@ bool TagPredicate::Evaluate(
...
+
+      // Create vector of iterators for ProximityIterator
+      std::vector<std::unique_ptr<indexes::text::TextIterator>> iterators;
+      iterators.push_back(std::move(lhs.filter_iterator));
+      iterators.push_back(std::move(rhs.filter_iterator));
+
+      // Create ProximityIterator to check proximity
+      auto proximity_iterator =
+          std::make_unique<indexes::text::ProximityIterator>(
+              std::move(iterators), *slop_, *inorder_, field_mask, nullptr);
```
KarthikSubbarao: If slop if `nullopt`, ProximityIterator needs to be provided with `nullopt`

src/coordinator/coordinator.proto
```diff
@@ -84,6 +84,8 @@ message ProximityPredicate {
 message AndPredicate {
   Predicate lhs = 1;
   Predicate rhs = 2;
+  optional uint32 slop = 3;
+  optional bool inorder = 4;
```
KarthikSubbarao: not optional

src/indexes/vector_base.h
```diff
@@ -255,10 +255,19 @@ class PrefilterEvaluator : public query::Evaluator {
   bool Evaluate(const query::Predicate& predicate,
                 const InternedStringPtr& key);
 
+  // Prefilter evalutor doesn't need a target key
```
KarthikSubbarao: The evaluation code for Prefiltering does have a specific key

src/query/predicate.cc
```diff
@@ -34,14 +39,45 @@ TermPredicate::TermPredicate(
...
+  }
+
+  auto postings = word_iter.GetTarget();
+  if (!postings) {
+    return EvaluationResult(false);
+  }
+
+  auto key_iter = postings->GetKeyIterator();
+  // Skip to the target key in the shared Postings object
+  if (!key_iter.SkipForwardKey(target_key)) {
```
KarthikSubbarao: Can we add a test for this?

src/query/predicate.cc
```diff
@@ -34,14 +39,45 @@ TermPredicate::TermPredicate(
...
+  if (!(key_iter.IsValid() && key_iter.ContainsFields(field_mask))) {
+    return EvaluationResult(false);
+  }
+
+  std::vector<indexes::text::Postings::KeyIterator> key_iterators;
+  key_iterators.emplace_back(std::move(key_iter));
+  auto iterator = std::make_unique<indexes::text::TermIterator>(
+      std::move(key_iterators), field_mask, nullptr);
+
+  if (iterator->DonePositions()) {
```
KarthikSubbarao, line 74: Would really suggest a more performant API here (HasCurrentPosition). This is O(N) for prefix search

It will be O(1) to just check if there is a position (first confirm that currentposition is Nullptr if there are no search results)

src/query/predicate.cc
```diff
@@ -52,13 +88,47 @@ PrefixPredicate::PrefixPredicate(
...
 }
 
-bool PrefixPredicate::Evaluate(const std::string_view& text) const {
-  if (text.empty()) return false;
-  return absl::StartsWith(text, term_);
+EvaluationResult PrefixPredicate::Evaluate(
+    const valkey_search::indexes::text::TextIndex& text_index,
+    const std::shared_ptr<valkey_search::InternedString>& target_key) const {
+  uint64_t field_mask = field_mask_;
+
```
KarthikSubbarao, line 97: Add a comment describing the approach used here

src/query/predicate.cc
```diff
@@ -69,13 +139,49 @@ SuffixPredicate::SuffixPredicate(
...
 
-bool SuffixPredicate::Evaluate(const std::string_view& text) const {
-  if (text.empty()) return false;
-  return absl::EndsWith(text, term_);
+EvaluationResult SuffixPredicate::Evaluate(
+    const valkey_search::indexes::text::TextIndex& text_index,
+    const std::shared_ptr<valkey_search::InternedString>& target_key) const {
+  uint64_t field_mask = field_mask_;
+
+  if (!text_index.suffix_.has_value()) {
```
KarthikSubbarao, line 149: There is a public API called `WithSuffixTrie` which we can use

src/query/predicate.cc
```diff
@@ -52,13 +88,47 @@ PrefixPredicate::PrefixPredicate(
...
+    std::string_view word = word_iter.GetWord();
+    if (!word.starts_with(term_)) break;
+
+    auto postings = word_iter.GetTarget();
+    if (postings) {
+      auto key_iter = postings->GetKeyIterator();
+      bool skip_result = key_iter.SkipForwardKey(target_key);
+      bool contains = skip_result && key_iter.ContainsFields(field_mask);
+
+      if (contains) {
```
KarthikSubbarao: Let's use the boolean check from the function below as it is more clearer

src/query/predicate.cc
```diff
@@ -126,10 +232,17 @@ ProximityPredicate::ProximityPredicate(
...
-bool ProximityPredicate::Evaluate(Evaluator& evaluator) const {
+EvaluationResult ProximityPredicate::Evaluate(Evaluator& evaluator) const {
   return evaluator.EvaluateText(*this);
 }
 
+EvaluationResult ProximityPredicate::Evaluate(
+    const valkey_search::indexes::text::TextIndex& text_index,
+    const std::shared_ptr<valkey_search::InternedString>& target_key) const {
+  // Temporarily make this pass through for tests.
+  return EvaluationResult(true);
```
KarthikSubbarao: Let's make this return false until ProximityPredicate is deleted

src/query/predicate.cc
```diff
@@ -193,41 +308,145 @@ bool TagPredicate::Evaluate(
...
-  return false;
+  return EvaluationResult(false);
 }
 
 ComposedPredicate::ComposedPredicate(std::unique_ptr<Predicate> lhs_predicate,
                                      std::unique_ptr<Predicate> rhs_predicate,
-                                     LogicalOperator logical_op)
+                                     LogicalOperator logical_op,
+                                     std::optional<uint32_t> slop,
+                                     std::optional<bool> inorder)
```
KarthikSubbarao: not optional

src/query/predicate.cc
```diff
@@ -193,41 +308,145 @@ bool TagPredicate::Evaluate(
...
+
+      VMSDK_LOG(WARNING, nullptr)
+          << "proximity iterator. current key"
+          << proximity_iterator.get()->CurrentKey()->Str();
+
+      // Return the proximity iterator for potential nested use.
+      return EvaluationResult(true, std::move(proximity_iterator));
+    }
+    // Both matched, non-proximity case
+    return EvaluationResult(true);
```
KarthikSubbarao, line 397: We are currently dropping the positions of either LHS / RHS if one of them does not have positions.

This should be changed to include the TextIterator of either one which has it.  We can continue to return true and still return the TextIterator.

Also, very important to add a test for this

src/query/predicate.h
```diff
@@ -294,15 +344,21 @@ class ComposedPredicate : public Predicate {
  public:
   ComposedPredicate(std::unique_ptr<Predicate> lhs_predicate,
                     std::unique_ptr<Predicate> rhs_predicate,
-                    LogicalOperator logical_op);
+                    LogicalOperator logical_op,
+                    std::optional<uint32_t> slop = std::nullopt,
+                    std::optional<bool> inorder = std::nullopt);
```
KarthikSubbarao: not optional

src/query/response_generator.cc
```diff
@@ -78,64 +79,120 @@ namespace valkey_search::query {
 
 class PredicateEvaluator : public query::Evaluator {
  public:
-  explicit PredicateEvaluator(const RecordsMap &records) : records_(records) {}
-  bool EvaluateTags(const query::TagPredicate &predicate) override {
+  explicit PredicateEvaluator(const RecordsMap &records)
+      : records_(records), per_key_indexes_(nullptr), target_key_(nullptr) {}
```
KarthikSubbarao, line 83: Is there any other way to feed this in?

src/query/response_generator.cc
```diff
@@ -78,64 +79,120 @@ namespace valkey_search::query {
...
-    // TODO: Implement TextPredicate evaluation.
-    return true;
+  EvaluationResult EvaluateText(
+      const query::TextPredicate &predicate) override {
+    // If we have per-key indexes, use them for text evaluation
+    if (per_key_indexes_ && target_key_) {
+      return EvaluateTextUsingIndex(predicate);
+    }
+    // Fallback: no per-key index available
+    return EvaluationResult(false);
```
KarthikSubbarao: Is this ever possible? I am trying to understand if this is unexpected. 

Some cases I can think of are if an index was deleted in between the FT.SEARCH and the time taken from background + yield back to mainthread. 

But you can confirm if this is the first validation using the per_key_indexes. if this is the first validation, we can keep this change

src/indexes/vector_base.h
daddaman-amz: yes this is temporary placeholder. will update the comment

src/query/response_generator.cc
daddaman-amz: Alternative approaches exist , but this follows the existing Evaluator interface pattern and avoids changing parameter changes for EvaluateNumeric() and EvaluateTags().

src/query/response_generator.cc
daddaman-amz: Yes this is first validation for per_key_indexes. Ideally this should never happen and is more like a defensive fallback. We can think about making this Check in future either here or somewhere higher

src/commands/filter_parser.cc
```diff
@@ -702,6 +713,7 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
     for (const auto& identifier : text_identifiers) {
       filter_identifiers_.insert(identifier);
     }
+
```
KarthikSubbarao: Nit: Extra new line can be removed

src/commands/filter_parser.cc
```diff
@@ -764,18 +775,22 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
...
-    }
-    // TODO: Swap ProximityPredicate with ComposedANDPredicate once it is
-    // flattened. Once that happens, we need to add slop and inorder properties
-    // to ComposedANDPredicate.
-    pred = std::make_unique<query::ProximityPredicate>(std::move(terms), slop,
-                                                       inorder);
+    // Exact phrase requires adjacent terms in order: slop=0, inorder=true
+    uint32_t slop = 0;
+    bool inorder = true;
+
```
KarthikSubbarao: Nit: Can we remove unnecessary new lines to help with readability across your PR? (not just here)

src/indexes/text/proximity.cc
```diff
@@ -14,7 +14,6 @@ ProximityIterator::ProximityIterator(
       current_position_(std::nullopt),
       field_mask_(field_mask) {
   CHECK(!iters_.empty()) << "must have at least one text iterator";
-  CHECK(slop_ >= 0) << "slop must be non-negative";
```
KarthikSubbarao, line 17: Can we update the check to validate that either SLOP OR INORDER=true should be set? (It does not require both)

src/query/predicate.cc
```diff
@@ -193,41 +306,103 @@ bool TagPredicate::Evaluate(
...
+  EvaluationResult rhs = rhs_predicate_->Evaluate(evaluator);
+  VMSDK_LOG(DEBUG, nullptr)
+      << "Inline evaluate OR predicate rhs: " << rhs.matches;
+
+  // TODO: Implement position-aware OR logic for nested proximity queries.
+  return EvaluationResult(lhs.matches || rhs.matches);
 }
 
-}  // namespace valkey_search::query
+}  // namespace valkey_search::query
```
KarthikSubbarao: Add new line

src/query/predicate.cc
```diff
@@ -16,13 +16,18 @@
 #include "absl/strings/string_view.h"
 #include "src/indexes/numeric.h"
 #include "src/indexes/tag.h"
+#include "src/indexes/text.h"
```
KarthikSubbarao, line 19: Can you take a look at comments on this file and address OR reply? Thanks

src/query/response_generator.cc
KarthikSubbarao: Let's discuss this

KarthikSubbarao: Isn't there a check in the first entry point in `VerifyFilter`?

src/query/response_generator.cc
```diff
@@ -78,64 +79,120 @@ namespace valkey_search::query {
...
+  InternedStringPtr target_key_;
+
+  EvaluationResult EvaluateTextUsingIndex(
+      const query::TextPredicate &predicate) {
+    auto it = per_key_indexes_->find(target_key_);
+    if (it == per_key_indexes_->end()) {
+      VMSDK_LOG(WARNING, nullptr)
+          << "Target key not found in index for predicate evaluation";
+      return EvaluationResult(false);
+    }
```
KarthikSubbarao: Observation: It seems odd to return false this far along the Predicate evaluation when we could have validated this upfront in VerifyFilter IF we knew that the query syntax tree contained a Text predicate. This might be something to think about / add a flag for if there are other uses for it.

src/query/predicate.cc
daddaman-amz: Will take this in prefilter as discussed

src/query/predicate.cc
daddaman-amz: If you are referring to the use of per key text index instead of records, I have added it in Verifyfilter()

```
  // For text predicates, evaluate using the text index instead of raw data.
  if (parameters.index_schema &&
      parameters.index_schema->GetTextIndexSchema()) {
    // TODO: Wait for any in-flight indexing operations to complete before
    // acquiring the lock, ensuring we evaluate against the latest index state.
    return parameters.index_schema->GetTextIndexSchema()->WithPerKeyTextIndexes(
        [&](auto &per_key_indexes) {
          PredicateEvaluator evaluator(records, &per_key_indexes, target_key);
          EvaluationResult result = predicate->Evaluate(evaluator);
          return result.matches;
        });
  }
  ```

File: src/query/predicate.cc (hunk 19 above)
Comment (daddaman-amz) [reply src/query/predicate.cc]: done

File: src/query/response_generator.cc (hunk 30 above)
Comment (daddaman-amz) [reply src/query/response_generator.cc]: I removed `EvaluateTextUsingIndex()` and inlined the code in `EvaluateText()`.

Regarding validation upfront in VerifyFilter: It may lead to changes in constructor signatures. The flag does seem useful, especially when we do changes for prefilter. I can explore these two aspects during prefilter work if it turns out useful.

Comment (daddaman-amz) [reply src/query/response_generator.cc]: Discussed offline

File: src/query/response_generator.cc (hunk 24 above)
Comment (daddaman-amz) [reply src/query/response_generator.cc]: resolved offline, removed redundant checks

File: src/query/predicate.cc (hunk 14 above)
Comment (daddaman-amz) [reply src/query/predicate.cc]: Discussed offline , right now there is no good way to test this as the initial search will send key to main thread only if it present

File: src/query/predicate.cc (hunk 17 above)
Comment (daddaman-amz) [reply src/query/predicate.cc:149]: Discussed offline. Per field index , which has the WithSuffixTrie  api , is not available currently in predicate.cc.

File: src/query/predicate.cc [hunk 31]
```diff
@@ -193,41 +306,107 @@ bool TagPredicate::Evaluate(
...
+      // Create ProximityIterator to check proximity
+      auto proximity_iterator =
+          std::make_unique<indexes::text::ProximityIterator>(
+              std::move(iterators), slop_, inorder_, field_mask, nullptr);
+
+      // Check if any valid proximity matches exist
+      if (proximity_iterator->DonePositions()) {
+        return EvaluationResult(false);
+      }
+      // Validate against original target key from evaluator
```
KarthikSubbarao, line 378:
```suggestion
      // Validate that we are on the original target key Evaluate call to ensure the proximity iterator has not moved it to a new key
```

PR #444, opened 2025-11-03
Stop Incrementally Updating Token Positions

BCathcart:
@Aksha1812 is working on a compressed format for the position map which won't be easy to incrementally update. The text indexes operate at the schema-level and the position map can contain information from multiple text attributes (aka "fields) in the key. With this change, each Text::AddRecord() call simply stages the new text data and it's committed at the end in one go after every field is processed. Even without serialization, this avoids traversing the radix tree to find the postings object and looking up the key in the sorted map for every word (aka "token") in every text field of the key. This is replaced by simply looking up the token in an unsorted map, and doing a single radix tree traversal for each unique token at the end to insert the key.

In more detail, the text index effectively maps tokens -> keys -> (positions -> fields) via a radix tree pointing to postings objects. While processing a key, I maintain an unsorted map of tokens -> (positions -> fields) for the key. After every field of the key is processed, I flush the data into the schema-level text index. Right now the positions -> fields (PositionMap) is put directly into the posting object for that word. Soon we'll serialize it first into a compressed structure.

### DIFF
--- a/src/indexes/text/text_index.cc
+++ b/src/indexes/text/text_index.cc
@@ -222,13 +222,8 @@ uint64_t TextIndexSchema::GetPostingsMemoryUsage() const {
 }
 
 uint64_t TextIndexSchema::GetRadixTreeMemoryUsage() const {
-  // TODO: not implemented
-  // if (!text_index_) {
-  //   return 0;
-  // }
-  // return metadata_.radix_memory_pool_.GetUsage() -
-  //        metadata_.posting_memory_pool_.GetUsage();
-  return 0;
+  // TODO: properly implement
+  return GetTotalTextIndexMemoryUsage() - GetPostingsMemoryUsage();
 }
 
 // Note: This is a subset of the memory reported by GetPostingsMemoryUsage(),
--- a/src/indexes/text/text_index.h
+++ b/src/indexes/text/text_index.h
@@ -85,8 +85,6 @@ class TextIndexSchema {
 
   // Access to metadata for memory pool usage
   TextIndexMetadata& GetMetadata() { return metadata_; }
-  // const TextIndexMetadata& GetMetadata() const { return metadata_; } - when
-  // is TextIndexSchema ever const?
 
  private:
   uint8_t num_text_fields_ = 0;
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -7,6 +7,7 @@ from ft_info_parser import FTInfoParser
 from valkeytestframework.util import waiters
 import threading
 import time
+import os
 from utils import IndexingTestHelper
 
 """
@@ -743,9 +744,9 @@ class TestFullTextDebugMode(ValkeySearchTestCaseDebugMode):
         assert info_after_delete["num_docs"] == 0, f"Expected 0 documents after deletion, got {info_after_delete['num_docs']}"
         
         # Verify complete cleanup for schema-level metrics
-        assert info_after_delete['posting_sz_bytes'] == 0, \
-            f"Posting memory should be zero after deletion: got {info_after_delete['posting_sz_bytes']}"
-        
+        # TODO: this won't be zero until serialized position map is added
+        # assert info_after_delete['posting_sz_bytes'] == 0, \
+        #     f"Posting memory should be zero after deletion: got {info_after_delete['posting_sz_bytes']}"
         
         # Note: Postings are currently stored in per-key indexes. When we implement shared
         # Posting objects across schema and key indexes, these metrics will reach zero on deletion.


src/indexes/text/posting.cc
```diff
@@ -164,75 +164,24 @@ template class FieldMaskImpl<uint64_t, 64>;
...
- public:
-  bool save_positions_;
-  size_t num_text_fields_;
-  std::map<Key, PositionMap> key_to_positions_;
-
-  Impl(bool save_positions, size_t num_text_fields)
-      : save_positions_(save_positions), num_text_fields_(num_text_fields) {}
-};
-
-Postings::Postings(bool save_positions, size_t num_text_fields)
```
Aksha1812: I would suggest keeping the Postings constructor. because otherwise it would allocate memory in the background and we wouldn't be able to track. See this https://github.com/valkey-io/valkey-search/pull/377/files#diff-bee98baab089d88ae05a6e4ce43ebbafcc5842de5223aeffcfda280cd2b712c4R186 Instead of Impl class you can initialize the key_to_positions map in the constructor. So memory scope can be put over it and we can track the allocation. Keep the destrctor for now as well. will handle the deallocation there as well. Ig you should do above after memory tracking PR gets merged, you will get some idea what exactly you would have to change.

src/indexes/text/posting.cc
```diff
@@ -241,7 +190,7 @@ size_t Postings::GetPostingCount() const {
 // Get total term frequency (sum of field occurrences across all positions)
 size_t Postings::GetTotalTermFrequency() const {
   size_t total_frequency = 0;
-  for (const auto& [key, positions] : impl_->key_to_positions_) {
+  for (const auto& [key, positions] : key_to_positions_) {
     for (const auto& [position, field_mask] : positions) {
```
Aksha1812, line 215: I am assuming you are still storing the old Position Map as it is in the key_to_positions_ which would allow us to do all this . If so that is correct . I can change other APIs with my PR. But we have to decide which PR gets merged first (probably yours ) since it sets the base for me add serialization. Note to self : Update all these helper functions as well, with Position Optimization PR

src/indexes/text/posting.h
```diff
@@ -171,8 +161,7 @@ struct Postings {
   };
 
  private:
-  class Impl;
-  std::unique_ptr<Impl> impl_;
+  std::map<Key, PositionMap> key_to_positions_;
```
Aksha1812, line 164: _Possible Future Optimization :_ Maybe in future we can think of a possibility of a Posting object like `using Postings = std::map<Key, PositionMap>` . After this PR it looks like the Postings class is just a wrapper container over the key_to_positions_ map taking unneccesary extra 8 bytes.

src/indexes/text/text_index.cc
```diff
@@ -64,93 +62,111 @@ absl::StatusOr<bool> TextIndexSchema::IndexAttributeData(
...
+        });
+    if (text_index_->suffix_.has_value()) {
+      std::string reverse_word(word.rbegin(), word.rend());
+      text_index_->suffix_.value().SetTarget(reverse_word, new_target);
     }
-  }
-
-  {
-    std::lock_guard<std::mutex> per_key_guard(per_key_text_indexes_mutex_);
-    per_key_text_indexes_.erase(key);
```
Aksha1812, line 182: where is this erase call happening now?

src/indexes/text/text_index.h
```diff
@@ -27,9 +27,11 @@ struct sb_stemmer;
 
 namespace valkey_search::indexes::text {
 
+// token -> (PositionMap, suffix support)
+using TokenPositions =
+    absl::flat_hash_map<std::string, std::pair<PositionMap, bool>>;
```
Aksha1812, line 33: little bit confused from above, can you clarify what is std::string here? i.e the key from this tokenpositions key value pair?

src/index_schema.cc
```diff
@@ -451,6 +451,11 @@ void IndexSchema::SyncProcessMutation(ValkeyModuleCtx *ctx,
                              std::move(attribute_data_itr.second.data),
                              attribute_data_itr.second.deletion_type);
   }
+  if (text_index_schema_) {
```
Aksha1812, line 455: Is there a batching mechanism that will stage certain keys and then we would proceed to commiting them , while other keys would be staging in the background? Curios how and when staging and commiting would happen

src/indexes/text/text_index.cc
```diff
@@ -16,23 +16,21 @@ namespace valkey_search::indexes::text {
...
 namespace {
 
-std::optional<std::shared_ptr<text::Postings>> AddWordToPostings(
-    std::optional<std::shared_ptr<text::Postings>> existing,
-    const InternedStringPtr& key, size_t text_field_number, uint32_t position,
-    bool save_positions, uint8_t num_text_fields) {
-  std::shared_ptr<text::Postings> postings;
+std::optional<std::shared_ptr<Postings>> AddKeyToPostings(
+    std::optional<std::shared_ptr<Postings>> existing,
+    const InternedStringPtr& key, PositionMap& pos_map) {
```
allenss-amazon: PositionMap &&, and you'll need std::move(..) at the call sites.

src/indexes/text/text_index.cc
```diff
@@ -64,93 +62,111 @@ absl::StatusOr<bool> TextIndexSchema::IndexAttributeData(
...
+    const auto& token = tokens.value()[i];
+    uint32_t position =
+        with_offsets_ ? i
+                      : 0;  // If positional info is disabled we default to 0
+    auto& [positions, suffix_eligible] = (*token_positions)[token];
+    if (suffix) suffix_eligible = true;
+    auto [pos_it, _] =
+        positions.try_emplace(position, FieldMask::Create(num_text_fields_));
+    pos_it->second->SetField(text_field_number);
+  }
```
allenss-amazon, line 96: I don't understand how this code avoids collisions on a per-field basis. I understand the locking for the key. But won't it be the case that multiple fields can be processed in parallel? How does that ensure integrity of in_progress_key_updates_ ??

src/index_schema.cc
BCathcart: `SyncProcessMutation()` gets called to process a key mutation. You can see above that each attribute of the key is processed one by one. `ProcessAttributeMutation()` then calls `AddRecord()/MutateRecord()/DeleteRecord()`. Any staging for the text index will happen here. Other keys can be processed concurrently.

src/indexes/text/posting.cc
BCathcart: I removed impl_ pattern here since I didn't see any benefit to having that layer of indirection. Nothing fundamentally changes because of this change.

src/indexes/text/posting.h
BCathcart: Not sure where the 8 bytes are coming from. I think using a C++ class here to encapsulate data with the functions that operate on it is the right idea.

src/indexes/text/text_index.cc
BCathcart: No the fields get processed sequentially in a loop in `SyncProcessMutation` which eventually gets called in the async mutation processing flow as well

src/indexes/text/text_index.cc
BCathcart: I call `extract()` now above which remove and returns it

src/indexes/text/text_index.h
BCathcart: It's the word/token

src/indexes/text/text_index.cc
BCathcart: Good catch. Fixed

src/indexes/text/posting.cc
BCathcart: I just wrapped the `AddKeyToPostings` and `RemoveKeyFromPostings` function calls for postings object memory tracking. It doesn't really work right now but should once you add the serialization. In also got rid of radix tree memory tracking for the moment since I think there are a couple details to think about and I just wanted to keep things as simple as possible for the moment. Are the total text index memory and postings object memory enough for the moment? The difference between them can be inferred to roughly be the radix tree overhead

.github/workflows/integration_tests.yml
```diff
@@ -33,17 +33,30 @@ jobs:
       run: docker build -t presubmit-image -f .devcontainer/Dockerfile .
 
     - name: Run Integration Tests
-      run: mkdir test-results && docker run --privileged --rm
-        --mount type=bind,source="$(pwd)/test-results",target=/workspace/test-results
-        -v "$(pwd):/workspace"
-        --user "ubuntu:ubuntu"
-        presubmit-image
-        sudo ci/build_ubuntu.sh --run-integration-tests --integration-output=/workspace/test-results
+      id: test-run
```
BCathcart, line 36: I ran into a permission error with the test result upload at first and then noticed that the job still passed even if the test failed due to `continue-on-error: true`. I'll make these changes in main for all test workflows when I have time

src/indexes/text/posting.cc
```diff
@@ -164,115 +164,45 @@ template class FieldMaskImpl<uint64_t, 64>;
...
-  if (impl_->save_positions_) {
-    // In positional mode, position must be explicitly provided
-    CHECK(position != UINT32_MAX)
-        << "Position must be provided in positional mode";
-    effective_position = position;
-  } else {
-    // For boolean search mode, always use position 0 regardless of input
-    effective_position = 0;
+// TODO: develop a better strategy to track terms
+unsigned int count_num_terms(const PositionMap& pos_map) {
```
Aksha1812, line 171: Can't you do this while staging? . I was incrementing total_term_frequency whenever we inserted a token. we can do the same whole staging right? there might be a inconsistency though between the time when it is staging and commiting. Is there any workaround that does not involve us looping again and again?

src/indexes/text/posting.cc
```diff
@@ -164,115 +164,45 @@ template class FieldMaskImpl<uint64_t, 64>;
...
-  }
-}
-
 // Check if posting list contains any documents
-bool Postings::IsEmpty() const { return impl_->key_to_positions_.empty(); }
-
-// Insert a posting entry for a key and field
-void Postings::InsertPosting(const Key& key, size_t field_index,
-                             Position position) {
-  NestedMemoryScope scope{
```
Aksha1812, line 208: can you add the NestedMemoryScopes back where they should be. currently looks like this PR negates all changes made in memory tracking PR and i might have to do it all again , rather than tweaking it

src/indexes/text/posting.cc
```diff
@@ -164,115 +164,45 @@ template class FieldMaskImpl<uint64_t, 64>;
...
- public:
-  bool save_positions_;
-  size_t num_text_fields_;
-  std::map<Key, PositionMap> key_to_positions_;
-
-  Impl(bool save_positions, size_t num_text_fields)
-      : save_positions_(save_positions), num_text_fields_(num_text_fields) {}
-};
-
-Postings::Postings(bool save_positions, size_t num_text_fields) {
```
Aksha1812, line 180: I need the explicit declaration of constructor with initialization of key_to_positions map to track memory allocated there

.github/workflows/integration_tests.yml
allenss-amazon: Oops. Good catch.

src/indexes/text/posting.cc
```diff
@@ -164,115 +164,45 @@ template class FieldMaskImpl<uint64_t, 64>;
...
-  NestedMemoryScope scope{
-      GetTextIndexSchema()->GetMetadata().posting_memory_pool_};
-  impl_ = std::make_unique<Impl>(save_positions, num_text_fields);
-  CHECK(impl_ != nullptr) << "Failed to create Postings implementation";
-}
-
-// Destructor with proper memory tracking and cleanup
-Postings::~Postings() {
```
Aksha1812, line 188: Can you move this back in , and make changes accordingly, all this was added so we can track memory being freed before Posting object destroys

src/indexes/text/text_index.cc
```diff
@@ -88,98 +77,128 @@ absl::StatusOr<bool> TextIndexSchema::IndexAttributeData(
...
     {
-      std::lock_guard<std::mutex> schema_guard(text_index_->mutex_);
-      NestedMemoryScope radix_scope{metadata_.radix_memory_pool_};
-      std::optional<std::shared_ptr<Postings>> new_target =
+      std::lock_guard<std::mutex> schema_guard(text_index_mutex_);
+      updated_target =
           text_index_->prefix_.MutateTarget(token, [&](auto existing) {
-            return AddWordToPostings(existing, key, text_field_number, position,
-                                     with_offsets_, num_text_fields_);
+            NestedMemoryScope scope{metadata_.posting_memory_pool_};
```
Aksha1812, line 137: you can move it back to AddKeyToPostings function. It doesn't matter though if we place it in either place,

src/indexes/text/text_index.h
```diff
@@ -86,7 +85,8 @@ class TextIndexSchema {
 
   // Access to metadata for memory pool usage
   TextIndexMetadata& GetMetadata() { return metadata_; }
-  const TextIndexMetadata& GetMetadata() const { return metadata_; }
+  // const TextIndexMetadata& GetMetadata() const { return metadata_; } - when
```
Aksha1812: yes you can delete this

src/indexes/text/posting.h
allenss-amazon: Ideally, we'd want PositionMap to be a first-class class that obeys all the rules of construction, destruction, copy and move constructors, etc. It's complicated because we want it to be able to efficiently address an inherently variable size amount of memory. I think we can have it both ways, i.e., a "wrapper" class that obeys the rules with an underyling implementation that's space efficient.

allenss-amazon: I can help with an outline of how to do that, if it helps.

src/indexes/text/text_index.cc
```diff
@@ -203,11 +222,13 @@ uint64_t TextIndexSchema::GetPostingsMemoryUsage() const {
...
-  }
-  return metadata_.radix_memory_pool_.GetUsage() -
-         metadata_.posting_memory_pool_.GetUsage();
+  // TODO: not implemented
+  // if (!text_index_) {
+  //   return 0;
+  // }
+  // return metadata_.radix_memory_pool_.GetUsage() -
+  //        metadata_.posting_memory_pool_.GetUsage();
+  return 0;
```
Aksha1812: this will fail the test_fulltext , where in debug mode i am expecting some value for this

src/indexes/text/text_index.cc
allenss-amazon: ack. I see.

src/indexes/text/text_index.cc
```diff
@@ -203,11 +222,13 @@ uint64_t TextIndexSchema::GetPostingsMemoryUsage() const {
 }
 
 uint64_t TextIndexSchema::GetRadixTreeMemoryUsage() const {
-  if (!text_index_) {
-    return 0;
-  }
-  return metadata_.radix_memory_pool_.GetUsage() -
-         metadata_.posting_memory_pool_.GetUsage();
+  // TODO: not implemented
```
Aksha1812: I know implementation for this is complete end to end , but currently these info fields are in debug mode and atleast memory allocation up to some extent is accurate. And i was using it to get a rough estimate. Can you move it back in

src/indexes/text/posting.cc
BCathcart: I wrap the AddKeyToPostings and RemoveKeyFromPostings calls so anything that is created or destroyed in that scope should be included. Posting objects wouldn't be created or destroyed anywhere else would they?

src/indexes/text/posting.cc
BCathcart: Since we already needed to fix this for the remove case, I thought might as well use it for the add case as well for the moment and then look at it holistically in a follow-up

src/indexes/text/text_index.cc
BCathcart: Good point. No need to wrap outside the functions

src/indexes/text/text_index.cc
BCathcart: I put it back a little earlier as the difference between total index memory and postings memory. Obviously not perfect and will include the per-key map overhead for example, but is that good enough for the moment?

Aksha1812 (approved):

PR #446, opened 2025-11-03
Aggregate Cleanup

allenss-amazon:
Restructures the integration of FT.AGGREATE. Now, FT.AGGREGATE and FT.SEARCH are derived from a common class "QueryCommand". This allows sharing of common code between the two commands, with virtual functions for handling per-command specific functions.


src/commands/commands.cc
```diff
@@ -0,0 +1,141 @@
...
+      blocked_client.SetReplyPrivateData(result.release());
+    };
+
+    if (ValkeySearch::Instance().UsingCoordinator() &&
+        ValkeySearch::Instance().IsCluster() && !parameters->local_only) {
+      auto mode = /* !vmsdk::IsReadOnly(ctx) ? query::fanout::kPrimaries ? */
+          ForceReplicasOnly.GetValue()
+              ? query::fanout::FanoutTargetMode::kReplicasOnly
+              : query::fanout::FanoutTargetMode::kRandom;
+      auto search_targets = query::fanout::GetSearchTargetsForFanout(ctx, mode);
```
boda26, line 124: This is using the old code in fanout_template.h. With the new cluster map, this can be replaced by ValkeySearch()::Instance().GetClusterMap()->GetRandomTargets()/GetReplicaTargets(). A switch clause might be needed here to determine the get target mode.
If the ctx is valid here (which I think should be the case), we should use ValkeySearch()::Instance().GetOrRefreshClusterMap()

boda26 (approved):

PR #452, opened 2025-11-05
Fix tag OR syntax in hybrid queries and add filter expression tests

### LINKED ISSUE #449: [BUG] (reopened) Unable to use {USA|GBR|CAN} syntax for tag filter in KNN hybrid query
**Describe the bug:**
According to documentation, filtering with syntax `@country:{USA|GBR|CAN}` should work for hybrid queries, but it does not. 

**Expected:**
This is expected to work as an equivalent for an OR filter, e.g. `(@country:{USA} | @country:{GBR} | @country:{CAN})`.

**Actual:**
The query does not return results or is not accepted, while the equivalent OR filter in parentheses works as expected.

**Example:**
This works (returns results):
```
FT.SEARCH idx_hnsw_cosine '(@country:{USA} | @country:{GBR} | @country:{CAN})=>[KNN 5 @text_embedding_vector $query_vector]' PARAMS 2 query_vector $'...'
```
This does not work (no results or error):
```
FT.SEARCH vector_index_x '@country:{USA|GBR|CAN}=>[KNN 5 @text_embedding_vector $query_vector]' PARAMS 2 query_vector $'...'
```

**Environment**

- Valkey Search Module Version: main branch (commit SHA 90124dc9)
- Valkey Core Version: build-debug (commit SHA if known)
- OS: Linux

zvi-code:
Fixes #449

## Summary
Fix tag OR syntax in hybrid queries and add comprehensive filter expression tests.

### Tag Filter Parsing Fix
- Use `'|'` separator for tag search queries instead of the index-defined separator
- Allows consistent `@field:{tag1|tag2|tag3}` syntax in all queries regardless of index configuration
- Previously used the separator from index creation, causing inconsistencies

### New Test Coverage
Add `test_filter_expressions.py` with 23 test cases covering:
- Tag filters with OR syntax (`@field:{tag1|tag2|tag3}`)
- All 9 numeric range variants (inclusive/exclusive bounds, infinity, equality)
- Logical negation (`-@field:{value}`)
- Operator precedence (negation > AND > OR)
- Hybrid queries (filters + vector search)
- Complex multi-filter queries

This ensures the tag OR syntax works correctly across all query patterns documented in COMMANDS.md.

### DIFF
--- a/src/indexes/tag.cc
+++ b/src/indexes/tag.cc
@@ -68,29 +68,65 @@ absl::StatusOr<bool> Tag::AddRecord(const InternedStringPtr& key,
   return true;
 }
 
+std::string Tag::UnescapeTag(absl::string_view tag) {
+  std::string result;
+  result.reserve(tag.size());
+  for (size_t i = 0; i < tag.size(); ++i) {
+    if (tag[i] == '\\' && i + 1 < tag.size()) {
+      // Escape sequence: consume next character literally
+      result += tag[++i];
+    } else {
+      result += tag[i];
+    }
+  }
+  return result;
+}
+
 absl::StatusOr<absl::flat_hash_set<absl::string_view>> Tag::ParseSearchTags(
     absl::string_view data, char separator) {
   absl::flat_hash_set<absl::string_view> parsed_tags;
-  std::vector<absl::string_view> parts = absl::StrSplit(data, separator);
-  for (const auto& part : parts) {
-    auto tag = absl::StripAsciiWhitespace(part);
-    if (tag.empty()) {
-      continue;
-    }
 
-    // Prefix tag is identified by a trailing '*'.
+  // Helper: validate and insert a single tag (handles prefix wildcards)
+  auto InsertTag = [&](absl::string_view raw) -> absl::Status {
+    auto tag = absl::StripAsciiWhitespace(raw);
+    if (tag.empty()) {
+      return absl::OkStatus();  // Empty tags are silently ignored
+    }
     if (tag.back() == '*') {
       if (!IsValidPrefix(tag)) {
         return absl::InvalidArgumentError(
             absl::StrCat("Tag string `", tag, "` ends with multiple *."));
       }
-      // Prefix tags that are shorter than min length are ignored.
-      if (tag.length() <= kDefaultMinPrefixLength) {
-        continue;
+      // Prefix tags shorter than min length are ignored
+      if (tag.length() > kDefaultMinPrefixLength) {
+        parsed_tags.insert(tag);
       }
+    } else {
+      parsed_tags.insert(tag);
     }
-    parsed_tags.insert(tag);
+    return absl::OkStatus();
+  };
+
+  // Parse respecting escape sequences:
+  // - \<separator> is NOT a real separator
+  // - \\ is an escaped backslash (not an escape prefix)
+  // - Unescaped separator splits tags
+  // Returns string_view positions; unescaping happens later at TagPredicate.
+  size_t tag_start = 0;
+  for (size_t i = 0; i < data.size(); ++i) {
+    if (data[i] == '\\' && i + 1 < data.size()) {
+      // Skip escaped character
+      ++i;
+    } else if (data[i] == separator) {
+      // Unescaped separator: extract tag
+      VMSDK_RETURN_IF_ERROR(InsertTag(data.substr(tag_start, i - tag_start)));
+      tag_start = i + 1;
+    }
   }
+
+  // Handle the last tag (after final separator or if no separator)
+  VMSDK_RETURN_IF_ERROR(InsertTag(data.substr(tag_start)));
+
   return parsed_tags;
 }
 
--- a/src/indexes/tag.h
+++ b/src/indexes/tag.h
@@ -11,6 +11,7 @@
 #include <cstdint>
 #include <memory>
 #include <optional>
+#include <string>
 
 #include "absl/base/thread_annotations.h"
 #include "absl/container/flat_hash_set.h"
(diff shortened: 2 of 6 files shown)


src/commands/filter_parser.cc
```diff
@@ -218,7 +218,10 @@ absl::StatusOr<absl::string_view> FilterParser::ParseTagString() {
 
 absl::StatusOr<absl::flat_hash_set<absl::string_view>> FilterParser::ParseTags(
     absl::string_view tag_string, indexes::Tag* tag_index) const {
-  return indexes::Tag::ParseSearchTags(tag_string, tag_index->GetSeparator());
+  // In search queries, the tag separator is always '|' regardless of the
+  // separator used when the index was created. This allows users to specify
+  // multiple tags using the syntax: @field:{tag1|tag2|tag3}
+  return indexes::Tag::ParseSearchTags(tag_string, '|');
```
allenss-amazon, line 224: While this is correct. It leaves open the situation of parsing a tag that itself contains a pipe character. I think the tag parsing code should be updated to handle escaping of the separator character.

integration/test_filter_expressions.py
```diff
@@ -0,0 +1,790 @@
...
+            "RETURN", "1", "country"
+        )
+        
+        # Should return up to 3 results (filtered by country tag)
+        assert result[0] >= 1 and result[0] <= 3
+        
+        # Verify all results match the country filter
+        for i in range(1, len(result), 2):
+            key = result[i].decode('utf-8')
+            assert key in ["doc:1", "doc:2", "doc:3"], f"Unexpected key {key}"
```
allenss-amazon, line 150: Yair pointed out that there are two paths for hybrid queries. I believe we can force the logic to use one way or the other configuring the decision threshold (and there are counters to validate which way it went). Can we update the tests that have hybrid queries to force both paths?

allenss-amazon: It would also be good to repeat the vector tests with both HNSW and FLAT, again different code paths.

allenss-amazon: Another issue that's been reported is that the order of the results is unexpected, it's possible that we're getting different ordering of the results for vector queries.

src/commands/filter_parser.cc
zvi-code: @allenss-amazon , Do we have somewhere definition of valid\invalid characters for tags? Can `"@country:{USA|GBR|CAN}=>[KNN 5 @embedding $vec]"` be a tag?

integration/test_filter_expressions.py
zvi-code: what do you mean? I have these 2:
1. "@country:{USA|GBR|CAN}=>[KNN 5 @embedding $vec]"
2. "@country:{USA} | @country:{GBR} | @country:{CAN}=>[KNN 5 @embedding $vec]"
Anything else?

My assumption was that the higher levels of parsing is done correctly, and the change does not affect those, but if I can extend the testing to check logically composed queries, like `query1 | query2 | query3`, where query-i is a valid filter by itself, is this your intent?

allenss-amazon: I'm trying to get alignment in capabilities between the ingestion and query subsystems. Assuming that this PR is merged more or less as-is, I believe we're in a state where a tag is forbidden from containing two characters: | and }. I note that this isn't an ingestion problem as we allow the redefinition of the tag separator character. This means that as long as your tags don't contain all 256 bytes, you can ingest all of them. However, you can't query for a tag that contains a pipe or a right brace -- because the parser has no way to treat these as part of a tag, they are hard-coded as delimiters.

I'm going to open an enhancement issue to specify that the query tag parsing logic be updated to support escaping of these two characters. We might roll the fix for this into the future Unicoding of the query logic.

zvi-code: @allenss-amazon, Our documentation explicitly specifies the syntax query operator and these have nothing todo with the tag delimiter. I agree we need to have complete query syntax, I don't think we should connect the 2. This is a fix for a bug with respect to current declared usage.

allenss-amazon: I agree that we can split this. I've opened a separate issue for the future enhancement to do the escape parsing.

src/commands/filter_parser.cc
yairgott: We're trying to stay aligned with RediSearch. Browsing their documentation, I found the [following](https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/tags/#handle-special-characters).

integration/test_filter_expressions.py
allenss-amazon: This is captured in the following issue https://github.com/valkey-io/valkey-search/issues/454

src/commands/filter_parser.cc
zvi-code: I addressed the escaping, the code now should be compatible with RediSearch, please review

src/indexes/tag.cc
```diff
@@ -68,29 +68,71 @@ absl::StatusOr<bool> Tag::AddRecord(const InternedStringPtr& key,
...
 absl::StatusOr<absl::flat_hash_set<absl::string_view>> Tag::ParseSearchTags(
     absl::string_view data, char separator) {
   absl::flat_hash_set<absl::string_view> parsed_tags;
-  std::vector<absl::string_view> parts = absl::StrSplit(data, separator);
-  for (const auto& part : parts) {
-    auto tag = absl::StripAsciiWhitespace(part);
+  absl::Status insert_status = absl::OkStatus();
+
+  // Helper: validate and insert a single tag (handles prefix wildcards)
+  auto InsertTag = [&](absl::string_view raw) -> bool {
```
yairgott: `InsertTag` may return `absl::Status` instead and `insert_status` can be removed followed by usage of `VMSDK_RETURN_IF_ERROR` .

zvi-code: makes sense, will fix

zvi-code: done

PR #455, opened 2025-11-07
ft_info_test and workflow files fix

Aksha1812:
## Fixes 

- Unit test : fixed test commands_test/ft_info_test. (Needed simple fix in ft_info_test, got confused earlier since tests were passing in workflow files and didn't notice the error)
- Unit tests asan : There was an error when updating Metadata in posting.cc , possibly due to index schema being in a different thread then call to posting. currently moved to metadata update to text_index.cc
- Integration tests asan : there was an issue in test_fulltext.py where i am checking for text index memory fields when asan build, memory tracking does not work in asan build, made test conditional 
- Workflow files fix : error where workflow files were returning success even though test failed. Reference : https://github.com/valkey-io/valkey-search/pull/444/files#diff-d917339141f036bb9e9a987412f09fc4c6780f8c0cd5ec1d89c4652c90032127


BCathcart:
I'd start by looking at the test logs from #377. The ASAN unit tests caught an issue and I just kicked off the ASAN integration tests which didn't previously run.

allenss-amazon (commented):
Fixes to the workflows need to be in main. Please split this into two PR, one for main and one for fulltext.

src/indexes/text/text_index.cc
```diff
@@ -35,6 +35,16 @@ std::optional<std::shared_ptr<Postings>> AddKeyToPostings(
     current_schema_->GetMetadata().num_unique_terms++;
     postings = std::make_shared<Postings>();
   }
+
+  // Track metadata before inserting
```
BCathcart, line 39: I'm not sure this is the fix we want to make. The issue may be testing related - we have tests that directly test the posting object API. InsertKey and RemoveKey fetch and access the thread-local schema object that wouldn't have been set. This is another indicator that we need to be careful with (and maybe rethink) this global thread-local pattern.

integration/test_fulltext.py
```diff
@@ -750,15 +750,25 @@ def test_ft_info_text_index_fields(self):
...
             f"Unique terms should decrease after deletion: {info_after_delete['num_unique_terms']} >= {initial_unique_terms}"
         assert info_after_delete['num_total_terms'] < initial_total_terms, \
             f"Total terms should decrease after deletion: {info_after_delete['num_total_terms']} >= {initial_total_terms}"
-        assert info_after_delete['position_sz_bytes'] < initial_position_memory, \
-            f"Position memory should decrease after deletion: {info_after_delete['position_sz_bytes']} >= {initial_position_memory}"
-        assert info_after_delete['total_text_index_sz_bytes'] < initial_total_memory, \
-            f"Total memory should decrease after deletion: {info_after_delete['total_text_index_sz_bytes']} >= {initial_total_memory}"
         
+        # Skip memory size checks when running with sanitizers (SAN_BUILD)
+        # as memory tracking may not be accurate in sanitizer builds
```
BCathcart, line 761: Hmm do you know why this is?

src/indexes/text/text_index.cc
Aksha1812: yes , will change global thread-local usage of text index schema in next PR. ig only other alternative is to pass it as a parameter.

integration/test_fulltext.py
Aksha1812: No, had pointed out to @allenss-amazon earlier . he wasn't surprised. we decided to just skip for now

src/indexes/text/text_index.cc
BCathcart: Ok let's get the workflow fix in. I'll follow up with you about the metadata tracking

BCathcart:
Workflows fixed in `main` in #457

PR #458, opened 2025-11-07
Change FanoutSearchTarget to NodeInfo

boda26:
1. Change FanoutSearchTarget to NodeInfo
2. Fix fanout_test accordingly
3. Add cancelled check in Reply() to fix testtimeout_CMD and CME

### DIFF
--- a/src/commands/commands.cc
+++ b/src/commands/commands.cc
@@ -15,6 +15,7 @@
 #include "src/schema_manager.h"
 #include "src/valkey_search.h"
 #include "valkey_search_options.h"
+#include "vmsdk/src/cluster_map.h"
 #include "vmsdk/src/debug.h"
 
 namespace valkey_search {
@@ -36,6 +37,15 @@ int Reply(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
   auto *res =
       static_cast<Result *>(ValkeyModule_GetBlockedClientPrivateData(ctx));
   CHECK(res != nullptr);
+
+  // Check if operation was cancelled and partial results are disabled
+  if (!options::GetEnablePartialResults().GetValue() &&
+      res->parameters->cancellation_token->IsCancelled()) {
+    ++Metrics::GetStats().query_failed_requests_cnt;
+    return ValkeyModule_ReplyWithError(
+        ctx, "Search operation cancelled due to timeout");
+  }
+
   if (!res->neighbors.ok()) {
     ++Metrics::GetStats().query_failed_requests_cnt;
     return ValkeyModule_ReplyWithError(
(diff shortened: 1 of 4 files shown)


allenss-amazon (approved):

src/commands/commands.cc
```diff
@@ -121,7 +131,12 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
           ForceReplicasOnly.GetValue()
```
yulazariy, line 131: the mode is no longer use and needs to be deleted

src/query/fanout.cc
```diff
@@ -201,7 +202,7 @@ absl::Status PerformSearchFanoutAsync(
   bool has_local_target = false;
   for (auto &node : search_targets) {
     auto detached_ctx = vmsdk::MakeUniqueValkeyDetachedThreadSafeContext(ctx);
-    if (node.type == FanoutSearchTarget::Type::kLocal) {
+    if (node.is_local) {
```
yulazariy, line 205: the is_local should be refer to the locality of the node that runs the query but in your case it means it is in the same shard.
that makes a node to always not fanout request to other nodes in the same shard.

you set it in the code 
```
std::optional<NodeInfo> ClusterMap::ParseNodeInfo(
    ValkeyModuleCallReply* node_arr, bool is_local_shard, bool is_primary) {
...
  return NodeInfo{.node_id = node_id_str,
                  .is_primary = is_primary,
                  .is_local = is_local_shard,
                  .socket_address = addr,
                  .additional_network_metadata = additional_network_metadata,
                  .shard = nullptr};
}
```

src/commands/commands.cc
```diff
@@ -121,7 +131,12 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
               ? query::fanout::FanoutTargetMode::kReplicasOnly
               : query::fanout::FanoutTargetMode::kRandom;
-      auto search_targets = query::fanout::GetSearchTargetsForFanout(ctx, mode);
+      // refresh cluster map if needed
+      ValkeySearch::Instance().GetOrRefreshClusterMap(ctx);
+      auto search_targets =
+          ForceReplicasOnly.GetValue()
+              ? ValkeySearch::Instance().GetClusterMap()->GetReplicaTargets()
```
yulazariy, line 138: `GetReplicaTargets` are all the replicas in cluster according to 
```
const std::vector<NodeInfo>& GetReplicaTargets() const {
    return replica_targets_;
  }; 
```
and 
```
std::shared_ptr<ClusterMap> ClusterMap::CreateNewClusterMap(
    ValkeyModuleCtx* ctx) {
...
// Populate target lists
  for (auto& [shard_id, shard] : new_map->shards_) {
    if (shard.primary.has_value()) {
      new_map->primary_targets_.push_back(shard.primary.value());
      new_map->all_targets_.push_back(shard.primary.value());
    }
    for (const auto& replica : shard.replicas) {
      new_map->replica_targets_.push_back(replica);
      new_map->all_targets_.push_back(replica);
    }
  }
...
}
```

PR #460, opened 2025-11-08
Fulltext string intern

allenss-amazon:
Substantial addtional status with FT._DEBUG STRINGPOOLSTATS

### DIFF
--- a/ci/build_ubuntu.sh
+++ b/ci/build_ubuntu.sh
@@ -106,6 +106,11 @@ function save_integration_output() {
     cp ${result_dir}/libsearch.so ${INTEGRATION_OUTPUT}
     cp -r -P ${result_dir}/integration/.valkey-test-framework ${INTEGRATION_OUTPUT}
     mv ${INTEGRATION_OUTPUT}/.valkey-test-framework ${INTEGRATION_OUTPUT}/valkey-test-framework
+    # Do the stest outputs too.
+    local stest_dir=${ROOT_DIR}/testing/integration/./build-release${san_suffix}
+    echo Stest Directory output is ${stest_dir}
+    cp -r -P ${stest_dir}/output ${INTEGRATION_OUTPUT}
+    cp -r -P ${stest_dir}/tmp ${INTEGRATION_OUTPUT}
 }
 
 function save_unittest_output() {


src/utils/string_interning.cc
```diff
@@ -20,78 +19,186 @@ namespace valkey_search {
...
+  str_to_interned_.erase(it);
+  return true;
 }
 
-std::shared_ptr<InternedString> StringInternStore::Intern(
-    absl::string_view str, Allocator* allocator) {
-  return Instance().InternImpl(str, allocator);
+InternedStringPtr StringInternStore::Intern(absl::string_view str,
+                                            Allocator* allocator) {
+  auto result = Instance().InternImpl(str, allocator);
```
yairgott, line 134: nit: `return Instance().InternImpl(str, allocator);`

src/utils/string_interning.cc
```diff
@@ -20,78 +19,186 @@ namespace valkey_search {
...
+InternedStringPtr StringInternStore::InternImpl(absl::string_view str,
+                                                Allocator* allocator) {
+  IsolatedMemoryScope scope{memory_pool_};
   absl::MutexLock lock(&mutex_);
-  auto it = str_to_interned_.find(str);
+  //
+  // Construct a fake InternedStringPtr to look up in the map.
+  // Doing it carefully to avoid modifying refcounts.
+  //
+  OutOfLineInternedString fake(str.data(), str.size());
```
yairgott, line 156: Minimize operations while the lock is acquired. Lines 156-159 should be placed before the lock is acquired.

src/utils/string_interning.cc
```diff
@@ -20,78 +19,186 @@ namespace valkey_search {
...
+  void* storage;
+  InternedStringPtr* ptr_ptr = MakeShadowInternPtrPtr(&fake, storage);
+  auto it = str_to_interned_.find(*ptr_ptr);
```
yairgott, line 159: It would be more efficient to call insert:
```
auto [itr, was_inserted] = = str_to_interned_.insert(...);
 if (!was_inserted) {
    return *itr;
}
...
 ```
 
 This is applicable to my original code as well :)

File: src/utils/string_interning.h [hunk 4]
```diff
@@ -23,89 +24,162 @@
...
 };
 
-class InternedString {
+static_assert(sizeof(InternedString) == 8,
+              "InternedString has unexpected padding");
+
+//
+// A smart pointer for InternedString that manages reference counting.
+//
+class InternedStringPtr {
```
yairgott, line 100: why can't we use IntrusiveRefCount?

src/utils/string_interning.cc
allenss-amazon: left over debugging.

src/utils/string_interning.cc
allenss-amazon: ack

src/utils/string_interning.cc
allenss-amazon: Not correct. The call to "MakeShadowInternPtrPtr", uses the on-stack declaration of "fake" to build a on-stack InternStringPtr that can be used to perform the find. But that address can't be put into the map as it's temporary. If the string is found, then no allocation was done (since fake was on-stack). But if the string is NOT found, then an appropriate malloc is done and the address of that malloc gets put into the map. What's really wanted here is the Rust equivalent of the "Entry", where the Entry might contain a previously value OR no-value in which case you can dynamically construct a value for insertion. THis is optimal as it performs only a single hash computation and preserves the state of the hashtable during the decision and conditional update. This is safe in Rust because of the ownership model, the compiler can prove that no other modification of the hashmap will be done why you're processing the "entry". I haven't seen either std or abseil offer a hash container which this capability.

src/utils/string_interning.h
allenss-amazon: Because InstrusiveRefCount has a virtual destructor, which means that there's a vtbl, that would double the overhead of an InternedString.

allenss-amazon: It's probably possible to modify IntrusiveRefCount to avoid the extra overhead and then they could be combined.

allenss-amazon:
replaced by #474

PR #464, opened 2025-11-11
FT.SEARCH partition and consistency controls

boda26:
Added FT.SEARCH partition and consistency control flags, including ALLSHARDS/SOMESHARDS and CONSISTENT/INCONSISTENT

### DIFF
--- a/integration/test_partition_consistency_controls.py
+++ b/integration/test_partition_consistency_controls.py
@@ -134,8 +134,11 @@ class TestPartitionConsistencyControls(ValkeySearchClusterTestCaseDebugMode):
         # do not force invalid slot fingerprint
         self.control_set("ForceInvalidSlotFingerprint", "no")
 
-        # force invalid index fingerprint
-        self.control_set("ForceInvalidIndexFingerprint", "yes")
+        # force invalid index fingerprint on one node only
+        client = self.new_client_for_primary(0)
+        client.execute_command(
+            "ft._debug", "CONTROLLED_VARIABLE", "set", "ForceInvalidIndexFingerprint", "yes"
+        )
 
         # disable consistency check, get valid results
         hnsw_result = search(client, "hnsw", False, enable_consistency=False)
--- a/src/commands/commands.cc
+++ b/src/commands/commands.cc
@@ -145,15 +145,13 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
       // get index fingerprint and version
       if (ForceInvalidIndexFingerprint.GetValue()) {
         // test only: simulate invalid index fingerprint and version
-        parameters->index_fingerprint_version.emplace();
-        parameters->index_fingerprint_version->set_fingerprint(404);
-        parameters->index_fingerprint_version->set_version(404);
+        parameters->index_fingerprint_version.set_fingerprint(404);
+        parameters->index_fingerprint_version.set_version(404);
       } else {
         // get fingerprint/version from IndexSchema
-        parameters->index_fingerprint_version.emplace();
-        parameters->index_fingerprint_version->set_fingerprint(
+        parameters->index_fingerprint_version.set_fingerprint(
             parameters->index_schema->GetFingerprint());
-        parameters->index_fingerprint_version->set_version(
+        parameters->index_fingerprint_version.set_version(
             parameters->index_schema->GetVersion());
       }
 
--- a/src/coordinator/coordinator.proto
+++ b/src/coordinator/coordinator.proto
@@ -101,8 +101,8 @@ message SearchIndexPartitionRequest {
   repeated ReturnParameter return_parameters = 13;
   bool enable_partial_results = 14;
   bool enable_consistency = 15;
-  optional IndexFingerprintVersion index_fingerprint_version = 16;
-  optional uint64 slot_fingerprint = 17;
+  IndexFingerprintVersion index_fingerprint_version = 16;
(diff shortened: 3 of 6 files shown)


src/coordinator/server.cc
```diff
@@ -127,59 +221,48 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
+      auto consistency_status =
+          this->PerformConsistencyChecks(request, *vector_search_parameters);
+      if (!consistency_status.ok()) {
+        reactor->Finish(consistency_status);
+        RecordSearchMetrics(true, std::move(latency_sample));
+        return;
+      }
+
+      // Consistency checks passed, now enqueue the search
+      auto status = query::SearchAsync(
```
yairgott: This seems like a code duplication of lines 259 till 267. I suggest extracting this code block into a dedicated reusable function.

integration/test_cancel.py
```diff
@@ -37,14 +39,24 @@ def search(
...
-    if not timeout:
-        return client.execute_command(*search_command(index, filter))
+    print("Search command: ", search_command(index, filter, enable_partial_results, enable_consistency))
+    if expect_consistency_error:
+        try:
+            x = client.execute_command(*search_command(index, filter, enable_partial_results, enable_consistency))
+            assert False, "Expected error, but got result: " + str(x)
+        except ResponseError as e:
+            print(e)
+            assert str(e) == "Index or slot consistency check failed"
```
allenss-amazon, line 53: Use "pytest.raises(ResponseError) here.

integration/test_cancel.py
```diff
@@ -37,14 +39,24 @@ def search(
...
+        return []
+    elif not timeout:
+        return client.execute_command(*search_command(index, filter, enable_partial_results, enable_consistency))
     else:
         try:
-            x = client.execute_command(*search_command(index, filter))
+            x = client.execute_command(*search_command(index, filter, enable_partial_results, enable_consistency))
             assert False, "Expected timeout, but got result: " + str(x)
         except ResponseError as e:
```
allenss-amazon, line 61: with pytest.raises(...)

integration/test_partition_consistency_controls.py
```diff
@@ -0,0 +1,147 @@
...
+    def execute_primaries(self, command: Union[str, list[str]]) -> list[Any]:
+        return [
+            self.client_for_primary(i).execute_command(*command)
+            for i in range(len(self.replication_groups))
+        ]
+
+    def config_set(self, config: str, value: str):
+        assert self.execute_primaries(["config set", config, value]) == [True] * len(
+            self.replication_groups
+        )
```
allenss-amazon, line 23: why only on primaries?

src/coordinator/server.cc
```diff
@@ -127,62 +239,35 @@ grpc::ServerUnaryReactor* Service::SearchIndexPartition(
...
+        reactor->Finish(consistency_status);
+        RecordSearchMetrics(true, std::move(latency_sample));
+        return;
+      }
+      // Consistency checks passed, now enqueue the search
+      this->EnqueueSearchRequest(std::move(vector_search_parameters),
+                                 reader_thread_pool, detached_ctx, response,
+                                 reactor, std::move(latency_sample));
+    });
+    return reactor;
```
allenss-amazon, line 245: This adds an extra transition to the mainthread. That's  a potentially serious performance degradataion. Let's talk about how to avoid this.

allenss-amazon (changes requested):
let's discuss

integration/test_partition_consistency_controls.py
boda26: I used the exact same code you created in test_cancel.py. I think config set only work on primaries, not on replicas. (please correct me if this is wrong)

src/coordinator/server.cc
boda26: Checking index fingerprint requires calling coordinator::MetadataManager::Instance().GetGlobalMetadata(), which would crash if not on main thread.

src/coordinator/coordinator.proto
```diff
@@ -93,6 +99,10 @@ message SearchIndexPartitionRequest {
   bool no_content = 11;
   optional Predicate root_filter_predicate = 12;
   repeated ReturnParameter return_parameters = 13;
+  bool enable_partial_results = 14;
+  bool enable_consistency = 15;
+  optional IndexFingerprintVersion index_fingerprint_version = 16;
+  optional uint64 slot_fingerprint = 17;
```
allenss-amazon: See comments in FT.INFO. IMO, these shouldn't be optional.

src/commands/commands.cc
```diff
@@ -145,15 +145,13 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
       // get index fingerprint and version
       if (ForceInvalidIndexFingerprint.GetValue()) {
         // test only: simulate invalid index fingerprint and version
-        parameters->index_fingerprint_version.emplace();
-        parameters->index_fingerprint_version->set_fingerprint(404);
-        parameters->index_fingerprint_version->set_version(404);
+        parameters->index_fingerprint_version.set_fingerprint(404);
```
allenss-amazon, line 148: 404 is risky, what if the correct value was 404? It might be better to take the correct value and modify that, i.e. make it negative or add 1 to or it or something like that? (You can 404 to it if you really like that number ;-))

allenss-amazon (approved):

PR #468, opened 2025-11-13
Update the fieldmask assignment for suffix predicates + Populate identifiers for CME Text predicate deserialization

KarthikSubbarao:
Update the fieldmask assignment for suffix predicates.

Changes made:
* For infix / suffix predicates with a field specification, we check if the field is enabled using WITHSUFFIXTRIE. If it is not, we reject it.
* For infix / suffix predicates without any field specification, we look across all fields to generate the field mask, field identifier, and min stem size. If no text field exists in the index that supports suffix, the query is rejected.
* Added tests for the case where an index has two fields. One enabled with suffix, one which is not. When we search for a suffix query that matches a word (which exists in the prefix tree), it should only return the document with the field containing the word matching the suffix search which was created using WITHSUFFIXTRIE 

* Populate identifiers for CME Text predicate deserialization by using a FieldMask to obtain the identifiers

### DIFF
--- a/src/commands/filter_parser.cc
+++ b/src/commands/filter_parser.cc
@@ -690,14 +690,17 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
     }
   } else {
     // Set identifiers to include all text fields in the index schema.
-    filter_identifiers_ = index_schema_.GetAllTextIdentifiers(with_suffix);
+    auto text_identifiers = index_schema_.GetAllTextIdentifiers(with_suffix);
     // Set field mask to include all text fields in the index schema.
     field_mask = index_schema_.GetAllTextFieldMask(with_suffix);
-    if (filter_identifiers_.size() == 0 || field_mask == 0ULL) {
+    if (text_identifiers.size() == 0 || field_mask == 0ULL) {
       if (with_suffix) {
         return absl::InvalidArgumentError("No fields support suffix search");
       }
       return absl::InvalidArgumentError("Index does not have any text field");
+    }
+    for (const auto& identifier : text_identifiers) {
+      filter_identifiers_.insert(identifier);
     }
     // When no field was specified, we use the min stem across all text fields
     // in the index schema. This helps ensure the root of the text token can be
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -322,8 +322,28 @@ std::optional<uint32_t> IndexSchema::MinStemSizeAcrossTextIndexes(
 // Returns the field mask including all the text fields.
 // If `with_suffix` is true, we only include fields that have suffix tree
 // enabled.
-uint64_t IndexSchema::GetAllTextFieldMask(bool with_suffix) const {
+FieldMaskPredicate IndexSchema::GetAllTextFieldMask(bool with_suffix) const {
   return with_suffix ? suffix_text_field_mask_ : all_text_field_mask_;
+}
+
+// Helper function to return the text identifiers based on the
+// FieldMaskPredicate.
+absl::flat_hash_set<std::string> IndexSchema::GetTextIdentifiersByFieldMask(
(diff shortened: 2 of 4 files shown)


integration/test_fulltext.py
```diff
@@ -696,6 +696,14 @@ def test_suffix_search(self):
...
         result = self.client.execute_command("FT.SEARCH", "idx", "@content:*xyz")
         assert result[0] == 0  # No matches
+        # Validate that we do not get results from fields which are not enabled with the suffix tree
+        result = self.client.execute_command("FT.SEARCH", "idx", "@extracontent:*ata1")
+        assert result[0] == 0  # No matches
+        result = self.client.execute_command("FT.SEARCH", "idx", "*ata1")
+        assert result[0] == 0  # No matches
+        # Validate that the default field search only includes results from the fields enabled with suffix.
+        result = self.client.execute_command("FT.SEARCH", "idx", "*unning")
+        assert result[0] == 1  # Only doc:1 is matched for "running"
```
BCathcart, line 727: Just curious, did you check that this assert failed before your change?

nit: you also may want to assert that the result actually contains the expected doc

src/index_schema.cc
```diff
@@ -271,29 +271,41 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexSchema::GetIndex(
...
+// If `with_suffix` is true, we only include the fields that have suffix tree
+// enabled.
+std::vector<std::string> IndexSchema::GetAllTextIdentifiers(
+    bool with_suffix) const {
   std::vector<std::string> identifiers;
   for (const auto &[alias, attribute] : attributes_) {
     auto index = attribute.GetIndex();
     if (index->GetIndexerType() == indexes::IndexerType::kText) {
-      identifiers.push_back(attribute.GetIdentifier());
+      // identifiers.push_back(attribute.GetIdentifier());
```
BCathcart: Leftover comment?

src/commands/filter_parser.cc
```diff
@@ -676,16 +690,22 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
     }
   } else {
     // Set identifiers to include all text fields in the index schema.
-    auto text_identifiers = index_schema_.GetAllTextIdentifiers();
+    auto text_identifiers = index_schema_.GetAllTextIdentifiers(with_suffix);
+    // Set field mask to include all text fields in the index schema.
+    field_mask = index_schema_.GetAllTextFieldMask(with_suffix);
```
BCathcart, line 695: As a small future optimization we can store some of this information in the TextIndexSchema on creation so we don't have to repeatedly re-compute

BCathcart (approved):

integration/test_fulltext.py
KarthikSubbarao: > Just curious, did you check that this assert failed before your change?


Yes

src/index_schema.cc
```diff
@@ -267,40 +267,83 @@ absl::StatusOr<std::shared_ptr<indexes::IndexBase>> IndexSchema::GetIndex(
...
+// Returns the field mask including all the text fields.
+// If `with_suffix` is true, we only include fields that have suffix tree
+// enabled.
+FieldMaskPredicate IndexSchema::GetAllTextFieldMask(bool with_suffix) const {
+  return with_suffix ? suffix_text_field_mask_ : all_text_field_mask_;
+}
+
+// Helper function to return the text identifiers based on the
+// FieldMaskPredicate.
+absl::flat_hash_set<std::string> IndexSchema::GetTextIdentifiersByFieldMask(
```
BCathcart, line 331: One thing that's crossed my mind is if we should worry about encapsulating all schema-level data and functionality specific to text in TextIndexSchema. Not sure if index_schema.cc will get too bloated otherwise. This isn't specific to your PR, just something I want to bring up.

PR #469, opened 2025-11-13
FT.INFO partition and consistency controls

boda26:
Add ALLSHARDS/SOMESHARDS and CONSISTENT/INCONSISTENT for FT.INFO commands.
ALLSHARDS: only return result if all shards responds
SOMESHARDS: always return some result, even if some shard did not respond
CONSISTENT: check for index consistency and return result if the index is consistent
INCONSISTENT: run the command without checking index consistency

### DIFF
--- a/integration/test_ft_info_partition_consistency_controls.py
+++ b/integration/test_ft_info_partition_consistency_controls.py
@@ -9,72 +9,9 @@ from valkeytestframework.util import waiters
 from valkey import ResponseError
 from test_fanout_base import MAX_RETRIES
 
-class TestFTInfoPartitionConsistencyControls(ValkeySearchClusterTestCaseDebugMode):
-
-    def execute_primaries(self, command: Union[str, list[str]]) -> list[Any]:
-        return [
-            self.client_for_primary(i).execute_command(*command)
-            for i in range(len(self.replication_groups))
-        ]
-
-    def config_set(self, config: str, value: str):
-        assert self.execute_primaries(["config set", config, value]) == [True] * len(
-            self.replication_groups
-        )
-
-    def control_set(self, key:str, value:str):
-        assert all(
-            [node.client.execute_command(*["ft._debug", "CONTROLLED_VARIABLE", "set", key, value]) == b"OK" for node in self.nodes]
-        )
-
-    def check_info(self, name: str, value: Union[str, int]):
-        results = self.execute_primaries(["INFO", "SEARCH"])
-        failed = False
-        for ix, r in enumerate(results):
-            if r[name] != value:
-                print(
-                    name,
-                    " Expected:",
-                    value,
-                    " Received:",
-                    r[name],
-                    " on server:",
-                    ix,
-                )
-                failed = True
-        assert not failed
-
-    def _check_info_sum(self, name: str) -> int:
-        """Sum the values of a given info field across all servers"""
-        results = self.execute_primaries(["INFO", "SEARCH"])
-        return sum([int(r[name]) for r in results if name in r])
-
-    def check_info_sum(self, name: str, sum_value: int):
-        """Sum the values of a given info field across all servers"""
-        waiters.wait_for_equal(
-          lambda: self._check_info_sum(name), 
-          sum_value, 
-          timeout=5
-        )
-
-    def sum_docs(self, index: Index) -> int:
-        return sum([index.info(self.client_for_primary(i)).num_docs for i in range(len(self.replication_groups))])
-
+class TestFTInfoPartitionConsistencyControls(ClusterTestUtils, ValkeySearchClusterTestCaseDebugMode):
     def run_info_command(self, client, index_name, enable_partial_results=False, require_consistency=False, expect_error=False):
-        if expect_error:
-            try:
-                x = client.execute_command(
-                    "FT.INFO",
-                    index_name,
-                    "PRIMARY",
-                    "SOMESHARDS" if enable_partial_results else "ALLSHARDS",
-                    "CONSISTENT" if require_consistency else "INCONSISTENT"
-                )
-                assert False, "Expected error, but got result: " + str(x)
-            except ResponseError as e:
-                assert str(e) == "Unable to contact all cluster members"
-            return []
-        else:
+        try:
             return client.execute_command(
                 "FT.INFO",
                 index_name,
@@ -82,6 +19,12 @@ class TestFTInfoPartitionConsistencyControls(ValkeySearchClusterTestCaseDebugMod
                 "SOMESHARDS" if enable_partial_results else "ALLSHARDS",
                 "CONSISTENT" if require_consistency else "INCONSISTENT"
             )
+        except ResponseError as e:
+            if expect_error:
+                assert str(e) == "Unable to contact all cluster members"
+                return []
+            else:
+                raise
 
     def test_ft_info_consistency_controls(self):
         self.execute_primaries(["flushall sync"])
@@ -92,7 +35,7 @@ class TestFTInfoPartitionConsistencyControls(ValkeySearchClusterTestCaseDebugMod
         index = Index(index_name, [Vector("v", 3, type="HNSW"), Numeric("n")])
         index.create(client)
         index.load_data(client, 1000)
-        waiters.wait_for_equal(lambda: self.sum_docs(index), 1000, timeout=3)
+        assert self.sum_docs(index) == 1000
 
         # normal result without consistency check
         normal_result = self.run_info_command(client, index_name, require_consistency=False)
@@ -111,6 +54,7 @@ class TestFTInfoPartitionConsistencyControls(ValkeySearchClusterTestCaseDebugMod
         
         # force invalid invalid index fingerprint and version
         self.control_set("ForceInfoInvalidSlotFingerprint", "yes")
+        self.config_set("search.ft-info-timeout-ms", "500")
 
         # enable consistency check, get error result
         cur_result = self.run_info_command(client, index_name, require_consistency=True, expect_error=True)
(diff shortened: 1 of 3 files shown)


src/commands/ft_dropindex.cc
```diff
@@ -48,6 +65,15 @@ class DropConsistencyCheckFanoutOperation
...
     req.set_index_name(index_name_);
+
+    if (enable_consistency_) {
+      req.set_enable_consistency(true);
+      if (expected_fingerprint_version_.has_value()) {
+        *req.mutable_index_fingerprint_version() =
+            expected_fingerprint_version_.value();
+      }
+    }
+
```
allenss-amazon: I'd set these values, always. No reason to be conditional. ```req.set_enable_true(enable_consistency_)``` should work fine. How can the fingerprint be optional? If the object isn't in the metadata store, that's differnt. Buf it is, how would it be possible to have a fingerpint?

src/commands/ft_info_parser.cc
```diff
@@ -57,6 +57,38 @@ vmsdk::KeyValueParser<InfoCommand> CreateInfoParser() {
             return absl::OkStatus();
           }));
 
+  parser.AddParamParser(
+      "ALLSHARDS",
+      std::make_unique<vmsdk::ParamParser<InfoCommand>>(
+          [](InfoCommand &cmd, vmsdk::ArgsIterator &itr) -> absl::Status {
+            cmd.enable_partial_results = false;
```
allenss-amazon, line 64: Why doesn't the GENERATE_FLAG_PARSER stuff work here?

src/commands/ft_info_parser.h
```diff
@@ -24,6 +24,8 @@ struct InfoCommand {
   std::shared_ptr<IndexSchema> index_schema;
   std::string index_schema_name;
   InfoScope scope{InfoScope::kLocal};
+  bool enable_partial_results{false};
+  bool enable_consistency{true};
```
allenss-amazon: I think "require_consistency" is probably a more descriptive name...

src/coordinator/server.cc
```diff
@@ -215,24 +215,60 @@ Service::GenerateInfoResponse(
...
+    const auto& entry_map = global_metadata->type_namespace_map().at(
+        kSchemaManagerMetadataTypeName);
+
+    // Check if index exists in entry map
+    if (!entry_map.entries().contains(index_name)) {
+      response.set_exists(false);
+      response.set_index_name(index_name);
+      response.set_error(status_or_schema.status().ToString());
+      response.set_error_type(coordinator::FanoutErrorType::INDEX_NAME_ERROR);
+      return std::make_pair(grpc::Status(grpc::StatusCode::NOT_FOUND,
```
allenss-amazon: I don't think the internals of the MetadataManager should be visible here. I think the proper layering is to consult the SchemaManager and expect that module to be able to deliver the right information. That may mean adding an API call, but that's fine. It's better to split that knowledge up so that future evolution of those data structures is easier.

src/query/cluster_info_fanout_operation.cc
```diff
@@ -9,21 +9,41 @@
...
+  if (enable_consistency_) {
+    // Get expected fingerprint/version from local metadata
+    auto global_metadata =
+        coordinator::MetadataManager::Instance().GetGlobalMetadata();
+    if (global_metadata->type_namespace_map().contains(
+            kSchemaManagerMetadataTypeName)) {
+      const auto& entry_map = global_metadata->type_namespace_map().at(
+          kSchemaManagerMetadataTypeName);
+      if (entry_map.entries().contains(index_name_)) {
+        const auto& entry = entry_map.entries().at(index_name_);
```
allenss-amazon: Same here, you should be talking to the SchemaManager, not the MetadataManager.

src/query/cluster_info_fanout_operation.cc
```diff
@@ -40,6 +60,15 @@ ClusterInfoFanoutOperation::GenerateRequest(
...
   req.set_db_num(db_num_);
   req.set_index_name(index_name_);
+
+  if (enable_consistency_) {
+    req.set_enable_consistency(true);
+    if (expected_fingerprint_version_.has_value()) {
+      *req.mutable_index_fingerprint_version() =
+          expected_fingerprint_version_.value();
+    }
+  }
```
allenss-amazon, line 60: Again, no reason not to remove the outer conditional.

src/query/fanout_operation_base.h
```diff
@@ -143,6 +149,10 @@ class FanoutOperationBase {
               if (status.error_code() == grpc::StatusCode::NOT_FOUND) {
                 resp.set_error_type(
                     coordinator::FanoutErrorType::INDEX_NAME_ERROR);
+              } else if (status.error_code() ==
+                         grpc::StatusCode::FAILED_PRECONDITION) {
+                resp.set_error_type(
+                    coordinator::FanoutErrorType::INCONSISTENT_STATE_ERROR);
```
allenss-amazon: This looks like a job for a switch statement.

src/query/primary_info_fanout_operation.cc
```diff
@@ -12,19 +12,46 @@
...
+      auto global_metadata =
+          coordinator::MetadataManager::Instance().GetGlobalMetadata();
+      if (global_metadata->type_namespace_map().contains(
+              kSchemaManagerMetadataTypeName)) {
+        const auto& entry_map = global_metadata->type_namespace_map().at(
+            kSchemaManagerMetadataTypeName);
+        if (entry_map.entries().contains(index_name_)) {
+          const auto& entry = entry_map.entries().at(index_name_);
+          expected_fingerprint_version_->set_fingerprint(entry.fingerprint());
+          expected_fingerprint_version_->set_version(entry.version());
```
allenss-amazon: SchemaManager. Seems to me that either the index is NOT FOUND OR you should expect there to be a version & fingerprint. Making this optional everywhere complicates the logic needlessly.

allenss-amazon (changes requested):
Didn't dig deeply into the tests, will do that on future reviews.

src/commands/ft_info_parser.cc
boda26: I defined GENERATE_NEGATE_FLAG_PARSER in FT.SEARCH pr, will change to that after it is merged

integration/test_ft_info_partition_consistency_controls.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def config_set(self, config: str, value: str):
+        assert self.execute_primaries(["config set", config, value]) == [True] * len(
+            self.replication_groups
+        )
+
+    def control_set(self, key:str, value:str):
+        assert all(
+            [node.client.execute_command(*["ft._debug", "CONTROLLED_VARIABLE", "set", key, value]) == b"OK" for node in self.nodes]
+        )
```
allenss-amazon: These feel like candidates for sharing with other tests. Perhaps putting them into a common file (like indexes.py)

integration/test_ft_info_partition_consistency_controls.py
```diff
@@ -0,0 +1,169 @@
...
+    def test_ft_info_consistency_controls(self):
+        self.execute_primaries(["flushall sync"])
+        self.config_set("search.info-developer-visible", "yes")
+        client: Valkey = self.new_cluster_client()
+
+        index_name = "hnsw"
+        index = Index(index_name, [Vector("v", 3, type="HNSW"), Numeric("n")])
+        index.create(client)
+        index.load_data(client, 1000)
+        waiters.wait_for_equal(lambda: self.sum_docs(index), 1000, timeout=3)
```
allenss-amazon: This should never wait. I'd turn it into an assert.

integration/test_ft_info_partition_consistency_controls.py
```diff
@@ -0,0 +1,169 @@
...
+            except ResponseError as e:
+                assert str(e) == "Unable to contact all cluster members"
+            return []
+        else:
+            return client.execute_command(
+                "FT.INFO",
+                index_name,
+                "PRIMARY",
+                "SOMESHARDS" if enable_partial_results else "ALLSHARDS",
+                "CONSISTENT" if require_consistency else "INCONSISTENT"
```
allenss-amazon: I'd combine these into a single path with a try/except block, if you're no expecting error you can re-throw (bare raise statement).

allenss-amazon (changes requested):
Hmm, with some more thought. A couple of issues occur to me.

1) If the index definition fingerprint doesn't match, then the remote can't be sure that the rest of the request parameters are valid (for example, the predicates might not be valid). Hence, I don't think that validating of the index should be optional, i.e., index definition validation should always happen. The lack of this will probably cause some transient failures in the current logic. 

2. As for slot validation, I think the logic gets simpler if the responder always includes the hash of the owned slots and the requestor simply decides whether or not to include that response. 

Hence, the contents of the request/response messages end up having no optional content. The request should have the index validation (fingerprint/version) and the response should have the owned slots.

The sender decides which nodes to contact and always includes the index validation information and the response has the owned slots information. The sender can decide what to do then. All of the logic is on the sender side and receiver has nothing optional.

integration/test_ft_info_partition_consistency_controls.py
allenss-amazon: ?

src/commands/ft_info_parser.cc
```diff
@@ -56,6 +56,30 @@ vmsdk::KeyValueParser<InfoCommand> CreateInfoParser() {
...
+            return absl::OkStatus();
+          }));
+
+  parser.AddParamParser(
+      "SOMESHARDS", GENERATE_FLAG_PARSER(InfoCommand, enable_partial_results));
+
+  parser.AddParamParser("CONSISTENT",
+                        GENERATE_FLAG_PARSER(InfoCommand, require_consistency));
+
+  // TODO: GENERATE_NEGATE_FLAG_PARSER in FT.SEARCH PR
```
allenss-amazon, line 74: Please handle this also.

PR #474, opened 2025-11-14
String intern redo.

allenss-amazon:
This is a replacement for PR https://github.com/valkey-io/valkey-search/pull/460 Please see that PR for the history.

### DIFF
--- a/src/indexes/text/textinfocmd.cc
+++ b/src/indexes/text/textinfocmd.cc
@@ -93,18 +93,20 @@ absl::Status IndexSchema::TextInfoCmd(ValkeyModuleCtx* ctx,
     VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, word));
     auto wi = index_schema->GetTextIndexSchema()
                   ->GetTextIndex()
-                  ->prefix_.GetWordIterator(word);
+                  ->GetPrefix()
+                  .GetWordIterator(word);
     bool with_keys = itr.PopIfNextIgnoreCase("WITHKEYS");
     bool with_positions = itr.PopIfNextIgnoreCase("WITHPOSITIONS");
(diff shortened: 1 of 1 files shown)


integration/test_fulltext.py
```diff
@@ -105,7 +105,10 @@ def validate_fulltext_search(client: Valkey):
         assert len(result) == 1
         assert result[0] == 0  # Number of documents found
     # Perform a wild card prefix operation with multiple matches
+    print(client.execute_command("FT._DEBUG textinfo products prefix ", "grea", "withkeys"))
```
VoletiRam, line 108: Can we assert the response of `FT._DEBUG textinfo` with an expected outcome? 

I think it is good to have tests verifying all the 3 textinfo variants.

src/commands/ft_debug.cc
```diff
@@ -131,6 +133,123 @@ absl::Status ControlledCmd(ValkeyModuleCtx *ctx, vmsdk::ArgsIterator &itr) {
...
+  }
+  //
+  // Put the stats into the log
+  //
+  VMSDK_LOG(NOTICE, ctx) << "<<<< Start InternStringPool Stats >>>>>";
+  VMSDK_LOG(NOTICE, ctx) << "Inline Total: " << stats.inline_total_stats_;
+  VMSDK_LOG(NOTICE, ctx) << "OutOfLine Total: "
+                         << stats.out_of_line_total_stats_;
+  VMSDK_LOG(NOTICE, ctx) << "ByRefCount Buckets: "
+                         << stats.by_ref_stats_.size();
```
VoletiRam, line 229: Wondering the need for these log lines as we get the command response with the same information.

allenss-amazon: It is redundant, but having a record might be useful.

PR #479, opened 2025-11-17
Add support for checking ACL key prefix permissions using the new Valkey module API.

eifrah-aws:
This change introduces a new function `AclValkeyCheckPermissions` that leverages Valkey's native ACL infrastructure to verify user permissions for key prefixes. It uses the new `ValkeyModule_ACLCheckKeyPrefixPermissions` API when available, falling back to the legacy `Call(ACL,...)` approach for older Valkey versions. 

Additionally, the build script is updated to properly detect and use the configured build tool (make or ninja) for CMake integration.

Areas touched:

* ACL permission checking: Added fast-path permission validation using ValkeyModule_ACLCheckKeyPrefixPermissions
* Build system: Improved build tool detection logic in build.sh
* Managed pointers: Added UniqueValkeyModuleUser for automatic ValkeyModuleUser cleanup
* Header reorganization: Fixed include paths in commands.h

** Generated by CodeLite. **

### DIFF
--- a/src/acl.cc
+++ b/src/acl.cc
@@ -188,6 +188,15 @@ absl::string_view CallReplyStringToStringView(ValkeyModuleCallReply *r) {
   return {r_c_str, l};
 }
 
+absl::string_view KeyAccessToStringView(acl::KeyAccess ka) {
+  switch (ka) {
+    case acl::KeyAccess::kWrite:
+      return kWriteCategory;
+    case acl::KeyAccess::kRead:
+      return kReadCategory;
+  }
+}
+
 /*
 The ACL rules provided by the Valkey Server ACL GETUSER command, are a list of
 rules contain commands and key patterns, which describe if the commands can run
@@ -195,7 +204,6 @@ against the key patterns. We only need the key patterns that the corresponding
 commands matches our module commands.
 */
 std::vector<absl::string_view> GetKeysWithAllowedCommands(
-    const absl::flat_hash_set<absl::string_view> &module_allowed_cmds,
     const std::vector<acl::ValkeyAclGetUserReplyView> &acl_views) {
   std::vector<absl::string_view> keys;
   for (const auto &acl_view : acl_views) {
@@ -279,6 +287,10 @@ void GetAclViewFromCallReplyImpl(
   }
 }
 
+bool IsKeyPrefixCheckSupported() {
+  return ValkeyModule_ACLCheckKeyPrefixPermissions != nullptr;
+}
+
 }  // namespace
 
 absl::StatusOr<std::vector<acl::ValkeyAclGetUserReplyView>>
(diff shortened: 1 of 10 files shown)


src/acl.cc
```diff
@@ -292,13 +293,92 @@ GetAclViewFromCallReply(ValkeyModuleCallReply *reply) {
...
+  }
+
+  // Convert "module_allowed_cmds" into `VALKEYMODULE_CMD_KEY*` bits.
+  unsigned int flags{0};
+  for (const auto &s : module_allowed_cmds) {
+    if (absl::EqualsIgnoreCase(kReadCategory, s)) {
+      flags |= VALKEYMODULE_CMD_KEY_ACCESS;
+    } else if (absl::EqualsIgnoreCase(kWriteCategory, s)) {
+      flags |= VALKEYMODULE_CMD_KEY_INSERT;
+    }
```
allenss-amazon: This seems like overkill, won't it always be ACCESS? I'd argue that at a minimum we shouldn't be recomputing this on every every call -- it's really constant in the code.

src/acl.cc
```diff
@@ -292,13 +293,92 @@ GetAclViewFromCallReply(ValkeyModuleCallReply *reply) {
...
+      ValkeyModule_GetModuleUserFromUserName(valkey_username.get()));
+
+  // We allow the user to execute command only if the user has permissions for
+  // ALL of the prefixes of the index
+  for (const auto &prefix : module_prefixes) {
+    if (ValkeyModule_ACLCheckKeyPrefixPermissions(
+            module_user.get(), prefix.c_str(), prefix.length(), flags) !=
+        VALKEYMODULE_OK) {
+      return absl::PermissionDeniedError(
+          "The user doesn't have a permission to execute a command");
```
allenss-amazon: Wrong text. If the user didn't have permission for the command, this would already have been handled by Valkeycore. Here it's the keyspace that's being denied.

src/acl.cc
```diff
@@ -292,13 +293,92 @@ GetAclViewFromCallReply(ValkeyModuleCallReply *reply) {
...
+    }
+  }
+  return absl::OkStatus();
+}
+
 absl::Status AclPrefixCheck(
     ValkeyModuleCtx *ctx,
     const absl::flat_hash_set<absl::string_view> &module_allowed_cmds,
```
allenss-amazon: Let's fix this to be a constant.

src/acl.cc
eifrah-aws: The current code passes a `absl::flat_hash_map` (which it needs since it deals with strings). I kept that logic for backward compatibility. Let's talk about this offline.

src/acl.cc
eifrah-aws: Not my first wording choice as well, HOWEVER, this is the error message that the module returns ATM - changing this will mean breaking existing tests and maybe user code...

see [here](https://github.com/valkey-io/valkey-search/blob/main/src/acl.cc#L314) and [here](https://github.com/valkey-io/valkey-search/blob/main/src/acl.cc#L322)

src/acl.cc
eifrah-aws: The current code requires the strings (`@read`, `@write`) I will need to check that this can be done without breaking the existing flow, otherwise, I will create an overloaded version for `AclPrefixCheck` that accepts constants.

src/acl.cc
eifrah-aws: fixed

src/acl.cc
eifrah-aws: fixed

PR #480, opened 2025-11-17
Populate fingerprint and version to IndexSchema

boda26:
Populate fingerprint and version to IndexSchema in both create new index and load from RDB case

### DIFF
--- a/integration/test_fanout_base.py
+++ b/integration/test_fanout_base.py
@@ -117,7 +117,7 @@ class TestFanoutBase(ValkeySearchClusterTestCaseDebugMode):
         version = int(info["index_version"])
         assert version == 0
 
-        # drop and create the index again with different definiton
+        # drop and create the index again with different definition
         assert node0.execute_command("FT.DROPINDEX", index_name) == b"OK"
 
         assert node0.execute_command(


integration/test_fanout_base.py
```diff
@@ -91,4 +92,87 @@ def test_fanout_timeout(self):
...
+            "FT.CREATE", index_name,
+            "ON", "HASH",
+            "PREFIX", "1", "doc:",
+            "SCHEMA", "price", "NUMERIC"
+        ) == b"OK"
+
+        raw = node0.execute_command("FT.INFO", index_name, "PRIMARY")
+        info = parser._parse_key_value_list(raw)
+        assert fingerprint == info["index_fingerprint"]
+        assert version + 2 == int(info["index_version"])
```
allenss-amazon: I think assert that the new version is larger is sufficient. I'll bet there are timing-dependent cases where "2" isn't' the right "larger" value.

integration/test_fanout_base.py
```diff
@@ -91,4 +92,87 @@ def test_fanout_timeout(self):
...
+        assert node0.execute_command("SAVE")
+
+        # flush the memory and reload from rdb
+        assert node0.execute_command("DEBUG RELOAD") == b"OK"
+
+        # Verify fingerprint and version persisted
+        raw = node0.execute_command("FT.INFO", index_name, "PRIMARY")
+        info = parser._parse_key_value_list(raw)
+        assert fingerprint == info["index_fingerprint"]
+        assert version + 2 == int(info["index_version"])
```
allenss-amazon: This doesn't prove that the RDB version is correct. What you need is the "restart" fixes that I have in my save/restore redo. Hopefully we can get that merged shortly. https://github.com/valkey-io/valkey-search/pull/401 Perhaps you can look at the test_saverestore.py integration test and model this code from that. You'll need the fixes to the test environment that are in the PR.

src/query/primary_info_fanout_operation.cc
```diff
@@ -138,6 +143,19 @@ int PrimaryInfoFanoutOperation::GenerateReply(ValkeyModuleCtx* ctx,
...
   ValkeyModule_ReplyWithCString(
       ctx, std::to_string(hash_indexing_failures_).c_str());
+
+  if (vmsdk::info_field::GetShowDeveloper()) {
+    auto status_or_schema =
+        SchemaManager::Instance().GetIndexSchema(db_num_, index_name_);
+    auto schema = std::move(status_or_schema.value());
+    ValkeyModule_ReplyWithSimpleString(ctx, "index_fingerprint");
+    std::string fingerprint_str = absl::StrCat(schema->GetFingerprint());
+    ValkeyModule_ReplyWithBigNumber(ctx, fingerprint_str.c_str(),
```
allenss-amazon: why not ReplyWithLongLong ?

src/schema_manager.cc
```diff
@@ -591,6 +596,8 @@ absl::Status SchemaManager::RemoveAll() {
 absl::Status SchemaManager::LoadIndex(
     ValkeyModuleCtx *ctx, std::unique_ptr<data_model::RDBSection> section,
     SupplementalContentIter &&supplemental_iter) {
+  VMSDK_LOG(NOTICE, nullptr) << "SchemaManager::LoadIndex is called";
```
allenss-amazon: do we want this in the code?

src/query/primary_info_fanout_operation.cc
boda26: I tried ReplyWithLongLong, but got an overflow of negative large number. uint_64 type might not fit in ReplyWithLongLong

allenss-amazon: just cast it to an int64_t and no overflow ;-)

integration/test_fanout_base.py
```diff
@@ -93,6 +94,43 @@ def test_fanout_timeout(self):
...
+            "FT.CREATE", index_name,
+            "ON", "HASH",
+            "PREFIX", "1", "doc:",
+            "SCHEMA", "price", "NUMERIC"
+        ) == b"OK"
+
+        raw = node0.execute_command("FT.INFO", index_name, "PRIMARY")
+        info = parser._parse_key_value_list(raw)
+        assert fingerprint == int(info["index_fingerprint"])
+        assert version < int(info["index_version"])
```
allenss-amazon, line 134: maybe drop and recreate with a different index definition, version should higher and fingerprint should mismatch.

integration/test_fanout_base.py
```diff
@@ -178,4 +215,93 @@ def test_sample_queue_size_config(self):
...
+        "FT.CREATE", index_name,
+        "ON", "HASH",
+        "PREFIX", "1", "doc:",
+        "SCHEMA", "price", "NUMERIC"
+    ) == b"OK"
+
+    # Save RDB on all nodes
+    for rg in test.replication_groups:
+        rg.primary.client.execute_command("SAVE")
+
```
allenss-amazon: I think you should save the version at this point and use that as the final comparison.

integration/test_fanout_base.py
```diff
@@ -178,4 +215,93 @@ def test_sample_queue_size_config(self):
...
+    waiters.wait_for_true(are_all_nodes_ready, timeout=10)
+    client = test.new_client_for_primary(0)
+    client.execute_command("CONFIG SET search.info-developer-visible yes")
+
+    raw = client.execute_command("FT.INFO", index_name, "PRIMARY")
+    info = parser._parse_key_value_list(raw)
+    assert fingerprint == int(info["index_fingerprint"])
+    assert version < int(info["index_version"])
+
```
allenss-amazon: I think you test all primaries & secondaries.

src/coordinator/metadata_manager.cc
```diff
@@ -649,6 +651,16 @@ void MetadataManager::OnLoadingEnded(ValkeyModuleCtx *ctx) {
...
   }
   is_loading_ = false;
+
+  // populate fingerprint and version to IndexSchema at the end of loading RDB
+  auto global_metadata = GetGlobalMetadata();
+  if (global_metadata->type_namespace_map().contains(
+          kSchemaManagerMetadataTypeName)) {
+    const auto &entries = global_metadata->type_namespace_map()
+                              .at(kSchemaManagerMetadataTypeName)
+                              .entries();
```
allenss-amazon, line 661: Let's not require the SchemaManager to know how to parse the outer header of a metadata object. In short, have this code crack open the name, dbnum, version and fingerprint and pass these down to the schema manager as parameters.

src/schema_manager.cc
```diff
@@ -680,6 +687,20 @@ void SchemaManager::OnServerCronCallback(ValkeyModuleCtx *ctx,
       ctx, options::GetBackfillBatchSize().GetValue());
 }
 
+void SchemaManager::PopulateFingerprintVersionFromMetadata(
+    const google::protobuf::Map<std::string, coordinator::GlobalMetadataEntry>
+        &entries) {
```
allenss-amazon: see previous comment.

allenss-amazon (changes requested):
We're close.

src/coordinator/metadata_manager.cc
```diff
@@ -659,7 +659,14 @@ void MetadataManager::OnLoadingEnded(ValkeyModuleCtx *ctx) {
...
                               .entries();
-    SchemaManager::Instance().PopulateFingerprintVersionFromMetadata(entries);
+    for (const auto &[name, entry] : entries) {
+      // In cluster mode, only support DB 0 for now
+      uint32_t db_num = 0;
+      uint64_t fingerprint = entry.fingerprint();
+      uint32_t version = entry.version();
+      SchemaManager::Instance().PopulateFingerprintVersionFromMetadata(
+          db_num, name, fingerprint, version);
+    }
```
allenss-amazon, line 669: Yuck, this didn't work out as hoped for. Probably the right way to do this is to create a callback that gets registered. Let's fix that in a subsequent review.

PR #481, opened 2025-11-18
Fix cluster map bugs

boda26:
Fix cluster map bugs
1. GetRandomReplicaPerShard: return one random replica per shard for ForceReplica in search
2. Fix the bug of NodeInfo, now node.is_local is true only for the local node instead of shard


src/commands/commands.cc
```diff
@@ -135,7 +135,9 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
       ValkeySearch::Instance().GetOrRefreshClusterMap(ctx);
       auto search_targets =
           ForceReplicasOnly.GetValue()
-              ? ValkeySearch::Instance().GetClusterMap()->GetReplicaTargets()
+              ? ValkeySearch::Instance()
```
allenss-amazon: the computation of mode isn't used now. 

We should settle on the interface to the clustermap. Is there one interface with an enum parameter or are there N interfaces, one for each of the enum-cases? I can support either of these -- but not both.

yulazariy: I agree with @allenss-amazon. and would vote for a single API and enum as usually most logic (like random) is the same but the filter of the targets depends on the enum

src/query/fanout_template.h
```diff
@@ -7,209 +7,17 @@
...
-          selected_targets.push_back(create_remote_target(
-              absl::StrCat(ip, ":", coordinator::GetCoordinatorPort(port))));
-        }
-      }
-    }
-    return selected_targets;
-  }
-
   template <typename RequestT, typename ResponseT, typename TrackerT>
   void PerformRemoteRequest(
```
yulazariy: is this in use?

vmsdk/src/cluster_map.h
```diff
@@ -37,10 +37,11 @@ static constexpr highwayhash::HHKey kHashKey{
...
 // Enumeration for fanout target modes
 enum class FanoutTargetMode {
-  kRandom,        // Default: randomly select one node per shard
-  kReplicasOnly,  // Select only replicas, one per shard
-  kPrimary,       // Select all primary (master) nodes
-  kAll            // Select all nodes (both primary and replica)
+  kRandom,              // Default: randomly select one node per shard
+  kOneReplicaPerShard,  // Select only replicas, one per shard
+  kPrimary,             // Select all primary nodes
+  kReplicas,            // Select all replica nodes
```
yulazariy, line 43: not in use so not sure if needed

vmsdk/src/cluster_map.h
```diff
@@ -111,11 +106,9 @@ class ClusterMap {
   // are all the slots assigned to some shard
   bool IsConsistent() const { return is_consistent_; }
 
-  // generate a random targets vector with one node from each shard
-  std::vector<NodeInfo> GetRandomTargets() const;
-
   // get a random node from a shard
-  const NodeInfo& GetRandomNodeFromShard(const ShardInfo& shard) const;
+  const NodeInfo& GetRandomNodeFromShard(const ShardInfo& shard,
```
yulazariy: should be private function now

vmsdk/src/cluster_map.h
boda26: kReplicas is used in cluster_map unit tests

PR #498, opened 2025-11-24
Optimize radix node target with InvasivePtr

BCathcart:
Previously, the `std::optional<std::shared_ptr<Postings>>` target would take up 24 bytes, even when it was `std::nullopt` (16 bytes for shared pointer plus the optional flag). I got rid of the optional so now the Target type itself determines existence based on if it evaluates to true and false. This means we could put `std::shared_ptr<Postings>` directly in the Node and get down to 16 bytes.

I went a step further and implemented `InvasivePtr` (typically called an "intrusive pointer" online.. looks like I got the adjectives mixed up but I'm keeping it). This moves the refcount alongside the underlying object in memory, removing the need for a separate control block and only requiring 8 bytes for a single underlying pointer. I'm not sure the exact size of the shared_ptr control block, but it's definitely larger than the simple 4-8bytes of refcount I have right now in the invasive pointer and it's nice having fewer underlying memory allocations for cache locality and memory fragmentation purposes.

### DIFF
--- a/src/indexes/text/invasive_ptr.h
+++ b/src/indexes/text/invasive_ptr.h
@@ -102,19 +102,19 @@ class InvasivePtr {
     explicit RefCountWrapper(Args&&... args)
         : data_(std::forward<Args>(args)...) {}
 
-    T data_;
     std::atomic<uint32_t> refcount_ = 1;
+    T data_;
   };
 
   void Release() {
-    if (ptr_ && --ptr_->refcount_ == 0) {
+    if (ptr_ && ptr_->refcount_.fetch_sub(1, std::memory_order_acq_rel) == 1) {
       delete ptr_;
     }
   }
 
   void Acquire() {
     if (ptr_) {
-      ptr_->refcount_++;
+      ptr_->refcount_.fetch_add(1, std::memory_order_relaxed);
     }
   }
 


src/indexes/text/invasive_ptr.h
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+ private:
+  struct RefCountWrapper {
+    template <typename... Args>
+    explicit RefCountWrapper(Args&&... args)
+        : data_(std::forward<Args>(args)...) {}
+
+    T data_;
+    std::atomic<uint32_t> refcount_ = 1;
```
allenss-amazon, line 105: Reorder these two, in case "T" only needs 4 byte alignment at the start. It means that future users of T only need to worry about the first declaration, rather than the last -- which changes frequently.

src/indexes/text/invasive_ptr.h
```diff
@@ -0,0 +1,126 @@
...
+
+  void Release() {
+    if (ptr_ && --ptr_->refcount_ == 0) {
+      delete ptr_;
+    }
+  }
+
+  void Acquire() {
+    if (ptr_) {
+      ptr_->refcount_++;
```
allenss-amazon: This one should override the default memory ordering. Since you don't care about the ordering of it.

src/indexes/text/invasive_ptr.h
BCathcart: Made the change. 

> It means that future users of T only need to worry about the first declaration, rather than the last -- which changes frequently.

Can you explain this a bit more though for my understanding?

src/indexes/text/invasive_ptr.h
BCathcart: Updated it to`memory_order_relaxed` and the decrement to `memory_order_acq_rel` based on AI's recommendation, which now that I'm reading up on the release/acquire semantics I think makes sense?

allenss-amazon: that's the right combination. For increment you don't actually care what the "correct" current value is as long as it gets larger. For decrement you do care about about the "sequentially" correct value.

src/indexes/text/invasive_ptr.h
allenss-amazon: It's about any padding that's required for "T". When the ordering you provided, it's the last member of "T" that affects the alignment. Since most people tend to add new data items to a struct at the end, this means that each modification potentially affects the alignment. By reversing them, only the first element matters and a quick comment addresses that.

BCathcart: With "T" coming second, wouldn't its alignment within `RefCountWrapper` be impacted if an 8-byte member is added to "T" that makes it 8-byte aligned when it was previously 4-byte aligned? I'm also not getting why the alignment within the private `RefCountWrapper` matters.

allenss-amazon: It's about the alignment at the start of T vs the end of T. With the ordering that I proposed, you can start T with a 4-byte thing and be certain that as long as it's at the start, there is no padding. With your ordering, then you have to end with a 4-byte thing to ensure no padding. From a maintenance perspective, I think it's easier to ensure that the first 1 or 2 declarations are correct to ensure padding rather than having to rely on the last 1 or 2 declarations.

BCathcart: Ah gotcha. I think both `RefCountWrapper` and `T` would have to be packed to ensure there is no padding though.

I'm playing around with these two structs:
```
    struct T {
        uint32_t a;
        uint64_t b;
    };
    struct A {
        uint32_t x;
        struct T t;
    };
```

Without packing, `T` inherits 8-byte alignment from `b`:
```
sizeof(A): 24
&(A.x): 6fcea610
&(A.t.a): 6fcea618
&(A.t.b): 6fcea620
```

If I just pack `RefCountWrapper`, `t` is packed against `x` but `b` is still made to be 8-byte aligned to the start of `T`, so we get this awkward memory layout:
```
sizeof(A): 20
&(A.x): c2211100
&(A.t.a): c2211104
&(A.t.b): c221110c
```

Everything looks good though if we pack both `A` and `T` in this case:
```
sizeof(A): 16
&(A.x): 9d6f2ed0
&(A.t.a): 9d6f2ed4
&(A.t.b): 9d6f2ed8
```

Although other formats of T would lead to 8-byte types being 4-byte aligned if we pack everything:
e.g.
```
    struct T {
        uint32_t a;
        uint32_t a2;
        uint64_t b;
    };
 ```

I'm thinking we shouldn't pack anything for the moment given `InvasivePtr` is positioned as a general use type. So anyways I already made the switch to have `T data_` come after `refcount_` which I think is totally fine, I'm just still not sure it makes much difference.

PR #499, opened 2025-11-24
Field Alignment in proximity searches based on (1) query field identifiers (2) search terms positional state

KarthikSubbarao:
Field Alignment in proximity searches based on (1) query field identifiers (2) search terms positional state

Added tests for validating field alignment in proximity with specific/default field identifier combinations

### DIFF
--- a/src/indexes/text/proximity.cc
+++ b/src/indexes/text/proximity.cc
@@ -217,6 +217,7 @@ bool ProximityIterator::NextPosition() {
       }
       continue;
     }
+    // No violations - so this positional combination is valid.
     if (!violating_iter.has_value()) {
       // Set the current field based on field mask intersection.
       current_field_mask_ = iters_[0]->CurrentFieldMask();


src/query/predicate.cc
```diff
@@ -53,25 +53,21 @@ EvaluationResult TermPredicate::Evaluate(
...
     return EvaluationResult(false);
   }
-
   std::vector<indexes::text::Postings::KeyIterator> key_iterators;
   key_iterators.emplace_back(std::move(key_iter));
   auto iterator = std::make_unique<indexes::text::TermIterator>(
       std::move(key_iterators), field_mask, nullptr);
-
-  if (iterator->DonePositions()) {
+  if (iterator->DoneKeys() || iterator->DonePositions()) {
```
daddaman-amz, line 70: was planning to change DonePositions() to HasCurrentPosition() api  which checks if it has position.

Do we need DoneKeys() check too for term iterator?

```
 bool HasCurrentPosition() const { return current_position_.has_value(); }
```

daddaman-amz: we discussed offline to have one api which will check all 3 - key, position and field. I will add in mY PR after rebase on this

PR #501, opened 2025-11-25
OR Proximity Iterator + Update to Inorder and Slop & violation based advancement algorithm

KarthikSubbarao:
This PR includes the following:
1) OrProximity Iterator - To find the smallest key among all iterators using NextKey and smallest position among all terms who are on that key,
1) Adding a config to allow the inorder checks / behavior to match other redis based search modules. This is less strict since it uses a single position for every term or intersection. By default, we will use the check which includes inorder and overlap checks by using a positional range
2) Updated slop algorithm to match redis based search modules , and updated the violation based advancement to move the group forward to the next possible match
3) Added tests for inorder and slop - (a) both the checks and the violation based advancement logic and (b) compat tests. Also includes complex nested searches with proximity AND and OR.


integration/test_fulltext.py
```diff
@@ -886,40 +886,44 @@ def test_proximity_predicate(self):
...
         # Test 2.3: Three terms With slop 0 but no order.
         result = client.execute_command("FT.SEARCH", "idx", 'gamma beta alpha', "slop", "0")
         assert (result[0], set(result[1::2])) == (2, {b"doc:1", b"doc:3"})
 
         # Test 2.4: Three terms With slop 1 and inorder
         result = client.execute_command("FT.SEARCH", "idx", 'alpha beta gamma', "slop", "1", "inorder")
+        assert (result[0], set(result[1::2])) == (1, {b"doc:1"})
+        result = client.execute_command("FT.SEARCH", "idx", 'alpha beta gamma', "slop", "2", "inorder")
         assert (result[0], set(result[1::2])) == (2, {b"doc:1", b"doc:2"})
         # Test 2.5: Three terms With slop 3 and inorder
```
daddaman-amz, line 915: Nit: need to update test comments to fit the changes

integration/test_fulltext.py
```diff
@@ -1020,6 +1025,370 @@ def test_proximity_predicate(self):
...
+            result = client.execute_command("FT.SEARCH", "idx", "apple purple one", "DIALECT", "2", "INORDER", "SLOP", "10")
+            assert (result[0], set(result[1::2])) == (1, {b"doc:1"})
+            # Distance from apple to purple is 6 (with banana in between)
+            result = client.execute_command("FT.SEARCH", "idx", "apple (banana) purple", "DIALECT", "2", "INORDER", "SLOP", "4")
+            assert result[0] == 0
+            result = client.execute_command("FT.SEARCH", "idx", "apple (banana) purple", "DIALECT", "2", "INORDER", "SLOP", "5")
+            assert (result[0], set(result[1::2])) == (1, {b"doc:1"})
+            # Distance from apple to purple is 6
+            # 7 - 0 - 1 = 6
+            # 6 - (n - 2) = 6 - 1 = 5
```
daddaman-amz, line 1116: Is this same as below?
Distance =  End - start - 1
Gap 1 (apple → banana): 3 - 0 - 1 = 2
Gap 2 (banana → purple): 7 - 3 - 1 = 3
Total: 2 + 3 = 5

KarthikSubbarao: `"apple (banana) purple"` is different from `"apple (purple) purple"`

integration/test_fulltext.py
```diff
@@ -1020,6 +1025,370 @@ def test_proximity_predicate(self):
...
+            result = client.execute_command("FT.SEARCH", "idx", 'term4 term1 term2 term3', "SLOP", "3")
+            assert (result[0], set(result[1::2])) == (4, {b"doc:2", b"doc:3", b"doc:4", b"doc:5"})
+            result = client.execute_command("FT.SEARCH", "idx", 'term1 term3 term4 term2', "SLOP", "2")
+            assert (result[0], set(result[1::2])) == (3, {b"doc:3", b"doc:4", b"doc:5"})
+            result = client.execute_command("FT.SEARCH", "idx", 'term4 term3 term1 term2', "SLOP", "1")
+            assert (result[0], set(result[1::2])) == (2, {b"doc:4", b"doc:5"})
+            result = client.execute_command("FT.SEARCH", "idx", 'term4 term1 term3 term2', "SLOP", "0")
+            assert (result[0], set(result[1::2])) == (1, {b"doc:5"})
+
+    def test_proximity_slop_compat(self):
```
daddaman-amz, line 1084: This test covers many cases with complex nested queries. I think splitting up the test would be good for readability and maintainability. what do you think?

integration/test_fulltext.py
KarthikSubbarao: This is the only one, right?

integration/test_fulltext.py
KarthikSubbarao: No, I wanted one test that focuses on slop compatibility

src/query/predicate.cc
```diff
@@ -83,6 +84,11 @@ EvaluationResult TermPredicate::Evaluate(
   key_iterators.emplace_back(std::move(key_iter));
   auto iterator = std::make_unique<indexes::text::TermIterator>(
       std::move(key_iterators), field_mask, nullptr, require_positions);
+  VMSDK_LOG(WARNING, nullptr)
```
daddaman-amz: Question: Do we plan to keep the comments or remove before final merge?

integration/test_fulltext.py
daddaman-amz: oh misread it. 

maybe you can clarify this  # 6 - (n - 2) = 6 - 1 = 5

KarthikSubbarao: I will address the comment update an upcoming PR

integration/test_fulltext.py
KarthikSubbarao: I will address the comment update an upcoming PR

Issue #369, opened 2025-09-04
[DOCUMENTATION] The list of arguments in FT.CREATE and FT.SEARCH json is incomplete


adam-fowler:
**If your request is related to existing documentation, please provide a link.**

[FT.CREATE](https://github.com/valkey-io/valkey-search/blob/main/src/commands/ft.create.json)
[FT.SEARCH](https://github.com/valkey-io/valkey-search/blob/main/src/commands/ft.search.json)

**For errors in existing documentation**

The json files for FT.CREATE and FT.SEARCH are incomplete. They do not include a complete list of the command arguments. Is there a plan to fill these out at some point? This isn't only a documentation issue. Some clients use the json files to generate code in particular the new Swift client.

PR #504, opened 2025-11-26
Update arguments for ft.create and ft.search json (#369)

eric-musliner:
- Command arguments for ft.create and ft.search were incomplete and blocking addition of ValkeySearch to the valkey-swift client


allenss-amazon:
How can we test this?

What about FT.AGGREGATE?

PR #506, opened 2025-11-27
Optimize term_iterator usage when positional checks are not needed

KarthikSubbarao:
Optimize term_iterator usage when positional checks are not needed and reject positional queries incase of NOOFFSETS

This PR does the following:
1. If the query used requires offsets (exact phrase OR if the FT.SEARCH command includes INORDER / SLOP), but the index does not support offsets, we reject the query.
2. On the background searches, we optimize the term iterator usage to skip position iterator creations and allocations because we are limiting the background searches (`EvaluatePrefilteredKeys`) to only keys (not positions).
3. In the main thread's PredicateEvaluator, we optimize the term iterator usage to skip positional iterator creations and allocations if the AND Intersection containing the term does not have slop/inorder. In this case, simply return a boolean based EvaluationResult (purely based on the whether the key exists given the term). If a positional iterator (based on the query) is needed, the EvaluationResult as usual will include the "Filter Iterator" that is initialized with positions since we would have constructed the TermIterator accordingly.
4. Integration Tests were updated to check for index creations with NOOFFSETS and then include various FT.SEARCH operations.

### DIFF
--- a/src/query/predicate.cc
+++ b/src/query/predicate.cc
@@ -189,6 +189,7 @@ EvaluationResult InfixPredicate::Evaluate(
     const valkey_search::indexes::text::TextIndex& text_index,
     const InternedStringPtr& target_key, bool require_positions) const {
   // TODO: Implement infix evaluation
+  CHECK(false) << "Infix Search - Not implemented";
   return EvaluationResult(false);
 }
 
@@ -209,6 +210,7 @@ EvaluationResult FuzzyPredicate::Evaluate(
     const valkey_search::indexes::text::TextIndex& text_index,
     const InternedStringPtr& target_key, bool require_positions) const {
   // TODO: Implement fuzzy evaluation
+  CHECK(false) << "Fuzzy Search - Not implemented";
   return EvaluationResult(false);
 }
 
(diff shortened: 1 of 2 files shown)


src/commands/filter_parser.cc
```diff
@@ -437,18 +437,24 @@ inline std::unique_ptr<query::Predicate> MayNegatePredicate(
...
     std::unique_ptr<query::Predicate> predicate, bool& negate,
     query::LogicalOperator logical_operator) {
   if (!prev_predicate) {
     return MayNegatePredicate(std::move(predicate), negate);
   }
+  // If INORDER OR SLOP, but the index schema does not support offsets, we
+  // reject the query.
+  if ((options_.inorder || options_.slop.has_value()) &&
+      !index_schema_.HasTextOffsets()) {
+    return absl::InvalidArgumentError("Index does not support offsets");
```
daddaman-amz, line 451: Nit: we can use a comon constant variable for this message as it is used in more than one place.

src/query/predicate.cc
```diff
@@ -197,12 +202,12 @@ FuzzyPredicate::FuzzyPredicate(
       distance_(distance) {}
 
 EvaluationResult FuzzyPredicate::Evaluate(Evaluator& evaluator) const {
-  return evaluator.EvaluateText(*this);
+  return evaluator.EvaluateText(*this, false);
```
daddaman-amz, line 206: I agree with this and currently can't think of other ways at the moment. But if you think of way where we can avoid changing signature everywhere, that will be great

KarthikSubbarao: This only changes the signature of text predicates. not non text predicates.

Also, text predicates need to have an option to operate in two modes - with positional searches and without

src/commands/filter_parser.cc
KarthikSubbarao: I did a search for `return absl::InvalidArgumentError` and did not find file in the package with constant error messages defined. We can consider starting this

PR #508, opened 2025-11-27
Optimize NextKey / NextPosition in TermIterator with priority-queue for min key/position checks

KarthikSubbarao:
Please do not review yet - I am working on this PR when I get time in between some other core changes, and waiting for the core changes to first get merged into the fulltext branch. After that, I will have this branch updated on latest changes and mark it as ready for review.

In this PR, I optimize NextKey / NextPosition in TermIterator by using a priority queue for efficiently finding the min key and min position when using the key iterator and position iterator

### DIFF
--- a/src/indexes/text/term.cc
+++ b/src/indexes/text/term.cc
@@ -14,9 +14,15 @@ TermIterator::TermIterator(std::vector<Postings::KeyIterator>&& key_iterators,
                            const InternedStringSet* untracked_keys)
     : query_field_mask_(query_field_mask),
       key_iterators_(std::move(key_iterators)),
+      pos_iterators_(),
+      current_key_(),
       current_position_(std::nullopt),
       current_field_mask_(0ULL),
-      untracked_keys_(untracked_keys) {
+      untracked_keys_(untracked_keys),
+      key_heap_(),
+      pos_heap_(),
+      current_key_indices_(),
+      current_pos_indices_() {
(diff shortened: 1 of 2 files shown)


src/indexes/text/term.cc
```diff
@@ -14,9 +14,15 @@ TermIterator::TermIterator(std::vector<Postings::KeyIterator>&& key_iterators,
                            const InternedStringSet* untracked_keys)
     : query_field_mask_(query_field_mask),
       key_iterators_(std::move(key_iterators)),
+      pos_iterators_(),
+      current_key_(),
       current_position_(std::nullopt),
```
yairgott, line 19: nit: No need to initialize with `std::nullopt`, it's the default value.

src/indexes/text/term.cc
```diff
@@ -14,9 +14,15 @@ TermIterator::TermIterator(std::vector<Postings::KeyIterator>&& key_iterators,
...
       current_field_mask_(0ULL),
-      untracked_keys_(untracked_keys) {
+      untracked_keys_(untracked_keys),
+      key_heap_(),
+      pos_heap_(),
+      current_key_indices_(),
+      current_pos_indices_() {
```
yairgott, line 25: nit: no need to explicitly call the default constructor

KarthikSubbarao:
Closed in place of my more recent PR: https://github.com/valkey-io/valkey-search/pull/813/changes

PR #510, opened 2025-11-28
Fix replica crash by using legacy ACL path during replication

### LINKED ISSUE #509: [BUG]
**Describe the bug**
Replica nodes crash with segmentation fault when processing FT.DROPINDEX commands replicated from primary nodes. The crash occurs in VM_GetModuleUserFromUserName 
during ACL permission checking, causing replicas to terminate and disrupting cluster operations.

**To Reproduce**
create index on primary -> drop that index -> the replica that replays ft.DROP crashes

**Expected behavior**

**Environment**

FT.DROPINDEX should replicate cleanly to all replica nodes without crashes.

- Valkey Core Version 

**Additional context**

eliastam:
Resolves #509


src/acl.cc
```diff
@@ -399,7 +399,7 @@ absl::Status AclPrefixCheck(ValkeyModuleCtx *ctx, acl::KeyAccess access,
 
 absl::Status AclPrefixCheck(ValkeyModuleCtx *ctx, acl::KeyAccess access,
                             const std::vector<std::string> &module_prefixes) {
-  if (IsKeyPrefixCheckSupported()) {
+  if (IsKeyPrefixCheckSupported() && vmsdk::IsRealUserClient(ctx)) {
```
yairgott, line 401: This would address the issue. That said, I suggest introducing the additional following changes:
1. Line 392: avoid constructing `module_prefixes` if the client is not a real user:
```
if (!vmsdk::IsRealUserClient(ctx)) {
    return absl::OkStatus();
}
```
2. Line 331: `if (!IsKeyPrefixCheckSupported()) {` should be replaced with a `CHECK(IsKeyPrefixCheckSupported());`.

eliastam: Added the suggested changes

PR #512, opened 2025-12-03
Compatibility test against Redisearch

boda26:


boda26:
12/3: Initial draft of create text index and text data

PR #521, opened 2025-12-05
Use new command "FT.Internal_update" to replicate index create/drop

eliastam:
@allenss-amazon This is a draft PR, I will raise a revision with unit test and integration tests as well 

changes: 
1. created ft_internal_update with admin flag. this command will be used to replicate create/drop with global version and index metadata version and fingerprint. Customers will not be allowed to call it 
2. gRPC broadcast only broadcasts to primaries. 
3. replicas will only receive one information via replication stream (ft_internal_update) (fixes dual source of truth)
4. for CMD cluster no change happen we still replicate create/ drop as is. 

# Failure modes behaviour
1. One Primary is down: No change in customer experience (we will get Unable to contact all cluster members). 
Once the node join back the cluster(or performed failover), it will get the index via gRPC (then it sends it to its replicas via replication stream)
2.  One Replica is down, No change in customer experience. create and drop will return ok. 
Replica will get the index from its primary later when it joins the cluster.
3. corrupted binary in ft.internal_update will crash the instance and increase metric where we can use to track those events. 
Added a config to allow skipping that command to avoid a poison pill situation as an action that the operator could do.

# Testing 
1. integration test passes 
2. manual testing 
Note: if you want to test the code you need to get this fix as well to fix replica crash during drop index
https://github.com/valkey-io/valkey-search/pull/510

CME
1. Show cluster structure
`echo "=== CME (Cluster Mode) Test ===" && echo "" && echo "Cluster Structure:" && /tmp/valkey/src/valkey-cli -c -p 7000 CLUSTER NODES`

```
=== CME (Cluster Mode) Test ===

Cluster Structure:
f78399f13c864d6c3d3ca4f6f3c300857102d419 127.0.0.1:7000@17000 myself,master - 0 0 1 connected 0-8191
10a60c6587d8f35b2e2be47241335e40b61294b9 127.0.0.1:7002@17002 slave 4519ea74c5aab5a91d58033be7dee8eaa714669d 0 1764965420000 2 connected
4519ea74c5aab5a91d58033be7dee8eaa714669d 127.0.0.1:7001@17001 master - 0 1764965420214 2 connected 8192-16383
8b37362676cd02c47eca8b5eb6ae46467e0d5614 127.0.0.1:7003@17003 slave f78399f13c864d6c3d3ca4f6f3c300857102d419 0 1764965421220 1 connected
```

2. show cluster initial status 

`echo "Initial Index Status:" && echo "Node 7000 (primary):" && /tmp/valkey/src/valkey-cli -c -p 7000 FT._LIST && echo "Node 7001 (primary):" && /tmp/valkey/src/valkey-cli -c -p 7001 FT._LIST && echo "Node 7002 (replica):" && /tmp/valkey/src/valkey-cli -c -p 7002 FT._LIST && echo "Node 7003 (replica):" && /tmp/valkey/src/valkey-cli -c -p 7003 FT._LIST`

```
Initial Index Status:
Node 7000 (primary):

Node 7001 (primary):

Node 7002 (replica):

Node 7003 (replica):

```

3. Creating an index in primay 1 (7000)
`/tmp/valkey/src/valkey-cli -c -p 7000 FT.CREATE cluster_demo SCHEMA field VECTOR HNSW 6 TYPE FLOAT32 DIM 128 DISTANCE_METRIC L2`

OK

4. FT.List after creating: 
```
After creating index on node 7000:
Node 7000 (primary):
cluster_demo
Node 7001 (primary):
cluster_demo
Node 7002 (replica):
cluster_demo
Node 7003 (replica):
cluster_demo
```

6. drop from the other primary (7001) and list

`/tmp/valkey/src/valkey-cli -c -p 7001 FT.DROPINDEX cluster_demo && echo "After dropping index:" && echo "Node 7000 (primary):" && /tmp/valkey/src/valkey-cli -c -p 7000 FT._LIST && echo "Node 7001 (primary):" && /tmp/valkey/src/valkey-cli -c -p 7001 FT._LIST && echo "Node 7002 (replica):" && /tmp/valkey/src/valkey-cli -c -p 7002 FT._LIST && echo "Node 7003 (replica):" && /tmp/valkey/src/valkey-cli -c -p 7003 FT._LIST`

```
OK
After dropping index:
Node 7000 (primary):

Node 7001 (primary):

Node 7002 (replica):

Node 7003 (replica):y

```

7.AOF files for reference: 
note 

Node 7000 AOF:
```
*2
$6
SELECT
$1
0
*4
$18
FT.INTERNAL_UPDATE
$12
cluster_demo
$137
�������"z
8type.googleapis.com/valkey_search.data_model.IndexSchema>

cluster_demo-�?2%
fieldfield                                                                                                                                                 � (�P�
@
$13                                                                                                                                                        ���������
*4
$18
FT.INTERNAL_UPDATE
$12
cluster_demo
$2
$13                                                                                                                                                        ���������
```

Node 7001 AOF:
```
*2
$6
SELECT
$1
0
*4
$18
FT.INTERNAL_UPDATE
$12
cluster_demo
$137
�������"z
8type.googleapis.com/valkey_search.data_model.IndexSchema>

cluster_demo-�?2%
fieldfield                                                                                                                                                 � (�P�
@
$0

*4
$18
FT.INTERNAL_UPDATE
$12
cluster_demo
$2
$12                                                                                                                                                        ������֩=
```

CMD test

1. create and list 
`
echo "=== CMD (Primary-Replica) Test ===" && echo "Creating cluster_demo on primary:" && /tmp/valkey/src/valkey-cli -p 8888 FT.CREATE cluster_demo SCHEMA field VECTOR HNSW 6 TYPE FLOAT32 DIM 128 DISTANCE_METRIC L2 && echo "Listing indexes:" && echo "Primary (8888):" && /tmp/valkey/src/valkey-cli -p 8888 FT._LIST && echo "Replica (8889):" && /tmp/valkey/src/valkey-cli -p 8889 FT._LIST       
`

```

=== CMD (Primary-Replica) Test ===
Creating cluster_demo on primary:
OK
Listing indexes:
Primary (8888):
cluster_demo
Replica (8889):
cluster_demo
```

2. drop 

`
/tmp/valkey/src/valkey-cli -p 8888 FT.DROPINDEX cluster_demo && echo "Primary (8888):" && /tmp/valkey/src/valkey-cli -p 8888 FT._LIST && echo "Replica (8889):" && /tmp/valkey/src/valkey-cli -p 8889 FT._LIST`

```
OK
Primary (8888):

Replica (8889):

```

AOF file: (same bevahiour)

```
*2
$6
SELECT
$1
0
*13
$9
FT.CREATE
$10
test_clean
$6
SCHEMA
$5
field
$6
VECTOR
$4
HNSW
$1
6
$4
TYPE
$7
FLOAT32
$3
DIM
$3
128
$15
DISTANCE_METRIC
$2
L2
*2
$12
FT.DROPINDEX
$10
test_clean
```

### DIFF
--- a/src/coordinator/metadata_manager.cc
+++ b/src/coordinator/metadata_manager.cc
@@ -565,6 +565,7 @@ absl::Status MetadataManager::ReconcileMetadata(const GlobalMetadata &proposed,
               << "Failed during reconciliation callback: %s"
               << result.message().data() << " for type " << type_name << ", id "
               << id << " from " << source;
+          return result;
         }
         auto status = CallFTInternalUpdateForReconciliation(id, proposed_entry);
         if (!status.ok()) {


src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,55 @@
...
+  size_t id_len;
+  const char *id_data = ValkeyModule_StringPtrLen(argv[1], &id_len);
+  std::string id(id_data, id_len);
+  
+  // Deserialize GlobalMetadataEntry
+  size_t metadata_len;
+  const char *metadata_data = ValkeyModule_StringPtrLen(argv[2], &metadata_len);
+  coordinator::GlobalMetadataEntry metadata_entry;
+  if (!metadata_entry.ParseFromArray(metadata_data, metadata_len)) {
+    return absl::InvalidArgumentError("ERR failed to parse GlobalMetadataEntry");
```
allenss-amazon: This merits a log entry and some kind of error counter. Either we have a logic problem, data corruption or a bad actor. In any case, this seems like something that wants to trigger an alarm.

src/commands/ft_create.cc
```diff
@@ -79,7 +79,11 @@ absl::Status FTCreateCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
     }
     ValkeyModule_ReplyWithSimpleString(ctx, "OK");
   }
-  ValkeyModule_ReplicateVerbatim(ctx);
+
+  // Replicate ft.create only for cmd clusters, whereas ft.internal_update will be replicated for CME clusters
+  if (!ValkeySearch::Instance().IsCluster()) {
```
allenss-amazon: The right test for this might be !coordinator::Enabled. Not 100% sure. It certainly won't matter for us.

src/commands/ft_internal_update.cc
allenss-amazon: You're probably in a poison-pill situation. This is VERY bad.

src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,55 @@
...
+  }
+  
+  // Deserialize GlobalMetadataVersionHeader
+  size_t header_len;
+  const char *header_data = ValkeyModule_StringPtrLen(argv[3], &header_len);
+  coordinator::GlobalMetadataVersionHeader version_header;
+  if (!version_header.ParseFromArray(header_data, header_len)) {
```
allenss-amazon: Same here.

src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,55 @@
...
+    return absl::InvalidArgumentError("ERR failed to parse GlobalMetadataVersionHeader");
+  }
+  
+  auto status = coordinator::MetadataManager::Instance().ProcessInternalUpdate(
+      ctx, kSchemaManagerMetadataTypeName, id, &metadata_entry, &version_header);
+  if (!status.ok()) {
+    return status;
```
allenss-amazon: Probably same here too.

src/coordinator/metadata_manager.cc
```diff
@@ -205,6 +206,23 @@ absl::StatusOr<IndexFingerprintVersion> MetadataManager::CreateEntry(
...
+  new_entry.SerializeToString(&metadata_binary);
+  metadata.version_header().SerializeToString(&header_binary);
+  
+  ValkeyModuleCallReply *reply = ValkeyModule_Call(detached_ctx_.get(), "FT.INTERNAL_UPDATE", "!Kcbb",
+                                                   std::string(id).c_str(),
+                                                   metadata_binary.data(), metadata_binary.size(),
+                                                   header_binary.data(), header_binary.size());
+  if (reply) {
+    ValkeyModule_FreeCallReply(reply);
+  }
```
allenss-amazon: else crash?

src/coordinator/metadata_manager.cc
```diff
@@ -284,10 +316,36 @@ void MetadataManager::BroadcastMetadata(
...
+      if (flags & VALKEYMODULE_NODE_MASTER) {
+        ValkeyModule_SendClusterMessage(ctx, node_id.c_str(),
+                                       kMetadataBroadcastClusterMessageReceiverId,
+                                       payload.c_str(), payload.size());
+      } else {
+        VMSDK_LOG_EVERY_N_SEC(NOTICE, ctx, 10)
+            << "Skipping metadata broadcast to non-master node " << node_id
+            << " (flags=" << flags << ")";
+      }
+    } else {
```
allenss-amazon: Please use the new ClusterMap abstraction for determining the addresses of all of the primaries.

src/coordinator/metadata_manager.cc
```diff
@@ -722,4 +814,37 @@ void MetadataManager::RegisterForClusterMessages(ValkeyModuleCtx *ctx) {
...
+
+absl::Status MetadataManager::ProcessInternalUpdate(ValkeyModuleCtx *ctx,
+                                                    absl::string_view type_name,
+                                                    absl::string_view id,
+                                                    const coordinator::GlobalMetadataEntry *metadata_entry,
+                                                    const coordinator::GlobalMetadataVersionHeader *global_version_header) {
+  if (metadata_entry) {
+    auto callback_status = TriggerCallbacks(type_name, id, *metadata_entry);
+    if (!callback_status.ok()) {
+      return callback_status;
```
allenss-amazon: Again, this feels like a crash situation.

eliastam: yes makes sense

src/coordinator/metadata_manager.cc
eliastam: Ack.

src/commands/commands.h
```diff
@@ -40,11 +40,14 @@ constexpr absl::string_view kListCommand{"FT._LIST"};
 constexpr absl::string_view kSearchCommand{"FT.SEARCH"};
 constexpr absl::string_view kDebugCommand{"FT._DEBUG"};
 constexpr absl::string_view kAggregateCommand{"FT.AGGREGATE"};
+constexpr absl::string_view kInternalUpdateCommand{"FT.INTERNAL_UPDATE"};
```
otherscase, line 43: Shouldn't this be `FT._INTERNAL_UPDATE` given it is meant to be internal/hidden. Perhaps `FT._UPDATE_INDEX`.

src/commands/commands.h
```diff
@@ -40,11 +40,14 @@ constexpr absl::string_view kListCommand{"FT._LIST"};
...
 
 const absl::flat_hash_set<absl::string_view> kCreateCmdPermissions{
     kSearchCategory, kWriteCategory, kFastCategory};
 const absl::flat_hash_set<absl::string_view> kDropIndexCmdPermissions{
     kSearchCategory, kWriteCategory, kFastCategory};
+const absl::flat_hash_set<absl::string_view> kInternalUpdateCmdPermissions{
+    kAdminCategory};
```
otherscase: Shouldn't this also be write and fast categories?

src/coordinator/metadata_manager.cc
otherscase: nit: this block is probably a good candidate for a separate function

src/coordinator/metadata_manager.cc
```diff
@@ -473,12 +531,46 @@ absl::Status MetadataManager::ReconcileMetadata(const GlobalMetadata &proposed,
       }
 
       if (trigger_callbacks) {
-        auto result = TriggerCallbacks(type_name, id, proposed_entry);
-        if (!result.ok()) {
-          VMSDK_LOG(WARNING, detached_ctx_.get())
-              << "Failed during reconciliation callback: %s"
-              << result.message().data();
-          return result;
+        // Use FT.INTERNAL_UPDATE for metadata updates
```
otherscase: Probably a good candidate for separate function given this function is already long

src/coordinator/metadata_manager.cc
otherscase: When does this return not ok? If the version we are writing is older and rejected, does it still return ok?

src/commands/commands.h
eliastam: I am not sure if it is required to start with _
I see ft._list as public api

src/coordinator/metadata_manager.cc
eliastam: the call back is triggered from reconcile after doing comparison. 
the callback itself delete previous entry and create a new one (similar to what it got). 

so it should not fail because of older/rejected version that part of the code did not change

src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,100 @@
...
+    return absl::InvalidArgumentError(
+        "ERR wrong number of arguments for FT_INTERNAL_UPDATE");
+  }
+
+  size_t id_len;
+  const char *id_data = ValkeyModule_StringPtrLen(argv[1], &id_len);
+  std::string id(id_data, id_len);
+
+  size_t metadata_len;
+  const char *metadata_data = ValkeyModule_StringPtrLen(argv[2], &metadata_len);
```
allenss-amazon: vmsdk::ToStringView is preferred over call VM_StringPtrLen

src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,100 @@
...
+
+  return absl::InvalidArgumentError("ERR failed to parse " + data_type);
+}
+
+absl::Status FTInternalUpdateCmd(ValkeyModuleCtx *ctx,
+                                 ValkeyModuleString **argv, int argc) {
+  if (argc != kFTInternalUpdateArgCount) {
+    return absl::InvalidArgumentError(
+        "ERR wrong number of arguments for FT_INTERNAL_UPDATE");
+  }
```
allenss-amazon: In theory, if you've registered the command with 4 args, this can't happen. Maybe a CHECK is the right thing here?

src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,100 @@
...
+
+  size_t id_len;
+  const char *id_data = ValkeyModule_StringPtrLen(argv[1], &id_len);
```
allenss-amazon: vmsdk::makeUniqueValkeyString ?

src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,100 @@
...
+  std::string id(id_data, id_len);
+
+  size_t metadata_len;
+  const char *metadata_data = ValkeyModule_StringPtrLen(argv[2], &metadata_len);
+  coordinator::GlobalMetadataEntry metadata_entry;
+  if (!metadata_entry.ParseFromArray(metadata_data, metadata_len)) {
```
allenss-amazon: I'd redo this tomake HandleParseFailure simply take a status result as an input. Then you can do:

```VMSDK_RETURN_IF_ERROR(HandleParseFailure(metadata_entry.ParseFromArray, "GlobalMetadataEntry", ...));```

This follows the style elsewhere better.

File: src/commands/ft_internal_update.cc [hunk 15]
```diff
@@ -0,0 +1,100 @@
...
+    return HandleParseFailure(ctx, "GlobalMetadataEntry", metadata_len, id);
+  }
+
+  size_t header_len;
+  const char *header_data = ValkeyModule_StringPtrLen(argv[3], &header_len);
+  coordinator::GlobalMetadataVersionHeader version_header;
+  if (!version_header.ParseFromArray(header_data, header_len)) {
+    return HandleParseFailure(ctx, "GlobalMetadataVersionHeader", header_len,
```
allenss-amazon: same here.

src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,100 @@
...
+    if (options::GetSkipCorruptedInternalUpdateEntries().GetValue()) {
+      VMSDK_LOG(WARNING, ctx)
+          << "SKIPPING corrupted AOF entry due to configuration";
+      Metrics::GetStats().ft_internal_update_skipped_entries_cnt++;
+      ValkeyModule_ReplyWithSimpleString(ctx, "OK");
+      return absl::OkStatus();
+    } else {
+      CHECK(false)
+          << "Protobuf parse failure during AOF loading - cannot continue";
+    }
```
allenss-amazon: Seems like this should be another invokation. Wouldn't HandleParseFailure work just as well here?

src/coordinator/metadata_manager.cc
```diff
@@ -246,6 +249,26 @@ absl::StatusOr<IndexFingerprintVersion> MetadataManager::CreateEntry(
...
+      detached_ctx_.get(), "FT.INTERNAL_UPDATE", "!Kcbb",
+      std::string(encoded_id).c_str(), metadata_binary.data(),
+      metadata_binary.size(), header_binary.data(), header_binary.size());
+  if (reply != nullptr &&
+      ValkeyModule_CallReplyType(reply) != VALKEYMODULE_REPLY_ERROR) {
+    ValkeyModule_FreeCallReply(reply);
+  } else {
+    CHECK(false) << "FT.INTERNAL_UPDATE failed during CreateEntry for index "
+                 << encoded_id;
+  }
```
allenss-amazon: This looks identical to the code in deleteentry, can't they be handled by a common function?

src/coordinator/metadata_manager.cc
```diff
@@ -328,10 +370,17 @@ void MetadataManager::BroadcastMetadata(
...
   version_header.SerializeToString(&payload);
-  // Nullptr for target means broadcast to all.
-  ValkeyModule_SendClusterMessage(ctx, /* target= */ nullptr,
-                                  kMetadataBroadcastClusterMessageReceiverId,
-                                  payload.c_str(), payload.size());
+
+  auto primary_nodes = GetPrimaryNodes(ctx);
+  if (primary_nodes.empty()) {
+    return;
+  }
```
allenss-amazon: This is unnecessary and probably a pessimization.

src/coordinator/metadata_manager.cc
```diff
@@ -328,10 +370,17 @@ void MetadataManager::BroadcastMetadata(
...
+
+  for (const auto &node : primary_nodes) {
+    ValkeyModule_SendClusterMessage(ctx, node.node_id.c_str(),
```
allenss-amazon: Ignoring the return value doesn't seem like the best choice.

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +927,120 @@ std::string ObjName::Encode() const {
...
+    VMSDK_LOG(WARNING, ctx)
+        << "ClusterMap not available, skipping metadata broadcast";
+    return {};
+  }
+
+  try {
+    return cluster_map->GetTargets(
+        vmsdk::cluster_map::FanoutTargetMode::kPrimary);
+  } catch (const std::exception &e) {
+    VMSDK_LOG(WARNING, ctx) << "Failed to get primary nodes: " << e.what();
```
allenss-amazon: What problem is this code solving? Why isn't this inside of the cluster-map maintenance code?

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +927,120 @@ std::string ObjName::Encode() const {
...
+    auto obj_name = ObjName::Decode(id);
+    auto result = TriggerCallbacks(type_name, obj_name, *metadata_entry);
+    if (!result.ok()) {
+      VMSDK_LOG(WARNING, ctx)
+          << "Failed during ProcessInternalUpdate callback for type "
+          << type_name << ", id " << id << " from " << "ProcessInternalUpdate";
+      Metrics::GetStats().process_internal_update_callback_failures_cnt++;
+      CHECK(false) << "ProcessInternalUpdate callback failed for type "
+                   << type_name << ", id " << id << ": " << result.message();
+    }
```
allenss-amazon: else? When would metadata_entry ever be null?

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +927,120 @@ std::string ObjName::Encode() const {
...
+  (*mutable_entries)[id] = *metadata_entry;
+
+  // Update global version header
+  const auto new_version = global_version_header->top_level_version();
+  const auto new_fingerprint =
+      ComputeTopLevelFingerprint(result.type_namespace_map());
+
+  result.mutable_version_header()->set_top_level_version(new_version);
+  result.mutable_version_header()->set_top_level_fingerprint(new_fingerprint);
+
```
allenss-amazon, line 940: Would a better name here be CreateEntryOnReplica ?

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +927,120 @@ std::string ObjName::Encode() const {
...
+    version_header.SerializeToString(&header_binary);
+
+    ValkeyModuleCallReply *reply = ValkeyModule_Call(
+        ctx, "FT.INTERNAL_UPDATE", "!Kcbb", id.c_str(), metadata_binary.data(),
+        metadata_binary.size(), header_binary.data(), header_binary.size());
+
+    if (reply == nullptr ||
+        ValkeyModule_CallReplyType(reply) == VALKEYMODULE_REPLY_ERROR) {
+      if (reply) ValkeyModule_FreeCallReply(reply);
+      promise.set_value(absl::InternalError("FT.INTERNAL_UPDATE failed"));
```
allenss-amazon: This needs a log message and should probably crash too.

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +927,120 @@ std::string ObjName::Encode() const {
...
+    cluster_map = future.get();
+  }
+
+  if (!cluster_map) {
+    VMSDK_LOG(WARNING, ctx)
+        << "ClusterMap not available, skipping metadata broadcast";
+    return {};
+  }
+
+  try {
```
allenss-amazon: don't use exceptions for failures, use error codes.

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +927,120 @@ std::string ObjName::Encode() const {
...
+
+    coordinator::GlobalMetadataVersionHeader version_header;
+    version_header.set_top_level_version(
+        metadata_.Get().version_header().top_level_version());
+    version_header.set_top_level_fingerprint(
+        metadata_.Get().version_header().top_level_fingerprint());
+
+    std::string metadata_binary, header_binary;
+    proposed_entry.SerializeToString(&metadata_binary);
+    version_header.SerializeToString(&header_binary);
```
allenss-amazon: This isn't the right way to do this. The right way is to create a threadsafe context for the ValkeyModule_Call.

src/coordinator/metadata_manager.cc
```diff
@@ -246,6 +248,13 @@ absl::StatusOr<IndexFingerprintVersion> MetadataManager::CreateEntry(
   VMSDK_ASSIGN_OR_RETURN(auto top_level_min_version, ComputeMinVersion());
   metadata.mutable_version_header()->set_top_level_min_version(
       top_level_min_version);
+
+  metadata_ = metadata;
```
allenss-amazon: isn't this redundant with line 213?

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +901,128 @@ std::string ObjName::Encode() const {
   return absl::StrCat("{", db_num_, "}", name_);
 }
 
+std::vector<vmsdk::cluster_map::NodeInfo> MetadataManager::GetPrimaryNodes(
+    ValkeyModuleCtx *ctx) {
```
allenss-amazon: ctx isn't used, right? I think it should be removed from the signature. Also, I'm still unclear on why you need to be doing a refresh of the cluster map here. If this is to work around an initialization problem, let's solve that in initialization.

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +901,128 @@ std::string ObjName::Encode() const {
...
+
+  return cluster_map->GetTargets(
+      vmsdk::cluster_map::FanoutTargetMode::kPrimary);
+}
+
+absl::Status MetadataManager::CreateEntryOnReplica(
+    ValkeyModuleCtx *ctx, absl::string_view type_name, absl::string_view id,
+    const coordinator::GlobalMetadataEntry *metadata_entry,
+    const coordinator::GlobalMetadataVersionHeader *global_version_header) {
+  auto obj_name = ObjName::Decode(id);
```
allenss-amazon, line 919: Let's put a CHECK for replica status.

src/coordinator/metadata_manager.cc
```diff
@@ -246,6 +248,13 @@ absl::StatusOr<IndexFingerprintVersion> MetadataManager::CreateEntry(
   VMSDK_ASSIGN_OR_RETURN(auto top_level_min_version, ComputeMinVersion());
   metadata.mutable_version_header()->set_top_level_min_version(
       top_level_min_version);
+
+  metadata_ = metadata;
+
+  // Call FT.INTERNAL_UPDATE for coordinator to ensure unified AOF replication
+  CallFTInternalUpdate(new_entry, metadata.version_header(), encoded_id,
+                       "CreateEntry");
```
murphyjacob4: nit, maybe we can also rename the function? I know right now it is just create, but if we later support alter, it would make sense to just support upsert.

```suggestion
                       "UpsertEntry");
```

src/coordinator/metadata_manager.cc
```diff
@@ -328,10 +343,18 @@ void MetadataManager::BroadcastMetadata(
   }
   std::string payload;
   version_header.SerializeToString(&payload);
-  // Nullptr for target means broadcast to all.
-  ValkeyModule_SendClusterMessage(ctx, /* target= */ nullptr,
-                                  kMetadataBroadcastClusterMessageReceiverId,
-                                  payload.c_str(), payload.size());
+
+  auto primary_nodes = GetPrimaryNodes(ctx);
```
murphyjacob4: If the local node's view is stale, I guess we could send it to a primary that is now demoted. I think we need to double check our primaryship when we receive a metadata message as well.

src/coordinator/metadata_manager.cc
```diff
@@ -539,14 +562,9 @@ absl::Status MetadataManager::ReconcileMetadata(const GlobalMetadata &proposed,
       }
```
murphyjacob4, line 558: Should we check we are a primary when we get broadcasted metadata? As a replica, we should never even request the metadata on a conflict since we are going to wait for our primary

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +901,128 @@ std::string ObjName::Encode() const {
...
+  ValkeyModuleCallReply *reply;
+  ValkeyModuleCtx *safe_context = detached_ctx_.get();
+  vmsdk::UniqueValkeyDetachedThreadSafeContext thread_safe_ctx;
+
+  if (!vmsdk::IsMainThread()) {
+    thread_safe_ctx =
+        vmsdk::MakeUniqueValkeyDetachedThreadSafeContext(detached_ctx_.get());
+    ValkeyModule_ThreadSafeContextLock(thread_safe_ctx.get());
+    safe_context = thread_safe_ctx.get();
+  }
```
murphyjacob4: IIRC we made it a constraint that metadata would only get processed by the main thread for simplicity. My mental model may be slightly out of date though. Is that no longer true?

src/coordinator/metadata_manager.cc
murphyjacob4: Is it just called on replica? It looks like we call it even on the primary since all metadata updates go through FT._internal_update which calls here. Or did I miss something?

murphyjacob4 (commented):
Overall makes a lot of sense. Converging these all into an internal command and then serializing that to the replication stream solves most of the problems. Mostly want to double check that we are properly guarding against BroadcastMetadata on replicas

We may want to start out with broadcasting to all nodes, and only accept the broadcast if we are a primary. This way we can guard against propagation delays.

It is okay to accept a metadata broadcast iff you are a primary locally. The remote view of whether that node is a primary won't matter (it could also be out of sync). If we reconcile metadata locally while being demoted, this is okay, but it may need to full sync to our new primary (our offset will be ahead of the new primary, which we can't recover from)

src/coordinator/metadata_manager.cc
eliastam: Right resolved it

src/coordinator/metadata_manager.cc
eliastam: done

src/commands/ft_internal_update.cc
```diff
@@ -0,0 +1,96 @@
...
+#include "src/schema_manager.h"
+#include "src/valkey_search_options.h"
+
+namespace valkey_search {
+
+constexpr int kFTInternalUpdateArgCount = 4;
+
+// Helper function to handle parse failures with poison pill recovery
+absl::Status HandleInternalUpdateFailure(
+    ValkeyModuleCtx *ctx, bool success, const std::string &operation_type,
```
murphyjacob4: Is success and status redundant? It seems like we set success == status.ok. Maybe we just use status?

src/coordinator/metadata_manager.cc
murphyjacob4: Nevermind, I reread it

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +907,100 @@ std::string ObjName::Encode() const {
...
+      << "CreateEntryOnReplica should only be called on replica nodes or "
+         "during loading";
+
+  auto obj_name = ObjName::Decode(id);
+  auto callback_result = TriggerCallbacks(type_name, obj_name, *metadata_entry);
+  if (!callback_result.ok()) {
+    VMSDK_LOG(WARNING, ctx)
+        << "Failed during CreateEntryOnReplica callback for type " << type_name
+        << ", id " << id << " from " << "CreateEntryOnReplica";
+    Metrics::GetStats().process_internal_update_callback_failures_cnt++;
```
murphyjacob4: Does incrementing the stat/logging really matter if we just crash next line?

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +907,100 @@ std::string ObjName::Encode() const {
...
+    const coordinator::GlobalMetadataEntry *metadata_entry,
+    const coordinator::GlobalMetadataVersionHeader *global_version_header) {
+  // Verify this is only called on replica nodes or during loading
+  CHECK((ValkeyModule_GetContextFlags(ctx) & VALKEYMODULE_CTX_FLAGS_SLAVE) ||
+        (ValkeyModule_GetContextFlags(ctx) & VALKEYMODULE_CTX_FLAGS_LOADING))
+      << "CreateEntryOnReplica should only be called on replica nodes or "
+         "during loading";
+
+  auto obj_name = ObjName::Decode(id);
+  auto callback_result = TriggerCallbacks(type_name, obj_name, *metadata_entry);
```
murphyjacob4, line 920: Should we just return this status if it isn't OK? The only other exit path is returning status OK. We can handle the error in the calling function

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +907,100 @@ std::string ObjName::Encode() const {
...
+
+void MetadataManager::CallFTInternalUpdate(
+    const coordinator::GlobalMetadataEntry &entry,
+    const coordinator::GlobalMetadataVersionHeader &header,
+    absl::string_view encoded_id, absl::string_view operation_name) {
+  std::string metadata_binary, header_binary;
+  entry.SerializeToString(&metadata_binary);
+  header.SerializeToString(&header_binary);
+
+  ValkeyModuleCallReply *reply = ValkeyModule_Call(
```
murphyjacob4: So this is only executed on primaries, right? Via FT.CREATE and FT.DROP. And FT._INTERNAL_UPDATE just replicates on primaries with no effect, right?

Should we replace this call with a simple call to VM_Replicate?

src/coordinator/metadata_manager.cc
```diff
@@ -883,4 +907,100 @@ std::string ObjName::Encode() const {
...
+
+  result.mutable_version_header()->set_top_level_version(new_version);
+  result.mutable_version_header()->set_top_level_fingerprint(new_fingerprint);
+
+  metadata_ = result;
+
+  return absl::OkStatus();
+}
+
+absl::Status MetadataManager::CallFTInternalUpdateForReconciliation(
```
murphyjacob4, line 946: I feel like this code is very similar to the code below. Can we unify to one function? It seems like reconciliation is basically identical to `CallFTInternalUpdate(proposed_entry, metadata_.Get().version_header())`. Am I missing something?

src/commands/ft_internal_update.cc
eliastam: correct, removed the redundant part

src/coordinator/metadata_manager.cc
eliastam: replaced it with ReplicateFTInternalUpdate

src/coordinator/metadata_manager.cc
eliastam: done

src/coordinator/metadata_manager.cc
murphyjacob4: Sorry if I'm dense - but why do we need to VM_Call instead of VM_Replicate the FT._INTERNAL_UPDATE in the case of reconciliation?

src/coordinator/metadata_manager.cc
```diff
@@ -546,7 +565,10 @@ absl::Status MetadataManager::ReconcileMetadata(const GlobalMetadata &proposed,
               << "Failed during reconciliation callback: %s"
               << result.message().data() << " for type " << type_name << ", id "
               << id << " from " << source;
-          return result;
```
murphyjacob4, line 549: I think our error handling could use some work. But for now - is there a reason we should ignore this error?

src/coordinator/metadata_manager.cc
eliastam: FT._INTERNAL_UPDATE is doing 2 things updaing metadata and then replicating itself, it would make more sense to keep it this way that taking updating metadata outside the functionality for primaries and accepting the functionality for replica

src/coordinator/metadata_manager.cc
eliastam: Added it back, it was removed by mistake

src/coordinator/metadata_manager.cc
eliastam: I reviewed the code `ValkeyModule_Replicate` requires an active client command context which is not the case here making it required to use ValkeyModule_Call with the ! flag is the only API supported for triggering replication from a background thread.

murphyjacob4: Ahhhh, right. Thanks!

src/coordinator/metadata_manager.cc
```diff
@@ -548,6 +567,10 @@ absl::Status MetadataManager::ReconcileMetadata(const GlobalMetadata &proposed,
               << id << " from " << source;
           return result;
         }
+        auto status = CallFTInternalUpdateForReconciliation(id, proposed_entry);
```
murphyjacob4, line 570: nit: Can we add a comment?
```suggestion
        // We cannot call VM_Replicate directly since reconciliation occurs
        // outside of a command execution. Instead, we call FT._INTERNAL_UPDATE
        // locally, which should have no effect, but will replicate verbatim to
        // our replicas and AOF.
        auto status = CallFTInternalUpdateForReconciliation(id, proposed_entry);
```

Issue #532, opened 2025-12-09
[FEATURE] Limit the neighbor deque size on background searches


KarthikSubbarao:
**The problem/use-case that the feature addresses**

What valkey-search does today is it collects X (can be very large) number of keys on the background search, and on the main thread, it limits this accordingly based on `LIMIT` / the default setting. If there has been any mutations, the count of neighbors (displayed in the FT.SEARCH response's first array element) will reflect it correctly since we don't include keys filtered out in the count.

However, the concerning part is that we allow the background search to collect keys without any limit which can be problematic for the performance and memory usage (potential OOMing)

An idea for addressing this is to compute the count (number of matched keys) on the background thread, but only collect the neighbor deque with `(1.5 * LIMIT)` number of keys IF sorting/scoring is not needed.

We can leverage a result wrapper for the neighbor deque that can be optimized as mentioned above when sorting is not needed. If it is needed (an FT.SEARCH that needs scoring in a CME case), then the entire neighbor deque can be preserved

Open question: Even if sorting / scoring based ordering is required in CME, should we still enforce a limit on the neighbor size (maybe a higher limit)? This will just help add a cap on how large the neighbors deque can get. Each FT.SEARCH/FT.AGGREATE inflight has its own neighbor deque.

If there are other ideas for how to reduce the neighbor deque size, it would be great to discuss them here

PR #537, opened 2025-12-11
Implement background search neighbor deque limiting optimization 

KarthikSubbarao:
Implement background search neighbor deque limiting optimization
- Add smart limiting for both vector and non-vector queries based on the query type (e.g. Search / Aggregate, SORTBY requirements, CME, etc)
- Added a `search.search-result-buffer-multiplier` to control the buffer multiplier described above:
- Implement total_count tracking to GRPC
- Add test coverage for non vector searches - with limit, offset, nocontent, cmd, cme variations

Implements Issue: https://github.com/valkey-io/valkey-search/issues/532

Here is the flow in CME with the PR:
1. FT.SEARCH is used on the original node with LIMIT <offset> <count>. Let's take as example of `LIMIT 100 50`
2. We fanout the request to other shards with `offset` as `0` and `count` as `offset + count`. LIMIT 0 150
3. In each individual node, in their background search of the fanned out request, we **conditionally** apply the trimming on the end using `COUNT * 1.5` from above. This is 150 (due to step 2) * 1.5 = 225. This trimming is done with the construction of a `SearchResult` which calls `TrimResults` **conditionally**. `SearchResult` is already aware of the command type and **skips** trimming when it detects that `RequiresCompleteResults` is true. This is a function implemented by FT.SEARCH and FT.AGGREGATE interfaces.
4. In the coordinator node, when processing all the replies in aggregation, we **conditionally** apply the original offset (100) and an buffer (1.5x) multiplied on the count (50). If we had 3 shards, we have 675 neighbors in total. So, we start from 100 and then apply the 1.5 on the original `count`. `1.5 * 50 = 75`. So, we take the first 75 neighbors starting from 100. Again, this trimming is done with the construction of a `SearchResult` which calls `TrimResults` **conditionally**. `SearchResult` is already aware of the command type and **skips** trimming when it detects that `RequiresCompleteResults` is true. This is a function implemented by FT.SEARCH and FT.AGGREGATE interfaces.
5. In the main thread of the original node, we then call `ProcessNonVectorNeighborsForReply` and `SerializeNonVectorNeighbors`.
6. `ProcessNonVectorNeighborsForReply` can drop any neighbor where `GetContent/VerifyFilter` fails. This is why the `1.5` multiplier buffer helps.
7.  `SerializeNonVectorNeighbors` applies the original range requested in the FT.SEARCH command. Only IF we already handled the offset in the coordinator, we skip applying the offset and use a starting position of 0. The original `<count>` value of 50 is applied here.  This gives us 50 results. We are able to detect whether an offset was applied because we have this information stored on the `SearchResult`. This will be false in case of an Aggregate command which had sorting

Here is the flow in CMD (standalone) with the PR:
1. FT.SEARCH is used on the original node with LIMIT <offset> <count>. Let's take as example of `LIMIT 100 50`
2. In the background search, we apply trimming in the front using the `offset` (to start at the 100th neighbor). Then, we apply the trimming on the end using `COUNT * 1.5`. This is 50 * 1.5 = 75. Thus, we get 75 neighbors starting from the 100th . This trimming is done with the construction of a `SearchResult` which calls `TrimResults` **conditionally**. `SearchResult` is already aware of the command type and **skips** trimming when it detects that `RequiresCompleteResults` is true. This is a function implemented by FT.SEARCH and FT.AGGREGATE interfaces.
3. In the main thread of the original node, we then call `ProcessNonVectorNeighborsForReply` and `SerializeNonVectorNeighbors`.
4. `ProcessNonVectorNeighborsForReply` can drop any neighbor where `GetContent/VerifyFilter` fails. This is why the `1.5` multiplier buffer helps.
5.  `SerializeNonVectorNeighbors` applies the original range requested in the FT.SEARCH command. Only IF we already handled the offset in the background search, we skip applying the offset and use a starting position of 0 in the neighbor deque. The original `<count>` value of 50 is applied here.  This gives us 50 results. We are able to detect whether an offset was applied because we have this information stored on the `SearchResult`. This will be false in case of an Aggregate command which had sorting

### DIFF
--- a/src/query/search.cc
+++ b/src/query/search.cc
@@ -464,11 +464,10 @@ void SearchResult::TrimResults(std::deque<indexes::Neighbor> &neighbors,
       range.end_index * options::GetSearchResultBufferMultiplier());
   // In standalone mode, we can optimize by trimming from front first.
   // Note: We cannot trim from the front in a Cluster Mode setting because
-  // each shard X results and we need to trim the OFFSET on the aggregated
-  // results. Thus, we can only trim from the end in searches for individual
-  // nodes.
-  // In cluster mode, the offset based trimming is applied after merging all
-  // results from shards at the coordinator level.
+  // each shard produces X results and we need to trim the OFFSET on the
+  // aggregated results. Thus, we can only trim from the end in searches for
+  // individual nodes. In cluster mode, the offset based trimming is applied
+  // after merging all results from shards at the coordinator level.
   if (!ValkeySearch::Instance().IsCluster()) {
     this->is_offsetted = true;
     // Trim from front (apply offset)
@@ -505,7 +504,8 @@ SerializationRange SearchResult::GetSerializationRange(
     if (parameters.IsVectorQuery()) {
       CHECK_GT(parameters.k, parameters.limit.first_index);
     }
-    start_index = std::min(neighbors.size(), parameters.limit.first_index);
+    start_index = std::min(neighbors.size(),
+                           static_cast<size_t>(parameters.limit.first_index));
   }
   // Determine end_index logic
   size_t limit_count = static_cast<size_t>(parameters.limit.number);


src/commands/ft_aggregate.cc
```diff
@@ -304,13 +304,22 @@ absl::Status SendReplyInner(ValkeyModuleCtx *ctx,
   return absl::OkStatus();
 }
 
+bool AggregateParameters::HasSortBy() const {
+  for (const auto &stage : stages_) {
+    if (dynamic_cast<const SortBy *>(stage.get()) != nullptr) {
```
allenss-amazon, line 323: probably not the right code. But ok for now. maybe make a TODO?

src/commands/ft_search.cc
```diff
@@ -35,49 +35,47 @@ namespace valkey_search {
...
+    const query::SearchParameters &parameters) {
+  const auto &neighbors = search_result.neighbors;
+  const size_t start_index = search_result.is_offsetted
+                                 ? 0
+                                 : query::CalcStartIndex(neighbors, parameters);
+  const size_t end_index =
+      std::min(start_index + query::CalcEndIndex(neighbors, parameters),
+               neighbors.size());
+  return {start_index, end_index};
+}
```
allenss-amazon: 1/ Maybe rename as PredictSerializationRange or GetPrelimSerializationRange ?
2/ I'd make this a virtual function of SearchParameters. The default implementation returns 0, Inf which supplies the answer from Rpc. For SearchCommand and AggregateCommand, they override the default as they see fit.

src/query/fanout.cc
```diff
@@ -51,8 +51,14 @@ CONTROLLED_BOOLEAN(ForceInvalidSlotFingerprint, false);
...
   bool operator()(indexes::Neighbor &a, indexes::Neighbor &b) const {
+    // Primary sort: by distance
     // We use a max heap, to pop off the furthest vector during aggregation.
-    return a.distance < b.distance;
+    if (a.distance != b.distance) {
+      return a.distance < b.distance;
+    }
+    // Secondary sort: by key ptr for stable ordering when distances are equal.
+    // e.g. primarily used in non vector queries without scores.
+    return a.external_id.get() > b.external_id.get();
```
allenss-amazon: I think here we need to actually compare the strings. That's because for non-local keys, there's no guarantee that the InternedStringPtr will actually be the same the next time it's executed., so external_id.get()->Str() here.

src/query/fanout.cc
```diff
@@ -132,10 +140,15 @@ struct SearchPartitionResultsTracker {
     }
   }
 
+  void AddTotalCount(size_t count) {
+    absl::MutexLock lock(&mutex);
+    accumulated_total_count += count;
```
allenss-amazon: atomic?

src/commands/ft_search.cc
KarthikSubbarao: > 1/ Maybe rename as PredictSerializationRange or GetPrelimSerializationRange ?

The serialization only happens on the main thread of the node where the original FT.SEARCH command was used. This is the final place where decide and apply a range (limit/offset) on the neighbor deque before forming the reply.

Because this is the final place, I am not sure if using the word `Prelim` is correct here as it indicates there is a subsequent range to be applied. `GetSerializationRange` made sense to me since it is used right before we add the reply.

> 2/ I'd make this a virtual function of SearchParameters. The default implementation returns 0, Inf which supplies the answer from Rpc. For SearchCommand and AggregateCommand, they override the default as they see fit.

We use the starting position as `0` only in case we have applied the offset based trimming. This can happen in the background search layer itself in case of CMD. In case of CME, this is applied in the coordinator.

There is no case where we use `inf` currently. The LIMIT is applied in the background search and also in the coordinator.


I am actually thinking that this function should be part of the `SearchResult` structure.

Do we expect it to be different for FT.Search and FT.AGGREGATE?

src/query/fanout.cc
KarthikSubbarao: Thanks, will use that

src/commands/ft_aggregate.cc
KarthikSubbarao: Sure, I will add a TODO

src/query/fanout.cc
KarthikSubbarao: Yes, I can see this being an issue if other shards had nodes which restarted or if failovers take place. I will use a string value compare here

src/commands/ft_search.cc
allenss-amazon: I misunderstood that this was for the foreground thread and not the background thread. So the renaming doesn't make sense.

Aggregation will definitely have cases where it want's 100% of the queries with no trimming. That's why I suggest that the computation of the limits be a virtual function so that Search and Aggregation will just apply logic that's right them.

allenss-amazon: I think that 100% == inf.

KarthikSubbarao: Here is the flow in CME currently:
1. FT.SEARCH is used on the original node with LIMIT <offset> <count>. Let's take as example of `LIMIT 100 50`
2. We fanout the request to other shards with `offset` as `0` and `count` as `offset + count`. LIMIT 0 150
3. In each individual node, in their background search of the fanned out request, we **conditionally** apply the trimming on the end using `COUNT * 1.5` from above. This is 150 (due to step 2) * 1.5 = 225. This trimming is done with the construction of a `SearchResult` which calls `TrimResults` **conditionally**. `SearchResult` is already aware of the command type and **skips** trimming when it detects that `NeedsSorting` is true. This is a function implemented by FT.SEARCH and FT.AGGREGATE interfaces.
4. In the coordinator node, when processing all the replies in aggregation, we **conditionally** apply the original offset (100) and an buffer (1.5x) multiplied on the count (50). If we had 3 shards, we have 675 neighbors in total. So, we start from 100 and then apply the 1.5 on the original `count`. `1.5 * 50 = 75`. So, we take the first 75 neighbors starting from 100. Again, this trimming is done with the construction of a `SearchResult` which calls `TrimResults` **conditionally**. `SearchResult` is already aware of the command type and **skips** trimming when it detects that `NeedsSorting` is true. This is a function implemented by FT.SEARCH and FT.AGGREGATE interfaces.
5. In the main thread of the original node, we then call `ProcessNonVectorNeighborsForReply` and `SerializeNonVectorNeighbors`.
6. `ProcessNonVectorNeighborsForReply` can drop any neighbor where `GetContent/VerifyFilter` fails. This is why the `1.5` multiplier buffer helps.
7.  `SerializeNonVectorNeighbors` applies the original range requested in the FT.SEARCH command. Only IF we already handled the offset in the coordinator, we skip applying the offset and use a starting position of 0. The original `<count>` value of 50 is applied here.  This gives us 50 results. We are able to detect whether an offset was applied because we have this information stored on the `SearchResult`. This will be false in case of an Aggregate command which had sorting

So, I think the `GetSerializationRange` function can be part of the `SearchResult` structure since it already has awareness of the command's sorting requirements

src/query/fanout.cc
allenss-amazon: It would also happen during normal operation. A key name for another node won't get preserved in the InternedStringPool across gRPC invokations, because there's no other reference to the name. Thus when the next gRPC operation comes in the string will get re-created, possible at a different memory address. The default comparison of InternedStringPtrs is done using the address of the allocated string.

src/commands/ft_search.cc
allenss-amazon: Hmm. we might be able to do a lot better than *1.5 With my sequence numbering code, it's now possible for a background thread to get a good measurement of the number of mutations that have been queued since it was started. Wouldn't it be the case that the *1.5 should really be +X where X is the # of mutations?

src/commands/ft_search.cc
```diff
@@ -173,28 +150,27 @@ void SerializeNonVectorNeighbors(ValkeyModuleCtx *ctx,
...
+    SendReplyNoContent(ctx, search_result, *this);
     return;
   }
+  size_t original_size = neighbors.size();
   // Support non-vector queries
   if (IsNonVectorQuery()) {
     query::ProcessNonVectorNeighborsForReply(
         ctx, index_schema->GetAttributeDataType(), neighbors, *this);
-    SerializeNonVectorNeighbors(ctx, neighbors, *this);
+    search_result.total_count -= (original_size - neighbors.size());
```
allenss-amazon, line 174: Inquiring minds want to know why we have to adjust total_count here.

src/query/fanout.cc
```diff
@@ -51,8 +51,14 @@ CONTROLLED_BOOLEAN(ForceInvalidSlotFingerprint, false);
...
 struct NeighborComparator {
   bool operator()(indexes::Neighbor &a, indexes::Neighbor &b) const {
+    // Primary sort: by distance
     // We use a max heap, to pop off the furthest vector during aggregation.
-    return a.distance < b.distance;
+    if (a.distance != b.distance) {
+      return a.distance < b.distance;
+    }
+    // Secondary sort: by key for consistent ordering when distances are equal.
+    // Primarily used in non vector queries without scores (distance = 0).
```
allenss-amazon, line 60: Let's add a comment that the full string compare is required because for external keys there's no guarantee of the stability of the InternedStringPtr across invokations.

src/query/search.cc
```diff
@@ -424,9 +425,131 @@ absl::StatusOr<std::deque<indexes::Neighbor>> DoSearch(
...
+                    neighbors.size());
+  }
+  // Vector query
+  return std::min(
+      static_cast<size_t>(parameters.k),
+      std::min(static_cast<size_t>(parameters.limit.number), neighbors.size()));
+}
+
+// Static buffer multiplier for search result trimming
+static constexpr double kSearchResultBufferMultiplier = 1.5;
```
allenss-amazon: This should be a configurable. I'd also recommend a configurable constant that's added.

src/query/search.cc
```diff
@@ -424,9 +425,131 @@ absl::StatusOr<std::deque<indexes::Neighbor>> DoSearch(
...
+                    const SearchParameters &parameters) {
+  if (parameters.IsNonVectorQuery()) {
+    return std::min(static_cast<size_t>(parameters.limit.number),
+                    neighbors.size());
+  }
+  // Vector query
+  return std::min(
+      static_cast<size_t>(parameters.k),
+      std::min(static_cast<size_t>(parameters.limit.number), neighbors.size()));
+}
```
allenss-amazon: I've never understood why these aren't a single function that returns a pair of numbers like your SerializationRange.

src/query/search.cc
```diff
@@ -424,9 +425,131 @@ absl::StatusOr<std::deque<indexes::Neighbor>> DoSearch(
...
+  // Clear neighbors if no results should be returned
+  if (ShouldReturnNoResults(parameters)) {
+    this->neighbors.clear();
+    return;
+  }
+  // Check if the command needs all results (e.g. for sorting). Trim otherwise.
+  if (RetainAllNeighbors(parameters)) {
+    this->neighbors = std::move(neighbors);
+  } else {
+    this->neighbors = std::move(neighbors);
```
allenss-amazon: move of neighbors belongs outside of the if?

src/query/search.cc
```diff
@@ -424,9 +425,131 @@ absl::StatusOr<std::deque<indexes::Neighbor>> DoSearch(
...
+    TrimResults(this->neighbors, parameters);
+  }
+}
+
+// Determine if we need full results or if we can optimize with trimming via
+// LIMIT offset & count.
+bool SearchResult::RetainAllNeighbors(const SearchParameters &parameters) {
+  return parameters.RequiresCompleteResults();
+}
```
allenss-amazon: This seems a bit baroque.

src/query/search.cc
```diff
@@ -424,9 +425,131 @@ absl::StatusOr<std::deque<indexes::Neighbor>> DoSearch(
...
+  // Note: We cannot trim from the front in a Cluster Mode setting because
+  // each shard X results and we need to trim the OFFSET on the aggregated
+  // results. Thus, we can only trim from the end in searches for individual
+  // nodes.
+  // In cluster mode, the offset based trimming is applied after merging all
+  // results from shards at the coordinator level.
+  if (!ValkeySearch::Instance().IsCluster()) {
+    this->is_offsetted = true;
+    size_t start_index = CalcStartIndex(neighbors, parameters);
+    // Trim from front (apply offset)
```
allenss-amazon: Here is where having the start/end index computation be a virtual function helps, because the specific cluster/non-cluster logic gets handled by that.

Today SearchParameters is a base class for SearchCommand and AggregateCommand. It should be simple to make another derived class (maybe call it RemoteSearch) that implements the scenario where the search is initiated remotely (cluster mode). Then making some of this virtual functions moves the logic out of here and into those places.

src/query/search.cc
```diff
@@ -424,9 +425,131 @@ absl::StatusOr<std::deque<indexes::Neighbor>> DoSearch(
...
+SerializationRange SearchResult::GetSerializationRange(
+    const SearchParameters &parameters) const {
+  CHECK(!ShouldReturnNoResults(parameters));
+  const size_t start_index =
+      is_offsetted ? 0 : query::CalcStartIndex(neighbors, parameters);
+  const size_t end_index =
+      std::min(start_index + query::CalcEndIndex(neighbors, parameters),
+               neighbors.size());
+  return {start_index, end_index};
+}
```
allenss-amazon, line 522: I think this should be the virtual function.

src/query/search.cc
KarthikSubbarao: Yes, I was looking into this earlier. However, we don't have a great double numeric value based config framework in Valkey.

We could store this as a string and convert it to 1.5. 

OR we could store this as a "percent" with min value of 100. Then we can divide it by when using it. So, 150 means 1.5.

allenss-amazon: You can also store it as a rational number, i.e., a numerator and a denominator.

KarthikSubbarao: A single config would be nicer. I added a string config which is converted to a double and stored on config set. We do something similar in core valkey to handle a double value.

KarthikSubbarao: Done

src/query/search.cc
KarthikSubbarao: I deleted these functions( CalculateStart and End) and just added logic in the single function called GetSerializationRange

src/query/search.cc
KarthikSubbarao: CalcStartIndex and CalcEndIndex have been deleted now and this logic is just contained within `GetSerializationRange`.

src/query/search.cc
KarthikSubbarao: I would feel more comfortable making this a virtual function that once we have divergent logic across two commands. For now, SearchResult is already making use of a virtual function named `RequiresCompleteResults` which is implemented in both Search and Aggregate commands, so we have awareness based on this to apply trimming or retain all the neighbors.

If the trimming logic itself is supposed to be divergent, then we can make this a virtual function. Until then and until we have tests for both cases (both command variations), I would prefer keeping a single function which is tested correctly and used from a single place

PR #538, opened 2025-12-11
Changes for Fuzzy search 

daddaman-amz:
This PR covers:
* Fuzzy search: Damerau-Levenshtein algorithm with configurable edit distance 
* Path iterator APIs : Used by Fuzzy search

### DIFF
--- a/src/indexes/text/fuzzy.h
+++ b/src/indexes/text/fuzzy.h
@@ -16,16 +16,20 @@
 #include "invasive_ptr.h"
 #include "posting.h"
 #include "radix_tree.h"
+#include "text.h"
 
 namespace valkey_search::indexes::text {
 
 // Fuzzy search using Damerau-Levenshtein distance on RadixTree
 struct FuzzySearch {
   // Returns KeyIterators for all words within edit distance <= max_distance
-  static std::vector<Postings::KeyIterator> Search(
-      const RadixTree<InvasivePtr<Postings>>& tree, absl::string_view pattern,
-      size_t max_distance) {
-    std::vector<indexes::text::Postings::KeyIterator> key_iterators;
+  static absl::InlinedVector<Postings::KeyIterator,
+                             kWordExpansionInlineCapacity>
+  Search(const RadixTree<InvasivePtr<Postings>>& tree,
+         absl::string_view pattern, size_t max_distance) {
+    absl::InlinedVector<indexes::text::Postings::KeyIterator,
+                        kWordExpansionInlineCapacity>
+        key_iterators;
 
     // Dynamic Programming matrix rows for Damerau-Levenshtein algorithm
     // Row i-2 (for transposition)
@@ -60,7 +64,8 @@ struct FuzzySearch {
           prev,  // Row i-1 of DP matrix (previous row)
       absl::InlinedVector<size_t, 32>&
           curr,  // Row i of DP matrix (current row being computed)
-      std::vector<indexes::text::Postings::KeyIterator>& key_iterators) {
+      absl::InlinedVector<indexes::text::Postings::KeyIterator,
+                          kWordExpansionInlineCapacity>& key_iterators) {
     // Iterate over children at current tree level
     while (!iter.Done()) {
       absl::string_view edge = iter.GetChildEdge();
--- a/src/query/predicate.cc
+++ b/src/query/predicate.cc
@@ -221,7 +221,9 @@ EvaluationResult FuzzyPredicate::Evaluate(
                                                       term_, distance_);
 
   // Filter to only include KeyIterators that match target_key and field_mask
-  std::vector<indexes::text::Postings::KeyIterator> filtered_key_iterators;
+  absl::InlinedVector<indexes::text::Postings::KeyIterator,
+                      indexes::text::kWordExpansionInlineCapacity>
+      filtered_key_iterators;
   for (auto& key_iter : key_iters) {
     if (key_iter.SkipForwardKey(target_key) &&
         key_iter.ContainsFields(field_mask)) {


src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+ private:
+static void SearchRecursive(
+    typename RadixTree<Target>::PathIterator iter,
+    absl::string_view pattern, 
+    size_t max_distance,
+    std::string word,  // Current word being built 
+    char prev_char,    // Previous character (for transposition detection)
+    std::vector<size_t>& prev_prev,  // Row i-2 of DP matrix (for transposition)
+    std::vector<size_t>& prev,       // Row i-1 of DP matrix (previous row)
+    std::vector<size_t>& curr,       // Row i of DP matrix (current row being computed)
```
BCathcart: Would it make sense to initialize the capacity to the size of the word?

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
-  virtual void Next() = override;
-  virtual const Key& operator*() const override;
-  virtual std::unique_ptr<WordIterator> clone() const override;
-};
+// Fuzzy search using Damerau-Levenshtein distance on RadixTree
+template <typename Target>
+struct FuzzySearch {
+  // Search for words within max_distance edits of pattern
+  // Returns vector of (word, target) pairs
+  static std::vector<std::pair<std::string, Target>> Search(
```
BCathcart: Looking at the existing pattern for other search types, this fuzzy search code would actually go into `BuildTextIterator()`defined on the fuzzy predicate. It would return a term iterator that holds a key iterator for every target we matched.

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+
+    // SAVE STATE: Each sibling must start with same parent state    
+    auto saved_prev_prev = prev_prev;
+    auto saved_prev = prev;
+    char saved_prev_char = prev_char;
+
+    // Process each character in the path
+    for (char ch : path) {
+      new_word += ch;
+      char lower_ch = std::tolower(static_cast<unsigned char>(ch));
```
BCathcart: Is it typical to ignore case in fuzzy search?

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+        // Damerau-Levenshtein: transposition
+        if (i > 1 && new_word.length() > 1 && 
+            lower_ch == std::tolower(
+                static_cast<unsigned char>(pattern[i - 2])) &&
+            lower_pat == std::tolower(
+                static_cast<unsigned char>(prev_char))) {
+          curr[i] = std::min(curr[i], prev_prev[i - 2] + cost);
+        }
+        
+        min_dist = std::min(min_dist, curr[i]);
```
BCathcart: Would we want to always take the value at `curr[pattern.length()]`? If we take the minimum couldn't it represent a point where we're only comparing a substring of the pattern to the word?

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+    typename RadixTree<Target>::PathIterator iter,
+    absl::string_view pattern, 
+    size_t max_distance,
+    std::string word,  // Current word being built 
+    char prev_char,    // Previous character (for transposition detection)
+    std::vector<size_t>& prev_prev,  // Row i-2 of DP matrix (for transposition)
+    std::vector<size_t>& prev,       // Row i-1 of DP matrix (previous row)
+    std::vector<size_t>& curr,       // Row i of DP matrix (current row being computed)
+    std::vector<std::pair<std::string, Target>>& results) {
+  // ITERATE OVER SIBLINGS at current tree level
```
BCathcart: Are we actually iterating over the children of the node?

A branching node will have a single character path to each child nodes. A compressed node will have a n-character path to its single child node. Do we handle both these cases?

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+  while (!iter.Done()) {
```
BCathcart: We may need hierarchical iterators. I'd guess you want to get a child iterator from from the current node the path iterator is at for this while loop. Otherwise the path iterator would need multiple different "done" states.

src/indexes/text/fuzzy.h
BCathcart: Would we want to always take the value at `curr[pattern.length()]`? If we take the minimum couldn't it represent a point where we're only comparing a substring of the pattern to the word?

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+        min_dist = std::min(min_dist, curr[i]);
+      }
+      
+      prev_prev.swap(prev);
+      prev.swap(curr);
+      prev_char = ch;
+    }
+    // NOW check if we should prune this subtree
+    size_t min_dist = *std::min_element(prev.begin(), prev.end());
+    bool should_prune = (min_dist > max_distance);
```
BCathcart: Will this cause us to exit too early? At the root the path is just the empty string, which will usually have a fuzzy distance greater than the max distance

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+
+    // Process each character in the path
+    for (char ch : path) {
+      new_word += ch;
+      char lower_ch = std::tolower(static_cast<unsigned char>(ch));
+      
+      curr[0] = new_word.length();
+      size_t min_dist = curr[0];
+      
+      for (size_t i = 1; i <= pattern.length(); ++i) {
```
BCathcart: Looks like we're building the matrix from scratch every time. Can we build on top of our parent's matrix?

src/indexes/text/fuzzy.h
BCathcart: Also are we using this `min_dist` variable right now? Looks like we re-compute it below

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+    // Descend to the child node that this path leads to
+    if (iter.CanDescend()) {
+      auto child_iter = iter.DescendNew();
+      std::cout << "DEBUG:     After descend: child_iter.IsWord()=" << child_iter.IsWord() 
+                << " new_word='" << new_word << "' distance=" << prev[pattern.length()] << std::endl;
+      
+      // Check if the child is a word
+      if (child_iter.IsWord() && prev[pattern.length()] <= max_distance) {
+        std::cout << "DEBUG:     ADDING RESULT: " << new_word << std::endl;
+        results.emplace_back(new_word, child_iter.GetTarget());
```
BCathcart: Are we adding the child as a valid result based on the parent's distance?

src/indexes/text/fuzzy.h
```diff
@@ -1,43 +1,202 @@
...
+    if (should_prune) {
+      // Restore state before moving to next sibling
+      prev_prev = saved_prev_prev;
+      prev = saved_prev;
+      prev_char = saved_prev_char;
+      iter.Next();
+      continue;
+    }
+    
+    // Descend to the child node that this path leads to
```
BCathcart: There can be up to 256 child nodes branching from this node. I guess the intention is to descend to the first child and then the next() calls will move across the children. It feels more intuitive to me to calculate the Damerau-Levenshtein distance to each child node when you're at the parent. Any children that are within the distance, you make a recursive call to as well as add to the result set if they have a target.

src/indexes/text/fuzzy.h
daddaman-amz: Yes the earlier version was more to show the intent of path iterator apis usage with the algorithm. I've since updated it to return KeyIterators directly.

src/indexes/text/fuzzy.h
daddaman-amz: Yeah I documented it in my doc's last section too. Also I realized during indexing we always make lowercase (via UnicodeNormalizer::CaseFold() in lexer.cc:106) so I think we may not need this conversion for the chars in tree but only for the user entered pattern.

src/indexes/text/fuzzy.h
daddaman-amz: Yes I removed that redundant value and use the one we compute here. Using min distance of the row helps us decide if we want to descend further. If min value is already more than target distance, we won't find any more matches down the tree.

src/indexes/text/fuzzy.h
daddaman-amz: We are building on the parent matrix. This loop is building the current row (using values from the previous row), which will be equal to the pattern length. The number of rows will depend on how much we descend, but if we prune at any stage we restore the state so the next child (sibling) can build on the existing parent rows.

src/indexes/text/fuzzy.h
daddaman-amz: We do a swap of the rows in the loop above so the curr row's values are now stored in prev row.  Therefore prev[pattern.length()] holds the latest computed value.   I missed adding a comment here, will do

src/indexes/text/fuzzy.h
daddaman-amz: we descend to the first child only if it satisfies the `Damerau-Levenshtein distance` . If we find that we have to prune (based on min_dist_, then we call Next() to move to next sibling.

src/commands/ft_create_parser.h
```diff
@@ -25,7 +25,7 @@
 namespace valkey_search {
 
 static constexpr absl::string_view kDefaultPunctuation =
-    ",.<>{}[]\"':;!@#$%^&*()?-+=~/\\|";
+    ",.<>{}[]\"':;!@#$%^&*()-+=~/\\|";
```
daddaman-amz, line 28: I removed the '?' for now for unit tests. But I want  to confirm if we can add ? as well

src/indexes/text/fuzzy.h
BCathcart: Ok that sounds like a good default to me but I wonder if we should have a config to enable case-sensitivity. We don't have to discuss that here though

src/indexes/text/fuzzy.h
BCathcart: I'm just going off of the fact the outer loop appears to be iterating over the whole path and the inner loop over the whole search pattern. Unless GetPath() is supposed to only return the single path segment from its parent?

src/indexes/text/fuzzy.h
BCathcart: Ok help me understand what the elements of the last row in the matrix represent. The last row represents the last character in the path. Each column is a different character of the search pattern. If we take the minimum of the last row and select an element other than the last one, are we not considering a distance that is only accounting for a substring of the search pattern?

src/indexes/text/fuzzy.h
BCathcart: It holds the latest computed value, but isn't that value for the current node and now we're descending to its child and immediately adding it if the current node satisfies the distance. Wouldn't we need to compute the distance at the child node before knowing it's safe to add? I might be fundamentally misunderstanding the algorithm you're going for here

src/indexes/text/fuzzy.h
BCathcart: My reading of this code is that we descend to the first child if some substring of the search pattern satisfies the Damerau-Levenshtein distance to the current node's path.

BCathcart: Oh are you assuming GetPath() returns the edge to a child of the current node? I think the code aligns closer to what I was expecting in that case. In the existing PathIterator API stub, GetPath() returns the path to the current node.. or maybe I misunderstood it. Anyways I'll take a look at your updated code shortly. I probably should have waited until you defined the APIs in hindsight.

BCathcart (commented):
.

src/indexes/text/fuzzy.h
daddaman-amz: I think you are right, I may have misunderstood the existing API's stub .   Yes the current GetPath() is returning just the edge but the stub comment indicates it should return the accumulated path till that point.  Let me re-check if original intent is more optimal for me to adapt.
If not I will rename the api to indicate "edge" like you suggested GetChildEdge() and have GetPath() be its own for its intended purpose.

integration/test_fulltext.py
```diff
@@ -190,6 +190,11 @@ def validate_fulltext_search(client: Valkey):
     result = client.execute_command("FT.SEARCH", "products", 'artificial intelligence research')
     assert result[0] == 1
     assert result[1] == b"product:6"
+    # Test fuzzy search
+    result = client.execute_command("FT.SEARCH", "products", '@desc:%wander%')
+    assert (result[0], set(result[1::2])) == (1, {b"product:4"})
+    result = client.execute_command("FT.SEARCH", "products", '@desc:%%greet%%')
+    assert (result[0], set(result[1::2])) == (3, {b"product:1", b"product:5", b"product:6"})
```
BCathcart, line 197: I don't think it necessarily hurts to add this here, but in general this seems like a vague test that validates some subset of our search functionality that isn't immediately clear. Not sure it's worth building on top of. I prefer having tests like `test_fuzzy_search` that comprehensively test a certain feature

src/indexes/text/fuzzy.h
daddaman-amz: Renamed this call as GetChildEdge() which is more apt and changed the GetPath() to just return accumulated path till the node

src/indexes/text/fuzzy.h
daddaman-amz: Renamed this call as GetChildEdge() which is more apt and changed the GetPath() to just return accumulated path till the node

integration/test_fulltext.py
```diff
@@ -1402,6 +1407,83 @@ def test_proximity_inorder_compat(self):
...
+        result = client.execute_command("FT.SEARCH", "idx1", 'drove the %car%', "SLOP", "0")
+        assert (result[0], set(result[1::2])) == (1, {b"doc:4"})
+        # Test with Inorder
+        result = client.execute_command("FT.SEARCH", "idx1", 'drove the %car%', "INORDER")
+        assert (result[0], set(result[1::2])) == (1, {b"doc:4"})
+        client.execute_command("DEL", "doc:11")
+        # Stemming case
+        result = client.execute_command("FT.SEARCH", "idx2", '%%drive%%')
+        assert (result[0], set(result[1::2])) == (1, {b"doc:4"})
+        # Not working. Expect driving to be treated as drive?
```
daddaman-amz: This is related to question mark and not stemming. will remove the comment

src/indexes/text/fuzzy.h
daddaman-amz: yeah worth discussing if such setting should be at index creation level or some global config

integration/test_fulltext.py
daddaman-amz: Will fix the "ignore question mark" as part of an other PR I have, as I see there is also may be some small  bug in parsing where, any special character towards the end is errored out as invalid syntax instead of ignoring the punctuation.
```
-> import pdb; pdb.set_trace()
(Pdb) client.execute_command("FT.SEARCH", "idx1", 'How are you doing?')
[0]
(Pdb) client.execute_command("FT.SEARCH", "idx1", 'How are you doing+')
*** valkey.exceptions.ResponseError: Invalid filter expression: `How are you doing+`. Invalid Query Syntax
(Pdb) client.execute_command("FT.SEARCH", "idx1", 'How are you doing!')
*** valkey.exceptions.ResponseError: Invalid filter expression: `How are you doing!`. Invalid Query Syntax
```

src/indexes/text/fuzzy.h
daddaman-amz: By this time we have processed all the characters in the edge up to this child node, and the edit distance is stored in `prev[pattern.length()]`. When we Descend, the iterator now points to that child node, and we can check if it has a word or not.

src/indexes/text/fuzzy.h
daddaman-amz: Each row represents the edit distance from the tree path processed so far to each prefix of the pattern.

For example, consider Pattern="cra" and tree word "car":
```
        ""  "c"  "cr"  "cra"    ← pattern
""   [ 0,  1,   2,   3  ]
"c"  [ 1,  0,   1,   2  ]
"ca" [ 2,  1,   1,   1  ]
"car"[ 3,  2,   1,   1  ]
 ↑
Tree
```
When computing the row for "ca", each column represents the cost of converting "ca" to a prefix of the pattern:

Column 0: "ca" → "" = 2 (delete both)
Column 1: "ca" → "c" = 1 (delete 'a')
Column 2: "ca" → "cr" = 1 (substitute a→r)
Column 3: "ca" → "cra" = 1 (insert 'r')

Above calculations are done by using the previous row (to account for processed chars ) + the char 'a' .  ( Added in this the code comment)

The last element (1) represents converting "ca" to the full pattern "cra".

For pruning: We use min_dist = 1 (the minimum across the row). If min_dist ≤ max_distance, we descend further. This works because the minimum represents the best possible alignment, suggesting potential for a good match deeper in the tree.

For matching: When we find a word, we check if the last column ≤ max_distance to determine if it's a valid match.

integration/test_fulltext.py
```diff
@@ -1402,6 +1407,84 @@ def test_proximity_inorder_compat(self):
...
+        # Simple Edit distance (ED) = 1
+        result = client.execute_command("FT.SEARCH", "idx1", '%car%')
+        assert (result[0], set(result[1::2])) == (2, {b"doc:2", b"doc:4"})
+        # Should be case insensitive
+        result = client.execute_command("FT.SEARCH", "idx1", '%CAR%')
+        assert (result[0], set(result[1::2])) == (2, {b"doc:2", b"doc:4"})
+        # Transposition (Damerau-Levenshtein) ED = 1
+        result = client.execute_command("FT.SEARCH", "idx1", '%crA%')
+        assert (result[0], set(result[1::2])) == (1, {b"doc:4"})
+        client.execute_command("HSET", "doc:5", "content", "abdc")
```
BCathcart: nit: what do you think about throwing all the document creation up top? Would make it a bit easier to keep track of for me

integration/test_fulltext.py
```diff
@@ -1402,6 +1407,84 @@ def test_proximity_inorder_compat(self):
...
+        result = client.execute_command("FT.SEARCH", "idx1", '%%bacd%%')
+        # race from doc:1 (ED = 2) and abcd from doc:4 (ED = 2, transposition)
```
BCathcart: > and abcd from doc:4 (ED = 2, transposition)

'abdc' from doc:5?

integration/test_fulltext.py
```diff
@@ -1402,6 +1407,84 @@ def test_proximity_inorder_compat(self):
...
+        assert (result[0], set(result[1::2])) == (2, {b'doc:1', b'doc:5'})
+        # In Composed AND
+        client.execute_command("HSET", "doc:11", "content", "Carl Weathers drove the big tramcar")
+        IndexingTestHelper.wait_for_backfill_complete_on_node(client, "idx1")
```
BCathcart: The backfill was already waited for up top

integration/test_fulltext.py
```diff
@@ -1402,6 +1407,84 @@ def test_proximity_inorder_compat(self):
...
 
+    def test_fuzzy_search(self):
+        client: Valkey = self.server.get_new_client()
+        # Create index with text fields
+        client.execute_command("FT.CREATE", "idx1", "ON", "HASH", "SCHEMA",
+                            "content", "TEXT", "NOSTEM")
+        client.execute_command("FT.CREATE", "idx2", "ON", "HASH", "SCHEMA",
+                    "content", "TEXT")
+        client.execute_command("FT.CREATE", "idx3", "ON", "HASH", "SCHEMA",
+                    "content", "TEXT", "NOSTEM", "content2", "TEXT", "NOSTEM")
```
BCathcart, line 1418: It looks like this index is unused

integration/test_fulltext.py
```diff
@@ -1402,6 +1407,84 @@ def test_proximity_inorder_compat(self):
...
+        IndexingTestHelper.wait_for_backfill_complete_on_node(client, "idx1")
+        result = client.execute_command("FT.SEARCH", "idx1", 'Driver drove the %Kar%')
+        assert (result[0], set(result[1::2])) == (1, {b"doc:4"})
+        result = client.execute_command("FT.SEARCH", "idx1", 'drove the %car%')
+        assert (result[0], set(result[1::2])) == (2, {b"doc:4", b"doc:11"})
+        # Test with slop
+        result = client.execute_command("FT.SEARCH", "idx1", 'drove the %car%', "SLOP", "0")
+        assert (result[0], set(result[1::2])) == (1, {b"doc:4"})
+        # Test with Inorder
+        result = client.execute_command("FT.SEARCH", "idx1", 'drove the %car%', "INORDER")
```
BCathcart, line 1459: nit: it might be more interesting if the tests stress the SLOP and INORDER features a bit more. You can start with a composed AND that only returns a result without them. Then test that the search fails with SLOP or INORDER and add a new search that does return a result.

integration/test_fulltext.py
```diff
@@ -1402,6 +1407,84 @@ def test_proximity_inorder_compat(self):
...
+        result = client.execute_command("FT.SEARCH", "idx1", '%%drive%%')
+        assert (result[0], set(result[1::2])) == (1, {b"doc:4"})
+        # stemming enabled should give doc:3 (with word 'driving')
+        # TODO: fails as ? is not ignored. enable after fix
+        # result = client.execute_command("FT.SEARCH", "idx2", '%%drive%%')
+        # assert (result[0], set(result[1::2])) == (2, {b"doc:3", b"doc:4"}) 
+        # Higher edit distance test (ED=10)
+        # Add a document with a word that requires high edit distance
+        client.execute_command("HSET", "doc:6", "content", "abcdefghij")
+        IndexingTestHelper.wait_for_backfill_complete_on_node(client, "idx1")
```
BCathcart: No need to wait for backfill

integration/test_fulltext.py
```diff
@@ -1402,6 +1407,84 @@ def test_proximity_inorder_compat(self):
...
+        # Increase max edit distance config
+        client.execute_command("CONFIG", "SET", "search.fuzzy-max-distance", "10")
+        result = client.execute_command("FT.SEARCH", "idx1", '%%%%%%%%%%z%%%%%%%%%%')
+        assert (result[0], set(result[1::2])) == (6, {b"doc:1", b"doc:2", b"doc:3", b"doc:4", b"doc:5", b"doc:6"})
+        # Long word test - add some prefix words for branching
+        client.execute_command("HSET", "doc:7", "content", "internationalization")
+        client.execute_command("HSET", "doc:8", "content", "ice")
+        client.execute_command("HSET", "doc:9", "content", "in")
+        client.execute_command("HSET", "doc:10", "content", "internet")
+        IndexingTestHelper.wait_for_backfill_complete_on_node(client, "idx1")
```
BCathcart: same here

src/indexes/text/fuzzy.h
```diff
@@ -2,42 +2,168 @@
...
-//
-struct FuzzyKeyIterator : public public indexes::EntriesFetcherIteratorBase {
-  FuzzyIterator(indexes::Text& index, absl::string_view word, size_t distance);
-  virtual bool Done() const override;
-  virtual void Next() = override;
-  virtual const Key& operator*() const override;
-  virtual std::unique_ptr<WordIterator> clone() const override;
-};
+// Fuzzy search using Damerau-Levenshtein distance on RadixTree
+template <typename Target>
```
BCathcart: I'm not sure this needs to be templated. The return type already exposes that a postings object is in the tree `std::vector<Postings::KeyIterator>`. I guess if we have different pointer types in trees pointing to postings objects at some point it could theoretically be useful, but I'm not sure why fuzzy search would operate on multiple different types of trees.

src/indexes/text/fuzzy.h
```diff
@@ -2,42 +2,168 @@
...
+    absl::InlinedVector<size_t, 32> curr(pattern.length() + 1);
+
+    // Initialize first row: distance from empty string to each pattern prefix.
+    // Example: for pattern "race", first row is [0, 1, 2, 3, 4]
+    for (size_t i = 0; i <= pattern.length(); ++i) {
+      prev[i] = i;
+    }
+
+    // Start traversal from root to explore all words in the tree
+    auto iter = tree.GetPathIterator("");
```
BCathcart, line 49: I'd like to think through if we can make the path iterator more intuitive. An iterator is typically pointing to a single element of a structure, and incrementing it takes it to the next logical element. The path iterator is a bit different right now wherein it points to a node in the tree and points to one of its children via a nested iterator.

> // Start traversal from root to explore all words in the tree

Logically we start at the root node.

> // Iterate over siblings at current tree level

Then immediately we're logically at the first child. If you look at the path iterator implementation you can understand what's happening, but without it last week I found it quite confusing.

I was thinking if we should expose the nested child iterator. At the top of `SearchRecursive` you would have `auto child_iter = path_iter.GetChildIterator()`. Then you can call `child_iter.GetEdge()` and `child_iter.Next()`. But then the descending API gets a bit messy so maybe it's fine to keep it hidden. In that case I'd suggest we commit to framing the path iterator from the perspective of the parent or children but not both.

Parent:
- GetPathIterator() // Start traversal from root to explore all words in the tree
- GetChildEdge()
- NextChild()
- Descend() // Descends to the child

Children:
- GetPathIterator() // Start traversal from first child of the root
- GetIncomingEdge()
- NextSibling()
- Descend()  // Descends to the first child of the current node

What do you think?

integration/test_fulltext.py
daddaman-amz: From my conversation with Karthik in earlier PRs, we are looking at this test more like a sanity test to validate all basic scenarios.  This sanity test runs with both CME and CMD modes.

src/indexes/text/fuzzy.h
daddaman-amz: Thank you.
As chatted offline I will rename the NextSibling() to NextChild() to be consistent

src/indexes/text/fuzzy.h
BCathcart: Thanks for the explanation.

This was the part I was trying to understand:
> For pruning: We use min_dist = 1 (the minimum across the row). If min_dist ≤ max_distance, we descend further. This works because the minimum represents the best possible alignment, suggesting potential for a good match deeper in the tree.

Makes sense now to me now. We use the best possible alignment for pruning and don't care what the actual distance is since the following characters could match the pattern and converge towards the best possible alignment distance. I guess if the best possible alignment was at the end of the pattern, going deeper wouldn't help.. anyways no need to worry about edge cases.

src/indexes/text/fuzzy.h
BCathcart: Makes sense once I understood the semantics of the PathIterator APIs

integration/test_fulltext.py
daddaman-amz: Yes there was a typo. The commented `Return clause` test will use it.  I will uncomment it in different PR

integration/test_fulltext.py
daddaman-amz: The test does that by showing doc:11 missing our in results with slop and inorder params. I have also now added extra test f to demo that doc:11 gets added to result with right parameters

BCathcart (approved):
Looks good! Note I didn't look closely at the path iterator implementation beyond the APIs since we're planning to replace them when we switch to Rax

PR #541, opened 2025-12-12
Block Search Response on Conflicting In-Flight Keys for fulltext queries

baswanth09:
### Full-text query blocking on in-flight mutations

Adds blocking mechanism for full-text queries when there are in-flight key mutations to ensure query consistency.

Blocking mechanism is implemented in three code paths:
1. commands.cc - Local standalone queries (CMD)
2. server.cc - Remote shard queries (gRPC) (CME)
3. fanout.cc - Local partition in fanout operations (CME)

#### Tests:
`test_fulltext_inflight_blocking.py` - Integration test for standalone (CMD) and cluster (CME) modes using PAUSEPOINTS testing infrastructure to simulate in-flight mutations.
Command: `TEST_PATTERN=test_fulltext_inflight_blocking ./build.sh --run-integration-tests`

### DIFF
--- a/src/commands/commands.cc
+++ b/src/commands/commands.cc
@@ -181,7 +181,7 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
       // Text predicate evaluation requires main thread to ensure text indexes
       // reflect current keyspace. Block if result keys have in-flight
       // mutations.
-      if (!result->parameters->no_content &&
+      if (result->search_result.ok() && !result->parameters->no_content &&
           query::QueryHasTextPredicate(*result->parameters)) {
         auto initiator_search = std::make_unique<async::InitiatorSearch>(
             std::move(blocked_client), std::move(result));


src/query/predicate.cc
```diff
@@ -435,4 +435,30 @@ EvaluationResult ComposedPredicate::Evaluate(Evaluator& evaluator) const {
   return EvaluationResult(true, std::move(or_proximity_iterator));
 }
 
+bool HasTextPredicate(const Predicate* predicate) {
```
BCathcart: Is it possible to flag the AST with whether or not it has a text component when we're building it so we don't have to traverse it at the end?

src/query/fanout.cc
```diff
@@ -256,6 +305,24 @@ absl::Status PerformSearchFanoutAsync(
         [tracker](absl::StatusOr<std::deque<indexes::Neighbor>> &neighbors,
                   std::unique_ptr<SearchParameters> parameters) {
           if (neighbors.ok()) {
+            // For pure full-text queries with content, check for in-flight key
+            // conflicts before adding results.
+            if (!parameters->no_content &&
+                query::IsPureFullTextQuery(*parameters)) {
```
BCathcart: Shouldn't we wait if the query has any text component? As long as there is some text evaluation happening on a key, we need to wait until its latest mutation is indexed

src/commands/commands.cc
```diff
@@ -29,6 +39,48 @@ struct Result {
...
+};
+
+void InFlightRetryTimerCallback(ValkeyModuleCtx *ctx, void *data);
+
+void CheckAndHandleInFlightConflicts(ValkeyModuleCtx *ctx,
+                                     InFlightRetryContext *retry_ctx) {
+  auto &result = retry_ctx->result;
+
+  if (result->parameters->cancellation_token->IsCancelled()) {
+    retry_ctx->blocked_client.SetReplyPrivateData(result.release());
```
BCathcart: Does this release the results when we timeout rather than an error response?

src/commands/commands.cc
```diff
@@ -128,6 +180,27 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
           .neighbors = std::move(neighbors),
           .parameters = std::move(upcast_parameters),
       });
+
+      // For pure full-text queries with content, check for in-flight key
+      // conflicts. Skip if no_content since we only return key names without
+      // fetching/evaluating data.
```
BCathcart: Did you consider waiting to evaluate for conflicts on the main thread? You could let the Reply callback run and then re-block the client and set up a timer if there are any conflicts. I'm curious if you've thought about the trade-offs

baswanth09: I think the current implementation makes more sense as we need to do the blocking mechanism at three code paths. Having a unified and consistent pattern across all of these code paths is better. Conflicts must be checked per-shard before aggregation: In cluster mode, each shard evaluates predicates against its local data. The conflict check must happen at each shard before results are sent/aggregated. For remote shards in `server.cc` we need to check before sending gRPC response - there is no Valkey Reply callback here, only `reactor->Finish()`

By checking in the background callback (before reaching each path's response mechanism), all three paths use the same pattern. Also, most queries won't have conflicts. Checking on the background thread allows non-conflicting queries to complete without extra main thread work.

src/commands/commands.cc
baswanth09: I think it is handled in the `Reply` function. I will double check
```
int Reply(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
  auto *res =
      static_cast<Result *>(ValkeyModule_GetBlockedClientPrivateData(ctx));
  CHECK(res != nullptr);

  // Check if operation was cancelled and partial results are disabled
  if (!res->parameters->enable_partial_results &&
      res->parameters->cancellation_token->IsCancelled()) {
    ++Metrics::GetStats().query_failed_requests_cnt;
    return ValkeyModule_ReplyWithError(
        ctx, "Search operation cancelled due to timeout");
  }
 ```

File: src/query/fanout.cc (hunk 2 above)
Comment (baswanth09) [reply src/query/fanout.cc]: You mean for the hybrid queries? I was on the impression that we need blocking mechanism only for pure full text queries as vector indexing takes more time in general when compared to full text. I will address this in the revision to handle all queries with a text component.

File: src/query/predicate.cc (hunk 1 above)
Comment (baswanth09) [reply src/query/predicate.cc]: Yes. I think it makes sense to have a flag like `has_text_predicate`.

File: integration/test_fulltext_inflight_blocking.py [hunk 5]
```diff
@@ -0,0 +1,152 @@
...
+            lambda: client.execute_command("FT._DEBUG PAUSEPOINT TEST mutation_processing") > 0,
+            timeout=5
+        )
+
+        search_thread, search_res, search_err = run_in_thread(
+            lambda: self.server.get_new_client().execute_command(
+                "FT.SEARCH", "idx", "@content:hello"
+            )
+        )
+        time.sleep(0.5)
```
BCathcart: Rather than sleeping here, can we check the `fulltext_query_blocked_count` metric?

integration/test_fulltext_inflight_blocking.py
```diff
@@ -0,0 +1,152 @@
...
+
+        waiters.wait_for_true(
+            lambda: all(nc.execute_command("FT._DEBUG PAUSEPOINT TEST mutation_processing") > 0
+                       for nc in primary_clients),
+        )
+
+        search_thread, search_res, search_err = run_in_thread(
+            lambda: client.execute_command("FT.SEARCH", "idx", "@content:hello")
+        )
+        time.sleep(0.5)
```
BCathcart: Same here

integration/test_fulltext_inflight_blocking.py
```diff
@@ -0,0 +1,152 @@
...
+from valkey_search_test_case import (
+    ValkeySearchTestCaseDebugMode,
+    ValkeySearchClusterTestCaseDebugMode
+)
+from valkeytestframework.util import waiters
+from valkeytestframework.conftest import resource_port_tracker
+from utils import IndexingTestHelper
+
+
+def run_in_thread(func):
```
BCathcart: Maybe throw this in the utils file for re-use?

integration/test_fulltext_inflight_blocking.py
```diff
@@ -0,0 +1,152 @@
...
+        try:
+            result[0] = func()
+        except Exception as e:
+            error[0] = e
+    t = threading.Thread(target=wrapper)
+    t.start()
+    return t, result, error
+
+
+class TestFullTextInFlightBlockingCMD(ValkeySearchTestCaseDebugMode):
```
BCathcart, line 15: We should also have a test for hybrid queries with a text component to make sure they are blocked and a test for a query without a text component on an index with a text field defined to make sure it doesn't block.

integration/test_fulltext_inflight_blocking.py
```diff
@@ -0,0 +1,152 @@
...
+        for slot_range in slots_info[:num_shards]:
+            for i in range(1000):
+                key = f"doc:test:{i}"
+                slot = client.execute_command("CLUSTER", "KEYSLOT", key)
+                if slot_range[0] <= slot <= slot_range[1]:
+                    shard_keys.append(key)
+                    break
+        return shard_keys
+
+    def test_fulltext_inflight_blocking_cluster_with_pausepoint(self):
```
BCathcart, line 205: In this test the evaluation will be blocked on every node and then unblocked at roughly the same time. A couple potentially interesting cases:
1. The coordinator evaluation isn't blocked but evaluation is blocked on at least one of the other nodes.
2. Evaluation is blocked on at least one of the other nodes but isn't blocked on the coordinator.
3. Let one of the remote searches time out with partial results enabled.

src/coordinator/server.cc
```diff
@@ -112,6 +115,85 @@ void SerializeNeighbors(SearchIndexPartitionResponse* response,
...
+                             std::vector<InternedStringPtr> keys)
+      : response(resp),
+        reactor(react),
+        latency_sample(std::move(sample)),
+        neighbors(std::move(nbrs)),
+        parameters(std::move(params)),
+        neighbor_keys(std::move(keys)) {}
+};
+
+void RemoteInFlightRetryTimerCallback(ValkeyModuleCtx* ctx, void* data);
```
BCathcart: I you move the definition above `CheckAndHandleRemoteInFlightConflicts` you shouldn't have to declare it here. Actually I'd suggest removing this functions entirely and just doing the retry_ctx cast in the callback along with all the logic.

src/query/inflight_retry.h
```diff
@@ -0,0 +1,59 @@
...
+  }
+  return false;
+}
+
+// Helper to schedule initial retry on main thread
+template <typename RetryContext>
+void ScheduleInFlightRetryOnMain(RetryContext* retry_ctx,
+                                 void (*timer_callback)(ValkeyModuleCtx*,
+                                                        void*)) {
+  vmsdk::RunByMain(
```
BCathcart: Just for my understanding here, is this queuing up a callback to run on the main thread and create a timer because the timer creation isn't threadsafe?

src/query/search.h
```diff
@@ -44,6 +44,9 @@ enum class SearchMode {
 
 constexpr int64_t kTimeoutMS{50000};
 constexpr size_t kMaxTimeoutMs{60000};
+
+// Retry interval for checking in-flight keys (in milliseconds)
+constexpr mstime_t kInFlightRetryIntervalMs = 3;
```
BCathcart: I suggest we make this configurable

src/query/fanout.cc
```diff
@@ -166,6 +169,52 @@ struct SearchPartitionResultsTracker {
...
+                            std::unique_ptr<SearchParameters> params,
+                            std::vector<InternedStringPtr> keys,
+                            std::shared_ptr<SearchPartitionResultsTracker> trk)
+      : neighbors(std::move(nbrs)),
+        parameters(std::move(params)),
+        neighbor_keys(std::move(keys)),
+        tracker(std::move(trk)) {}
+};
+
+void LocalInFlightRetryTimerCallback(ValkeyModuleCtx *ctx, void *data);
```
BCathcart: Same thing here

src/coordinator/server.cc
```diff
@@ -112,6 +115,85 @@ void SerializeNeighbors(SearchIndexPartitionResponse* response,
   }
 }
 
+namespace {
+
+// Context for timer-based retry when waiting for in-flight keys on remote shard
+struct RemoteInFlightRetryContext {
```
BCathcart: What do you think about moving all this retry logic for all the different cases into a separate file? I see you already have inflight_retry.h which you could build upon. I personally like the idea of keeping this feature's logic grouped together and making the touchpoints into it very clear. I was getting a bit confused jumping between commands.cc, server.cc, and fanout.cc. There are also similar patterns between these different cases so maybe you can re-use some more functionality?

src/commands/commands.cc
BCathcart: In all three cases you need to check on the main thread regardless. It could be that the background thread sees that there are no conflicting keys, and then a conflicting key update arrives in the meantime before the response logic gets a hold of the main thread. Let's make sure we're not releasing responses based solely on the background check.

What I was getting at before is wondering if we need to do the background check at all. The benefit I see is if you do a check on the background thread and don't see a conflict, we can ask the main thread to execute the response logic asap. If we do see a conflict, then we can set the timer to wait a bit since there's no point having the main thread check again immediately. This will be a bit more efficient than always executing asap on the main thread.

src/coordinator/server.cc
baswanth09: Agreed. I will try to move the re-usable parts into inflight_retry.h

src/query/inflight_retry.h
baswanth09: Yes, exactly. The codebase has a dedicated helper StartTimerFromBackgroundThread in utils.cc that wraps ValkeyModule_CreateTimer in RunByMain, confirming this pattern. Timer creation must happen on the main thread.

src/query/inflight_retry.h
```diff
@@ -0,0 +1,106 @@
...
+    return neighbor_keys_;
+  }
+
+ private:
+  std::vector<InternedStringPtr> neighbor_keys_;
+};
+
+// Process a retry attempt. Handles cancellation, conflict checking, and
+// completion. This is the main entry point called by timer callbacks.
+template <typename RetryContext>
```
BCathcart: It's great how you've condensed the number of code paths! I actually don't think you even need the template argument. You should be able to rely on polymophism:

```
void ProcessRetry(ValkeyModuleCtx* ctx, InFlightRetryContextBase* retry_ctx,
                  void (*timer_callback)(ValkeyModuleCtx*, void*),
                  const char* log_prefix);
```

Then you wouldn't need the separate `InFlightRetryCallback`, `RemoteInFlightRetryCallback`, and `LocalInFlightRetryCallback` functions. You do use them to define separate description strings used in error messages, but those can be moved into the retry classes themselves (e.g. define a virtual void `GetDesc()` in `InFlightRetryContextBase`)

src/query/inflight_retry.h
```diff
@@ -0,0 +1,106 @@
...
+  }
+  vmsdk::RunByMain(
+      [retry_ctx, timer_callback, has_conflicts]() {
+        auto ctx = vmsdk::MakeUniqueValkeyThreadSafeContext(nullptr);
+        if (has_conflicts) {
+          ValkeyModule_CreateTimer(
+              ctx.get(), options::GetInFlightRetryIntervalMs().GetValue(),
+              timer_callback, retry_ctx);
+        } else {
+          timer_callback(ctx.get(), retry_ctx);
```
BCathcart: Nice!

integration/test_fulltext_inflight_blocking.py
```diff
@@ -0,0 +1,334 @@
...
+        )
+
+        waiters.wait_for_true(
+            lambda: client.execute_command("FT._DEBUG PAUSEPOINT TEST mutation_processing") > 0,
+            timeout=5
+        )
+
+        search_thread, search_res, search_err = run_in_thread(
+            lambda: self.server.get_new_client().execute_command(
+                "FT.SEARCH", "idx", "@content:hello"
```
BCathcart, line 62: Let's validate the response is correct in these tests just in case

integration/test_fulltext_inflight_blocking.py
```diff
@@ -0,0 +1,334 @@
...
+            )
+            hset_threads.append(t)
+
+        waiters.wait_for_true(
+            lambda: all(nc.execute_command("FT._DEBUG PAUSEPOINT TEST mutation_processing") > 0
+                       for nc in remote_clients),
+        )
+
+        search_thread, search_res, search_err = run_in_thread(
+            lambda: coordinator.execute_command("FT.SEARCH", "idx", "@content:hello")
```
BCathcart, line 301: Let's validate the search response in these CME tests as well

BCathcart (approved):
Looks good! Just a couple minor comments

I'd recommend touching base with @allenss-amazon. He's been working on avoiding evaluation on the main thread when it's unnecessary by tracking if a key has been mutated since the search executed in the background. Your mechanism here is of course still needed if a key has been mutated and that mutation is in flight, but there could be minor conflicts so I'd recommend both being aware of each other's incoming changes

src/query/inflight_retry.h
```diff
@@ -0,0 +1,69 @@
...
+// Usage pattern:
+// 1. Create derived context with case-specific data
+// 2. Call ScheduleOnMainThread() to start the retry loop
+// 3. The timer callback calls ProcessRetry() which handles:
+//    - Cancellation check -> `OnCancelled()`
+//    - Conflict check -> schedule retry timer
+//    - No conflicts -> `OnComplete()`
+class InFlightRetryContextBase {
+ public:
+  explicit InFlightRetryContextBase(std::vector<InternedStringPtr> keys)
```
allenss-amazon: If your is to move rather than copy, you'll usually use "&&" in the signature.

src/query/inflight_retry.h
```diff
@@ -0,0 +1,69 @@
...
+ private:
+  std::vector<InternedStringPtr> neighbor_keys_;
+};
+
+// Timer callback for retry scheduling
+void InFlightRetryCallback(ValkeyModuleCtx* ctx, void* data);
+
+// Process a retry attempt. Handles cancellation, conflict checking, and
+// completion. This is the main entry point called by timer callbacks.
+void ProcessRetry(ValkeyModuleCtx* ctx, InFlightRetryContextBase* retry_ctx);
```
allenss-amazon: The function signature suggests it might want to be a member function of InFlightRetryContextBase, why not?

src/query/inflight_retry.h
```diff
@@ -0,0 +1,69 @@
...
+
+// Schedule the retry context to be processed on the main thread.
+// - If `has_conflicts` is true: schedules with delay (background saw conflicts)
+// - If `has_conflicts` is false: schedules immediately (background saw no
+//   conflicts, but main thread must still verify before completing)
+void ScheduleOnMainThread(InFlightRetryContextBase* retry_ctx,
```
allenss-amazon: same

src/commands/commands.cc
```diff
@@ -29,6 +39,37 @@ struct Result {
   std::unique_ptr<QueryCommand> parameters;
 };
 
+// Context for timer-based retry when waiting for in-flight keys
+struct InFlightRetryContext : public query::InFlightRetryContextBase {
+  vmsdk::BlockedClient blocked_client;
+  std::unique_ptr<Result> result;
+
+  InFlightRetryContext(vmsdk::BlockedClient bc, std::unique_ptr<Result> res,
```
allenss-amazon: For documentation purpose, I'd use a "&&" here on the parameter declaration.

src/commands/commands.cc
```diff
@@ -128,6 +169,25 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
           .neighbors = std::move(neighbors),
           .parameters = std::move(upcast_parameters),
       });
+
+      // Text predicate evaluation requires main thread to ensure text indexes
+      // reflect current keyspace. Block if result keys have in-flight
+      // mutations.
+      if (!result->parameters->no_content &&
+          query::QueryHasTextPredicate(*result->parameters)) {
```
allenss-amazon, line 185: I'd like to see a better name for QueryHasTextPredicate . In the future, we may have other data types that need this functionality. Something that's referencing the underlying conflict would be better, e.g., SharesIndexData or something like that.

allenss-amazon: Also, why is nocontent a special case? That doesn't seem right.

src/commands/commands.cc
```diff
@@ -128,6 +169,25 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
+        auto neighbor_keys =
+            query::CollectNeighborKeys(result->neighbors.value());
```
allenss-amazon: Not sure why we need to duplicate this information. Aren't we retaining the result struct too?

src/commands/commands.cc
BCathcart: For nocontent we are just returning the key names so there is no need to re-evaluate the data

BCathcart:
>3. IMO using time-based retry logic is actually more complicated than necessary and will offer inferior latencies on a conflict. What I'd recommend instead is that when you add a HashSet<shared_ptr<query::SearchParameters>> to the mutation queue entry. When you check for conflicts, you can insert yourself on the first conflict. Then the ingestion engine when it's done with a mutation queue entry checks the pointer and if non-null, calls a method on that base class to start the retry.

Good idea. Triggering it based on ingestion minimizes the latency and the chance of re-blocking.

> 4. This polling code is susceptible to live-lock. I believe that once `#3` is implemented that solving live-lock becomes relatively simple. Specifically, if a new mutation arrives and finds that there is already a waiting command (i.e., there is existing contention), then the new mutation should be deferred. Deferring the mutation would be implemented by having each mutation queue entry have a pending value (i.e, a pending value) and then simply skipping this entry until the next update cycle. If an update cycle finds the entry conflicted then it just skips it. This also requires that at the end of a conflicted query that it unblock the associated mutation queue entry, i.e., remove the pending operation pointer (actually this is a set of pending query::SearchParameters and if the set is empty, it just copies the deferred value into the main value and lets the next write period treat this as a normal non-conflict mutation.

I'm not sure I follow the logic around deferring a mutation. A mutation arrives to the search ingestion after it has been applied to the keyspace. We're going to be returning the key data from the keyspace (unless "nocontent" is used), so we would always need to wait for all relevant mutations to be ingested before evaluating.

src/commands/commands.cc
baswanth09: We retain result, but keys are stored as `Neighbor.external_id`. Collecting once avoids re-iterating on each retry

src/query/inflight_retry.h
baswanth09: Makes sense. I will update

allenss-amazon:
w.r.t. "deferring a mutation". I'm proposing one solution to the livelock problem, I'm sure there are others (for example, you could grab the mrmwmutex [for writing]) . But fundamentally, with the indexes being shared between background threads and the mainthread there will have to be some exclusion.

src/commands/commands.cc
allenss-amazon: 1. In this situation, we expect retries to be rare. So doing work in the non-retry code path to accelerate the retry code path degrades overall performance.
2. Duplicating data adds cognitive load to people that are reading the code. They end up spending a lot time searching the code with the question: "What's different about this list?", only to find out that the answer is "nothing".
3. Duplicating data is always a long-term maintenance risk.

src/commands/commands.cc
allenss-amazon: I may be out of date. Have we moved to doing full proximity scanning in the background threads? If so, they you're right that nocontent is a special case.

src/commands/commands.cc
```diff
@@ -27,6 +39,38 @@ struct Result {
...
+  InitiatorSearch(vmsdk::BlockedClient &&bc, std::unique_ptr<Result> &&res)
+      : query::SearchParameters(0, nullptr, res->parameters->db_num),
+        blocked_client(std::move(bc)),
+        result(std::move(res)) {}
+
+  const char *GetDesc() const override { return "initiator"; }
+
+  query::SearchParameters &GetParameters() override {
+    return *result->parameters;
+  }
```
allenss-amazon: Why is this needed, this class inherits from query::SearchParameters ?

src/commands/commands.cc
```diff
@@ -27,6 +39,38 @@ struct Result {
...
+
+  std::vector<indexes::Neighbor> &GetNeighbors() override {
+    return result->search_result->neighbors;
+  }
+
+  void OnComplete(std::vector<indexes::Neighbor> &neighbors) override {
+    blocked_client.SetReplyPrivateData(result.release());
+  }
```
allenss-amazon, line 66: Why pass in neighbors if it's not used??

src/coordinator/server.cc
```diff
@@ -112,6 +114,78 @@ void SerializeNeighbors(SearchIndexPartitionResponse* response,
...
+// SearchParameters subclass for remote responder (remote shard in fanout).
+// Handles in-flight retry completion by processing neighbors and sending gRPC
+// response.
+class RemoteResponderSearch : public query::SearchParameters {
+ public:
+  SearchIndexPartitionResponse* response;
+  grpc::ServerUnaryReactor* reactor;
+  std::unique_ptr<vmsdk::StopWatch> latency_sample;
+  std::vector<indexes::Neighbor> neighbors;
+  std::unique_ptr<query::SearchParameters> parameters;
```
allenss-amazon: Don't understand why this is needed if you're inheriting from query::SearchParameters

allenss-amazon (commented):
Didn't get very far in the review. I like the idea of classes for the different initiator/retry types. But I don't under the class structure that's present here. Let's talk on Monday.

src/commands/commands.cc
baswanth09: We use it in `fanout.cc`
```
 void OnComplete(std::vector<indexes::Neighbor> &neighbors) override {
    tracker->AddResults(neighbors);
    tracker->AddTotalCount(total_count);
  }
  ```

File: src/coordinator/server.cc (hunk 27 above)
Comment (baswanth09) [reply src/coordinator/server.cc]: We need these members to do the `OnComplete` method in `server.cc` (handling response on remote nodes).

Comment (allenss-amazon) [reply src/coordinator/server.cc]: You have two parameters blocks floating around. This is what I dont' understand the need for.

Comment (baswanth09) [reply src/coordinator/server.cc]: My bad. You are right. I overlooked query::SearchParameters and assumed we are taking about all the members in general. I have fixed the parameters duplication in the revision

File: src/index_schema.cc [hunk 28]
```diff
@@ -1597,6 +1599,20 @@ vmsdk::BlockedClientCategory IndexSchema::GetBlockedCategoryFromProto() const {
   }
 }
 
+bool IndexSchema::RegisterWaitingQuery(
+    const std::vector<indexes::Neighbor> &neighbors,
+    std::shared_ptr<query::InFlightRetryContext> query_ctx) {
+  absl::MutexLock lock(&mutated_records_mutex_);
```
allenss-amazon: How about an early return if tracked_mutated_records is empty?

src/commands/commands.cc
allenss-amazon: From reading later in the code, the problem is that InitiatorSearch shouldn't derive from query::SearchParameters. It fails the "is a" test for inheritance. It "contains" a query::SearchParameters (indirect through the result member).

src/query/search.h
```diff
@@ -128,11 +128,28 @@ struct SearchParameters {
...
+  // used in retry contexts).
+  virtual void OnComplete(std::vector<indexes::Neighbor>& neighbors) {}
+  virtual void OnCancelled() {}
+  virtual std::vector<indexes::Neighbor>& GetNeighbors() {
+    static std::vector<indexes::Neighbor> empty;
+    return empty;
+  }
+  // Description for debugging in-flight retry code paths.
+  virtual const char* GetDesc() const { return "base"; }
+  virtual SearchParameters& GetParameters() { return *this; }
```
allenss-amazon: Again, this should probably be pure virtual.

src/query/search.cc
```diff
@@ -720,4 +720,8 @@ absl::Status SearchAsync(std::unique_ptr<SearchParameters> parameters,
   return absl::OkStatus();
 }
 
+bool QueryHasTextPredicate(const SearchParameters &parameters) {
```
allenss-amazon: Lets' rename, What this function is really telling the caller is that it uses the per-key TextIndex so that locking is required -- right? So let's name it something like RequiresTextIndexLocking() ??

src/query/search.h
```diff
@@ -128,11 +128,28 @@ struct SearchParameters {
...
+  // In-flight retry completion callbacks. Override in derived classes to
+  // handle completion of searches that were blocked waiting for in-flight
+  // mutations. Default implementations do nothing (for basic SearchParameters
+  // used in retry contexts).
+  virtual void OnComplete(std::vector<indexes::Neighbor>& neighbors) {}
+  virtual void OnCancelled() {}
+  virtual std::vector<indexes::Neighbor>& GetNeighbors() {
+    static std::vector<indexes::Neighbor> empty;
+    return empty;
+  }
```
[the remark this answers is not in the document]
allenss-amazon: Based on what I wrote above this should be gone from here.

src/query/inflight_retry.cc
```diff
@@ -0,0 +1,62 @@
...
+#include "vmsdk/src/log.h"
+
+namespace valkey_search::query {
+
+InFlightRetryContext::InFlightRetryContext(
+    std::unique_ptr<SearchParameters> params)
+    : parameters_(std::move(params)) {}
+
+void InFlightRetryContext::ProcessRetry() {
+  auto& params = parameters_->GetParameters();
```
allenss-amazon: This makes no sense. This object has a parameters pointer, you just use it directly. There's no need for the "GetParameters" machinery that you have.

src/commands/commands.cc
```diff
@@ -27,6 +39,38 @@ struct Result {
...
+
+  void OnCancelled() override {
+    // Let Reply callback handle the cancellation
+    blocked_client.SetReplyPrivateData(result.release());
```
allenss-amazon: Shouldn't this check for PartialResults??

src/commands/commands.cc
allenss-amazon: Right, but you're not saving it.

src/coordinator/search_converter.cc
```diff
@@ -117,10 +120,11 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> GRPCPredicateToPredicate(
       VMSDK_ASSIGN_OR_RETURN(
           auto predicate,
           GRPCPredicateToPredicate(predicate.negate().predicate(), index_schema,
-                                   attribute_identifiers));
+                                   attribute_identifiers, has_text_predicate));
       return std::make_unique<query::NegatePredicate>(std::move(predicate));
     }
     case Predicate::kTerm: {
+      has_text_predicate = true;
```
KarthikSubbarao, line 127: We do not need this. Please see the QueryOperations which is also stored in search parameters

```
enum class QueryOperations : uint64_t {
  kNone = 0,
  kContainsOr = 1 << 0,
  kContainsAnd = 1 << 1,
  kContainsNumeric = 1 << 2,
  kContainsTag = 1 << 3,
  kContainsNegate = 1 << 4,
  kContainsText = 1 << 5,
  kContainsProximity = 1 << 6,
  kContainsNestedComposed = 1 << 7,
  kContainsTextTerm = 1 << 8,
  kContainsTextPrefix = 1 << 9,
  kContainsTextSuffix = 1 << 10,
  kContainsTextFuzzy = 1 << 11,
};
```

baswanth09: Got it. I believe these were added after the creation of this PR. I will fix it in a follow PR.

baswanth09: #734

PR #543, opened 2025-12-12
Remove double locking in prefilter locking

daddaman-amz:
Remove redundant lock in PrefilterEvaluator::EvaluateText()

PrefilterEvaluator::EvaluateText() was acquiring per_key_text_indexes_mutex_, but DoSearch() (3 levels higher in the call stack) already holds a time-sliced reader lock protecting this data. This double-locking caused performance degradation.

Added GetPerKeyTextIndexes() direct accessor and updated prefilter evaluation to use it instead of WithPerKeyTextIndexes().

### DIFF
--- a/src/indexes/vector_base.cc
+++ b/src/indexes/vector_base.cc
@@ -106,19 +106,15 @@ query::EvaluationResult PrefilterEvaluator::EvaluateText(
     const query::TextPredicate &predicate, bool require_positions) {
   CHECK(key_);
   // Evaluate using per-key text index
-  // This acquires the lock and looks up the key in per_key_text_indexes
   auto text_index_schema = predicate.GetTextIndexSchema();
-  return text_index_schema->WithPerKeyTextIndexes(
-      [&](auto &per_key_indexes) -> query::EvaluationResult {
-        auto it = per_key_indexes.find(*key_);
-        if (it == per_key_indexes.end()) {
-          VMSDK_LOG(WARNING, nullptr)
-              << "Target key not found in index for pre-filter evaluation";
-          return query::EvaluationResult(false);
-        }
-        // Evaluate predicate against this key's text index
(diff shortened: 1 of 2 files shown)


src/indexes/vector_base.cc
```diff
@@ -106,19 +106,15 @@ query::EvaluationResult PrefilterEvaluator::EvaluateText(
...
-        auto it = per_key_indexes.find(*key_);
-        if (it == per_key_indexes.end()) {
-          VMSDK_LOG(WARNING, nullptr)
-              << "Target key not found in index for pre-filter evaluation";
-          return query::EvaluationResult(false);
-        }
-        // Evaluate predicate against this key's text index
-        return predicate.Evaluate(it->second, *key_, require_positions);
-      });
+  auto &per_key_indexes = text_index_schema->GetPerKeyTextIndexes();
```
KarthikSubbarao, line 110: Just curious about the differences in the flows for the PredicateEvaluator and PrefilterEvaluator.

The per-key-index was obtained one time in the VerifyFilter (PredicateEvaluator flow) before getting into the evaluate methods.

How come we cannot have a similar flow in the Prefilter case?

daddaman-amz: In predicate evaluator,  VerifyFilter() is the place where the evaluator is intialized with data source which is the recordsmap for non-text. So I used the same place to set the source for text. Since we need to acquire the lock via WithPerKeyTextIndexes() anyway, it made sense to acquire it once at this higher level and pass the reference down to the evaluator, avoiding multiple lock acquisitions during evaluation.

For prefilter I followed exisitng pattern where each index was handled separately in their own evaluate methods. Earlier when there was lock I should have accounted for same problem of multiple lock acquisition and acquired at higher level, but now since we removed the lock anyway , I think its okay to keep it as is.

KarthikSubbarao (approved):

PR #544, opened 2025-12-12
Add reserve in vector population in text parsing for default cases + AND Intersection wrapping cases

KarthikSubbarao:
Add reserve in vector population in text parsing for default cases + AND Intersection wrapping cases.

In perf tests with default field searches with an index with 50 text fields, `SetupTextFieldConfiguration` was becoming a bottleneck in parsing without this vector reserve

### DIFF
--- a/src/commands/filter_parser.cc
+++ b/src/commands/filter_parser.cc
@@ -468,12 +468,13 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::WrapPredicate(
       not_rightmost_bracket &&
       new_predicate->GetType() == query::PredicateType::kComposedOr) {
     std::vector<std::unique_ptr<query::Predicate>> new_children;
-    if (prev_predicate) {
-      new_children.push_back(std::move(prev_predicate));
-    }
     auto* new_composed =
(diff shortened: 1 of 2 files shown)


src/commands/filter_parser.cc
```diff
@@ -468,12 +468,13 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::WrapPredicate(
       not_rightmost_bracket &&
       new_predicate->GetType() == query::PredicateType::kComposedOr) {
     std::vector<std::unique_ptr<query::Predicate>> new_children;
-    if (prev_predicate) {
-      new_children.push_back(std::move(prev_predicate));
-    }
     auto* new_composed =
         dynamic_cast<query::ComposedPredicate*>(new_predicate.get());
     auto children = new_composed->ReleaseChildren();
+    new_children.reserve(1 + children.size());
```
daddaman-amz, line 474: any merit in checking   if prev_predicate exist or not before reserving + 1?

KarthikSubbarao: Good question, but probably not too much of a gain. We see issues when there are too many expansions due to not reserving enough
In the perf tests, the identifiers vector was expanding a lot due to large number of text fields

PR #545, opened 2025-12-14
Eliminate redundant mainthread filter invokations

allenss-amazon:
Tracks sequence number for each key. Two maps<Key, SequenceNumber> one for the DB and one for the Index. See the design doc note.

Basically one map tracks the Key DB, the other map tracks the indexes. When a query is completed, the value of the index map is saved for each of the keys that are part of the result. Then back in the mainthread the KeyDB is checked for each of the resulting keys. If the two sequence numbers are equal, then the predicate filter is skipped, otherwise it's executed.


allenss-amazon:
Karthik grabbed this an put it into one of his PRs which is now merged.

PR #547, opened 2025-12-15
Add getter in config for the default value

yanamolo:
Implemented default value getter in configuration infrastructure and added String config default.

### DIFF
--- a/vmsdk/src/module_config.cc
+++ b/vmsdk/src/module_config.cc
@@ -324,8 +324,10 @@ absl::Status String::Register(ValkeyModuleCtx *ctx) {
 }
 
 absl::Status String::FromString(std::string_view value) {
-  default_ = value.data();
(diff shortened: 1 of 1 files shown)


vmsdk/src/module_config.cc
```diff
@@ -324,7 +324,8 @@ absl::Status String::Register(ValkeyModuleCtx *ctx) {
 }
 
 absl::Status String::FromString(std::string_view value) {
-  SetValueOrLog(value.data(), WARNING);
+  default_ = value.data();
```
eifrah-aws: if `value.data()` is `nullptr` - it will crash.

eifrah-aws (changes requested):

vmsdk/src/module_config.cc
yanamolo: Will add a check for `value.data()`

PR #553, opened 2025-12-18
Shard the memory tracking lock

allenss-amazon:


vmsdk/src/memory_allocation_overrides.cc
```diff
@@ -77,10 +80,21 @@ class SystemAllocTracker {
...
 
-  absl::Mutex mutex_;
-  absl::flat_hash_set<void*, absl::Hash<void*>, std::equal_to<void*>,
+  struct Shard {
+    absl::Mutex mutex_;
+    absl::flat_hash_set<void*, absl::Hash<void*>, std::equal_to<void*>,
                       RawSystemAllocator<void*>>
       tracked_ptrs_ ABSL_GUARDED_BY(mutex_);
+  };
+  static const size_t kShardBits = 7; // 128 Shards
```
KarthikSubbarao, line 89: 128 is good enough, I'd imagine

vmsdk/src/memory_allocation_overrides.cc
```diff
@@ -77,10 +80,21 @@ class SystemAllocTracker {
...
+  static const size_t kShardCount = 1 << kShardBits;
+  static const size_t kShardMask = (kShardCount - 1) << 4; // Since all allocations are 16 byte aligned
```
zvi-code, line 91: @allenss-amazon , zmalloc is only 8 bytes aligned not 16, are these (eventually) zmalloc allocations?

allenss-amazon: Well, it doesn't matter in this case, I'm using some bits for hash to select a lock, as long as those bits vary decently it doesn't matter.

PR #561, opened 2025-12-25
Endurance test

Nivesh-01:
This PR is for endurance testing. It uses Memtier Benchmark for testing. Added support and tests for Text search as well.


daddaman-amz:
I think this PR needs to be rebased again with latest fulltext as the history was corrected and force pushed by Brennan recently.    Right now it shows 136 file changes

testing/integration/utils.py
```diff
@@ -155,7 +155,8 @@ def start_valkey_cluster(
 
     for port in ports:
         stdout_path = os.path.join(stdout_directory, f"{port}_stdout.txt")
-        stdout_file = open(stdout_path, "w")
+        # Open file without buffering to avoid ResourceWarnings
```
daddaman-amz, line 158: comment says "without buffering" but buffering is enabled.   

Also is that responsible for removing `ResourceWarnings` or is it the file pattern you used below?

testing/integration/utils.py
```diff
@@ -738,6 +761,8 @@ def periodic_flushdb_task(
         logging.info("<FLUSHDB> Invoking flush DB")
         try:
             client.flushdb()
+            # Wait for cluster to stabilize after FLUSHDB
+            time.sleep(2)
```
daddaman-amz, line 765: nit: was this needed after adding full text tests or its needed for all tests in general?

Nivesh-01: fulltext

daddaman-amz (commented):
.

PR #562, opened 2025-12-27
Add text negation querying with per-field tracked/untracked keys

VoletiRam:
Add text negation querying with per-field tracked/untracked keys. Aggregate tracked/untracked keys across relevant fields and remove matched term of negation ((tracked_keys - matched_keys) + untracked_keys = negation result)

### DIFF
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -12,8 +12,10 @@
 #include "absl/container/inlined_vector.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/string_view.h"
+#include "src/index_schema.h"
 #include "src/index_schema.pb.h"
 #include "src/indexes/text/lexer.h"
+#include "src/indexes/text/negation_iterator.h"
 
 namespace valkey_search::indexes {
 
@@ -34,11 +36,25 @@ Text::Text(const data_model::TextIndex& text_index_proto,
 
 absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr& key,
                                      absl::string_view data) {
-  // TODO: Key Tracking
+  auto result = text_index_schema_->StageAttributeData(
+      key, data, text_field_number_, !no_stem_, min_stem_size_,
+      with_suffix_trie_);
 
-  return text_index_schema_->StageAttributeData(key, data, text_field_number_,
-                                                !no_stem_, min_stem_size_,
-                                                with_suffix_trie_);
+  if (!result.ok()) {
+    return result;
+  }
+
+  absl::MutexLock lock(&index_mutex_);
+  if (result.value()) {
+    // Has indexable content
+    tracked_keys_.insert(key);
+    untracked_keys_.erase(key);
+  } else {
+    // No indexable content
+    untracked_keys_.insert(key);
+  }
+
+  return result;
 }
 
 absl::StatusOr<bool> Text::RemoveRecord(const InternedStringPtr& key,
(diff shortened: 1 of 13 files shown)


src/indexes/text.cc
```diff
@@ -122,17 +193,49 @@ std::unique_ptr<EntriesFetcherIteratorBase> Text::EntriesFetcher::Begin() {
...
-void* TextPredicate::Search(bool negate) const {
+void* TextPredicate::Search(bool negate,
+                            const IndexSchema* index_schema) const {
   size_t estimated_size = EstimateSize();
+
+  // Get aggregated tracked/untracked keys for negation support
+  const InternedStringSet* tracked_keys_ptr = nullptr;
+  const InternedStringSet* untracked_keys_ptr = nullptr;
+  std::unique_ptr<InternedStringSet> aggregated_tracked;
+  std::unique_ptr<InternedStringSet> aggregated_untracked;
```
BCathcart, line 204: Building new sets by looping over all the tracked and untracked keys for every predicate isn't going to lead to the performance we want. We should think in terms of iterating over the existing structures.

Also won't there be some overlap between the tracked and untracked keys. For example, a key is tracked for one text field but untracked for another. If there is a text search across all text fields, isn't it possible that the same key shows up in the normal and negated version of the search?

I think this comes back to the benefits of simplifying the equation from `not-matching keys + untracked keys = negation result` to `total keys - matching keys = negation result` where total keys is tracked at the schema level. We're already finding the keys on the non-negated query rather than searching directly for the non-matching keys, so might as well subtract them from the total keys rather than the matching keys at the attribute level.

We probably eventually want to move to this pattern for all attribute types, but if we want to just focus on text in the meantime, we already have a list of tracked keys at the schema level for just text fields in the form of the per-key tree map. Might be easiest to just add untracked keys to the TextIndexSchem so we have the total keys at the schema-level that we can for use every negated text query and not worry about tracking on a per-text attribute level.

VoletiRam: Thanks for suggesting this. I agree about performance and I came to this realization while implementing the current per-field negation but I didn't get answer on value of keeping the current per-field tracking and also you said Allen had plans to change this future so I continued with per-field aggregation logic.

>  For example, a key is tracked for one text field but untracked for another. If there is a text search across all text fields, isn't it possible that the same key shows up in the normal and negated version of the search?

Aah missed it. We should somehow do term match even across all untracked keys to eliminate it which would further increase complexity. 


To keep it simple while having performance gain we expect, it is better to switch to schema level for negation.

allenss-amazon: While I do have plans to change the handling of tracked/untracked keys. Don't assume that this will get done by this release.

VoletiRam: Understood. Will switch to schema-level.

VoletiRam: Created the PR for negation support with schema level tracked/untracked keys here: https://github.com/valkey-io/valkey-search/pull/577

Will close this PR as it will be clean to have the schema based negation in a new PR for the review.

BCathcart: Tracked/untracked has come up again. One thing to note is we have a contract to provide the number of tracked keys per attribute. Tracking only at the schema level is fine for the negation logic, but then how do we maintain the number of tracked keys per attribute?

BCathcart: We could store a field mask with each tracked key at the schema level and use it to increment/decrement per-attribute counts

PR #564, opened 2025-12-28
Endurance failover

Nivesh-01:
Endurance testing only for failover

### DIFF
--- a/testing/integration/utils.py
+++ b/testing/integration/utils.py
@@ -859,22 +859,26 @@ class MemtierProcess:
 
     def process_logs(self):
         for line in self._process_memtier_subprocess_output():
+            is_acceptable_error = False
             if line.error is not None:
-                # If error_predicate is provided and returns True, this is an acceptable error - skip it
+                # If error_predicate is provided and returns True, this is an acceptable error
                 if self.error_predicate is not None and self.error_predicate(line.error):
                     logging.debug(
                         "<%s> encountered expected error (ignored): %s", self.name, line.error
                     )
-                    continue
-                # This is an unexpected error - log it
-                logging.error(
-                    "<%s> encountered error: %s", self.name, line.error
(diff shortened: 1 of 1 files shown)


testing/integration/stability_test.py
```diff
@@ -197,6 +203,8 @@ def tearDown(self):
                 use_coordinator=False,
                 replica_count=1,
                 repl_diskless_load="disabled",
+                failover_interval_sec=45,
```
daddaman-amz, line 206: Interval is set at 45 sec so for a 60 second test we will see one failover. I think it's also worth writing a dedicated test which: 

1. Fills cluster to significant load (say 100K or 1M keys)
2. Performs multiple failovers with longer time periods (e.g., every 2-3 minutes over 15+ minutes)
3. Validates data consistency and search accuracy throughout

This need not run as part of CI for now if we think it can slow down CI significantly.

testing/integration/stability_test.py
```diff
@@ -289,6 +311,24 @@ def test_valkeyquery_stability(self, config):
             # BGSAVE will fail if another is ongoing.
             if result.name == "BGSAVE":
                 pass
+            # During failover testing, some background tasks may fail transiently
```
daddaman-amz, line 314: I noticed these tests are not part of the CI run. 
https://github.com/valkey-io/valkey-search/actions/runs/20556797899/job/59042109155?pr=564

Can we see the results of these tests somewhere after your changes?

Nivesh-01: Yes these tests are currently not run in github because it needs additional packages to run. I am not sure if this is the reason why they were originally not added in the test runs but I am guessing this is the reason. Moreover these tests do take a very long time to run upward of 25 mins so for now to see the test results you will need to set everything on your machine and then run the tests there.

PR #570, opened 2025-12-30
Rax Integration

BCathcart:
TODO

### DIFF
--- a/src/indexes/text/rax/rax_malloc.h
+++ b/src/indexes/text/rax/rax_malloc.h
@@ -1,8 +1,31 @@
-/*
- * Copyright (c) 2025, valkey-search contributors
- * All rights reserved.
- * SPDX-License-Identifier: BSD 3-Clause
+/* Rax -- A radix tree implementation.
  *
+ * Copyright (c) 2017, Redis Ltd.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the name of Redis nor the names of its contributors may be used
+ *     to endorse or promote products derived from this software without
+ *     specific prior written permission.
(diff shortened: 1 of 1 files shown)


src/indexes/text/rax/rax_malloc.h
```diff
@@ -1,45 +1,23 @@
-/* Rax -- A radix tree implementation.
- *
- * Copyright (c) 2017, Redis Ltd.
```
eifrah-aws: You can't really delete license of the original author - I am pretty sure it is not legal.

src/indexes/text/rax/rax_malloc.h
```diff
@@ -1,45 +1,23 @@
...
+/*
+ * Copyright (c) 2025, valkey-search contributors
  * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
  *
- * Redistribution and use in source and binary forms, with or without
```
eifrah-aws: Same, do not change the original license.

src/indexes/text/rax/rax_malloc.h
BCathcart: Ok good to know. I'll definitely keep the original header but do you think we also need to specify that the Valkey license applies to the changes/additions we're making or is it fine to keep everything under the Redis license?

BCathcart:
> The correct way of doing this (IMO) is to create a PR in valkey that exposes new module API for RAX and then use it here.

That was what I was thinking for the long term, but in the short term I'm going to be adding APIs and playing around with the internals a bit. Having the files here should allow us to move faster and avoids exposing the core valkey components to these changes. I plan to merge the code with sub-optimal changes at first to unblock things and then iterate on it.

src/indexes/text/rax/rax_malloc.h
eifrah-aws: Since Redis license (in their pre-valkey fork) is permissive, I believe that we should only retain the original copyright notice ("Copyright (c) 2017, Redis Ltd.") and you can apply your own license (VSS license)

allenss-amazon: We should start with the 7.2 files as is (including the license). Then modify from there to suit our purposes.

While I agree that exposing the rax data type as module APIs is the best way to handle this. We're not operating a lock-step mode with the core release cycles. So I think that copying the files is fine for now and that when rax module APIs become available in some future version of the core, we can remove the files and switch to using that interface.

PR #571, opened 2025-12-30
Handle escape characters in ingestion

VoletiRam:
Handle escape characters in ingestion. Fix the issue where we are handling backslash as punctuation character and tokenize the words with escape characters. Special handle `\\` for escape characters in tokenization.

### DIFF
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -199,7 +199,7 @@ def validate_fulltext_search(client: Valkey):
 class TestFullText(ValkeySearchTestCaseDebugMode):
 
     def test_escape_sequences(self):
-        """Escape sequences preserved literally during indexing, matching Redis."""
+        """Escape sequences are stripped during indexing, matching Redis."""
         client: Valkey = self.server.get_new_client()
         client.execute_command("FT.CREATE", "idx", "ON", "HASH", "SCHEMA", "content", "TEXT", "NOSTEM")
         
@@ -212,8 +212,9 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert client.execute_command("FT.SEARCH", "idx", "@content:line1")[0] == 0
         assert client.execute_command("FT.SEARCH", "idx", "@content:tab")[0] == 0
 
-        assert client.execute_command("FT.SEARCH", "idx", r'@content:line1\\nline2')[0] == 1
-        assert client.execute_command("FT.SEARCH", "idx", r'@content:tab\\there')[0] == 1
+        # Backslashes are stripped during indexing, so search without them
+        assert client.execute_command("FT.SEARCH", "idx", "@content:line1nline2")[0] == 1
+        assert client.execute_command("FT.SEARCH", "idx", "@content:tabthere")[0] == 1
 
     @pytest.mark.parametrize("prefilter_eval_enabled", [pytest.param(False, id="Prefilter_eval=False"), pytest.param(True, id="Prefilter_eval=True")])
     def test_text_search(self, prefilter_eval_enabled):
--- a/src/indexes/text/lexer.cc
+++ b/src/indexes/text/lexer.cc
@@ -108,9 +108,9 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
     // Build word, handling backslash escape sequences
     while (pos < text.size()) {
       if (text[pos] == '\\' && pos + 1 < text.size()) {
-        // Escape sequence - include both backslash and next character
-        word_buffer.push_back(text[pos++]);
-        word_buffer.push_back(text[pos++]);
+        // Escape sequence - strip backslash, keep only next character
+        pos++;                               // Skip the backslash
+        word_buffer.push_back(text[pos++]);  // Keep the escaped character
       } else if (IsPunctuation(text[pos])) {
         // Regular punctuation - end of word
         break;


src/indexes/text/lexer.cc
```diff
@@ -91,19 +91,37 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
...
-    while (pos < text.size() && !IsPunctuation(text[pos])) {
-      pos++;
+    std::string word_buffer;
+
+    // Build word, handling backslash escape sequences
+    while (pos < text.size()) {
+      if (text[pos] == '\\' && pos + 1 < text.size()) {
+        // Escape sequence - include both backslash and next character
+        word_buffer.push_back(text[pos++]);
+        word_buffer.push_back(text[pos++]);
```
BCathcart: Is the purpose of the backslash to escape punctuation characters? If so, should we check if the next character is a punctuation character and if it is, drop the backslash and include the punctuation character in the string. If it's not a punctuation character, then we just treat the backslash normally.

integration/test_fulltext.py
```diff
@@ -198,6 +198,23 @@ def validate_fulltext_search(client: Valkey):
...
+        client.execute_command("HSET", "doc:1", "content", r"line1\nline2")
+        client.execute_command("HSET", "doc:2", "content", r"tab\there")
+        
+        IndexingTestHelper.wait_for_backfill_complete_on_node(client, "idx")
+        
+        # Escape sequences create unified tokens
+        assert client.execute_command("FT.SEARCH", "idx", "@content:line1")[0] == 0
+        assert client.execute_command("FT.SEARCH", "idx", "@content:tab")[0] == 0
+
+        assert client.execute_command("FT.SEARCH", "idx", r'@content:line1\\nline2')[0] == 1
```
BCathcart: Why are we interpreting query input differently than ingestion input? I assume '\\' is a punctuation character, so in the query case it needs to be escaped so that it is treated as part of the word. If '\\' is also valid punctuation in the ingestion case, then I'd assume it would be treated as such unless it is escaped.

VoletiRam: Thanks for catching this. I just came to know redis skips the backlash and keeps the escape character for ingestion. Will update the test to match the behavior

src/indexes/text/lexer.cc
VoletiRam: No. Previously, we used to tokenize the `line1\nline2` into two words `line1` and `line2`. With this change, we are matching with redis which is:
```
- line1\nline2  → 1 term (backslash preserved)
- test\xvalue   → 1 term (even with non-escape char)
- test\\value   → 1 term (double backslash)
- test"value    → 2 terms (quote IS punctuation)
```

BCathcart: Based on your new change, we now simply keep whichever character is following a backslash regardless if it is punctuation or not:
```
        // Escape sequence - strip backslash, keep only next character
        pos++;                               // Skip the backslash
        word_buffer.push_back(text[pos++]);  // Keep the escaped character
```

In C, the escape sequence needs to have some meaning or else it throws an error. I guess this preserves the ability to add new escape sequences in the future without breaking backwards compatibility. In my mind, I'm breaking the escape sequences into two categories:
1. There is a syntactically significant character that we want to treat like a normal character (e.g. quotations - \\")
2. We want the string to be interpreted in a way that is a pain or impossible to express otherwise (e.g. newline - \\n, alert - \\a)

In our case, I don't know if the latter is useful in any way for us. For the former, punctuation characters are the only syntactically significant characters right now and a backslash followed by a punctuation character does have meaning: treat the punctuation character as a regular character. The punctuation characters can be user defined though, so we can't reserve special escape sequences to have other meaning. Always treating whatever comes after a backslash as a normal character regardless if it is a normal character or punctuation seems like a logical simplification. The one caveat though about using backslashes to signal this, is that backslashes themselves now have syntactic significance outside of being punctuation. Does that mean '\\' can't ever be a punctuation character itself? If the user tries to make '\\' a punctuation character, then it is either treated like punctuation and the user can't escape other punctuation characters (what we have today) or it's treated as the start of an escape sequence and won't ever be treated as punctuation (what we have currently in this PR).

src/indexes/text/lexer.cc
```diff
@@ -91,19 +91,37 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
...
     }
 
     size_t word_start = pos;
-    while (pos < text.size() && !IsPunctuation(text[pos])) {
-      pos++;
+    std::string word_buffer;
+
+    // Build word, handling backslash escape sequences
+    while (pos < text.size()) {
+      if (text[pos] == '\\' && pos + 1 < text.size()) {
```
KarthikSubbarao, line 110: Can we test the following?
1. Single Backslash character with punctuation next to it
2. Single Backslash character with normal character next to it
3. Multiple backslash characters in a text field


I am specifically curious to understand what characters are retained in the ingested tokens and where the delimitation takes place

src/indexes/text/lexer.cc
```diff
@@ -91,19 +91,37 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
...
+        // Escape sequence - include both backslash and next character
```
KarthikSubbarao: We don't include the escape char when the next char is punctuation

src/indexes/text/lexer.cc
KarthikSubbarao: I confirmed... The important behavior we need to implement is that when we are able to escape a punctuation character, we should ensure that we do not break into a new token. Skipping the backslash is not sufficient

src/indexes/text/lexer.cc
BCathcart: >All punctuation marks and whitespace (besides underscores) separate the document and queries into tokens. For example, any character of ,.<>{}[]"':;!@#$%^&*()-+=~ will break the text into terms, so the text foo-bar.baz...bag will be tokenized into [foo, bar, baz, bag]

- https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/escaping/

Redis doesn't allow punctuation to be customized IIRC and doesn't include the backslash as punctuation. Maybe we should disallow backslash as punctuation or consider making it fixed as well.

src/indexes/text/lexer.cc
BCathcart: Agreed this is the behaviour we want. It looks like we keep the escaped character and continue in the while loop, so I thought this is what was happening. In any case we should have a test if we don't already.

allenss-amazon:
We should also add to the compatibility test suite for these cases.

src/indexes/text/lexer.cc
VoletiRam: Here is the behavior of redis search during the ingestion with various backslash count:
backslash + alphabet:
Single backslash:
```
10.189.160.64:6380> HSET doc1 content 'hello\world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 1
15) num_records
16) (integer) 1
```
Double backslash:
```
10.189.160.64:6380> HSET doc1 content 'hello\\world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 1
15) num_records
16) (integer) 1
```
three backslash:
```
10.189.160.64:6380> HSET doc1 content 'hello\\\world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 1
15) num_records
16) (integer) 1
```
Four backslash
```
10.189.160.64:6380> HSET doc1 content 'hello\\\\world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 1
15) num_records
16) (integer) 1
```
Five backslash
```
10.189.160.64:6380> HSET doc1 content 'hello\\\\\world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 1
15) num_records
16) (integer) 1
```
six backslash
```
10.189.160.64:6380> HSET doc1 content 'hello\\\\\\world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 1
15) num_records
16) (integer) 1
```

No tokenization happens during the ingestion irrespective of the backslash count if it is followed by an alphabet. It is clear Backlash is not an Punctuation.

Now backslash + punctuation:

Single backslash and punctuation:
```
10.189.160.64:6380> HSET doc1 content 'hello\,world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 1
15) num_records
16) (integer) 1
```
double backslash and punctuation:
```
10.189.160.64:6380> HSET doc1 content 'hello\\,world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 2
15) num_records
16) (integer) 2
```
two backslash and punctuation:
```
10.189.160.64:6380> HSET doc1 content 'hello\\,world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 2
15) num_records
16) (integer) 2
```
three backslashes and punctuation:
```
10.189.160.64:6380> HSET doc1 content 'hello\\\,world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 1
15) num_records
16) (integer) 1
```
four backslash and punctuation:
```
10.189.160.64:6380> HSET doc1 content 'hello\\\\,world'
(integer) 1
10.189.160.64:6380> ft.info idx
13) num_terms
14) (integer) 2
15) num_records
16) (integer) 2
```
From the above, we can say whatever follows after blackslash, if it is a punctuation character, we retain it and not tokenize. If it is not (another backslash), we don't know if they retain backslash or discard or retain only single backslash and discard rest of the count.

Sharing query side backslash parsing with ft.explaincli with various backslash:
```
10.189.160.64:6380> ft.explaincli idx 'hello\world'
 1) INTERSECT {
 2)   UNION {
 3)     hello
 4)     +hello(expanded)
 5)   }
 6)   UNION {
 7)     world
 8)     +world(expanded)
 9)   }
10) }
11) 
10.189.160.64:6380> ft.explaincli idx 'hello\\world'
1) UNION {
2)   hello\world
3)   +hello\world(expanded)
4) }
5) 
10.189.160.64:6380> ft.explaincli idx 'hello\\\world'
 1) INTERSECT {
 2)   UNION {
 3)     hello\
 4)     +hello\(expanded)
 5)   }
 6)   UNION {
 7)     world
 8)     +world(expanded)
 9)   }
10) }
11) 
10.189.160.64:6380> ft.explaincli idx 'hello\\\\world'
1) UNION {
2)   hello\world
3)   +hello\world(expanded)
4) }
5) 
10.189.160.64:6380> ft.explaincli idx 'hello\\\\\world'
 1) INTERSECT {
 2)   UNION {
 3)     hello\
 4)     +hello\(expanded)
 5)   }
 6)   UNION {
 7)     world
 8)     +world(expanded)
 9)   }
10) }
11) 
10.189.160.64:6380> ft.explaincli idx 'hello\\\\\\world'
1) UNION {
2)   hello\world
3)   +hello\world(expanded)
4) }
5) 
10.189.160.64:6380> 
```
Important thing is, they consider backslash as delimiter in case of querying and splitting into tokens if it is followed by an alphabet. This is changing if there are multiple backslashes and they happen to split if there are odd number of backslash and only retain one backslash if the count of backslash is greater than or equal to 2.


now backslash + punctuation character in querying:
```
10.189.160.64:6380> ft.explaincli idx 'hello\,world'
1) UNION {
2)   hello,world
3)   +hello,world(expanded)
4) }
5) 
10.189.160.64:6380> ft.explaincli idx 'hello\\,world'
 1) INTERSECT {
 2)   UNION {
 3)     hello\
 4)     +hello\(expanded)
 5)   }
 6)   UNION {
 7)     world
 8)     +world(expanded)
 9)   }
10) }
11) 
10.189.160.64:6380> ft.explaincli idx 'hello\\\,world'
1) UNION {
2)   hello,world
3)   +hello,world(expanded)
4) }
5) 
10.189.160.64:6380> ft.explaincli idx 'hello\\\\,world'
 1) INTERSECT {
 2)   UNION {
 3)     hello\
 4)     +hello\(expanded)
 5)   }
 6)   UNION {
 7)     world
 8)     +world(expanded)
 9)   }
10) }
11) 
10.189.160.64:6380> ft.explaincli idx 'hello\\\\\,world'
1) UNION {
2)   hello,world
3)   +hello,world(expanded)
4) }
5) 
10.189.160.64:6380> ft.explaincli idx 'hello\\\\\\,world'
 1) INTERSECT {
 2)   UNION {
 3)     hello\
 4)     +hello\(expanded)
 5)   }
 6)   UNION {
 7)     world
 8)     +world(expanded)
 9)   }
10) }
```
Here, we can see, punctuation is retained after backslash and not split and we can retain upto one backslash if there are multiple backslashes. Similarly, the odd count of backslashes decides if the punctuation can be retained. With this, we could assume, they might retain upto one backslash during ingestion if there are multiple backslash(more than or equal to 2)

 
 
 The conflicting behavior here with our feature to include backslash as punctuation (custom or default):

1.  Backslash is not a delimiter during ingestion irrespective of count. (we don't know if they discard or retain one or leave as is)
2. Backslash is a delimiter during querying if there is an alphabet follows (splits into tokens) on odd count. One Backslash is retained with multiple backslashes in querying.
3. Backslash is not a delimiter if there is a punctuation character follows on odd count. One Backslash is retained with multiple backslashes in querying.

Gives us two options:
    a. We should remove backslash from our default punctuation list and not allow it for custom punctuation. We follow same behavior silently where we use backslash as delimiter or not based on if a punctuation character follows and retain upto 1 backslash on multiple backslashes.
    b.  We keep backslash in punctuation list as well as use as escape separator. Use first backslash as escape separator if it followed by punctuation. Else it will be considered as regular punctuation character. Every second character on multiple backslash will act as punctuation.

PR #573, opened 2025-12-30
Backport fixes into 1.1

allenss-amazon:


Aksha1812:
What changes are we backporting here ?  FYI : valkey-search would be using valkey-ci-agent for backporting changes . Actively working with Sarthak on it. After that we would just need to PRs to respective project boards with status to be backported and agent will auto raise backport PR. That said we should close all other backporting changes for now if that sounds fine . If the changes here are already added to the project board we can close this

allenss-amazon:
abandoned in preference to CI-agent

PR #574, opened 2026-01-01
Better error handling in thread CPU calculations

yanamolo:
Fix thread monitoring CPU percentage calculation and add negative elapsed time handling

- Refactor GetThreadCPUPercentage() to handle negative CPU elapsed time gracefully
- Simplify function logic by removing redundant error conditions and early updates
- Add comprehensive test coverage with mocked clock_gettime using linker wrapping
- Fix error handling in valkey_search.cc CPU percentage fields to log errors instead of using .value_or()

The negative elapsed time condition can occur when system clock adjustments happen between measurements, and should return 0.0 rather than throwing errors.

### DIFF
--- a/vmsdk/src/thread_monitoring.cc
+++ b/vmsdk/src/thread_monitoring.cc
@@ -21,8 +21,6 @@
 
 namespace vmsdk {
 
-static std::atomic_uint64_t thread_monitoring_cpu_error_cnt{0};
-
 #ifdef __APPLE__
 namespace {
 thread_inspect_t ConvertToMachThread(pthread_t tid) {
@@ -35,6 +33,14 @@ ThreadMonitor::ThreadInfoFunc ThreadMonitor::thread_info_func = thread_info;
 #endif
 
 ThreadMonitor::ThreadMonitor(pthread_t thread_id) { thread_id_ = thread_id; }
+
+std::string ThreadMonitor::ThreadIdToString() const {
+#ifdef __APPLE__
+  return absl::StrFormat("%p", thread_id_);
+#else
+  return absl::StrFormat("%u", static_cast<unsigned int>(thread_id_));
+#endif
+}
 
 absl::StatusOr<double> ThreadMonitor::GetThreadCPUPercentage() {
   // First call, initializing values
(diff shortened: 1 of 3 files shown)


src/valkey_search.cc
```diff
@@ -126,14 +126,32 @@ static vmsdk::info_field::Float used_read_cpu(
     "thread-pool", "used_read_cpu",
     vmsdk::info_field::FloatBuilder().App().Computed([]() -> double {
       auto reader_thread_pool = ValkeySearch::Instance().GetReaderThreadPool();
-      return reader_thread_pool->GetAvgCPUPercentage().value_or(-1);
+      auto result = reader_thread_pool->GetAvgCPUPercentage();
+      if (!result.ok()) {
+        VMSDK_LOG(WARNING, nullptr) << "Failed to get read CPU percentage: "
+                                    << result.status().message();
+        return 0.0;
```
zvi-code, line 134: Is there a risk we will flood the log? Given we have metric I suggest to limit the number of logs can be emitted.

src/valkey_search.cc
```diff
@@ -126,14 +126,32 @@ static vmsdk::info_field::Float used_read_cpu(
     "thread-pool", "used_read_cpu",
     vmsdk::info_field::FloatBuilder().App().Computed([]() -> double {
       auto reader_thread_pool = ValkeySearch::Instance().GetReaderThreadPool();
-      return reader_thread_pool->GetAvgCPUPercentage().value_or(-1);
+      auto result = reader_thread_pool->GetAvgCPUPercentage();
+      if (!result.ok()) {
+        VMSDK_LOG(WARNING, nullptr) << "Failed to get read CPU percentage: "
```
zvi-code: Worth adding the thread-id and additional data that would help understand the case where this occured

vmsdk/src/thread_monitoring.cc
```diff
@@ -46,23 +52,30 @@ absl::StatusOr<double> ThreadMonitor::GetThreadCPUPercentage() {
...
-  if (wall_time_elapsed > 0) {
-    return (static_cast<double>(cpu_time_elapsed_us) / wall_time_elapsed) *
-           100.0;
+  if (wall_time_elapsed == 0) {
+    return 0.0;
+  }
+
+  if (cpu_time_elapsed_us < 0) {
+    // Increment the monitoring counter for negative values
+    thread_monitoring_cpu_error_cnt.fetch_add(1, std::memory_order_relaxed);
```
zvi-code: Why is this incremented internally and not in the upper level? Is this because we don't have the value? should we return the value? [have this function be a utility and use explicit api to update state?] - My concern is that metric is not covering all cases and we can't have log every time we have an error

src/valkey_search.cc
yanamolo: The log messages can be removed.

allenss-amazon: Rather than removing them, perhaps the VMSDK_LOG_EVERY_N_SECS macro would be better, so that the message is rates limited to say once every 10 seconds or every 100 seconds or once an hour, ....

src/valkey_search.cc
yanamolo: I will add more data in the function where we throw the initial error

src/valkey_search.cc
yanamolo: Yes, that is a better approach

vmsdk/src/thread_monitoring.cc
yanamolo: The function will return only error, and the metric will be updated in the upper level function

PR #576, opened 2026-01-04
Fix string-interned race in multi-thread

allenss-amazon:
The new interned string has a race which shows up in multi-threading. The problem was that the invariant around the map entry and non-zero reference counts wasn't properly handled.


BCathcart:
Moved to #613 on top of main

PR #577, opened 2026-01-04
 Negation querying with schema level tracked/untracked keys

VoletiRam:
Negation querying with schema level tracked/untracked keys. Negation logic: `total keys - matching keys = negation result `

### DIFF
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -128,20 +128,16 @@ namespace valkey_search::query {
 void* TextPredicate::Search(bool negate) const {
   size_t estimated_size = EstimateSize();
 
+  auto fetcher = std::make_unique<indexes::Text::EntriesFetcher>(
+      estimated_size, GetTextIndexSchema()->GetTextIndex(), nullptr,
+      GetFieldMask());
+  fetcher->predicate_ = this;
+
   if (!negate) {
-    auto fetcher = std::make_unique<indexes::Text::EntriesFetcher>(
-        estimated_size, GetTextIndexSchema()->GetTextIndex(), nullptr,
-        GetFieldMask());
-    fetcher->predicate_ = this;
     return fetcher.release();
   }
 
-  auto positive_iterator = BuildTextIterator(
-      std::make_unique<indexes::Text::EntriesFetcher>(
-          estimated_size, GetTextIndexSchema()->GetTextIndex(), nullptr,
-          GetFieldMask())
-          .get());
-
+  auto positive_iterator = BuildTextIterator(fetcher.get());
   auto negation_iterator =
       std::make_unique<indexes::text::NegationTextIterator>(
           std::move(positive_iterator),
--- a/src/indexes/text/negation_iterator.cc
+++ b/src/indexes/text/negation_iterator.cc
@@ -19,7 +19,6 @@ NegationTextIterator::NegationTextIterator(
     : schema_tracked_keys_(schema_tracked_keys),
       schema_untracked_keys_(schema_untracked_keys),
       query_field_mask_(query_field_mask),
-      phase_(TRACKED),
       tracked_iter_(schema_tracked_keys.begin()),
       untracked_iter_(schema_untracked_keys.begin()) {
   if (positive_iterator) {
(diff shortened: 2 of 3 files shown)


src/indexes/text.cc
```diff
@@ -125,11 +127,30 @@ namespace valkey_search::query {
...
+        GetFieldMask());
+    fetcher->predicate_ = this;
+    return fetcher.release();
+  }
+
+  auto positive_iterator = BuildTextIterator(
+      std::make_unique<indexes::Text::EntriesFetcher>(
+          estimated_size, GetTextIndexSchema()->GetTextIndex(), nullptr,
+          GetFieldMask())
+          .get());
```
allenss-amazon: This would make more sense, if you setup the positive iterator unconditionally (since it's used in all paths) and then conditionally wrapped it in the negative iterator.

Also, I notice that one path assigns fetch->predicate_, but the other path doesn't. This asymmetry sticks out as a potential issue.

src/indexes/text/negation_iterator.cc
```diff
@@ -0,0 +1,127 @@
...
+
+  while (tracked_iter_ != schema_tracked_keys_.end() &&
+         matched_keys_.contains(*tracked_iter_)) {
+    ++tracked_iter_;
+  }
+
+  if (tracked_iter_ == schema_tracked_keys_.end()) {
+    phase_ =
+        (untracked_iter_ != schema_untracked_keys_.end()) ? UNTRACKED : DONE;
+  }
```
allenss-amazon: Isn't the point of this code to align the iterator to the first valid key? Wouldn't it be true that if you setup the correct initialization of iterators/phases you should be able to just call NextKey here, right?

src/indexes/text/negation_iterator.cc
```diff
@@ -0,0 +1,127 @@
...
+          break;
+        }
+      } while (matched_keys_.contains(*tracked_iter_));
+    } else {
+      ++untracked_iter_;
+      if (untracked_iter_ == schema_untracked_keys_.end()) {
+        phase_ = DONE;
+      }
+    }
+  }
```
allenss-amazon: Why isn't the loop a call to NextKey() ?

allenss-amazon (commented):
Overall, it's definitely advantageous to think of the problem as a state machine: Phase1, Phase2 and Done -- and it should be documented as such. When I read the code, I believe it's true that the phase variable is a redundantly encoded version of the tracked and untracked iterators. In other words:

1. Phase1 <=> tracked_iterator != end.
2. Phase2 <=> tracked_iterator == end && untracked_iterator != end.
3. Done <=> tracked_iterator == end && untracked_iterator == end.

Generally, having redundant state is undesirable because:

1. It increases the cognitive load on the future code reader/maintainer as there are more pieces of state floating around.
2. It increases the complexity of the code as there's extra code to maintain the duplicate state.
3. It's error prone because the duplicate state can get out of sync without careful coding.

One hybrid solution would be to create a few simple boolean functions like IsPhase1() and IsPhase2(), etc.

src/indexes/text/negation_iterator.cc
VoletiRam: Sentinel logic doesn't help here because with empty sets iterator cannot distinguish between begin(uninitialized) and end(exhausted) but we need to position at first valid key at initialization. Will add initialization variable and we can move this logic to nextKey.

src/indexes/text/negation_iterator.cc
VoletiRam: Aah yes.

KarthikSubbarao:
When I read this code earlier, I saw it was missing handling from the PredicateEvaluator (predicate.cc) from the main thread verification. Can we include this as well? There can be mutations since the time of the background search and the main thread verification + reply stage

src/indexes/text.cc
```diff
@@ -88,7 +92,7 @@ bool Text::IsTracked(const InternedStringPtr& key) const {
   return false;
 }
 
-uint64_t Text::GetRecordCount() const {
+size_t Text::GetTrackedKeyCount() const {
   // TODO: keep track of number of keys indexed for this attribute
   return 0;
```
daddaman-amz, line 97: @VoletiRam    
Currently ft.info 's  `num_records` in using this function and since we return zero, text is not being accounted for in that

Will your PR be implementing tracked key count or is this something not relevant for text since we have shared structure?

PR #580, opened 2026-01-06
Fix ASAN failure for unit tests commands_test and query_test

baswanth09:
1. Fix `commands_test`
  a. Cleanup of `TextParsingOptions` dangling references in tests.
  b. Memory leak in `ft_aggregate_exec_test.cc`. Do cleanup during teardown.

2. Fix `query_test`
Make `entries_range` `static` to prevent dangling reference when mock lambdas execute after InitIndexSchema() returns.

### DIFF
--- a/testing/search_test.cc
+++ b/testing/search_test.cc
@@ -181,40 +181,32 @@ void InitIndexSchema(MockIndexSchema *index_schema) {
       "numeric_index_100_20", "numeric_index_100_20", numeric_index_100_20));
   VMSDK_EXPECT_OK(index_schema->AddIndex(
       "numeric_index_100_40", "numeric_index_100_40", numeric_index_100_40));
-  indexes::Numeric::EntriesRange entries_range;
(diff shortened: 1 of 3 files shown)


allenss-amazon (approved):

testing/search_test.cc
```diff
@@ -240,7 +232,8 @@ TEST_P(EvaluateFilterAsPrimaryTest, ParseParams) {
   const EvaluateFilterAsPrimaryTestCase &test_case = GetParam();
   auto index_schema = CreateIndexSchema(kIndexSchemaName).value();
   InitIndexSchema(index_schema.get());
-  FilterParser parser(*index_schema, test_case.filter, {});
+  TextParsingOptions options{};
+  FilterParser parser(*index_schema, test_case.filter, options);
```
BCathcart, line 236: Just FYI, you could change FilterParser constructor to copy `options` param instead of saving a reference. No worries though since I see it's already merged

PR #583, opened 2026-01-06
Switch from deque of neighbors to vector

KarthikSubbarao:
Switch from deque of neighbors to vector. We are now able to apply a vector reserve to handle searches in a more performant manner

### DIFF
--- a/src/indexes/vector_base.cc
+++ b/src/indexes/vector_base.cc
@@ -252,9 +252,10 @@ absl::StatusOr<bool> VectorBase::ModifyRecord(const InternedStringPtr &key,
 }
 
 template <typename T>
-absl::StatusOr<std::deque<Neighbor>> VectorBase::CreateReply(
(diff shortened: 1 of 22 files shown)


src/indexes/vector_base.cc
```diff
@@ -263,7 +264,7 @@ absl::StatusOr<std::deque<Neighbor>> VectorBase::CreateReply(
       continue;
     }
     // Sorting in asc order.
-    ret.emplace_front(Neighbor{vector_key.value(), ele.first});
+    ret.emplace(ret.begin(), Neighbor{vector_key.value(), ele.first});
```
daddaman-amz: will this be more expensive to insert at front now that we changed the type and does it matter ?

allenss-amazon: That's the right question, does it matter? If assume that we know how many elements we have we should be able to resize the array and then insert from the back to the front.

PR #588, opened 2026-01-08
Fix for handling trailing punctuations in query

daddaman-amz:
Fix to not throw syntax error when there is a punctuation towards end of the query


src/commands/filter_parser.cc
```diff
@@ -811,7 +811,22 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
       terms.push_back(std::move(result.predicate));
       // For unquoted text, stop after first token. For exact phrases, continue
       // parsing all tokens.
-      if (!exact_phrase) break;
+      if (!exact_phrase) {
+        // Consume all consecutive trailing punctuation (not query syntax)
```
KarthikSubbarao: Why is this needed?

Can you share query strings where this change is needed?

daddaman-amz: Below is the case where trailing punctuations in query show up as syntax error instead of consuming them. 
```
(Pdb) client.execute_command("FT.SEARCH", "idx1", '!!!!car ')   // works if they are at beginning
[1, b'doc:4', [b'content', b'Driver drove the car!']]

(Pdb) client.execute_command("FT.SEARCH", "idx1", 'car!!!!')
*** valkey.exceptions.ResponseError: Invalid filter expression: `car!!!!`. Invalid Query Syntax
```

KarthikSubbarao: Does this get handled already in quoted (exact phrases) text case? I see the handling was only added in the unquoted text case?

Also, as discussed earlier, the handling added seems like duplicated code which would be nice to avoid by re-entering (covering with a single pass)

src/commands/filter_parser.cc
```diff
@@ -665,6 +665,30 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseUnquotedTextToken(
...
+    while (!IsEnd()) {
+      char ch = Peek();
+      if (lexer.IsPunctuation(ch)) {
+        // Reject reserved characters in unquoted text
+        if (ch == '{' || ch == '}' || ch == '[' || ch == ']' || ch == ':' ||
+            ch == ';' || ch == '$') {
+          return absl::InvalidArgumentError(
+              absl::StrCat("Unexpected character at position ", pos_ + 1, ": `",
+                           expression_.substr(pos_, 1), "`"));
+        }
```
KarthikSubbarao: Why is this needed? Does it not get handled through the normal flow? We should try to not add redundant handling in isolated code paths, and try to leverage the overall query parsing flow to handle different paths

PR #591, opened 2026-01-09
Addressing memory deallocation overhead by minimizing the mutex contension

yairgott:
The proposed approach utilizes two layers of address trackers: an accurate address tracker and a snapshot tracked addresses which is initialized as a read-only snapshot of accurate address tracker when switching to the Valkey allocator.

Determining whether an address is tracked involves a two-step process. First, the system checks whether the address resides in the tracked addresses snapshot. While this improves performance, it may introduce false positives if another thread has already removed the address. Therefore Any positive match MUST be verified against the accurate address tracker.

Since deallocations of addresses allocated before the module load completes are extremely rare, the accurate tracker is accessed infrequently and therefor its overall overhead is negligible.

### DIFF
--- a/vmsdk/src/memory_allocation_overrides.cc
+++ b/vmsdk/src/memory_allocation_overrides.cc
@@ -97,9 +97,6 @@ class SystemAllocTracker {
   }
 
  private:
-  inline bool ValkeyAllocQuickCheck(void* ptr) const {
-    return IsUsingValkeyAlloc() && !tracked_ptrs_snapshot_.contains(ptr);
-  }
   // RawSystemAllocator implements an allocator that will not go through
   // the SystemAllocTracker, for use by the SystemAllocTracker to prevent
   // infinite recursion when tracking pointers.


vmsdk/src/memory_allocation_overrides.cc
```diff
@@ -37,23 +36,39 @@ class SystemAllocTracker {
...
+    tracked_ptrs_.insert(ptr);
   }
-  bool IsTracked(void* ptr) {
-    auto& shard = GetShardForPtr(ptr);
-    absl::MutexLock lock(&shard.mutex_);
-    return shard.tracked_ptrs_.contains(ptr);
+
+  bool IsTracked(void* ptr) const {
+    if (IsUsingValkeyAlloc() && !tracked_ptrs_snapshot_.contains(ptr)) {
+        return false;
```
allenss-amazon, line 69: This code suggests that it's possible for a tracked address to be in tracked_ptrs but not in tracked_ptrs_snapshot. How is that possible?

yairgott: `tracked_ptrs_snapshot_` is a read-only snapshot captured immediately before switching to the Valkey allocator. Since this snapshot is not updated when addresses are freed, it contains stale data. It effectively acts as a pre-filter that may yield false positives where any hit must be validated against the live `tracked_ptrs_`.

yairgott:
> What about realloc? Isn't the code assuming that once we switch to Valkey all new allocations are done from valkey? in case of realloc could it be that the new pointer is from non-valkey allocator but the thread snapshot will not contain it?

Good point! I fixed this based on your suggested code. PTAL

vmsdk/src/memory_allocation_overrides.cc
```diff
@@ -205,22 +225,43 @@ void* __wrap_calloc(size_t __nmemb, size_t size) noexcept {
...
+    void* new_ptr = vmsdk::PerformAndTrackMalloc(AlignSize(size),
+                                                 ValkeyModule_Alloc,
+                                                 ValkeyModule_MallocUsableSize);
+    if (new_ptr == nullptr) {
+      // Valkey allocation failed, keep system buffer and restore tracking
+      vmsdk::SystemAllocTracker::GetInstance().TrackPointer(ptr);
+    }
+    else {
+      // Step 2: Copy data and free system buffer
+      memcpy(new_ptr, ptr, size);
```
zvi-code: The problem is ptr can be smaller allocation than requested size (it most cases of realloc to grow) so you can get `seg-fault` if accessing it out of range.
The other problem is that you don't know the size, this is why I proposed to first realloc with system allocator and only then copy to a new buffer and free the temp

zvi-code: I suggest to replace the `else` implementation with:
```  
// Bootstrap path: still using system allocator
auto tmp_ptr = vmsdk::PerformAndTrackRealloc(ptr, size, __real_realloc,
                                               empty_usable_size);
if (tmp_ptr == nullptr) {
  // Valkey allocation failed, keep system buffer and restore tracking
  vmsdk::SystemAllocTracker::GetInstance().TrackPointer(ptr);
  vmsdk::PerformAndTrackFree(new_ptr, ValkeyModule_Free,
                                          ValkeyModule_MallocUsableSize);
  return nullptr;
}
memcpy(new_ptr, tmp_ptr, size);
vmsdk::PerformAndTrackFree(tmp_ptr, __real_free, empty_usable_size);
```

yairgott: Makes sense!

zvi-code (approved):
LGTM

PR #596, opened 2026-01-10
Apply clang format to text_index.*

BCathcart:
I see @eifrah-aws added `PointerAlignment: Right` to `.clang-format` but from what I found that's overridden by `DerivePointerAlignment: true` from the Google style which aims to keep the pointer style consistent to what it currently is. I made some changes to `text_index.cc` that caused the derived pointer alignment to switch from left aligned to right aligned. I'm adding a separate formatting commit here first so the git history is clear.


BCathcart:
The MacOS build is failing because of our fulltext merge. We didn't have this check running in the fulltext branch and didn't prioritize fixing it. Let's fix this up next week.

PR #597, opened 2026-01-10
Update QueryOperations to be sent via protobuf allowing remote nodes to benefit from optimizations

KarthikSubbarao:
Update QueryOperations to be sent via protobuf allowing remote nodes to benefit from optimizations


KarthikSubbarao:
There seems to be a warning linter in a full unrelated to this PR. This is why the MacOs build is red. Will look at it this week

PR #598, opened 2026-01-11
Implement single slot indexes (resolves #141)

### LINKED ISSUE #141: Add Single Slot index support.
Today, indexes cover data in all slots. This requires query operations to be broadcast to all slots (shards actually) and to integrate their results. This architecture fundamentally limits the horizontal scalability of searching. This is a good solution for applications that have only a few indexes that cover the vast majority of the Valkey data.

However, there are use-cases where the desire is to create a large number of indexes, where each index covers only a small portion of the Valkey data corpus. For example, a common use-case for Valkey is session management, i.e., a user signs-on to a web-site and this implicitly creates an http session and Valkey is used to store the per-session data, allowing stateless web and business logic servers. 

In this use case it's not uncommon to want to execute "search" operations on the per-session data. Not very likely with vector, but with the upcoming full-text enhancements this can be a useful capability.  To implement this facility today, would require that each session have a distinguishing field, i.e., a TAG or NUMERIC field added and that each search would filter on this particular session. The problem with this is that search QPS is limited by the broadcast architecture and can only be scaled vertically (more CPUs) and through replicas, which can be impracticable or very inefficient.

Single slot indexes is a straight-forward solution to this problem and aligns with the Valkey clustering architecture. Single slot indexes are defined to only cover the data within one slot. This eliminates the need for cross-slot broadcast operations, meaning that query operations would now scale horizontally.

At the user API level, a Single slot index is created with the usual FT.CREATE command by decorating the index name and the keyspace prefix list with a hash-tag. Because the prefix list contains a hash-tag, it only covers keys in that slot. By decorating the index name with the same hash-tag, clients will easily be able to route search commands to that same slot. Thus these indexes are totally confined to the single slot, no data for this slot need be present on any other member of the cluster. This is not a breaking change, because the current implementation of FT.CREATE was specifically updated to reject index names or prefixes which contain a hash-tag -- in anticipation of supporting this feature.

The state of how clients route FT.xxxx commands isn't consistent. Today, those commands can be sent to any node in the cluster but for single slot indexes, these commands can only be executed by a specific node in the cluster. There are several possible solutions to this problem, but the best UX would be to add code in the search module such that it would proxy single slot FT.xxxx commands that were received by a non-owning node. Over the long term, it would be advantageous to teach clients how to route FT.xxxx commands based on the presence/absence of a hash-tag in the index name.

allenss-amazon:
With this PR, "single-slot" indexes are implemented.

A single-slot index is an index that contains data only from a single slot. Query operations on a single-slot index are handling only by the shard that owns that slot. By avoiding involving every shard, query operation throughput will scale horizontally.

A single-slot index is created by using a hash-tag in the index name. Since that index can only contain data from that slot, it is required that the index be defined with at least one prefix. It is also required that all prefixes defined for this index must have the same hash-tag as the index name.

Query operations can be sent to any node in the cluster and will automatically be forwarded to the owning shard. Of course, maximum efficiency is obtained when the query operations are dispatched the shard that owns the slot.

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -1,66 +1,69 @@
 # Command List
-- [`FT.CREATE`](#ftcreate) 
+
+- [`FT.CREATE`](#ftcreate)
 - [`FT.DROPINDEX`](#ftdropindex)
 - [`FT.INFO`](#ftinfo)
 - [`FT._LIST`](#ft_list)
 - [`FT.SEARCH`](#ftsearch)
+
 #
 
 ## FT.CREATE
 
 The `FT.CREATE` command creates an empty index and initiates the backfill process. Each index consists of a number of field definitions. Each field definition specifies a field name, a field type and a path within each indexed key to locate a value of the declared type. Some field type definitions have additional sub-type specifiers.
 
-For indexes on HASH keys, the path is the same as the hash member name. The optional `AS` clause can be used to rename the field if desired. 
+For indexes on HASH keys, the path is the same as the hash member name. The optional `AS` clause can be used to rename the field if desired.
 Renaming of fields is especially useful when the member name contains special characters.
 
 For indexes on JSON keys, the path is a JSON path to the data of the declared type. Because the JSON path always contains special characters, the `AS` clause is required.
-
 
 ```bash
 FT.CREATE <index-name>
     ON HASH
     [PREFIX <count> <prefix> [<prefix>...]]
-    SCHEMA 
+    [SKIPINITIALSCAN]
+    SCHEMA
         (
-            <field-identifier> [AS <field-alias>] 
-                  NUMERIC 
-                | TAG [SEPARATOR <sep>] [CASESENSITIVE] 
+            <field-identifier> [AS <field-alias>]
+                  NUMERIC
+                | TAG [SEPARATOR <sep>] [CASESENSITIVE]
                 | VECTOR [HNSW | FLAT] <attr_count> [<attribute_name> <attribute_value>]+
             [SORTABLE]
         )+
 ```
 
+- **\<index-name\>** (required): This is the name you give to your index. If an index with the same name exists already, an error is returned.
 
-- **\<index-name\>** (required): This is the name you give to your index. If an index with the same name exists already, an error is returned.  
-    
-- **ON HASH | JSON** (optional): Only keys that match the specified type are included into this index. If omitted, HASH is assumed.  
-    
+- **ON HASH | JSON** (optional): Only keys that match the specified type are included into this index. If omitted, HASH is assumed.
+
 - **PREFIX \<prefix-count\> \<prefix\>** (optional): If this clause is specified, then only keys that begin with the same bytes as one or more of the specified prefixes will be included into this index. If this clause is omitted, all keys of the correct type will be included. A zero-length prefix would also match all keys of the correct type.
-    
+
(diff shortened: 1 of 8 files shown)


vmsdk/src/cluster_map.cc
```diff
@@ -139,6 +139,47 @@ std::vector<NodeInfo> ClusterMap::GetTargets(FanoutTargetMode mode,
   }
 }
 
+std::vector<NodeInfo> ClusterMap::GetTargetsForSlot(FanoutTargetMode mode,
+                                                    bool prefer_local,
+                                                    uint16_t slot) const {
```
zvi-code, line 139: Can we make this logic subset of the `ClusterMap::GetTargets` logic? 
If we think of 2 phase, 1 choose the shards 2 choose the nodes in shard, than we can make the shard selection be the only change

src/commands/commands.cc
```diff
@@ -123,6 +143,18 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
+    bool do_fanout = ValkeySearch::Instance().UsingCoordinator() &&
+                     ValkeySearch::Instance().IsCluster() &&
+                     !parameters->local_only;
+    if (do_fanout) {
+      search_targets = ComputeSearchTargets(ctx, *parameters);
+      if (search_targets.empty()) {
+        return absl::InternalError("No available nodes to execute the query");
+      }
+    }
+
```
zvi-code, line 201: should we check that `parameters->local_only` is only valid if slot is owned by local?

src/commands/commands.cc
```diff
@@ -165,6 +186,18 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
 
+    std::vector<vmsdk::cluster_map::NodeInfo> search_targets;
+
+    bool do_fanout = ValkeySearch::Instance().UsingCoordinator() &&
+                     ValkeySearch::Instance().IsCluster() &&
+                     !parameters->local_only;
+    if (do_fanout) {
+      search_targets = ComputeSearchTargets(ctx, *parameters);
+      if (search_targets.empty()) {
+        return absl::InternalError("No available nodes to execute the query");
```
KarthikSubbarao, line 198: Should we use a consistent error message for this case for search operations? What message do we return in other search fanout failures due to this reason?

src/commands/commands.cc
```diff
@@ -111,6 +108,30 @@ void Free([[maybe_unused]] ValkeyModuleCtx *ctx, void *privdata) {
...
+
+std::vector<vmsdk::cluster_map::NodeInfo> ComputeSearchTargets(
+    ValkeyModuleCtx *ctx, const QueryCommand &parameters) {
+  auto mode = /* !vmsdk::IsReadOnly(ctx) ? query::fanout::kPrimaries ? */
+      ForceReplicasOnly.GetValue()
+          ? vmsdk::cluster_map::FanoutTargetMode::kOneReplicaPerShard
+          : vmsdk::cluster_map::FanoutTargetMode::kRandom;
+
+  // refresh cluster map if needed
+  auto cluster_map = ValkeySearch::Instance().GetOrRefreshClusterMap(ctx);
```
KarthikSubbarao, line 122: How efficient is this function?

I was just looking at it and it is using deprecated APIs.

A std::mutex path or modern atomic load with std::memory_order_acquire might be better than what it does

allenss-amazon: Well, the idea is that this provides the opportunity to refresh the cluster map. But that's only done if the map is more out of date than a fixed value, i.e., the map is updated no more often than X times per second.

PR #441, opened 2025-10-31
Position Optimization

Aksha1812:
FlatPositionMap is a space-optimized serialized representation of position data,
replacing std::map<Position, FieldMask> which consumes 80+ bytes per position
with a byte array achieving 1-8 bytes per position. This is critical for memory
efficiency as millions of these structures exist across the full-text corpus.

During document ingestion, positions accumulate in std::map for efficient
random insertion. Upon completion, the map is serialized into FlatPositionMap
and the map is destroyed. The FlatPositionMap is read-only thereafter and used
by search queries.

Structure Layout:
  [Variable header] [optional partition map] [position/field data]

Header layout (variable length):
First byte (8 bits):
  Bit 0:     Header selection (0=standard, 1=special)
  Bits 1-2:  Encoding scheme (2 bits) - reserved for future use
  Bits 3-4:  Number of bytes to store position count (0-3 = 1-4 bytes)
  Bits 5-6:  Number of bytes to store partition count (0-3 = 1-4 bytes)
  Bit 7:     Reserved

After first byte:
  - N bytes for number of positions (N determined by bits 3-4)
  - M bytes for number of partitions (M determined by bits 5-6, can be 0)

Encoding scheme:
- Single general case with byte-based partitions
- Partitions created every 128 bytes (PARTITION_SIZE) of serialized data
- Each partition entry stores the partition start byte offset and cumulative sum of deltas
- Position/ field Mask have one bit reserved for continue/ stop bit bit .
- Value is encoded into the other 7 bits . The position/ field mask bit is composited with the value
- Field masks optimized: if num_fields=1, no field mask bytes stored
- Field masks only stored when they change or at partition start (when
num_fields > 1)

Delta encoding stores position differences not absolutes.

A FlatPositionMapIterator provides sequential iteration and skip-forward with
minimal state overhead, maintaining cumulative position for delta decoding.

PR #599, opened 2026-01-12
Followup of #441

Aksha1812:


Aksha1812:
Moving changes to new PR with main as target

PR #601, opened 2026-01-12
Compatibility Tests

boda26:


integration/compatibility/data_sets.py
```diff
@@ -302,9 +404,115 @@ def make_field_definition(key_type, name, typ, i):
...
+    if data_source is None:
+        data_source = "text" if data_set in TEXT_DATASETS else "vector"
+
+    match data_source:
+        case "vector":
+            data = compute_data_sets()
+        case "text":
+            data = compute_text_data_sets(data_set)
+        case _:
+            raise ValueError(f"Unknown data source: {data_source}")
```
allenss-amazon, line 515: why do we have separate data sets? Why not just merge them?

allenss-amazon (approved):

PR #602, opened 2026-01-12
fixing #449 tag query parsing

zvi-code:
Complete the fix in https://github.com/valkey-io/valkey-search/pull/452

The Fix was validated manually on CME


allenss-amazon (approved):
Should we backport this?

PR #603, opened 2026-01-13
Followup of #441 Position Optimization

Aksha1812:
Followups addressed.

- > This loop could use some more thought. Inherently it doesn't make any sense to me why the linear forward walking should be limited by partition size. Since we're logically at some unknown place in the partition. It would seem to me the choice to abandon linear forward search can be bounded much better. Specifically, if keep track of the partition index then we can cheaply test against the start of the next partition to determine whether to commence forward scanning from the current point or to switch to the partition-table based lookup.

Added current_partition_idx. I feel like instead of keeping track of current_partition_idx it would be simpler to store next parittion's start offset. Let me know if that sounds good will change it quickly.

- > This function is likely the most time-critical. With some thought, you can combine the IsValid and the End condition into the same test. There's no reason to treat these differently. Since you know at the time that the iterator is constructed where the last byte of data is, you don't need a terminator byte -- rather treat this like an STL iterator that has an "end()" value, which is classically pointing one element past that last valid element. Thus the IsInvalid and terminator byte can be replaced by a single == comparison to the end() value.
I don't understand why the iterator has both current_start_ptr_ and current_end_ptr_. The start of this routine shows that current_start_ptr is actually ignored. Why not just a "current_ptr" that's pointed to the next byte of data to be read? Do you actually need to go backwards?

Removed current_end_ptr_ . I am not sure about "Since you know at the time that the iterator is constructed where the last byte of data is, you don't need a terminator byte "  don't think I agree with this. 

Rest are smaller fixes and test changes which are addressed in this PR  https://github.com/valkey-io/valkey-search/pull/441#pullrequestreview-3617178123.

### DIFF
--- a/src/indexes/text/flat_position_map.cc
+++ b/src/indexes/text/flat_position_map.cc
@@ -238,6 +238,7 @@ PositionIterator::PositionIterator(const FlatPositionMap& flat_map)
       cumulative_position_(0),
       num_partitions_(0),
       current_partition_idx_(0),
+      next_partition_offset_(UINT32_MAX),
       header_size_(0),
       current_field_mask_(1) {
   CHECK(flat_map_)
@@ -270,14 +271,22 @@ void PositionIterator::NextPosition() {
     return;
   }
 
-  // Update partition index if we've crossed into next partition
-  if (current_partition_idx_ < num_partitions_) {
-    const char* partition_map = flat_map_ + header_size_;
-    uint32_t next_offset = ReadVarUint(
-        partition_map + (current_partition_idx_ * kPartitionDeltaBytes * 2),
-        kPartitionDeltaBytes);
-    if (current_ptr_ == data_start_ + next_offset) {
+  // Check if we need to load/update next partition offset
+  if (next_partition_offset_ == UINT32_MAX ||
+      current_ptr_ == data_start_ + next_partition_offset_) {
+    // If we crossed a partition (not initial load), increment index
+    if (next_partition_offset_ != UINT32_MAX) {
       current_partition_idx_++;
+    }
+
+    // Load next partition offset (or UINT32_MAX if no more partitions)
+    if (current_partition_idx_ < num_partitions_) {
+      const char* partition_map = flat_map_ + header_size_;
+      next_partition_offset_ = ReadVarUint(
+          partition_map + (current_partition_idx_ * kPartitionDeltaBytes * 2),
+          kPartitionDeltaBytes);
+    } else {
+      next_partition_offset_ = UINT32_MAX;
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/indexes/text/flat_position_map.cc
```diff
@@ -251,42 +253,47 @@ PositionIterator::PositionIterator(const FlatPositionMap& flat_map)
...
+  // Update partition index if we've crossed into next partition
+  if (current_partition_idx_ < num_partitions_) {
+    const char* partition_map = flat_map_ + header_size_;
+    uint32_t next_offset = ReadVarUint(
+        partition_map + (current_partition_idx_ * kPartitionDeltaBytes * 2),
+        kPartitionDeltaBytes);
+    if (current_ptr_ == data_start_ + next_offset) {
+      current_partition_idx_++;
+    }
+  }
```
Aksha1812, line 291: I think I can just keep of this next_offset instead of current_partition_idx_ so that I don't have to get it everytime. Let me know if that sounds good.

allenss-amazon: Next_offset or current_partition_idx are effectively the same, i.e., O(1) determination of whether to scan forward or to profitably do a partition table lookup. Actually, next_offset is an epsilon more efficient as it avoids the need to compute the offset of the table before querying it. So I like that.

src/indexes/text/flat_position_map.cc
```diff
@@ -316,12 +323,15 @@ bool PositionIterator::SkipForwardPosition(Position target) {
...
-  while (IsValid() && (current_start_ptr_ - partition_start) < kPartitionSize) {
+  // Scan within current partition until we reach target or partition end
+  while (IsValid()) {
     if (cumulative_position_ >= target) return cumulative_position_ == target;
+
+    uint32_t prev_partition_idx = current_partition_idx_;
     NextPosition();
+
+    // If we crossed into next partition, use partition table to jump
+    if (IsValid() && current_partition_idx_ > prev_partition_idx) break;
```
allenss-amazon: Is there a way to this above NextPosition and avoid the IsValid re-test? We would need the know the end of the current partition, but I think this is knowable when a table lookup is done, as it's that start of the next partition, right?

allenss-amazon (approved):

src/indexes/text/flat_position_map.cc
Aksha1812: yup thought so. Actually i would need both because we need either the partition idx or pointer on the partition map to go to the next partition entry. we cannot just use next offset to go to the next partition entry. If we use only next offset we would have to find that offset in the partition map and then go to the next one. but yeah we would atleast avoid cimputing next_offset each time. which is better than the previous.

src/indexes/text/flat_position_map.cc
```diff
@@ -270,15 +271,23 @@ void PositionIterator::NextPosition() {
...
+    // If we crossed a partition (not initial load), increment index
+    if (next_partition_offset_ != UINT32_MAX) {
       current_partition_idx_++;
     }
+
+    // Load next partition offset (or UINT32_MAX if no more partitions)
+    if (current_partition_idx_ < num_partitions_) {
+      const char* partition_map = flat_map_ + header_size_;
+      next_partition_offset_ = ReadVarUint(
+          partition_map + (current_partition_idx_ * kPartitionDeltaBytes * 2),
```
allenss-amazon, line 286: This code fragment (and other similar fragments) is implementing the equivalent of an array. I think having a function that hides Varint and the pointer arithmetic will make it much easier for readers as it will allow reasoning at the logical level. I'd like to see this all similar instances in the code replaced with inline function calls.

PR #605, opened 2026-01-13
Rax Integration

BCathcart:
This PR replaces the rough custom `RadixTree` with `Rax` in our text index structure. RadixTree was naturally becoming more and more aligned with Rax, but without the space efficiency implemented yet. It makes more sense to switch to the well-used Rax and modify it as needed. Everything in this PR was done with some care except the `PathIterator` which I intend to redo with a follow-up in a week or so. Please ignore it. There are also opportunities for follow-up optimization:
-  `raxMutate` uses the existing APIs to find the existing target and then update the target if needed with two traversals. It shouldn't be hard to only traverse once in every case.
- Use SIMD to compare outgoing edges to the search character while traversing (motivated by ART paper)
- Optimize the iterator. There were some lessons from implementing the RadixTree iterator I think we could use to get a small efficiency gain in the Rax iterator.

### DIFF
--- a/src/indexes/text/text_index.cc
+++ b/src/indexes/text/text_index.cc
@@ -63,7 +63,8 @@ InvasivePtr<Postings> RemoveKeyFromPostings(
   return existing_postings;
 }
 
-// Factory for target mutate callback with memory tracking and target copying
+// Factory for target mutate callback with memory tracking and new target
+// copying to the outer scope
 template <typename Target, typename MutateFn>
 std::function<void *(void *)> CreateTargetMutateFn(
     MemoryPool &memory_pool, MutateFn mutate_fn,
@@ -106,7 +107,7 @@ std::function<void *(void *)> CreateTargetSetFn(
   };
 }
 
-// Factory for simple target mutation (no memory tracking, no target copying)
+// Factory for simple target mutation
 template <typename Target, typename MutateFn>
 std::function<void *(void *)> CreateSimpleTargetMutateFn(MutateFn mutate_fn) {
   return [mutate_fn = std::move(mutate_fn)](void *old_val) -> void * {
--- a/src/indexes/text/text_index.h
+++ b/src/indexes/text/text_index.h
@@ -23,7 +23,6 @@
 #include "src/indexes/text/invasive_ptr.h"
 #include "src/indexes/text/lexer.h"
 #include "src/indexes/text/posting.h"
-#include "src/indexes/text/radix_tree.h"  // TODO (remove)
 #include "src/indexes/text/rax_wrapper.h"
 
 struct sb_stemmer;


src/indexes/text/rax/rax.c
```diff
@@ -917,6 +917,52 @@ int raxFind(rax *rax, unsigned char *s, size_t len, void **value) {
...
+ * callback function. The callback receives the current value (NULL if the
+ * key doesn't exist) and caller context, and returns the new value (NULL
+ * to delete the key).
+ * 
+ * OWNERSHIP: The callback takes ownership of the old value. If returning a
+ * new value, the callback MUST clean up the old value.
+ * 
+ * Returns 1 on success, 0 on error (errno will be set to ENOMEM
+ * on out of memory). */
+// TODO: implement a lower-level version that doesn't traverse the tree twice.
```
Aksha1812, line 931: we can just add this with this pr. maybe we just need another helper function for remove and insert which uses already found node

src/indexes/text/rax/rax_malloc.h
```diff
@@ -28,18 +28,19 @@
...
 #define RAX_ALLOC_H
-#include "zmalloc.h"
-#define rax_malloc zmalloc
-#define rax_realloc zrealloc
-#define rax_free zfree
-#define rax_ptr_alloc_size zmalloc_size
+#include <stddef.h>
+
+/* Override with the wrappers provided by VMSDK. */
+extern void* __wrap_malloc(size_t size);
```
Aksha1812, line 36: do we know if these work well to track memory allocated by rax. maybe a small test. Also currently radix_memory_pool is unused. maybe you can add it to this pr.

src/indexes/text/rax_wrapper.h
```diff
@@ -0,0 +1,292 @@
...
+  // provided by the RadixTree itself, so if the target objects are disjoint
+  // (which is normal) then no locking is required within the mutate function
+  // itself.
+  //
+  void MutateTarget(absl::string_view word,
+                    absl::FunctionRef<void*(void*)> mutate);
+
+  // TODO: Replace with GetWordCount("") once it's implemented
+  // Get the total number of words in the RadixTree.
+  size_t GetTotalWordCount() const;
```
Aksha1812: does this return same thing as in num_unique_terms counter from ft info.

src/indexes/text/rax_wrapper.h
```diff
@@ -0,0 +1,292 @@
...
+   *
+   *                  [compressed]
+   *                  "te" |
+   *                   [branching]
+   *                "s" /     \ "a"
+   *          [compressed]   [compressed]
+   *          "ting" /           \ "m"
+   *   Target <- [leaf]           [leaf] -> Target
+   *
+   *  would become...
```
Aksha1812: what will happen if more words are added into one of the subtrees ? example teas , teach etc.. does it become 

```
   *                  [compressed]
   *                  "te" |
   *                   [branching]
   *                "s" /     \ "a"
   *          [compressed]   [compressed]        
   *          "ting" /           \ "m".          \ "s"           \ "ch"
   *   Target <- [leaf]           [leaf] -> Target.    [leaf] -> Target.      [leaf] -> Target. 
   ``` 
   will the other subtrees also change or no?

   or
   
   ```
   *                  [compressed]
   *                  "te" |
   *                   [branching]
   *                "sting" /     \ "a"
   *          [leaf]              [compressed]        
   *                             \ "m".                          \ "s"                           \ "ch"
   *                             [leaf] -> Target.    [leaf] -> Target.      [leaf] -> Target. 
   ```

File: src/indexes/text/rax_wrapper.h [hunk 5]
```diff
@@ -0,0 +1,292 @@
...
+    absl::string_view GetPath() const;
+
+    // Get the edge label for the current child being iterated
+    absl::string_view GetChildEdge();
+
+    // Get the target for this word, will assert if !IsWord()
+    void* GetTarget() const;
+
+    // Postings-specific accessor. Caller is responsible for tracking the type.
+    InvasivePtr<Postings> GetPostingsTarget() const;
```
Aksha1812, line 234: Is a separate API needed for each Target used e.g stem_mappings?

src/indexes/text/rax_wrapper.cc
```diff
@@ -0,0 +1,392 @@
...
+  unsigned char* c_word = const_cast<unsigned char*>(
+      reinterpret_cast<const unsigned char*>(word.data()));
+  void* opaque_callback = reinterpret_cast<void*>(&mutate);
+  int res = raxMutate(rax_, c_word, word.size(), MutateCallbackWrapper,
+                      opaque_callback);
+  CHECK(res) << "Rax mutation failed for word: " << word << ", errno: " << errno
+             << " (" << strerror(errno) << ")";
+}
+
+size_t Rax::GetTotalWordCount() const { return raxSize(rax_); }
```
Aksha1812: rename GetUniqueWordCount()

src/indexes/text/rax_wrapper.cc
```diff
@@ -0,0 +1,392 @@
...
+
+size_t Rax::GetWordCount(absl::string_view prefix) const {
+  // TODO: Implement word counting
```
Aksha1812: Is this supposed to return word count of the word we are at or all the words under that prefix tree. if just the word count of the word we are at we can add the CountTermFrequency() api call here. which would be later updated to accommodate efficient total term calculation.

build.sh
```diff
@@ -281,7 +281,7 @@ function build() {
 function format() {
     cd "${ROOT_DIR}"
     printf "Formatting...\n"
-    find src testing vmsdk/src vmsdk/testing -name "*.h" -o -name "*.cc" | xargs clang-format -i
+    find src testing vmsdk/src vmsdk/testing -name "*.h" -o -name "*.cc" | grep -v '^src/indexes/text/rax/' | xargs clang-format -i
```
Aksha1812, line 284: what about some new minor changes in rax folder. do we ignore formatting on those. any simple alternative ?

src/indexes/text/rax_wrapper.cc
```diff
@@ -0,0 +1,392 @@
...
+  CHECK(!Done()) << "Out of range";
+
+  raxNext(&iter_);
+  if (raxEOF(&iter_)) {
+    done_ = true;
+    return;
+  }
+}
+
+// TODO: currently unused and untested
```
Aksha1812, line 129: skipping review for this and PathIterator since not final implementation pending.

src/indexes/text/rax_wrapper.h
Aksha1812: Ignore

src/indexes/text/text_index.cc
```diff
@@ -150,35 +168,56 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
                      std::string(token.rbegin(), token.rend()))
                : std::nullopt;
 
+    // The updated target gets set in target_add_fn and later used in
+    // target_set_fn, so that all trees point to the same postings object
     InvasivePtr<Postings> updated_target;
 
+    auto target_add_fn = [&](void *old_val) {
+      // Note: Right now this won't include the position map memory since
```
Aksha1812: forgot to remove this while implementing serialization of position map. can you remove it in this PR?

src/indexes/text/text_index.cc
```diff
@@ -201,24 +240,45 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
       return;
     }
   }
-
   TextIndex &key_index = node.mapped();
 
-  auto iter = key_index.GetPrefix().GetWordIterator("");
+  // The updated target gets set in target_remove_fn and later used in
+  // target_set_fn, so that all trees point to the same postings object
+  InvasivePtr<Postings> updated_target;
```
Aksha1812, line 324: most of the code looks very similar to target_add_fn above except the API call to Postings. maybe a helper fn?

testing/rax_wrapper_test.cc
```diff
@@ -0,0 +1,473 @@
+/*
```
Aksha1812, line 1: do we also want to delete the old radix_test file ?

testing/rax_wrapper_test.cc
```diff
@@ -0,0 +1,473 @@
...
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
```
Aksha1812, line 3: Also some WordIterator and PathIterator tests missing from the old radix file are they no longer needed , I guess you will add pathiterator tests later when you implement it but are the missing WordIterator tests no longer needed? also i remember we used to validate tree structure earlier . can we add it again.

build.sh
BCathcart: I don't want to apply formatting to the existing code since in all likelihood it will make it harder to read, especially for someone that is familiar with the Rax code as is. For new code, it's best to try to match the formatting of the surrounding code as best as possible but I'm not going to be too picky about code entering these files. It's the same approach we use for the code in the valkey core

src/indexes/text/rax/rax.c
BCathcart: You're right it shouldn't be that much work but in terms of managing priorities right now, I think it's best to get something functionally correct in now while I work on other stuff. I'm tracking this as a follow-up item for myself along with the path iterator, although optimizing this is definitely lower on my list

src/indexes/text/rax/rax_malloc.h
BCathcart: Good point. I added `RaxMallocMemoryTracking` test to make sure we're routing through the allocator wrappers as expected.

The text index memory tracking we added including `radix_memory_pool` is currently completely broken. Let's tackle this in a separate PR.

src/indexes/text/rax_wrapper.cc
BCathcart: It was part of the original API definition with the description:
```
// Get the number of words that have the specified prefix in O(len(prefix))
// time.
```

I was assuming it was the number of words (i.e. nodes with targets) in that subtree, but I think it's probably related to the size estimation now that I read the description.

src/indexes/text/rax_wrapper.h
BCathcart: Looks like `num_unique_terms` is the number of posting objects in the text index, so yes it should be the same as the number of entries in the prefix tree (or suffix tree if it's enabled for all fields).

src/indexes/text/rax_wrapper.h
BCathcart: The RadixTree followed the same node abstraction as Rax and I never changed it, so it's the latter. I'll remove these comments - they are leftover from when I copied this API from radix_tree.h.

src/indexes/text/rax_wrapper.h
BCathcart: Yes or the caller will have to call `GetTarget` and do its own casting. I remember I originally tried to make the Rax class templated so `GetTarget` would return the actual InvasivePtr type but I ran into issues around the mutation callback passed to the C-layer being typed and it ended up getting more complex than was worthwhile.

src/indexes/text/text_index.cc
BCathcart: We didn't actually do what it says. Let's remove this in the follow-up when we fix the memory tracking.

testing/rax_wrapper_test.cc
BCathcart: I'm going to keep the RadixTree in the project for the moment. Once we've tested and stabilized everything on top of Rax, I'll remove them.

testing/rax_wrapper_test.cc
BCathcart: The WordIterator tests I just condensed and removed redundancy. No coverage should be lost there. The path iterator didn't exist when I started on this. I copied the test over now.

The tree structure validation was because I implemented the RadixTree code from scratch. Rax is mature and has it's own testing so the tests here are focused on the new functionality and making sure the rax wrapper APIs work as expected. I don't think we should worry about the internals of the rax implementation that we haven't touched.

src/indexes/text/text_index.cc
BCathcart: I remember having the same intuition originally but then it got a bit convoluted because we're maintaining references to variables in the calling scope in the closures and `RemoveKeyFromPostings` and `AddKeyToPostings` don't have the same signature.

Here's a generated draft of the refactor:

```
// Helper function template that handles the common boilerplate
template <typename MutationFn>
auto MakePostingsMutator(MemoryPool& memory_pool, 
                         InvasivePtr<Postings>& updated_target,
                         MutationFn&& mutate_fn) {
  return [&memory_pool, &updated_target, mutate_fn = std::forward<MutationFn>(mutate_fn)](void* old_val) {
    NestedMemoryScope scope{memory_pool};

    // Take ownership of the existing postings object reference if there is one.
    // It will be deconstructed at the end of this scope.
    auto existing_postings =
        old_val ? InvasivePtr<Postings>::AdoptRaw(
                      static_cast<InvasivePtrRaw<Postings>>(old_val))
                : InvasivePtr<Postings>{};

    // Mutate the postings using the provided function
    InvasivePtr<Postings> new_postings = mutate_fn(std::move(existing_postings));

    // Copy the new postings to the outer scope
    updated_target = new_postings;

    // Pass ownership of the new postings object reference to the tree
    return static_cast<void*>(std::move(new_postings).ReleaseRaw());
  };
}

auto target_add_fn = MakePostingsMutator(
    metadata_.posting_memory_pool_, 
    updated_target,
    [&](InvasivePtr<Postings> existing) {
      return AddKeyToPostings(existing, key, std::move(pos_map),
                              &metadata_, num_text_fields_);
    });

auto target_remove_fn = MakePostingsMutator(
    metadata_.posting_memory_pool_,
    updated_target,
    [&](InvasivePtr<Postings> existing) {
      return RemoveKeyFromPostings(existing, key, &metadata_);
    });
```

It's a little more concise and re-uses code but harder to follow IMO. What do you think?

src/indexes/text/rax_wrapper.h
Aksha1812: Ok . Note : This will require special handling for stem tree as the target is different.

src/indexes/text/text_index.cc
Aksha1812: I agree its a bit tricky to understand but better than having two similar code blocks and having to understand both and difference between them . Maybe a slightly elaborate function description than `// Helper function template that handles the common boilerplate` would help .

src/indexes/text/rax/rax_malloc.h
Aksha1812: where is this added . couldn't find it .

src/indexes/text/rax_wrapper.h
Aksha1812: Do we need this then . since // TODO: Replace with GetWordCount("") once it's implemented . Also this is not used currently anywhere is it? . Also i would still rename it GetUniqueWordCount. It might be misinterpreted as total number of words which it is not, it is total number of unique words , which sounds more consistent to the info we have

src/indexes/text/rax_wrapper.cc
Aksha1812: I guess later we can discuss about if we want to use the 
1) number of unique words in that sub tree with the number of positions, which is not that expensive 
2) or if we want to get the total term frequency , which involves going through the position map and getting total numbe ropf terms across all fields (this can also be simplified with some changes in the flat position map which might just increase per key memory a little bit) 

for the size estimation.

src/indexes/text/rax_wrapper.h
BCathcart: Looks like I already removed the TODO and renamed it to `GetTotalUniqueWordCount`

src/indexes/text/rax/rax_malloc.h
BCathcart: You should be able to see it in the diff since your last review. Otherwise check the latest commit.

src/indexes/text/text_index.cc
BCathcart: Refactored

src/indexes/text/rax_wrapper.h
Aksha1812: you can leave out Total i guess. upto you.

KarthikSubbarao:
I see failing Integration ASAN tests from the runs after the merge: https://github.com/valkey-io/valkey-search/actions/runs/21653491847/job/62423258968

Just wanted to check if the same was observed locally as well with this PR.

On all open PRs, after the merge from main, I see Integration ASAN tests started failing

BCathcart:
There's some timing issue. It passes locally consistently. Looks like we need to update the workflows to prevent merging when one doesn't run, since everything looked green on the PR.

PR to revert this change until the fix: #677.

PR #607, opened 2026-01-13
Endurance test

Nivesh-01:
Adding in testing for Text, Tag, and Numeric index for endurance testing in the stability tests. These tests also use memtier benchmark.


testing/integration/utils.py
```diff
@@ -155,7 +155,8 @@ def start_valkey_cluster(
 
     for port in ports:
         stdout_path = os.path.join(stdout_directory, f"{port}_stdout.txt")
-        stdout_file = open(stdout_path, "w")
+        # Open file without buffering to avoid ResourceWarnings
+        stdout_file = open(stdout_path, "w", buffering=1)
```
daddaman-amz: comment says "without buffering" but buffering is enabled.

Also is that responsible for removing ResourceWarnings or is it the file pattern you used below?

daddaman-amz: Question:

In the log file you shared, isee below lines related to ResourceWarnings. Is that still expected?

```
/local/home/tuwanivu/valkey-search/testing/integration/stability_test.py:540: ResourceWarning: unclosed file <_io.BufferedReader name=30>
  results = stability_runner.StabilityRunner(config).run()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/local/home/tuwanivu/valkey-search/testing/integration/stability_test.py:540: ResourceWarning: unclosed file <_io.BufferedReader name=35>
  results = stability_runner.StabilityRunner(config).run()
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```

testing/integration/stability_test.py
```diff
@@ -199,6 +199,294 @@ def tearDown(self):
...
+                ftdropindex_interval_sec=10,
+                flushdb_interval_sec=20,
+                randomize_bg_job_intervals=True,
+                num_memtier_threads=10,
+                num_memtier_clients=10,
+                num_search_clients=10,
+                insertion_mode="time_interval",
+                test_time_sec=60,
+                test_timeout=120,
+                keyspace_size=1000000,
```
daddaman-amz, line 220: Question:   Does the code check if we reached this keyspace size   or is this just a max limit for ingestion and test may or may not hit it?

daddaman-amz (commented):
Overall you mentioned that all new non-backfill tests pass  but backfill tests fails.   We can get this PR in if you think they are not test related failures

integration/compatibility/aggregate-answers.pickle.gz
daddaman-amz, line 1: Can you let me know what this is and what is new?  Thanks

Nivesh-01: Nothing is. In each run a new file is generated so whenever those tests are run a new pickle file will be generated and hence it is being shown in the commit. I asked Miles(author of pickle file) and he said it is expected.

testing/integration/stability_test.py
Nivesh-01: max limit

testing/integration/utils.py
Nivesh-01: without buffering is a typo, I am trying to remove the resource warnings through it but have not yet gotten much success in it.

daddaman-amz (approved):
you can confirm on the broader comment on test results and check DCO before merge

integration/compatibility/aggregate-answers.pickle.gz
allenss-amazon: No, the idea is that this file is generated once and checked-in so that for each regression run we don't regenerate it, but just run it. This eliminates the dependency on docker and the Redis container in the regression test flow.

PR #608, opened 2026-01-13
Endurance failover

Nivesh-01:
Endurance testing for failover. Currently only for vectors

### DIFF
--- a/src/index_schema.h
+++ b/src/index_schema.h
@@ -163,6 +163,11 @@ class IndexSchema : public KeyspaceEventSubscription,
   uint64_t CountRecords() const;
 
   int GetAttributeCount() const { return attributes_.size(); }
+  int GetTagAttributeCount() const;
+  int GetNumericAttributeCount() const;
+  int GetVectorAttributeCount() const;
+  int GetTextAttributeCount() const;
+  int GetTextItemCount() const;
 
   virtual absl::Status RDBSave(SafeRDB *rdb) const;
   absl::Status SaveIndexExtension(RDBChunkOutputStream output) const;
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -253,7 +253,6 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
     def test_text_search(self):
         """
         Test FT.SEARCH command with a text index.
-        Tests with both prefilter disabled and enabled.
         """
         client: Valkey = self.server.get_new_client()
         # Create the text index on Hash documents
@@ -1795,7 +1794,7 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert result[1] == b"hash:10"
 
     def test_hybrid_non_vector_query(self):
-        """Test hybrid non-vector queries"""
+        """Test hybrid non-vector queries with deeply nested combinations"""
         client: Valkey = self.server.get_new_client()
         client.execute_command("FT.CREATE", "idx", "ON", "HASH", "PREFIX", "1", "hash:", "SCHEMA",
                              "title", "TEXT", "NOSTEM",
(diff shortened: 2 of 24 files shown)


testing/integration/stability_test.py
```diff
@@ -123,6 +123,8 @@ def tearDown(self):
                 use_coordinator=True,
                 replica_count=1,
                 repl_diskless_load="swapdb",
+                failover_interval_sec=45,
```
daddaman-amz: <placing comment from prev review>

Interval is set at 45 sec so for a 60 second test we will see one failover. I think it's also worth writing a dedicated test which:

Fills cluster to significant load (say 100K or 1M keys)
Performs multiple failovers with longer time periods (e.g., every 2-3 minutes over 15+ minutes)
Validates data consistency and search accuracy throughout
This need not run as part of CI for now if we think it can slow down CI significantly.

Let me know what you think about this

testing/integration/utils.py
```diff
@@ -1017,3 +1062,423 @@ def connect_to_valkey_cluster(
...
 
     assert False
+
+# Cluster Failover Functions
+class ClusterNode(NamedTuple):
+    """Represents a node in the cluster topology."""
+    node_id: str
+    addr: str  # host:port
+    is_master: bool
+    master_id: str | None  # For replicas, the ID of their master
```
daddaman-amz: nit:  you can use primary instead of master wherever possible for all variables

testing/integration/utils.py
```diff
@@ -1017,3 +1062,423 @@ def connect_to_valkey_cluster(
...
+        logging.info("Sending SHUTDOWN NOSAVE to node %s", addr)
+        node_client.execute_command("SHUTDOWN", "NOSAVE")
+    except Exception as e:
+        # Connection drop is EXPECTED and means shutdown succeeded
+        logging.info("Node %s connection dropped (expected after SHUTDOWN): %s", addr, e)
+        return True
+    
+    # If we reach here without exception, something unexpected happened
+    logging.warning("SHUTDOWN command completed without connection drop - unexpected")
+    return True
```
daddaman-amz, line 1297: Should we send false or do a retry?

testing/integration/utils.py
```diff
@@ -1017,3 +1062,423 @@ def connect_to_valkey_cluster(
...
+        stdout_path = os.path.join(stdout_dir, f"{port}_restart_stdout.txt")
+        
+        if not os.path.exists(node_dir):
+            logging.error("Node directory %s does not exist", node_dir)
+            return None
+        
+        logging.info("Restarting node on port %d", port)
+        
+        # Open stdout file for logging
+        stdout_file = open(stdout_path, "w")
```
daddaman-amz: you can use with statement here or ensure the file handle is properly closed

testing/integration/stability_test.py
Nivesh-01: For some of the tests I have made it so that we can do multiple failovers.
I think that some of the tests that other people are doing already handles and with what we are currently trying to do with the tests it do not think we should validate search results

testing/integration/utils.py
Nivesh-01: done

testing/integration/utils.py
Nivesh-01: As of now it does retry and does another failover automatically since the failovers are periodic.

testing/integration/utils.py
Nivesh-01: did not cause the warnings to go away. Moreover this is code already in production so not changing it

testing/integration/stability_test.py
```diff
@@ -165,12 +169,14 @@ def tearDown(self):
                 num_memtier_clients=10,
                 num_search_clients=10,
                 insertion_mode="time_interval",
-                test_time_sec=60,
-                test_timeout=120,
+                test_time_sec=90,
```
daddaman-amz, line 172: Q1) Should they be changed for `hnsw_with_backfill_coordinator_replica`  and `hnsw_with_backfill_no_coordinator_replica` as well? 

Q2)  How many failovers do these tests see in your run?

testing/integration/stability_runner.py
```diff
@@ -140,51 +152,101 @@ def run(self) -> StabilityRunResult:
...
+                    client, self.config.index_name, index_state, self.failover_state,
+                    entry_point_port=self.config.ports[0]
+                ),
+                failover_state=self.failover_state,
             )
+            task.run()
+            threads.append(task)
 
         if self.config.flushdb_interval_sec != 0:
+            task = utils.RandomIntervalTask(
```
daddaman-amz, line 226: Were you able to verify that all existing tests run successfully after these changes  as I understand  that these tests dont run in PR worklfow

src/index_schema.h
```diff
@@ -220,8 +220,11 @@ class IndexSchema : public KeyspaceEventSubscription,
   };
 
   MutationSequenceNumber GetIndexMutationSequenceNumber(const Key &key) const {
+    absl::MutexLock lock(&mutated_records_mutex_);
     auto itr = index_key_info_.find(key);
-    CHECK(itr != index_key_info_.end()) << "Key not found: " << key->Str();
+    if (itr == index_key_info_.end()) {
```
daddaman-amz: Q1) So this is a fix based on what failover testing found?

testing/integration/stability_test.py
Nivesh-01: 2-3 failovers per test work.
I do not think we need to change this to increase time.

testing/integration/stability_runner.py
Nivesh-01: They do not have a workflow yet a PR is out for that as well. As of now none of these have any failures in my 10 or so times I ran after these changes were made.

src/index_schema.h
Nivesh-01: Yes

daddaman-amz: pls double check with brennan on this before merge

BCathcart: This code was added by @KarthikSubbarao recently. Asserts validate the assumptions on which the code is written. If these assumptions are wrong, the code may be fundamentally broken. Removing an assert to get tests passing without understanding the details is very dangerous. Please check with Karthik.

daddaman-amz: @Nivesh-01  pls check on this

KarthikSubbarao: The check here was added in "Eliminate redundant mainthread filter invokations".

See: https://github.com/valkey-io/valkey-search/commit/5a230b259b7a75e83c2b840f9c04dccfb52c8c5f#diff-e8751cf92c30681c53b75ee94cd3f9fa8d2b0c777f0364a705275140dc79aafe based on Allen's PR: https://github.com/valkey-io/valkey-search/pull/545/changes

@allenss-amazon , is this check expected to hold under all circumstances? (ie - have we uncovered a bug during the failover with the Mutation Sequence Number related code or is it safe to make this change?)

src/index_schema.h
```diff
@@ -228,6 +228,11 @@ class IndexSchema : public KeyspaceEventSubscription,
   };
 
   MutationSequenceNumber GetIndexMutationSequenceNumber(const Key &key) const {
+    // absl::MutexLock lock(&mutated_records_mutex_);
```
daddaman-amz: remove commented code?

Nivesh-01: done

src/index_schema.h
KarthikSubbarao: @Nivesh-01 , this thread is not yet resolved.


Are we still seeing occasional failures due to this assert? 

If this, that is the purpose of the endurance test, and we don't want to silently hide issues. 

If this issue is continuing to be observed lets discuss this here and open up an issue to track a fix and raise awareness on any real issue found during stability tests

BCathcart: I hit this failure once running the following endurance test:

```
        dict(
            testcase_name="text_with_backfill_coordinator_replica",
            config=stability_runner.StabilityTestConfig(
                index_name="text_with_backfill",
                ports=(7051, 7052, 7053, 7054),
                index_type="TEXT",
                vector_dimensions=10,  # Dummy value for TEXT index
                bgsave_interval_sec=0,
                ftcreate_interval_sec=0,
                ftdropindex_interval_sec=0,
                flushdb_interval_sec=0,
                randomize_bg_job_intervals=True,
                num_memtier_threads=10,
                num_memtier_clients=10,
                num_search_clients=10,
                insertion_mode="time_interval",
                test_time_sec=60,
                test_timeout=120,
                keyspace_size=1000000,
                use_coordinator=True,
                replica_count=1,
                repl_diskless_load="swapdb",
            ),
        ),
```

I accidentally re-ran the test before saving the engine logs. I'll create an issue when I come across it again.

Nivesh-01: This code was added as a fix to the error index_key_info_ not found and the numops error that were happening. This was not enough and needed a further deep dive. The following PR is trying to fix the crash - https://github.com/valkey-io/valkey-search/pull/793. This is issue for the numops issue - https://github.com/valkey-io/valkey-search/issues/643. Resolved this over here becasue this is going to be tracked separately. Since the tests were working correctly we will be merging it in and tracking the issue separately

PR #613, opened 2026-01-14
String Intern race fix, This is a recreation of PR #576.

allenss-amazon:
Needed to be recreated because of fulltext merge into main.


BCathcart:
Will making all the atomic operations memory_order_seq_cst slow things down? I'm assuming this is defensive programming and may not be strictly necessary in all cases

Looks like the ASAN integration tests are flaky which is probably unrelated to your change

PR #614, opened 2026-01-14
Stemming Ingestion and Search  (resolves #651)

### LINKED ISSUE #651: [FEATURE]  Stemming Support
Implement stemming on ingestion. Verify query operation of stemming is compatible.

Aksha1812:
## Ingestion 

```
happy and happiness stem to -> happi
personal and person stem to -> person 

→ payload: {doc:1 happy, 
            doc:2 happiness, 
            doc:3 happi,
            doc:4 personal,
            doc:5 person
            }
            
Prefix Trie path:                  ()
                                /       \
                           happ      person   
                          /     \          \
                      iness      y          al  
                  
Posting objects to be stored :

 happy : Posting {doc:1},
 happi : Posting {doc:3} ,
 happiness : Posting {doc:2}, 
 personal: Posting {doc:4}, 
 person: Posting {doc:5}, 

Stem Trie path:                    ()
                                /       \
                           happi      person   

Stem Targets to be stored :

happi -> {happy, happiness} 
person -> {personal}

(even 'happi' stems to 'happi' but mapping a word to itslef doesnt make sense)
```

## Search

```
A non exact term search should return 

EXACT (original word) + NONEXACT (stem variants + stem root) PS :  we would only look for stem variants and stem root in the stem enabled fields)

```

## Search Behaviour 

_Exact Term, Prefix, Suffix, fuzzy , verbatime behaviour unchanged. Also earlier during ingestion we used to stem and store the word so we were searching for the stemmed word which was wrong, this has been corrected._

### NonExact Examples

--------------

**Field config** : field1 - STEM enabled
**Sample documents** :  
- product : 1  field1 "happy"
- product : 2  field1 "happi"
- product : 3  field1 "happiness"

**Stem Tree** :  happi -> {happy, happiness} 
**Query / Results** : 

   Query :  FT.SEARCH idx 'happi' 
   Result : product : 1 , product : 2, product : 3  // all stem results returned

--------------

**Field config** : field1 - NOSTEM
**Sample documents** :  
- product : 1  field1 "happy"
- product : 2  field1 "happi"
- product : 3  field1 "happiness"

**Stem Tree** :  Empty
**Query / Results** : 

   Query :  FT.SEARCH idx 'happy' 
   Result : product : 1  // only exact term returned, stem tree empty
   
   📕 **Note** : If we don't specify a field RedisSearch appears to be looking for exact search for stem root as well. i.e for above search it would return both product : 1 and product:2. This would apply for all below examples. We don't match this behaviour. @boda26 

--------------

**Field config** : field1 - STEM enabled, field2 - NOSTEM
**Sample documents** :  
- product : 1  field1 "abc"  field1 "happy"
- product : 2  field1 "happy"  field1 "abc"
- product : 3  field1 "happiness"  field1 "abc"

**Stem Tree** :  happi -> {happy, happiness} 
**Query / Results** : 

   Query :  FT.SEARCH idx 'happy' 
   Result : product : 1 , product : 2, product : 3 // exact search  of happy + stem variants in stem enabled fields 
   
    Query :  FT.SEARCH idx 'happiness' 
   Result : product : 2, product : 3    // exact search  of happiness + stem variants in stem enabled fields 
   
Query :  FT.SEARCH idx 'happi' 
   Result : product : 2, product : 3    // exact search  of happi + stem variants in stem enabled fields 
   

--------------

**Field config** : field1 - STEM enabled , MINSTEMSIZE = 6 (Earlier this was per field but now per schema so pretty straightforward)
**Sample documents** :  
- product : 1  field1 "happy"
- product : 2  field1 "happiness"

**Stem Tree** :  happi -> {happiness} // happy doesn't get stemmed 
**Query / Results** : 

   Query :  FT.SEARCH idx 'happi' 
   Result : product : 2 // only happiness would have been stemmed.
   
 --------------

**Field config** : field1 - STEM enabled
**Sample documents** :  
- product : 1  field1 "very happy today"

**Stem Tree** :  happi -> {happiness} , veri -> {very} ..
**Query / Results** : 

   Query :  FT.SEARCH idx 'very happi today' 
   Result : product : 1 // happy stems to happi , so when we do a non exact phrase search we also look for all stem variants.

### DIFF
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -526,10 +526,6 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         # Verify "happier" stems differently - should only match doc:3, not doc:1 or doc:2
         result = client.execute_command("FT.SEARCH", "idx", '@content:happier')
         assert result[0] == 1 and result[1] == b'doc:3'
-
-        # Test 1b: "happier" can be found via prefix search
-        assert (client.execute_command("FT.SEARCH", "idx", '@content:happ*')[0],
-                set(client.execute_command("FT.SEARCH", "idx", '@content:happ*')[1::2])) == (3, {b"doc:1", b"doc:2", b"doc:3"})
         
         # Test 2: Term search with different stem group (run/running/runs vs runner)
         assert (client.execute_command("FT.SEARCH", "idx", '@content:run')[0],
@@ -547,27 +543,12 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         
         assert client.execute_command("FT.SEARCH", "idx", '@nostem_field:happi')[0] == 0
         
-        result = client.execute_command("FT.SEARCH", "idx", '@nostem_field:happiness')
-        assert result[0] == 1 and result[1] == b'doc:2'
-        assert set(result[2]) == {b'content', b'Happiness is key to success', b'nostem_field', b'happiness'}
-        
-        assert client.execute_command("FT.SEARCH", "idx", '@nostem_field:run')[0] == 0
-        
-        result = client.execute_command("FT.SEARCH", "idx", '@nostem_field:running')
-        assert result[0] == 1 and result[1] == b'doc:4'
-        assert set(result[2]) == {b'content', b'Running every day improves health', b'nostem_field', b'running'}
-        
         # Test 4: Exact phrase (quotes) - NO stem expansion
         result = client.execute_command("FT.SEARCH", "idx", '@content:"very happy today"')
         assert result[0] == 1 and result[1] == b'doc:1'
         assert set(result[2]) == {b'content', b'I am very happy today', b'nostem_field', b'happy'}
         
         assert client.execute_command("FT.SEARCH", "idx", '@content:"very happi today"')[0] == 0
-        
-        # Test exact phrase without stopwords
-        result = client.execute_command("FT.SEARCH", "idx", '@content:"happiness key"')
-        assert result[0] == 1 and result[1] == b'doc:2'
-        assert set(result[2]) == {b'content', b'Happiness is key to success', b'nostem_field', b'happiness'}
         
         # Test 5: Non-exact phrase (no quotes) - DOES use stem expansion
         result = client.execute_command("FT.SEARCH", "idx", '@content:very happi today')
@@ -579,12 +560,7 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert result[0] == 1 and result[1] == b'doc:1'
         assert set(result[2]) == {b'content', b'I am very happy today', b'nostem_field', b'happy'}
         
-        result = client.execute_command("FT.SEARCH", "idx", '@content:happi key success')
-        assert result[0] == 1 and result[1] == b'doc:2'
-        assert set(result[2]) == {b'content', b'Happiness is key to success', b'nostem_field', b'happiness'}
-        
         # Test 5a: Stem expansion with INORDER - verifies stem variants maintain position ordering
-        # Query uses stemmed form "run" which expands to "running", "runs" - should match in order
         result = client.execute_command("FT.SEARCH", "idx", '@content:run every day', "INORDER")
         assert result[0] == 1 and result[1] == b'doc:4'
         assert set(result[2]) == {b'content', b'Running every day improves health', b'nostem_field', b'running'}
@@ -605,29 +581,10 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert result[0] == 1 and result[1] == b'doc:4'
         assert set(result[2]) == {b'content', b'Running every day improves health', b'nostem_field', b'running'}
         
-        # Test 5d: Multiple stem variants with INORDER and SLOP
-        # "running" and "runs" both stem to "run", testing cross-variant proximity
-        # doc:5: "He runs [very] fast" has gap of 1 word in correct order
-        result = client.execute_command("FT.SEARCH", "idx", '@content:run fast', "SLOP", "0", "INORDER")
-        assert result[0] == 0  # Gap too large for SLOP=0
-        result = client.execute_command("FT.SEARCH", "idx", '@content:run fast', "SLOP", "1", "INORDER")
-        assert result[0] == 1 and result[1] == b'doc:5'
-        assert set(result[2]) == {b'content', b'He runs very fast', b'nostem_field', b'runs'}
-        
-        result = client.execute_command("FT.SEARCH", "idx", '@content:run every day')
-        assert result[0] == 1 and result[1] == b'doc:4'
-        assert set(result[2]) == {b'content', b'Running every day improves health', b'nostem_field', b'running'}
-        
-        result = client.execute_command("FT.SEARCH", "idx", '@content:run fast')
-        assert result[0] == 1 and result[1] == b'doc:5'
-        assert set(result[2]) == {b'content', b'He runs very fast', b'nostem_field', b'runs'}
-        
         # Test 6: Prefix wildcard - NO stem expansion (matches literal prefix)
         # "happi*" matches "happier" and "happiness" (both start with "happi")
         assert (client.execute_command("FT.SEARCH", "idx", '@content:happi*')[0],
                 set(client.execute_command("FT.SEARCH", "idx", '@content:happi*')[1::2])) == (2, {b"doc:2", b"doc:3"})
-        assert (client.execute_command("FT.SEARCH", "idx", '@content:run*')[0],
-                set(client.execute_command("FT.SEARCH", "idx", '@content:run*')[1::2])) == (3, {b"doc:4", b"doc:5", b"doc:6"})
         
         # Test 7: Fuzzy search - NO stem expansion (matches by edit distance)
         result = client.execute_command("FT.SEARCH", "idx", '@content:%happy%')
@@ -638,18 +595,12 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert (client.execute_command("FT.SEARCH", "idx", '@content:%%happi%%')[0],
                 set(client.execute_command("FT.SEARCH", "idx", '@content:%%happi%%')[1::2])) == (2, {b"doc:1", b"doc:3"})
         
-        result = client.execute_command("FT.SEARCH", "idx", '@content:%running%')
-        assert result[0] == 1 and result[1] == b'doc:4'
-        assert set(result[2]) == {b'content', b'Running every day improves health', b'nostem_field', b'running'}
-        
         # Test 8: VERBATIM mode - NO stem expansion (exact match only)
         assert client.execute_command("FT.SEARCH", "idx", '@content:happi', "VERBATIM")[0] == 0
         
         result = client.execute_command("FT.SEARCH", "idx", '@content:happy', "VERBATIM")
         assert result[0] == 1 and result[1] == b'doc:1'
         assert set(result[2]) == {b'content', b'I am very happy today', b'nostem_field', b'happy'}
-        
-        assert client.execute_command("FT.SEARCH", "idx", '@content:run', "VERBATIM")[0] == 0
         
         # Test 9: Complex - stem term in AND/OR queries
         result = client.execute_command("FT.SEARCH", "idx", '@content:very @content:happi')
@@ -659,18 +610,13 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert (client.execute_command("FT.SEARCH", "idx", '@content:happi | @content:run')[0],
                 set(client.execute_command("FT.SEARCH", "idx", '@content:happi | @content:run')[1::2])) == (4, {b"doc:1", b"doc:2", b"doc:4", b"doc:5"})
         
-        # Test 10: Stem + prefix combination
-        result = client.execute_command("FT.SEARCH", "idx", '@content:happi @content:succ*')
-        assert result[0] == 1 and result[1] == b'doc:2'
-        assert set(result[2]) == {b'content', b'Happiness is key to success', b'nostem_field', b'happiness'}
-        
-        # Test 11: Validate fix for stem variants NOT matching in NOSTEM fields
+        # Test 10: Validate fix for stem variants NOT matching in NOSTEM fields
         result = client.execute_command("FT.SEARCH", "idx", "swim")
         assert result[0] == 1, f"Expected 1 result for 'swim', got {result[0]}"
         assert result[1] == b"doc:0", f"Expected doc:0, got {result[1]}"
         assert b"doc:9" not in set(result[1::2]), "doc:9 should NOT match (has 'swimming' only in NOSTEM field)"
         
-        # Test 12: Mixed stemming behavior with TEXT and TEXT NOSTEM fields
+        # Test 11: Mixed stemming behavior with TEXT and TEXT NOSTEM fields
         # This test validates default field search (no field specifier) when schema has mixed field types
         client.execute_command("FT.CREATE", "testindex", "ON", "HASH", "PREFIX", "1", "product:", "SCHEMA", "title", "TEXT", "NOSTEM", "desc", "TEXT")
         
@@ -700,7 +646,7 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert result[0] == 3, f"Expected 3 results for 'running' after update, got {result[0]}"
         assert set(result[1::2]) == {b"product:3", b"product:4", b"product:5"}
         
-        # Test 13: Schema-level MINSTEMSIZE - stemming applies to all fields
+        # Test 12: Schema-level MINSTEMSIZE - stemming applies to all fields
         client.execute_command("FT.CREATE", "testindex2", "ON", "HASH", "PREFIX", "1", "product:", 
                              "MINSTEMSIZE", "6",
                              "SCHEMA", "title", "TEXT", "desc", "TEXT")
@@ -723,7 +669,7 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert result[0] == 2, f"Expected 2 results for 'happy', got {result[0]}"
         assert set(result[1::2]) == {b"product:4", b"product:5"}, f"Expected {{product:4, product:5}}, got {set(result[1::2])}"
         
-        # Test 14: NOSTEM fields with stem expansion from other fields
+        # Test 13: NOSTEM fields with stem expansion from other fields
         client.execute_command("FT.CREATE", "testindex_nostem", "ON", "HASH", "PREFIX", "1", "product:", "SCHEMA", "title", "TEXT", "NOSTEM", "desc", "TEXT", "NOSTEM")
         
         client.execute_command("HSET", "product:3", "title", "abc", "desc", "happiness")


src/indexes/text/text_index.h
```diff
@@ -33,6 +33,11 @@ namespace valkey_search::indexes::text {
 using TokenPositions =
     absl::flat_hash_map<std::string, std::pair<PositionMap, bool>>;
 
+// Stem tree target: maps stem root to set of parent words that stem to it
+// Example: "happi" → {"happy", "happiness", "happily"}
+// Using optional so RadixTree can check emptiness via boolean conversion
+using StemParents = std::optional<absl::flat_hash_set<std::string>>;
```
BCathcart: Does this need to be an optional?

src/indexes/text/text_index.cc
```diff
@@ -182,6 +202,20 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
     }
   }
 
+  // Populate stem tree with mappings
+  {
+    std::lock_guard<std::mutex> schema_guard(text_index_mutex_);
```
BCathcart: I'm planning to improve the locking anyways, but in the meantime we could use a separate lock for the prefix/suffix trees and the stem tree

src/commands/filter_parser.cc
```diff
@@ -583,7 +583,8 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseQuotedTextToken(
                                                     field_or_default, false));
   return FilterParser::TokenResult{
       std::make_unique<query::TermPredicate>(text_index_schema, field_mask,
-                                             std::move(token), true),
+                                             std::move(token), true,
+                                             min_stem_size.has_value()),
```
BCathcart: Have we considered what happens when one text field has no stemming while another one does and we are doing a search across all text fields?

src/query/predicate.cc
```diff
@@ -61,27 +62,47 @@ EvaluationResult TermPredicate::Evaluate(
...
-  if (!require_positions) {
-    return EvaluationResult(true);
+
+  // Collect all words to search (original term + stem variants if applicable)
+  std::vector<std::string> words_to_check;
+  words_to_check.push_back(term_);
+
+  // Add stem variants if stemming is enabled
+  if (!exact_ && isstem_) {
+    text_index_schema_->GetAllStemVariants(term_, words_to_check);
```
BCathcart: If this code is for main thread evaluation, we need to consider locking because the stem tree could be getting updated in the background by ingestion at the same time

src/commands/filter_parser.cc
Aksha1812: If that field has no stemming the words inside it wouldn't be stemmed. but you are right I am not checking this while searching. i assume you are talking about a scenario like this 

```
127.0.0.1:6380> FT.CREATE testindex ON HASH PREFIX 1 "product:" SCHEMA title TEXT NOSTEM desc TEXT
OK
127.0.0.1:6380> HSET product:3 title "running" desc "abc"
(integer) 2
127.0.0.1:6380> HSET product:4 title "abc" desc "run"
(integer) 2
127.0.0.1:6380> HSET product:5 title "abc" desc "running"
(integer) 2
127.0.0.1:6380> FT.SEARCH testindex 'run' dialect 2
1) (integer) 3
2) "product:3"
3) 1) "title"
   2) "running"
   3) "desc"
   4) "abc"
4) "product:4"
5) 1) "title"
   2) "abc"
   3) "desc"
   4) "run"
6) "product:5"
7) 1) "title"
   2) "abc"
   3) "desc"
   4) "running"
```

redis handles it but explaincli doesn't paint the full picture 

```
127.0.0.1:6379> FT.CREATE testindex ON HASH PREFIX 1 "product:" SCHEMA title TEXT NOSTEM desc TEXT
OK
127.0.0.1:6379>  HSET product:3 title "running" desc "abc"
(integer) 2
127.0.0.1:6379> HSET product:4 title "abc" desc "run"
(integer) 2
127.0.0.1:6379> HSET product:5 title "abc" desc "running"
(integer) 2
127.0.0.1:6379> FT.SEARCH testindex 'run' dialect 2
1) (integer) 2
2) "product:4"
3) 1) "title"
   2) "abc"
   3) "desc"
   4) "run"
4) "product:5"
5) 1) "title"
   2) "abc"
   3) "desc"
   4) "running"
127.0.0.1:6379> FT.explaincli testindex 'run' dialect 2
1) UNION {
2)   run
3)   +run(expanded)
4) }
5)
```

Aksha1812: maybe maintaining a field mask indicating stemming enabled fields might be a simple solution . I am thinking if there is any other way. predicate Evaluator should also be able to check this.

BCathcart: Yeah that was exactly the situation I was thinking of. I was talking to @boda26 last week and I'm thinking it might be good to add compatibility testing with stemming in this PR since he says it should be easy to enable. Might as well catch these issues up front. Right now we do compatibility testing with NOSTEM. I expect we can run all the existing tests with simply dropping the NOSTEM flag. We'd also want to test when there is one text field with stemming and one without (the situation here) which might take a bit more setup.

src/indexes/text/text_index.h
Aksha1812: yes it is the same as Target for prefix tree. when empty optional returns false then we delete the stem root from the stem tree.

src/commands/filter_parser.cc
Aksha1812: this will be fixed with next change. adding test for this case as well.

src/indexes/text/text_index.cc
Aksha1812: Are the locks for prefix and suffix tree already implemented.  I don't see them.

src/indexes/text/text_index.h
BCathcart: The problem with in optional is that when it's none, it still needs to reserve enough space for a flat_hash_map (plus the flag indicating if it's set or not). I'd use a unique pointer here. It's only 8 bytes of overhead when it's none. A separate memory allocation will be needed anyways when we move to Rax which opaquely holds a `void *` to the provided target. Rax is also smarter and only uses a single bit to indicate "no target" instead of a nullptr.

src/indexes/text/lexer.cc
```diff
@@ -168,14 +175,14 @@ std::string Lexer::StemWord(const std::string& word, bool stemming_enabled,
...
-  DCHECK(stemmer) << "Stemmer is null";
+  CHECK(stemmer) << "Stemmer is not initialized";
 
   const sb_symbol* stemmed = sb_stemmer_stem(
       stemmer, reinterpret_cast<const sb_symbol*>(word.c_str()), word.length());
 
-  DCHECK(stemmed) << "Stemming failed for word: " + word;
-
   int stemmed_length = sb_stemmer_length(stemmer);
+  CHECK(stemmed_length < min_stem_size || stemmed) << "Stemming failed";
```
BCathcart: Shouldn't the stemmed length be longer than the minimum stem size? This check in generally doesn't make much sense to me. I assume you're going for asserting that there is a valid value AND it's a reasonable length?

src/indexes/text/lexer.cc
```diff
@@ -168,14 +175,14 @@ std::string Lexer::StemWord(const std::string& word, bool stemming_enabled,
     return word;
   }
 
-  DCHECK(stemmer) << "Stemmer is null";
+  CHECK(stemmer) << "Stemmer is not initialized";
 
   const sb_symbol* stemmed = sb_stemmer_stem(
       stemmer, reinterpret_cast<const sb_symbol*>(word.c_str()), word.length());
 
-  DCHECK(stemmed) << "Stemming failed for word: " + word;
```
BCathcart, line 176: I wouldn't remove this check here. I'm not sure what happens when you pass a null `stemmer` pointer to `sb_stemmer_length`. Might as well fail fast.

src/indexes/text/lexer.cc
```diff
@@ -140,8 +143,12 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
         continue;  // Skip stop words
       }
 
-      word = StemWord(word, stemming_enabled, min_stem_size, stemmer);
+      std::string stemmed_word =
+          StemWord(word, stemming_enabled, min_stem_size, stemmer);
 
+      if (stem_mappings && word != stemmed_word) {
```
BCathcart: If we have a null `stem_mappings` here we have a big coding bug. I wouldn't silently drop the stem mapping.

src/indexes/text/text_index.cc
```diff
@@ -220,6 +259,27 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
       std::string reverse_word(word.rbegin(), word.rend());
       suffix_opt.value().get().SetTarget(reverse_word, new_target);
     }
+
+    // If the postings are now empty, remove from stem tree if it was a parent
+    if (!new_target) {
+      // Check if this word has a stem mapping
+      std::string stemmed =
+          lexer_.StemWord(std::string(word), true, 4, lexer_.GetStemmer());
```
BCathcart: Why are we using a hardcoded `min_stem_size`?

I think we need to reverse engineer the lookup. An easy solution could be to loop over all the fields that have stemming enabled, and call `StemWord` for each different `min_stem_size`.

src/indexes/text/text_index.cc
```diff
@@ -220,6 +259,27 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
...
+      if (stemmed != word) {
+        // This word was a stem parent, remove it from stem tree
+        std::lock_guard<std::mutex> stem_guard(stem_tree_mutex_);
+        stem_tree_.MutateTarget(stemmed, [&](StemParents existing) {
+          if (existing && !existing->empty()) {
```
BCathcart: It shouldn't ever be empty should it?

src/indexes/text/text_index.cc
```diff
@@ -258,4 +318,34 @@ uint64_t TextIndexSchema::GetTotalTextIndexMemoryUsage() const {
   return metadata_.text_index_memory_pool_.GetUsage();
 }
 
+std::string TextIndexSchema::GetAllStemVariants(
+    const std::string &search_term, std::vector<std::string> &words_to_search) {
+  // Stem the search term
+  std::string stemmed =
+      lexer_.StemWord(search_term, true, 4, lexer_.GetStemmer());
```
BCathcart: Same thing here. A hard-coded `min_stem_size` isn't what we want.

src/query/predicate.cc
```diff
@@ -61,27 +61,57 @@ EvaluationResult TermPredicate::Evaluate(
...
+    while (!word_iter.Done()) {
+      if (word_iter.GetWord() == word) {
+        auto postings = word_iter.GetTarget();
+        if (postings) {
+          auto key_iter = postings->GetKeyIterator();
+          // Skip to target key and verify it contains the appropriate fields
+          // Original term uses all fields, stem variants only use stemmable
+          // fields
+          if (key_iter.SkipForwardKey(target_key) &&
+              key_iter.ContainsFields((word == term_ || word == stem_root)
```
BCathcart: I'm going to look closer at the search-side logic later, but can you quickly summarize to me what's going on here?

src/indexes/text.cc
```diff
@@ -179,20 +179,47 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
       key_iterators;
-  while (!word_iter.Done()) {
-    if (word_iter.GetWord() == GetTextString()) {
-      key_iterators.emplace_back(word_iter.GetTarget()->GetKeyIterator());
+  absl::InlinedVector<FieldMaskPredicate,
+                      indexes::text::kWordExpansionInlineCapacity>
+      field_masks;
+
+  // Collect all words to search
+  std::vector<std::string> words_to_search;
```
KarthikSubbarao: Any reason why this can't be `std::string_view`? This comment applies to all the `std::string` usages below

src/indexes/text.cc
```diff
@@ -179,20 +179,47 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
                       indexes::text::kWordExpansionInlineCapacity>
       key_iterators;
-  while (!word_iter.Done()) {
-    if (word_iter.GetWord() == GetTextString()) {
-      key_iterators.emplace_back(word_iter.GetTarget()->GetKeyIterator());
+  absl::InlinedVector<FieldMaskPredicate,
+                      indexes::text::kWordExpansionInlineCapacity>
+      field_masks;
+
+  // Collect all words to search
```
KarthikSubbarao: The new code added here seems common to what is added in predicate.cc. Is there part of it which can be extracted into a common file?

src/indexes/text.cc
```diff
@@ -179,20 +179,47 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
+                : stem_variant_field_mask);
+      }
+      word_iter.Next();
     }
-    word_iter.Next();
   }
   return std::make_unique<indexes::text::TermIterator>(
       std::move(key_iterators), fetcher->field_mask_, fetcher->untracked_keys_,
-      fetcher->require_positions_);
+      fetcher->require_positions_, std::move(field_masks));
```
KarthikSubbarao: Question: Is there a benefit to providing both `fetcher->field_mask_` and `std::move(field_masks)`?

src/indexes/text.cc
```diff
@@ -179,20 +179,47 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
+  // Search for all words and track which field mask to use for each
+  for (const auto& word : words_to_search) {
+    auto word_iter = fetcher->text_index_->GetPrefix().GetWordIterator(word);
+    while (!word_iter.Done()) {
+      if (word_iter.GetWord() == word) {
+        key_iterators.emplace_back(word_iter.GetTarget()->GetKeyIterator());
+        // For original term and stem root, use full query field mask
+        // For stem variants, use only fields with stemming enabled
+        field_masks.emplace_back(
+            (word == std::string(GetTextString()) || word == stem_root)
```
KarthikSubbarao: Why are we creating a new string every time we are comparing against `word`?

src/indexes/text/text_index.cc
KarthikSubbarao: @Aksha1812 - Please see the code that was used in the filter_parser.cc. It correctly calculated the min stem size which should be used here

src/indexes/text/text_index.cc
```diff
@@ -258,4 +318,34 @@ uint64_t TextIndexSchema::GetTotalTextIndexMemoryUsage() const {
...
+
+  // If stemmed form is different, add it to search list
+  if (stemmed != search_term) {
+    words_to_search.push_back(stemmed);
```
KarthikSubbarao: Question: Is there any hint on how many words to search? If yes, it will help to `reserve`

src/indexes/text.cc
```diff
@@ -179,20 +179,47 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
+  if (!IsExact() && stem_variant_field_mask != 0) {
+    stem_root = GetTextIndexSchema()->GetAllStemVariants(
+        std::string(GetTextString()), words_to_search);
+  }
+
+  // Search for all words and track which field mask to use for each
+  for (const auto& word : words_to_search) {
+    auto word_iter = fetcher->text_index_->GetPrefix().GetWordIterator(word);
+    while (!word_iter.Done()) {
+      if (word_iter.GetWord() == word) {
```
KarthikSubbarao: 1. Will we not be on the correct word on the first iteration in this while loop? If no, can we seek?

2. Once we finished the iteration, why do we need to continue iterating until the word iterator is done? Can we not break from the while loop?

src/indexes/text/text_index.cc
```diff
@@ -258,4 +318,34 @@ uint64_t TextIndexSchema::GetTotalTextIndexMemoryUsage() const {
   return metadata_.text_index_memory_pool_.GetUsage();
 }
 
+std::string TextIndexSchema::GetAllStemVariants(
```
KarthikSubbarao, line 321: This is an odd function name. It returns the StemRoot, right? Should it not be `GetStemRoot`?

src/indexes/text/text_index.cc
```diff
@@ -258,4 +318,34 @@ uint64_t TextIndexSchema::GetTotalTextIndexMemoryUsage() const {
...
+    if (stem_iter.GetWord() == stemmed) {
+      const auto &parents_opt = stem_iter.GetTarget();
+      if (parents_opt.has_value()) {
+        const auto &parents = parents_opt.value();
+        words_to_search.insert(words_to_search.end(), parents.begin(),
+                               parents.end());
+      }
+      break;
+    }
+    stem_iter.Next();
```
KarthikSubbarao: Will we not be at the correct word at the first iteration? If no, should we seek instead of repeated next calls?

src/indexes/text/text_index.cc
```diff
@@ -258,4 +318,34 @@ uint64_t TextIndexSchema::GetTotalTextIndexMemoryUsage() const {
...
+      lexer_.StemWord(search_term, true, 4, lexer_.GetStemmer());
+
+  // If stemmed form is different, add it to search list
+  if (stemmed != search_term) {
+    words_to_search.push_back(stemmed);
+  }
+
+  // Look up parent words in stem tree
+  // Lock to prevent concurrent updates during read
+  std::lock_guard<std::mutex> stem_guard(stem_tree_mutex_);
```
KarthikSubbarao: How well does this locking work if it is being used from both main thread and reader threads / writer threaders?

src/query/predicate.cc
```diff
@@ -61,27 +61,57 @@ EvaluationResult TermPredicate::Evaluate(
...
-  auto key_iter = postings->GetKeyIterator();
-  // Skip to target key and verify it contains the required fields
-  if (!key_iter.SkipForwardKey(target_key) ||
-      !key_iter.ContainsFields(field_mask)) {
-    return EvaluationResult(false);
-  }
-  if (!require_positions) {
-    return EvaluationResult(true);
+
+  // Collect all words to search (original term + stem variants if applicable)
```
KarthikSubbarao: I commented on text.cc where this code is implemented based on. The comments are the same for this file , but we can have our discussion there first

src/indexes/text/text_index.cc
KarthikSubbarao: Yes, we should not be using a hard coded value. this is field specific, unless we change the API contract and make it schema specific

src/indexes/text/lexer.cc
Aksha1812: actually stemmed_length has no correlation with min_stem_size. it is only the original word that should be greater than min_stem_size to continue stemming. I guess a better sanity check for the stemmed_length would be if it is greater than 0 and less than equal to word.length(). will fix it.

src/indexes/text/lexer.cc
Aksha1812: replaced with a CHECK

src/indexes/text/text_index.cc
Aksha1812: my bad. actually this is a good point i hadn't considered different fields having different min_stem_size, since this clause is not in redis. examples like below could be very tricky.

```
127.0.0.1:6380> FT.CREATE testindex ON HASH PREFIX 1 "product:" SCHEMA title TEXT MINSTEMSIZE 6 desc TEXT
OK
127.0.0.1:6380> HSET product:3 title "happiness" desc "abc"
(integer) 2
127.0.0.1:6380> HSET product:4 title "happy" desc "run"
(integer) 2
127.0.0.1:6380> FT.SEARCH testindex 'happi' dialect 2
1) (integer) 1
2) "product:3"
3) 1) "desc"
   2) "abc"
   3) "title"
   4) "happiness"
127.0.0.1:6380> HSET product:5 title "abc" desc "happy"
(integer) 2
127.0.0.1:6380> FT.SEARCH testindex 'happi' dialect 2
1) (integer) 3
2) "product:3"
3) 1) "desc"
   2) "abc"
   3) "title"
   4) "happiness"
4) "product:4"
5) 1) "desc"
   2) "run"
   3) "title"
   4) "happy"
6) "product:5"
7) 1) "desc"
   2) "happy"
   3) "title"
   4) "abc"
127.0.0.1:6380> 
```

'title' has minstemsize 6 and 'desc' has minstemsize 4 (default). In the first search result it works well since 'happy' in 'title' is not stemmed and not added to the stem tree. but when we add 'happy' to 'desc' field. It gets stemmed and added to the stem tree. And when we search we get both results because we look for the stemmed word on all fields having stem enabled.

For this just fixing the StemWord API call wouldn't work since we cannot call it again and again for minstemsizes per field and create termiterators for them for every field. 

- What we can do is just get all the stem variants with minstemsize = minstemsize among all fields. 
- Instead of a uint64_t stemming_enabled_fields_ = 0; maintain a array with min stem sizes for each field.
- use it to generate a more smarter stem_variant_field_mask so that it only looks for the fields having stem_variant.length() > min_stem_size for that field.

src/indexes/text/text_index.cc
Aksha1812: It doesn't just return Stem root it updates words_to_search with all stem parents of that root. returning stem root separately because when searched for non exact term redis search looks for original and stem root in all fields and other stem parents in corresponding stem fields . So this technically returns stem root + all stem parents.

src/query/predicate.cc
Aksha1812: we can discuss offline tomorrow

src/indexes/text/text_index.cc
Aksha1812: we can get the length of the Target to get number of Stem Parents. If you are asking generally , in common usage in English <10 words. more extreme case 10-20 stem parents could be there. more than that is pretty rare. this is for actual words. will have to check what is upper bound base don snowball rules if we enter gibberish

src/indexes/text/text_index.cc
Aksha1812: how do i test that. perf tests?

src/indexes/text.cc
Aksha1812: This part is the same it was before. I just added loop over word_to_search and update logic inside. I can update this if you want.

src/indexes/text.cc
Aksha1812: actually for most cases TermIterator doesn't need std::move(field_masks)  . only fetcher->field_mask_ is used. this std::move(field_masks) is used only for stemming. and optional to provide.

src/indexes/text.cc
```diff
@@ -179,20 +179,47 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
+      fetcher->field_mask_ & GetTextIndexSchema()->GetStemmingFieldMask();
+
+  std::string stem_root;
+  if (!IsExact() && stem_variant_field_mask != 0) {
+    stem_root = GetTextIndexSchema()->GetAllStemVariants(
+        std::string(GetTextString()), words_to_search);
+  }
+
+  // Search for all words and track which field mask to use for each
+  for (const auto& word : words_to_search) {
```
KarthikSubbarao: Let's limit the words to search by the max expansions config to avoid too many words. See how the prefix/suffix predicate handle this.

src/indexes/text.cc
KarthikSubbarao: So, does an Enum help? Just meant that we don't need two separate constructor members. They serve the same purpose

src/indexes/text/text_index.cc
KarthikSubbarao: If we know the exact size of `words_to_search`, then we should be able `reserve` here to avoid expansions then, right?

src/indexes/text.cc
KarthikSubbarao: Yeah, if you can think about the two questions wrt to the current code in this PR and update/reply here, that would be great

src/indexes/text.cc
Aksha1812: I changed the vector words_to_search to accomodate string_view. GetAllStemVariants func still returns a string and then we add its string view to words_to_search because when a word is stemmed regardless of it being in the stem tree or not we look for exact term of the query_term itself and the stemmed_root . In example below none of the words should be stemmed but when we search for one of the stem parents we also get stemmed root in results.
```
127.0.0.1:6379> flushall
OK
127.0.0.1:6379> FT.CREATE testindex ON HASH PREFIX 1 "product:" SCHEMA title TEXT NOSTEM desc TEXT
OK
127.0.0.1:6379>  HSET product:3 title "running" desc "abc"
(integer) 2
127.0.0.1:6379> HSET product:5 title "run" desc "abc"
(integer) 2
127.0.0.1:6379> FT.SEARCH testindex 'running' dialect 2
1) (integer) 2
2) "product:3"
3) 1) "title"
   2) "running"
   3) "desc"
   4) "abc"
4) "product:5"
5) 1) "title"
   2) "run"
   3) "desc"
   4) "abc"
```

So in this case there is no ownership for the stem_root to transfer. We can sync if not clear

src/indexes/text/text_index.cc
Aksha1812: no i can replace with a CHECK

src/commands/filter_parser.cc
```diff
@@ -756,9 +756,6 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseUnquotedTextToken(
     }
     VMSDK_RETURN_IF_ERROR(SetupTextFieldConfiguration(field_mask, min_stem_size,
                                                       field_or_default, false));
-    if (!exact && min_stem_size.has_value()) {
-      token = lexer.StemWord(token, true, *min_stem_size, lexer.GetStemmer());
```
KarthikSubbarao: `min_stem_size` is unused now in filter_parser.cc

We can actually remove it from the function parameter here, and have `SetupTextFieldConfiguration` cleaned up to not calculate and return this

src/indexes/text.cc
```diff
@@ -201,17 +201,58 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
+  // views
+  absl::string_view stem_root;
+  if (!IsExact()) {
+    uint64_t stem_enabled_mask =
+        fetcher->field_mask_ & GetTextIndexSchema()->GetStemmingEnabledFields();
+
+    stemmed = GetTextIndexSchema()->GetAllStemVariants(
+        GetTextString(), words_to_search,
+        GetTextIndexSchema()->GetMinStemSize(), stem_enabled_mask, false);
+    if (!stemmed.empty() && stemmed != std::string(GetTextString())) {
```
KarthikSubbarao: can we remove the `std::string()` call here?

src/indexes/text.cc
```diff
@@ -201,17 +201,58 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
                       indexes::text::kWordExpansionInlineCapacity>
       key_iterators;
-  if (!word_iter.Done() && word_iter.GetWord() == GetTextString()) {
-    key_iterators.emplace_back(word_iter.GetTarget()->GetKeyIterator());
+  absl::InlinedVector<FieldMaskPredicate,
+                      indexes::text::kWordExpansionInlineCapacity>
+      field_masks;
+
+  // stemmed word
+  std::string stemmed;
```
KarthikSubbarao: There does not seem to be any usage of this in the overall function scope. Can we remove this and move it into the `if (!IsExact())` block?

src/indexes/text/text_index.cc
```diff
@@ -258,4 +329,52 @@ uint64_t TextIndexSchema::GetTotalTextIndexMemoryUsage() const {
   return metadata_.text_index_memory_pool_.GetUsage();
 }
 
+std::string TextIndexSchema::GetAllStemVariants(
+    absl::string_view search_term,
+    std::vector<absl::string_view> &words_to_search, uint32_t min_stem_size,
```
KarthikSubbarao: I had commented earlier if we know  (from the stem tree) the size needed for `words_to_search` to help us reserve a capacity for the vector before multiple push_back calls.

src/indexes/text/text_index.cc
```diff
@@ -93,7 +94,30 @@ absl::StatusOr<bool> TextIndexSchema::StageAttributeData(
     size_t text_field_number, bool stem, size_t min_stem_size, bool suffix) {
   NestedMemoryScope scope{metadata_.text_index_memory_pool_};
 
-  auto tokens = lexer_.Tokenize(data, stem, min_stem_size);
+  // Track which fields have stemming enabled and their min stem size
```
KarthikSubbarao: Is there some duplicated responsibilities of this and UpdateTextFieldMasksForIndex?
Also, why do we have to calculate the min stem size each time we ingest a field of a document? UpdateTextFieldMasksForIndex caches the min stem size during index creation

integration/test_fulltext.py
```diff
@@ -469,8 +468,8 @@ def test_default_tokenization(self):
             ("effect*", True, "Case insensitivity - lowercase matches uppercase"),
             ("\"The quick-running searches are finding EFFECTIVE results!\"", False, "Stop word cannot be used in exact phrase searches"),
             # TODO: Change to True once the stem tree is supported and ingestion is updated.
```
BCathcart: Let's remove the TODO

integration/test_fulltext.py
```diff
@@ -483,27 +482,229 @@ def test_default_tokenization(self):
...
-
-        # We cannot find stems on 'content' with NOSTEM
-        assert client.execute_command("FT.SEARCH", "idx", '@content:"run"') == [0]
-
-        # We can find original words in both cases
-        assert client.execute_command("FT.SEARCH", "idx", '@title:"running"') == expected # TODO: fails here
-        assert client.execute_command("FT.SEARCH", "idx", '@content:"running"') == expected
+        
+        # Test 1: Term search finds multiple stem parents (happy, happiness share stem "happi")
+        # Note: "happier" stems differently and is not in the "happi" stem group
```
BCathcart, line 518: I'd prove this in the code with a test that a "happier" search doesn't return doc1 or doc2 rather than just leaving a comment

integration/test_fulltext.py
```diff
@@ -483,27 +482,229 @@ def test_default_tokenization(self):
...
+        assert set(result[2]) == {b'content', b'Running every day improves health', b'nostem_field', b'running'}
+        
+        # Test 4: Exact phrase (quotes) - NO stem expansion
+        result = client.execute_command("FT.SEARCH", "idx", '@content:"very happy today"')
+        assert result[0] == 1 and result[1] == b'doc:1'
+        assert set(result[2]) == {b'content', b'I am very happy today', b'nostem_field', b'happy'}
+        
+        assert client.execute_command("FT.SEARCH", "idx", '@content:"very happi today"')[0] == 0
+        
+        # Test exact phrase without stopwords
```
BCathcart: Just for my understanding, we ignore stop words when ingesting positions? I assume we've tested this against redis behaviour?

integration/test_fulltext.py
```diff
@@ -483,27 +482,229 @@ def test_default_tokenization(self):
...
+        result = client.execute_command("FT.SEARCH", "idx", '@content:"happiness key"')
+        assert result[0] == 1 and result[1] == b'doc:2'
+        assert set(result[2]) == {b'content', b'Happiness is key to success', b'nostem_field', b'happiness'}
+        
+        # Test 5: Non-exact phrase (no quotes) - DOES use stem expansion
+        result = client.execute_command("FT.SEARCH", "idx", '@content:very happi today')
```
BCathcart, line 554: Can you also add a test for `@content:very happiness today` where you search for a different term but with the same stem?

integration/test_fulltext.py
```diff
@@ -483,27 +482,229 @@ def test_default_tokenization(self):
...
+        assert result[0] == 1 and result[1] == b'doc:2'
+        assert set(result[2]) == {b'content', b'Happiness is key to success', b'nostem_field', b'happiness'}
+        
+        # Test 11: Validate fix for stem variants NOT matching in NOSTEM fields
+        result = client.execute_command("FT.SEARCH", "idx", "swim")
+        assert result[0] == 1, f"Expected 1 result for 'swim', got {result[0]}"
+        assert result[1] == b"doc:0", f"Expected doc:0, got {result[1]}"
+        assert b"doc:9" not in set(result[1::2]), "doc:9 should NOT match (has 'swimming' only in NOSTEM field)"
+        
+        # Test 12: Mixed stemming behavior with TEXT and TEXT NOSTEM fields
```
BCathcart: I'd be clear in this comment you're ~~not~~ now testing the search across all text fields at once

integration/test_fulltext.py
```diff
@@ -483,27 +482,229 @@ def test_default_tokenization(self):
...
+        client.execute_command("HSET", "product:4", "title", "abc", "desc", "run")
+        client.execute_command("HSET", "product:5", "title", "abc", "desc", "running")
+        
+        IndexingTestHelper.wait_for_backfill_complete_on_node(client, "testindex")
+        
+        result = client.execute_command("FT.SEARCH", "testindex", "run", "DIALECT", "2")
+        assert result[0] == 3, f"Expected 3 results for 'run', got {result[0]}"
+        assert set(result[1::2]) == {b"product:3", b"product:4", b"product:5"}
+        
+        result = client.execute_command("FT.SEARCH", "testindex", "running", "DIALECT", "2")
```
BCathcart, line 633: It feels a bit odd that when we search for "running, we get a match on "run" in a NOSTEM field. I know you're matching Redis behaviour here but maybe we should reconsider if this makes sense.

Also just a thought, but we could add a word that stems to itself to the stem tree (e.g. run -> [running, runs, run]) and then we'd only have to bother searching for the stem if we know it actually exists in the corpus. Quite a lot of times the stem won't be a valid word and will never lead to a match in the corpus. It's a time-space tradeoff.

integration/test_fulltext.py
```diff
@@ -483,27 +482,229 @@ def test_default_tokenization(self):
...
+        assert result[0] == 3, f"Expected 3 results for 'running', got {result[0]}"
+        assert set(result[1::2]) == {b"product:3", b"product:4", b"product:5"}
+        
+        # Update product:3 to have "running" in NOSTEM title field
```
BCathcart, line 637: Do we need this section? Looks like it might be duplicating what's covered in test 11 and first part of test 12

src/indexes/text.cc
```diff
@@ -201,17 +201,58 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
+        GetTextString(), words_to_search,
+        GetTextIndexSchema()->GetMinStemSize(), stem_enabled_mask, false);
+    if (!stemmed.empty() && stemmed != std::string(GetTextString())) {
+      stem_root = stemmed;
+      words_to_search.push_back(stem_root);
+    }
+  }
+
+  // Search for all words and track which field mask to use for each
+  for (const auto& word : words_to_search) {
```
BCathcart: On index creation, you could store a separate field mask for each min-stem-size bucket. e.g. if you have text fields with min-stem-size 2, 4, and 5, you would have three field masks with each one being a subset of the previous. Then you can look up the field-mask based on the word size rather than looping over the fields and creating it each time.

src/valkey_search_options.cc
```diff
@@ -326,6 +326,19 @@ static auto max_term_expansions =
                           kMaximumMaxTermExpansions)  // max limit (100k)
         .Build();
 
+/// Register the "--max-stem-expansions" flag. Controls the maximum number of
+/// stem parent words to expand in term search
+constexpr absl::string_view kMaxStemExpansionsConfig{"max-stem-expansions"};
+constexpr uint32_t kDefaultMaxStemExpansions{10};  // Default 10 parents
```
BCathcart: @KarthikSubbarao I see the default number of terms for prefix/suffix/fuzzy above is 200. Would we want it to be the same for stem expansion? The number of words we get from the stem tree (plus the original word) is the number of key iterators we get so it's pretty similar right? I wouldn't ever expect us to hit this max anyways, but maybe we could hit 10 in certain cases and I'm not really sure why we'd limit it there

src/indexes/text.cc
```diff
@@ -201,17 +201,58 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
+  // Get stem variants - GetAllStemVariants returns owned string, adds parent
+  // views
+  absl::string_view stem_root;
+  if (!IsExact()) {
+    uint64_t stem_enabled_mask =
+        fetcher->field_mask_ & GetTextIndexSchema()->GetStemmingEnabledFields();
+
+    stemmed = GetTextIndexSchema()->GetAllStemVariants(
+        GetTextString(), words_to_search,
+        GetTextIndexSchema()->GetMinStemSize(), stem_enabled_mask, false);
```
BCathcart: Should we filter out the fields based on the length of the search term as well? Right now we look up the term variants that match the stem of this word based on the smallest min_stem_size across all text fields. For example, there is a search for "happy" and we find "happiness" in the stem tree. There is text field with min_stem_size of 6 and a doc has the word "happiness" in it, but the min_stem_size across all text fields is 4. Currently we would match on "happiness" for that doc because "happiness".len() >= 6 and we looked up "happiness" in the stem tree because min_stem_length across all fields is 4. However, the original search term "happy" wouldn't be stemmed and we wouldn't match if min_stem_size was 6 across all text fields.

If we're going to keep the min_stem_size on a per field basis, I think we need to think on a per-field basis across the board. That means only looking up terms in the stem tree to search on that field if the search term meets the min stem requirement, and then also filtering out the terms found in the stem tree that don't meet the min stem requirement.

src/indexes/text/text_index.cc
Aksha1812: Above conversation is irrelevant now . moved minstemsize to schema level

src/indexes/text/lexer.cc
```diff
@@ -140,8 +143,13 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
         continue;  // Skip stop words
       }
 
-      word = StemWord(word, stemming_enabled, min_stem_size, stemmer);
+      std::string stemmed_word =
+          StemWord(word, stemming_enabled, min_stem_size, stemmer);
 
+      if (word != stemmed_word) {
```
BCathcart: I'd personally move the check for stemming enabled outside of `StemWord`. We're doing a function call and a string comparison to determine if we need to do something we already have a boolean for. Or maybe keep keep the check in there alongside the min stem size check but change the return type to `std::optional<std::string> ` so it's easy to know if stemming was rejected

src/indexes/text.h
```diff
@@ -10,13 +10,10 @@
...
 #include <memory>
 
-namespace valkey_search::indexes::text {
-// Inline capacity for word expansion key iterators
-constexpr size_t kWordExpansionInlineCapacity = 200;
-// Inline capacity for proximity terms
-constexpr size_t kProximityTermsInlineCapacity = 64;
-}  // namespace valkey_search::indexes::text
-
+// need to import text index before other files but clang reorders it
```
BCathcart: Why is this required?

src/indexes/text.cc
```diff
@@ -187,17 +187,47 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
                       indexes::text::kWordExpansionInlineCapacity>
       key_iterators;
-  if (!word_iter.Done() && word_iter.GetWord() == GetTextString()) {
-    key_iterators.emplace_back(word_iter.GetTarget()->GetKeyIterator());
+
+  // Collect all words to search: original word first
+  absl::InlinedVector<absl::string_view,
+                      indexes::text::kWordExpansionInlineCapacity>
+      words_to_search;
+  words_to_search.push_back(GetTextString());
```
BCathcart: Let's call `GetTextString()` once in this function and assign it to a variable

src/indexes/text.cc
```diff
@@ -187,17 +187,47 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
...
+    }
   }
+
+  // Search for all words (original word first, then stem variants)
+  bool found_original = false;
+  for (const auto& word : words_to_search) {
+    auto word_iter = fetcher->text_index_->GetPrefix().GetWordIterator(word);
+    if (!word_iter.Done() && word_iter.GetWord() == word) {
+      key_iterators.emplace_back(word_iter.GetTarget()->GetKeyIterator());
+      if (word == GetTextString()) {  // First word is the original
```
BCathcart: Should we separate the logic for the original word and the stem variant? Might make it easier to read and we do have different assumptions we could validate. The original word may or may not exist in the corpus, but all the stem variants are expected to exist and therefore something has gone wrong on the ingestion side or with the word iterator if it hasn't found a postings object.

src/query/predicate.cc
```diff
@@ -61,29 +61,65 @@ EvaluationResult TermPredicate::Evaluate(
...
-  auto key_iter = postings->GetKeyIterator();
-  // Skip to target key and verify it contains the required fields
-  if (!key_iter.SkipForwardKey(target_key) ||
-      !key_iter.ContainsFields(field_mask)) {
-    return EvaluationResult(false);
-  }
-  if (!require_positions) {
-    return EvaluationResult(true);
+
+  // Collect all words to search: original word first
```
BCathcart: It's a similar pattern to text.cc, can we re-use some more code between them?

src/indexes/text.h
Aksha1812: I had to move some capacity constants to text_index.h from text.h . We import text_index.h in text.h . but other files imported in text.h do so as well. but clang orders the imports alphabetically. we causes these constants to be not imported in other files. I know it looks a bit suspicious but i saw other places in the code follow similar pattern when similar problem occurs

integration/test_fulltext.py
```diff
@@ -483,27 +479,233 @@ def test_default_tokenization(self):
...
+        assert set(result[2]) == {b'content', b'I am very happy today', b'nostem_field', b'happy'}
+        
+        assert client.execute_command("FT.SEARCH", "idx", '@content:"very happi today"')[0] == 0
+        
+        # Test exact phrase without stopwords
+        result = client.execute_command("FT.SEARCH", "idx", '@content:"happiness key"')
+        assert result[0] == 1 and result[1] == b'doc:2'
+        assert set(result[2]) == {b'content', b'Happiness is key to success', b'nostem_field', b'happiness'}
+        
+        # Test 5: Non-exact phrase (no quotes) - DOES use stem expansion
```
BCathcart, line 553: Can we test with INORDER and SLOP as well?

src/query/predicate.cc
Aksha1812: it is similar pattern but they use of key iterators are different also the stemmed string is created here . if it is declared in a helper function. we wouldn't be able to pass a string view  for it.

BCathcart:
We're going to merge this shortly after we add testing with positional aspects along with some other minor changes.

Follow-up items:
- Compatibility testing
- Draft PR where we split term and stem variants into separate term iterators

integration/test_fulltext.py
```diff
@@ -483,27 +479,264 @@ def test_default_tokenization(self):
...
+        # Test 5c: Stem expansion with both INORDER and SLOP - most restrictive case
+        # "run" with "improves" in doc:4: "Running [every day] improves" has gap of 2 words in correct order
+        result = client.execute_command("FT.SEARCH", "idx", '@content:run improv', "SLOP", "1", "INORDER")
+        assert result[0] == 0  # Gap too large for SLOP=1
+        result = client.execute_command("FT.SEARCH", "idx", '@content:run improv', "SLOP", "2", "INORDER")
+        assert result[0] == 1 and result[1] == b'doc:4'
+        assert set(result[2]) == {b'content', b'Running every day improves health', b'nostem_field', b'running'}
+        
+        # Test 5d: Multiple stem variants with INORDER and SLOP
+        # "running" and "runs" both stem to "run", testing cross-variant proximity
```
BCathcart: What does cross-variant proximity mean here? This section seems pretty similar to 5c except you also have a test with SLOP == 0. I'd suggest a quick audit of this entire test case when you do the refactor to make sure every section has some clear purpose, especially since we changed our approach with min stem size part way through

src/indexes/text/lexer.cc
```diff
@@ -162,20 +170,23 @@ sb_stemmer* Lexer::GetStemmer() const {
   return it->second.get();
 }
 
-std::string Lexer::StemWord(const std::string& word, bool stemming_enabled,
-                            uint32_t min_stem_size, sb_stemmer* stemmer) const {
-  if (word.empty() || !stemming_enabled || word.length() < min_stem_size) {
+std::string Lexer::StemWord(const std::string& word, uint32_t min_stem_size,
+                            sb_stemmer* stemmer) const {
+  if (word.empty() || word.length() < min_stem_size) {
     return word;
```
BCathcart, line 176: When you do the refactor, I'd still consider returning nullopt if the stemming wasn't done to avoid the string compare

integration/test_fulltext.py
Aksha1812: Sure I might just refer from Miles' tests afterwards. those would be more comprehensive and detailed

src/indexes/text/lexer.cc
Aksha1812: sure will keep that in mind

integration/test_fulltext.py
```diff
@@ -199,7 +196,7 @@ def validate_fulltext_search(client: Valkey):
     result = client.execute_command("FT.SEARCH", "products", '@desc:%wander%')
     assert (result[0], set(result[1::2])) == (1, {b"product:4"})
     result = client.execute_command("FT.SEARCH", "products", '@desc:%%greet%%')
-    assert (result[0], set(result[1::2])) == (3, {b"product:1", b"product:5", b"product:6"})
+    assert (result[0], set(result[1::2])) == (3, {b"product:1", b"product:5", b"product:6"}) , f"{set(result[1::2])}"
```
KarthikSubbarao: what is this? what is it testing?
```
set(result[1::2])
```

integration/test_fulltext.py
```diff
@@ -181,12 +179,11 @@ def validate_fulltext_search(client: Valkey):
     # 2. Stemming is not done on words. (`words` is not stemmed and the ingestion is not)
     # 3. Punctuation is applied (removal of `,` in this example).
     result1 = client.execute_command("FT.SEARCH", "products", '@desc:"random words, these are not correct"')
-    result2 = client.execute_command("FT.SEARCH", "products", '@desc:"random word, these are not correct"')
+    result2 = client.execute_command("FT.SEARCH", "products", '@desc:"random word, correct"')
```
KarthikSubbarao, line 182: I am fine with keeping this case.

But `2.` in the comment is referring to `words`. 


So, in addition to what we have, Can you add this test?
```
'@desc:"random words, correct"')
```

integration/test_fulltext.py
Aksha1812: This is checking if the result of FT.SEARCH products '@desc:%%greet%%' is {product:1", b"product:5", b"product:6"}

src/indexes/text.cc
```diff
@@ -183,21 +183,70 @@ void* TextPredicate::Search(bool negate) const {
...
       static_cast<const indexes::Text::EntriesFetcher*>(fetcher_ptr);
-  auto word_iter =
-      fetcher->text_index_->GetPrefix().GetWordIterator(GetTextString());
   absl::InlinedVector<indexes::text::Postings::KeyIterator,
                       indexes::text::kWordExpansionInlineCapacity>
       key_iterators;
-  if (!word_iter.Done() && word_iter.GetWord() == GetTextString()) {
-    key_iterators.emplace_back(word_iter.GetTarget()->GetKeyIterator());
+
+  absl::string_view text_string = GetTextString();
```
KarthikSubbarao, line 212: Can we remove all the new lines here? I generally prefer compacted code with comments added when needed for separation. New lines after every line of code makes it hard to read

Same comment on GetAllStemVariants and others new fns added

integration/test_fulltext.py
Aksha1812: that is already 3. right ? or am i reading your comment wrong. 2. is referring that actual term is 'words' but we are looking for "word" which is a stemmed version of it and since it is exact phrase search we don't get any results.

src/indexes/text.cc
Aksha1812: ok removing

integration/test_fulltext.py
KarthikSubbarao: i meant why are we adding , `, f"{set(result[1::2])}"`

Aksha1812: oh i just added it for debugging something. I can remove it.

PR #620, opened 2026-01-16
Add SeekForwardPosition API to TextIterator interface and all the derived iterators

KarthikSubbarao:
Add SeekForwardPosition API to TextIterator interface and all the derived iterators.

This will help in cases where there is an ordering violation identified based on the initial primed positions of the text iterators. To resolve the ordering violation, we try to seek when possible to avoid repetitive NextPosition calls

Note: It will only ever be used by an search query with an AND Intersection with some proximity property thus requiring a ProximityIterator

### DIFF
--- a/src/indexes/text/proximity.cc
+++ b/src/indexes/text/proximity.cc
@@ -292,9 +292,7 @@ bool ProximityIterator::NextPosition() {
       } else {
         // No violations: advance the term appearing first physically
         size_t first_idx = in_order_ ? 0 : pos_with_idx_[0].second;
-        if (!iters_[first_idx]->DonePositions()) {
-          iters_[first_idx]->NextPosition();
-        }
+        iters_[first_idx]->NextPosition();
       }
       continue;
     }


src/indexes/text/proximity.cc
```diff
@@ -259,27 +266,31 @@ bool ProximityIterator::NextPosition() {
...
       } else {
-        // No violations, advance first non-done iterator
-        for (size_t i = 0; i < n; ++i) {
-          if (!iters_[i]->DonePositions()) {
-            iters_[i]->NextPosition();
-            break;
-          }
+        // No violations: advance the term appearing first physically
+        size_t first_idx = in_order_ ? 0 : pos_with_idx_[0].second;
+        if (!iters_[first_idx]->DonePositions()) {
```
daddaman-amz: What would happen if the iterator is exhausted?

KarthikSubbarao: If it was exhausted, then we should not try advancing, since it is against the contract (and can crash). There is a check at the top in the loop condition which will exit if any iterator is done once the curr iteration is done

daddaman-amz (approved):

PR #622, opened 2026-01-21
Introduce ShardedAtomic for zero-contention, high-concurrency counters

yairgott:
This PR introduces ShardedAtomic<T>, a component optimized for capturing statistic counters in multi-threaded environments where writes are frequent and reads are infrequent. The implementation maintains a Thread-Local Storage (TLS) for the counter; individual threads update their local copies, while read operations aggregate these values. This technique effectively eliminates cache line bouncing and addresses the performance bottlenecks associated with std::atomic, which often suffers from significant degradation due to cache coherency traffic as thread counts increase.

Key Changes:
* Core Implementation: Added ShardedAtomic<T>, using TLS for lock-free, non-contention writes on the hot path.
* Registry Pattern: Implemented a global registry to track active thread-local nodes, enabling aggregation (GetTotal) and resetting.
* Benchmark Integration: Added Google Benchmark support to the CMake build system and implemented sharded_atomic_benchmark.cc.

Performance Analysis: Benchmarks demonstrate that ShardedAtomic achieves linear scalability for writes, whereas std::atomic degrades as thread count increases. At 48 threads, ShardedAtomic is approximately 380x faster per operation than std::atomic:

Benchmark                                           Time             CPU   Iterations
-------------------------------------------------------------------------------------
BM_StandardAtomic_Add/real_time/threads:1        6.19 ns         6.19 ns    111485462
BM_StandardAtomic_Add/real_time/threads:2        14.9 ns         29.9 ns     46299232
BM_StandardAtomic_Add/real_time/threads:4        14.5 ns         58.0 ns     45147756
BM_StandardAtomic_Add/real_time/threads:8        14.4 ns          115 ns     49008584
BM_StandardAtomic_Add/real_time/threads:16       14.8 ns          237 ns     48227248
BM_StandardAtomic_Add/real_time/threads:32       13.5 ns          431 ns     52890400
BM_StandardAtomic_Add/real_time/threads:48       14.9 ns          692 ns     48483456
BM_ShardedAtomic_Add/real_time/threads:1         1.95 ns         1.95 ns    371462750
BM_ShardedAtomic_Add/real_time/threads:2        0.972 ns         1.94 ns    713690864
BM_ShardedAtomic_Add/real_time/threads:4        0.485 ns         1.94 ns   1448671348
BM_ShardedAtomic_Add/real_time/threads:8        0.244 ns         1.96 ns   2831339160
BM_ShardedAtomic_Add/real_time/threads:16       0.121 ns         1.94 ns   5784267360
BM_ShardedAtomic_Add/real_time/threads:32       0.059 ns         1.88 ns   12028487104
BM_ShardedAtomic_Add/real_time/threads:48       0.039 ns         1.84 ns   17671151088
BM_StandardAtomic_GetTotal                      0.194 ns        0.194 ns   1000000000
BM_ShardedAtomic_GetTotal                        22.8 ns         22.8 ns     30634040


vmsdk/src/sharded_atomic.h
```diff
@@ -0,0 +1,137 @@
...
+  // Cost: ~1-2 CPU Cycles (MOV + ADD). No LOCK prefix.
+  inline void Add(T n) {
+    ThreadLocalNode& node = GetLocalNode();
+
+    // Optimistic Load -> Add -> Store
+    // Since this thread is the exclusive writer to 'node', we don't need atomic
+    // RMW. We use relaxed ordering to avoid memory fences, relying on the
+    // single-writer invariant.
+    T current = node.value.load(std::memory_order_relaxed);
+    node.value.store(current + n, std::memory_order_relaxed);
```
allenss-amazon, line 42: What is the point of using atomics here? Why not just use a normal variable?

vmsdk/src/sharded_atomic.h
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    T GetTotal(std::memory_order order) const {
+      T total = 0;
+      // We lock purely to ensure the vector doesn't change size (iterators
+      // validity)
+      absl::ReaderMutexLock lock(&mutex_);
+
+      for (const auto* node : nodes_) {
+        total += node->value.load(order);
```
allenss-amazon, line 94: I don't see the memory ordering as providing any meaningful improvement in the semantics of this. It's going to be eventually consistent no matter what you do.

vmsdk/src/sharded_atomic.h
yairgott: Without atomic we're risking that the value has a stale value as a result of reading it from the CPU cache

vmsdk/src/sharded_atomic.h
yairgott: Yeah, but I don't think that it really hurts, the API has a default value of relaxed.

vmsdk/src/sharded_atomic.h
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
```
allenss-amazon, line 2: 2026

vmsdk/testing/sharded_atomic_benchmark.cc
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
```
allenss-amazon, line 2: 26

allenss-amazon (approved):

PR #623, opened 2026-01-21
Add Compatibility Tests and Index Schema Selection

boda26:
Added unescaped and escaped punctuation, fuzzy search compatibility tests.
Added index schema selection as pytest parameters, enabling default (stemming) and nostem indexes.


integration/compatibility/data_sets.py
```diff
@@ -110,6 +110,72 @@
             ],
             'price': (0, 50)
         }
+    },
+    # ,.<>{}[]"':;!@#$%^&*()-+=~
```
VoletiRam, line 114: Checking our default list of punctuation
```
 ",.<>{}[]\"':;!@#$%^&*()-+=~/\\|";
```

I see forward slash (/) and pipe(|). Do we need to test this as well? If we know we handle as punctuation unlike redis, we need to document them somewhere for behavior difference.

integration/compatibility/generate_text.py
```diff
@@ -58,6 +298,38 @@ def _run_test(self, builder_fn, data_set_name, key_type, dialect, inorder=False,
...
                     seen.add(query_str)
+
+                    # check if Redis parsing is correct by calling FT.EXPLAINCLI, 
+                    # and compare the words order with the query string
+                    if is_redis and check_parsing:
+                        # words_from_explaincli = self.extract_words_from_explaincli(f"{key_type}_idx1", query_str, dialect)
+                        # words_from_query = self.extract_words_from_query(query_str)
+                        # if words_from_explaincli != words_from_query:
+                        #     print(f"Redis parsing is inconsistent for query {query_str}, skipping this query")
+                        #     continue
```
VoletiRam: Do we need this commented code?

integration/compatibility_test.py
```diff
@@ -352,7 +352,7 @@ def do_answer(client, expected, data_set):
         data_set = (expected['data_set_name'], expected['key_type'])
 
     # Set Valkey-specific config for inorder tests
-    if 'inorder' in expected['testname'] or 'slop' in expected['testname']:
```
VoletiRam, line 391: Why removed slop?

integration/compatibility/text_query_builder.py
```diff
@@ -163,6 +175,7 @@ def gen_suffix(vocab: List[str], rng: random.Random) -> List[SuffixTerm]:
         result.append(SuffixTerm("*" + word[-suffix_len:]))
     return result
 
+    op = rng.choice(["G", "AND", "OR"])
```
VoletiRam: Added by accident?

integration/compatibility/data_sets.py
boda26: Redisearch does not include these two, so they are not included in compatibility tests. We should document the difference.

integration/compatibility_test.py
boda26: The config deals with overlap words and is for INORDER only, it would cause wrong behavior in SLOP only cases.

allenss-amazon (approved):

PR #624, opened 2026-01-22
Optimize all proximity queries and pure OR queries

KarthikSubbarao:
Optimize all proximity queries and pure OR queries.

In this PR, query operations are updated to help us track nested composed operations and proximity (not just exact phrase).

Next, the search logic is updated in two ways:
1. If the operation is a pure composed OR, we do not need to enter the prefilter evaluator because the entries fetcher results are accurate once we deduplicate the keys. The "optimized" (non prefilter evaluator) code path for non vector search was updated to perform this  deduplication.
2. The `EvaluateFilterAsPrimary` fn was updated to have the "optimized" text proximity logic get executed if there is an AND, Text, proximity flag, but no nested flag and no numeric/tag. 

With this PR, we have now optimized the following:
1. Flat Composed OR queries (of any type)
2. Flat Proximity Text queries

The following were already optimized from my earlier PRs:
1. Single Term (Numeric / Text/ Tag)
2. Exact Phrase

A subsequent optimization I will look into next is Flat Composed AND Text (non proximity)

### DIFF
--- a/src/query/search.cc
+++ b/src/query/search.cc
@@ -163,24 +163,12 @@ inline bool IsTextProximityOnlyNonNested(QueryOperations query_operations) {
 // Helper fn to identify composed AND predicates that we cannot optimize
 // currently.
 inline bool IsUnsolvedComposedAnd(QueryOperations query_operations) {
-  // If there are nested composed predicates, we cannot optimize currently.
-  if (query_operations & QueryOperations::kContainsNestedComposed) {
-    return true;
+  // AND composed predicates are only unsolved if they are not pure text
+  // proximity predicates without nesting.
+  if (query_operations & QueryOperations::kContainsAnd) {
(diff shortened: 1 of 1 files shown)


src/commands/filter_parser.cc
```diff
@@ -475,14 +475,19 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::WrapPredicate(
   // Only extend AND nodes when we're adding with AND operator
   if (prev_predicate->GetType() == query::PredicateType::kComposedAnd &&
       logical_operator == query::LogicalOperator::kAnd && !no_prev_grp) {
+    if (query_operations_ &
```
daddaman-amz: Would   `A AND B AND C` be flagged as  Nested here?

PR #629, opened 2026-01-26
Avoid deadlock while processing  MULTI/EXEC mutations

yairgott:
Processing of the MULTI/EXEC queued mutation results in a deadlock if a worker thread attempts to acquire the WriterLock after its current time slice quota has expired. Exceeding the quota forces the worker thread to wait for the next scheduling quota cycle. However, the main thread holds the mutex while waiting for the worker threads to complete the queue (via the blocking_counter). This leads to a circular dependency where the worker thread waits for a new cycle that cannot start until the main thread releases the lock.

This is addressed by introducing a control to ignore the time quota to allow the ingestion of all the MULTI/EXEC tasks before allowing the switch.

### DIFF
--- a/vmsdk/src/time_sliced_mrmw_mutex.cc
+++ b/vmsdk/src/time_sliced_mrmw_mutex.cc
@@ -89,15 +89,17 @@ void TimeSlicedMRMWMutex::Unlock(bool may_prolong, bool ignore_time_quota) {
   absl::MutexLock lock(&mutex_);
   CHECK_GT(active_lock_count_, 0L);
   --active_lock_count_;
-  if (ABSL_PREDICT_FALSE(ignore_time_quota)) {
-    CHECK_GT(ignore_time_quota_count_, 0L);
-    --ignore_time_quota_count_;
-  }
+
   if (ABSL_PREDICT_FALSE(may_prolong)) {
     CHECK_GT(may_prolong_count_, 0L);
     --may_prolong_count_;
   }
-
+  if (ABSL_PREDICT_FALSE(ignore_time_quota)) {
+    CHECK_GT(ignore_time_quota_count_, 0L);
+    --ignore_time_quota_count_;
+  }
+  CHECK(active_lock_count_ > 0 ||
+        (may_prolong_count_ == 0 && ignore_time_quota_count_ == 0));
   if (ShouldSwitch() && GetWaiters(GetInverseMode(current_mode_)) > 0) {
     InitSwitch();
   }


src/index_schema.cc
```diff
@@ -747,6 +744,9 @@ void IndexSchema::ScheduleMutation(bool from_backfill, const Key &key,
         }
       },
       priority);
+  if (ABSL_PREDICT_FALSE(!scheduled && blocking_counter)) {
+    blocking_counter->DecrementCount();
+  }
```
murphyjacob4, line 756: If we can't schedule, should we return an error absl::Status upward? I guess just dropping it is a consistency problem, right?

src/index_schema.cc
```diff
@@ -731,7 +728,7 @@ void IndexSchema::ScheduleMutation(bool from_backfill, const Key &key,
       ++stats_.backfill_inqueue_tasks;
     }
   }
-  mutations_thread_pool_->Schedule(
+  auto scheduled = mutations_thread_pool_->Schedule(
       [from_backfill, weak_index_schema = GetWeakPtr(),
        ctx = detached_ctx_.get(), delay_capturer = CreateQueueDelayCapturer(),
        key_str = std::move(key), blocking_counter]() mutable {
```
murphyjacob4, line 734: Should we still wrap it in an absl::CleanUp? Either that, or we need to turn `if (!index_schema) return` into a `CHECK(index_schema)`

src/index_schema.cc
yairgott: `scheduled`  equals to false should be happening just on shutdown so I'm not too worried about consistency but I agree that returning error upward is the cleaner approach.

src/index_schema.cc
yairgott: `index_schema ` could be nullptr if the schema was destructed but because `blocking_counter` is declare on the stack while a strong reference to the `index_schema` is held, if `index_schema` is not null, its guaranteed that the `blocking_counter` is valid.

murphyjacob4: If it's guaranteed I think a CHECK is better. That way if the invariant fails, we crash instead of entering deadlock (and requiring someone to SSH and use gdb/lldb to figure out what is going on) 😄

yairgott: But `blocking_counter ` is null unless the flow is issued by the main thread multi-exec processing.

murphyjacob4: Yeah I agree that it won't be a problem. Was just thinking that a CHECK would ensure this is the case. It is a good use for a CHECK since if it isn't true, we will deadlock

vmsdk/src/time_sliced_mrmw_mutex.cc
```diff
@@ -82,6 +87,7 @@ void TimeSlicedMRMWMutex::Unlock(bool may_prolong) {
   absl::MutexLock lock(&mutex_);
   CHECK_GT(active_lock_count_, 0L);
   --active_lock_count_;
+  ignore_time_quota_ = false;
```
murphyjacob4: Would reentrant locking be a problem?

E.g.

```
Lock() // 1
SetIgnoreTimeQuota()
...
Lock() // 2
Unlock() // 2 - ignore_time_quota now false
...
Unlock() // 1
```

yairgott: Yeah, I've already fixed this.

src/index_schema.cc
yairgott: Maybe I'm missing what you're saying, where do you suggest to add `CHECK(index_schema)` ?

murphyjacob4: I am thinking `CHECK(!blocking_counter)` on line 738 (before we return)

vmsdk/src/time_sliced_mrmw_mutex.cc
```diff
@@ -51,21 +51,25 @@ void TimeSlicedMRMWMutex::IncMayProlongCount() {
...
-  Lock(Mode::kLockRead, may_prolong);
+void TimeSlicedMRMWMutex::ReaderLock(bool& may_prolong,
+                                     bool ignore_time_quota) {
+  Lock(Mode::kLockRead, may_prolong, ignore_time_quota);
 }
 
-void TimeSlicedMRMWMutex::WriterLock(bool& may_prolong) {
-  Lock(Mode::kLockWrite, may_prolong);
+void TimeSlicedMRMWMutex::WriterLock(bool& may_prolong,
+                                     bool ignore_time_quota) {
```
murphyjacob4, line 60: May be more readable as a flag? WDYT?

e.g.
```
vmsdk::WriterMutexLock lock(&time_sliced_mutex_, MRMW_LOCK_IGNORE_TIME_QUOTA);
vmsdk::WriterMutexLock lock(&time_sliced_mutex_, MRMW_MAY_PROLONG);
vmsdk::WriterMutexLock lock(&time_sliced_mutex_, MRMW_MAY_PROLONG | MRMW_LOCK_IGNORE_TIME_QUOTA);


```

PR #630, opened 2026-01-26
Add fulltext metrics to info

daddaman-amz:
Add following metrics to info search and will go as part of 1.2 release
```
search_used_text_memory_bytes   :  Total memory used by TEXT field indexes across all indexes
search_used_text_memory_human  :  human readable format

search_number_of_text_attributes   :  Total count of TEXT field attributes across all indexes
search_number_of_tag_attributes :  Total count of VECTOR field attributes across all indexes
search_number_of_numeric_attributes : Total count of NUMERIC field attributes across all indexes
search_number_of_vector_attributes : Total count of TAG field attributes across all indexes

search_corpus_num_text_items : Number of text values indexed in the corpus
search_ingest_field_text : Number of text fields ingestions across all indexes - counter 
search_vector_requests_count: Number of requests that are vector search queries exclusively
search_nonvector_requests_count :  Number of requests that are non-vector search queries exclusively
search_text_requests_count    :    Number of query requests  that needed text search 
search_query_text_term_count :  Number of query requests  that needed term search
search_query_text_prefix_count :  Number of query requests  that needed prefix search
search_query_text_suffix_count : Number of query requests  that needed suffix search
search_query_text_fuzzy_count : Number of query requests  that needed fuzzy search
search_query_text_proximity_count : Number of query requests  that needed proximity search
search_query_numeric_count : Number of query requests  that needed numeric search
search_query_tag_count : Number of query requests  that needed tag search
```


src/valkey_search.cc
```diff
@@ -355,6 +374,18 @@ static vmsdk::info_field::Integer query_prefiltering_requests_cnt(
...
 
+static vmsdk::info_field::Integer nonvector_requests_count(
+    "query", "nonvector_requests_count",
+    vmsdk::info_field::IntegerBuilder().App().Computed([]() -> long long {
+      return Metrics::GetStats().query_nonvector_requests_cnt;
+    }));
+
+static vmsdk::info_field::Integer text_requests_count(
+    "query", "text_requests_count",
+    vmsdk::info_field::IntegerBuilder().Dev().Computed([]() -> long long {
```
KarthikSubbarao, line 421: Note: If this is the marked as Dev, we cannot suggest this metric to customers for monitoring or include it in documentation

src/commands/filter_parser.cc
```diff
@@ -322,6 +330,7 @@ FilterParser::ParseNumericPredicate(const std::string& attribute_alias) {
   auto numeric_index =
       dynamic_cast<const indexes::Numeric*>(index.value().get());
   query_operations_ |= QueryOperations::kContainsNumeric;
+  query_numeric_count.Increment();
```
KarthikSubbarao: By incrementing the counter here, you are overcounting. What if an FT.SEARCH command's query string has 1000 numeric predicates? Currently, we increment it 1000 times per command.

I believe we had discussed incremented this at most once per command.

You can add an efficient function that checks flags after parsing is done and increments metrics just once.

Same comment applies to the rest of this file.

src/query/search.cc
```diff
@@ -533,8 +533,15 @@ absl::StatusOr<std::vector<indexes::Neighbor>> DoSearch(
   auto &time_sliced_mutex = parameters.index_schema->GetTimeSlicedMutex();
   vmsdk::ReaderMutexLock lock(&time_sliced_mutex);
   ++Metrics::GetStats().time_slice_queries;
+  if (parameters.filter_parse_results.query_operations &
+      QueryOperations::kContainsText) {
+    ++Metrics::GetStats().query_text_requests_cnt;
```
KarthikSubbarao: What does this metric track? Are you including the fanned out requests? From the code it seems like it. 

It seems like some metrics, e.g. `query_text_term_count` and `query_nonvector_requests_cnt`, are tracked only on the originating/local node. And this metric is including fanned out requests. Any reason why they are not consistent?

src/metrics.h
```diff
@@ -32,6 +32,8 @@ class Metrics {
     uint64_t query_failed_requests_cnt{0};
     uint64_t query_result_record_dropped_cnt{0};
     uint64_t query_hybrid_requests_cnt{0};
+    std::atomic<uint64_t> query_nonvector_requests_cnt{0};
```
KarthikSubbarao, line 35: do we have a vector related metric?

src/valkey_search.cc
```diff
@@ -104,6 +104,19 @@ static vmsdk::info_field::Integer used_memory(
         .Computed(vmsdk::GetUsedMemoryCnt)
         .CrashSafe());
 
+static vmsdk::info_field::Integer used_text_memory_bytes(
+    "memory", "used_text_memory_bytes",
+    vmsdk::info_field::IntegerBuilder().Dev().Computed([]() -> uint64_t {
```
KarthikSubbarao, line 109: Trying to understand why these are `Dev`. These would be useful for customers to monitor (we can only recommend them to do this if we mark this as `App`)

Are the numeric/tag or vector memory usage metrics DEV as well?

src/schema_manager.cc
```diff
@@ -405,6 +405,62 @@ uint64_t SchemaManager::GetNumberOfAttributes() const {
   }
   return num_attributes;
 }
+
+uint64_t SchemaManager::GetNumberOfTextAttributes() const {
```
KarthikSubbarao: All these functions seem the same. Can we use an ENUM as an input to help reuse a single function across all the functions below? They do the same type of calculation

Applies to all attribute count metrics

src/valkey_search.cc
daddaman-amz: There aren't separate numeric/tag or vector memory usage metrics.  
So for this reason I made it dev only for now

src/metrics.h
daddaman-amz: There is a `search_hybrid_requests_count` .  
We talked about adding separate `nonvector_requests_count`  to include only tag, text and numeric

src/query/search.cc
daddaman-amz: Yes we don't want to include fanned out request.  Let me recheck this , I may have missed checking for local  during some refactor

src/valkey_search.cc
daddaman-amz: Yes we are keeping this as Dev for now.

src/commands/filter_parser.cc
daddaman-amz: will change this to just count once per operation instead of predicate.

src/query/search.cc
```diff
@@ -555,8 +599,13 @@ absl::StatusOr<std::vector<indexes::Neighbor>> DoSearch(
   auto &time_sliced_mutex = parameters.index_schema->GetTimeSlicedMutex();
   vmsdk::ReaderMutexLock lock(&time_sliced_mutex);
   ++Metrics::GetStats().time_slice_queries;
+  IncrementQueryOperationMetrics(
+      parameters.filter_parse_results.query_operations, search_mode);
```
KarthikSubbarao: How did we choose this line of code for incrementing metrics? (What does this represent? Attempted requests, successful requests?)

It would be good to check where other query/search metrics are currently being incremented

src/schema_manager.cc
```diff
@@ -415,6 +463,18 @@ uint64_t SchemaManager::GetTotalIndexedDocuments() const {
   }
   return num_hash_keys;
 }
+uint64_t SchemaManager::GetTotalTextMemoryUsage() const {
```
KarthikSubbarao, line 466: @BCathcart - Does this seem accurate?

src/query/search.cc
```diff
@@ -569,6 +618,9 @@ absl::StatusOr<std::vector<indexes::Neighbor>> DoSearch(
   }
 
   if (!parameters.filter_parse_results.root_predicate) {
+    if (search_mode == SearchMode::kLocal) {
+      ++Metrics::GetStats().query_vector_requests_cnt;
+    }
```
KarthikSubbarao: This seems like the wrong place to increment the metric:
```
    if (parameters.filter_parse_results.root_predicate) {
      ++Metrics::GetStats().query_hybrid_requests_cnt;
    }
```

KarthikSubbarao: These metrics are incremented in the main thread in the command parsing

src/commands/ft_search_parser.cc
```diff
@@ -398,6 +444,14 @@ absl::Status PreParseQueryString(query::SearchParameters &parameters) {
     }
   }
   // TODO: Return Temp Error for unsupported predicates.
```
KarthikSubbarao: This comment can be removed

src/query/search.cc
```diff
@@ -39,6 +39,7 @@
 #include "src/valkey_search.h"
 #include "src/valkey_search_options.h"
 #include "third_party/hnswlib/hnswlib.h"
+#include "vmsdk/src/info.h"
```
KarthikSubbarao: Can we remove this?

daddaman-amz: yes missed it

src/schema_manager.cc
Aksha1812: This API GetTotalTextMemoryUsage() gives the Text Index memory usage metadata_.text_index_memory_pool_.GetUsage(); . 

The text_index_memory_pool is scoped around 

```
absl::StatusOr<bool> TextIndexSchema::StageAttributeData(
    const InternedStringPtr &key, absl::string_view data,
    size_t text_field_number, bool stem, bool suffix) {
  NestedMemoryScope scope{metadata_.text_index_memory_pool_};
  ```
  
  ```
  void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
  NestedMemoryScope scope{metadata_.text_index_memory_pool_};
  ```
  
  ```
void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
  NestedMemoryScope scope{metadata_.text_index_memory_pool_};
```

which are the only entry/exit points for memory to be added / deleted from the Text Index. 

Earlier when i was running some tests it was pretty near to the info search memory. which we consider source of truth. only difference would be just some smaller data structures in the overall code and memory occupied by the TextIndexSchema structures by itself. 

To be extra sure I would add a few thousand documents in an index and compare this against what we have in info search memory

daddaman-amz: We can add that 1000 doc test as follow up .   But as we are using the right available function for text memory aggregation I think we should be good for this PR

src/valkey_search.cc
```diff
@@ -385,6 +404,24 @@ static vmsdk::info_field::Integer query_prefiltering_requests_cnt(
...
+    }));
+
+static vmsdk::info_field::Integer vector_requests_count(
+    "query", "vector_requests_count",
+    vmsdk::info_field::IntegerBuilder().App().Computed([]() -> long long {
+      return Metrics::GetStats().query_vector_requests_cnt;
+    }));
+
+static vmsdk::info_field::Integer text_requests_count(
+    "query", "text_requests_count",
```
KarthikSubbarao, line 420: We only added Text? (No Numeric and Tag?)

daddaman-amz: We have 
search_query_numeric_count : Number of query requests  that needed numeric search
search_query_tag_count : Number of query requests  that needed tag search

KarthikSubbarao (approved):

integration/test_fulltext.py
```diff
@@ -1982,3 +1981,155 @@ def test_fulltext_search_cluster(self):
...
+        shards_incremented = 0
+        for i, c in enumerate(all_clients):
+            info = c.info("search")
+            text_delta = int(info.get("search_text_requests_count", 0)) - initial_metrics[i]['text']
+            nonvector_delta = int(info.get("search_nonvector_requests_count", 0)) - initial_metrics[i]['nonvector']
+            if text_delta > 0 or nonvector_delta > 0:
+                assert text_delta == nonvector_delta == 1, \
+                    f"Shard {i}: Both metrics must increment together by 1, got text={text_delta}, nonvector={nonvector_delta}"
+                shards_incremented += 1
+        assert shards_incremented == 1, f"Expected exactly 1 shard to increment, got {shards_incremented}"
```
KarthikSubbarao, line 2134: How can it fanout without being in cluster mode?

integration/test_fulltext.py
```diff
@@ -1982,3 +1981,155 @@ def test_fulltext_search_cluster(self):
...
+                    "tags", f"tag{i}|category{i % 3}",
+                    "vector", vec
+                )
+        info_search = client.info("search")
+        # Validate memory metrics
+        ft_memory_fulldata = int(info_search.get("search_used_text_memory_bytes", 0))
+        is_san_build = os.environ.get('SAN_BUILD', 'no') != 'no'
+        if not is_san_build:
+            assert ft_memory_fulldata > 0
+            assert info_search.get("search_used_text_memory_human", 0) != 0
```
KarthikSubbarao, line 2040: I am wondering why this does not work in ASAN builds

KarthikSubbarao (approved):

integration/test_fulltext.py
allenss-amazon: ASAN monkeys with the allocator chain, this means that some of the memory stats won't work in ASAN.

PR #640, opened 2026-01-29
Adding Endurance tests to github workflow

Nivesh-01:
Adding Endurance tests to github workflow to run daily at a fixed time.

### DIFF
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -250,6 +250,44 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert client.execute_command("FT.SEARCH", "idx_no_bs", r'test4\\\word4')[0] == 1
         assert client.execute_command("FT.SEARCH", "idx_no_bs", r'@content:test5\\\\word5')[0] == 1
 
+    def test_aggregate_with_text_search(self):
+        """Test FT.AGGREGATE with text search query."""
+        client: Valkey = self.server.get_new_client()
+        client.execute_command(
+            "FT.CREATE", "books", "ON", "HASH", "PREFIX", "1", "book:",
+            "SCHEMA", "title", "TEXT", "author", "TEXT", "year", "NUMERIC"
+        )
+        client.execute_command("HSET", "book:1", "title", "The Great Gatsby", "author", "F. Scott Fitzgerald", "year", "1925")
+        client.execute_command("HSET", "book:2", "title", "The Catcher in the Rye", "author", "J.D. Salinger", "year", "1951")
+        client.execute_command("HSET", "book:3", "title", "The Grapes of Wrath", "author", "John Steinbeck", "year", "1939")
+        client.execute_command("HSET", "book:4", "title", "Great Expectations", "author", "Charles Dickens", "year", "1861")
+        client.execute_command("HSET", "book:5", "title", "The Great Adventure", "author", "Unknown Author", "year", "2020")
+        IndexingTestHelper.wait_for_backfill_complete_on_node(client, "books")
+        # Prefix: gre* matches great (books 1,4,5)
+        result = client.execute_command("FT.AGGREGATE", "books", "gre*", "LOAD", "1", "title")
+        assert result[0] == 3
+        assert {result[i][1] for i in range(1, 4)} == {b"The Great Gatsby", b"Great Expectations", b"The Great Adventure"}
+        # Fuzzy: %greet% matches great (ED=1)
+        result = client.execute_command("FT.AGGREGATE", "books", "%greet%", "LOAD", "1", "title")
+        assert result[0] == 3
+        assert {result[i][1] for i in range(1, 4)} == {b"The Great Gatsby", b"Great Expectations", b"The Great Adventure"}
+        # Exact phrase
+        result = client.execute_command("FT.AGGREGATE", "books", '"great gatsby"', "LOAD", "1", "title")
+        assert result[0] == 1
+        assert set(result[1]) == {b"title", b"The Great Gatsby"}
+        # Prefix with GROUPBY and SORTBY
+        result = client.execute_command(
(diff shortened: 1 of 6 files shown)


.github/workflows/endurance_tests.yml
```diff
@@ -0,0 +1,181 @@
...
+    branches:
+      - main
+      - enduranceGitHubRun
+      - enduranceMemory
+    paths:
+      - 'src/**'
+      - 'testing/**'
+      - 'vmsdk/src/**'
+      - 'vmsdk/testing/**'
+      - '.github/workflows/**'
```
VoletiRam: Is it needed? we want to trigger the workflow irrespective of changes in location in the file. I would remove this.

.github/workflows/endurance_tests.yml
```diff
@@ -0,0 +1,181 @@
...
+  schedule:
+    - cron: "0 9 * * *" # Daily endurance tests at 2 AM UTC
```
VoletiRam: Having comments on the same line would make the entire line comment in yaml. I would suggest removing the comment or add above. Btw its 2am pst not utc

.github/workflows/endurance_tests.yml
```diff
@@ -0,0 +1,181 @@
...
+          -e REQUEST_COUNT="0" \
+          -e THREADS="${THREADS}" \
+          -e CLIENTS="${CLIENTS}" \
+          -e PIPELINE_DEPTH="${PIPELINE_DEPTH}" \
+          -e DATA_SIZE="${DATA_SIZE}" \
+          -e WORKLOAD_TYPE="${WORKLOAD_TYPE}" \
+          -e KEYSPACE_SIZE="${KEYSPACE_SIZE}" \
+          --user "ubuntu:ubuntu" \
+          endurance-test-image \
+          sudo -E /workspace/scripts/benchmark/run_endurance_test.sh
```
VoletiRam: I believe this is your local clouddesk path. Do you need to update this?

.github/workflows/endurance_tests.yml
```diff
@@ -0,0 +1,181 @@
...
+    - cron: "0 9 * * *" # Daily endurance tests at 2 AM UTC
+
+concurrency:
+  group: endurance-tests-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  endurance-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 7200
```
VoletiRam: I believe github default runners have 360 minutes hard limit. Unless you have self hosted machine, you cannot run more than 360 minutes. I would advice check this and see if you need to setup self hosted machine.

.github/workflows/endurance_tests.yml
```diff
@@ -0,0 +1,181 @@
...
+          -e ENABLE_TLS="${ENABLE_TLS}" \
+          -e TEST_DURATION="${TEST_DURATION}" \
+          -e REQUEST_COUNT="0" \
+          -e THREADS="${THREADS}" \
+          -e CLIENTS="${CLIENTS}" \
+          -e PIPELINE_DEPTH="${PIPELINE_DEPTH}" \
+          -e DATA_SIZE="${DATA_SIZE}" \
+          -e WORKLOAD_TYPE="${WORKLOAD_TYPE}" \
+          -e KEYSPACE_SIZE="${KEYSPACE_SIZE}" \
+          --user "ubuntu:ubuntu" \
```
VoletiRam, line 133: Why changing user here to ubuntu and doing sudo below. This is redundant permission assignment. You can remove ubuntu and sudo if you plan to run this as root user and it will work.

VoletiRam: Can you try this something similar along this lines? I guess root is not required to run this.
```
```yaml
docker run --rm \
  --mount type=bind,source="$(pwd)/endurance-results",target=/workspace/results \
  --mount type=bind,source="$(pwd)/tls",target=/workspace/tls \
  -v "$(pwd):/workspace" \
  -e SERVER_TYPE="${SERVER_TYPE}" \
  -e BRANCH_VERSION="${BRANCH_VERSION}" \
  # ... other env vars ...
  --user "$(id -u):$(id -g)" \
  endurance-test-image \
  /../run_endurance_test.sh
```

.github/workflows/endurance_tests.yml
```diff
@@ -0,0 +1,181 @@
...
+          endurance-test-image \
+          sudo -E /workspace/scripts/benchmark/run_endurance_test.sh
+
+    - name: Update Results Permissions
+      if: always()
+      run: sudo chmod -R 755 endurance-results
```
VoletiRam, line 139: You can remove this workarounds with sudo if you run with right user?

.github/workflows/endurance_tests.yml
Nivesh-01: This is how it is done in the other tests so I believe that keeping it the same would be better for consistency

.github/workflows/endurance_tests.yml
Nivesh-01: same as above

.github/workflows/endurance_tests.yml
Nivesh-01: you are right, this is changed

.github/workflows/endurance_tests.yml
Nivesh-01: Done

.github/workflows/endurance_tests.yml
Nivesh-01: Done

.github/workflows/endurance_tests.yml
Nivesh-01: Done

VoletiRam (approved):

PR #644, opened 2026-01-30
[FEATURE] Implement SKIPINITIALSCAN option on FT.CREATE command

allenss-amazon:
This option is useful when you want to create an index in an existing cluster and know that you want to skip the backfill.

### DIFF
--- a/integration/test_saverestore.py
+++ b/integration/test_saverestore.py
@@ -169,7 +169,7 @@ class TestSaveRestore_v2_v1(ValkeySearchTestCaseDebugMode):
 
     @pytest.mark.parametrize("parameters", [
         [index, [5, KEY_COUNT, 0], [5, 1, 0, 0]],
-        [vector_only_index, [3, 0, 0], [3, 1, 0, 0]],
+        [vector_only_index, [3, KEY_COUNT, 0], [3, 1, 0, 0]],
         [non_vector_index, [3, KEY_COUNT, 0], [3, 1, 0, 0]],
         ])
     def test_saverestore_v2_v1(self, parameters):
@@ -183,7 +183,7 @@ class TestSaveRestore_v2_v2(ValkeySearchTestCaseDebugMode):
 
     @pytest.mark.parametrize("parameters", [
         [index, [5, KEY_COUNT, 0], [5, 0, KEY_COUNT, 0]],
-        [vector_only_index, [3, 0, 0], [3, 0, 0, 0]],
+        [vector_only_index, [3, KEY_COUNT, 0], [3, 0, KEY_COUNT, 0]],
         [non_vector_index, [3, KEY_COUNT, 0], [3, 0, KEY_COUNT, 0]],
         ])
     def test_saverestore_v2_v2(self, parameters):
@@ -341,7 +341,7 @@ class TestMutationQueue(ValkeySearchTestCaseDebugMode):
         self.client.execute_command("save")
 
         i = self.client.info("search")
-        assert i["search_rdb_save_keys"] == 0
+        assert i["search_rdb_save_keys"] == len(records)
         assert i["search_rdb_save_mutation_entries"] == 0
         assert i["search_rdb_save_backfilling_indexes"] == 1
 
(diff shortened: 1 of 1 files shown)


src/index_schema.cc
```diff
@@ -284,7 +285,9 @@ IndexSchema::IndexSchema(ValkeyModuleCtx *ctx,
 
 absl::Status IndexSchema::Init(ValkeyModuleCtx *ctx) {
   VMSDK_RETURN_IF_ERROR(keyspace_event_manager_->InsertSubscription(ctx, this));
-  backfill_job_ = std::make_optional<BackfillJob>(ctx, name_, db_num_);
+  if (!skip_initial_scan_) {
+    backfill_job_ = std::make_optional<BackfillJob>(ctx, name_, db_num_);
+  }
```
KarthikSubbarao: I am trying to understand the behavior as a result of this code.

When the index is created, if `skip_initial_scan_`, we skip backfill.

But also, after a save and restore on a different server, it will skip backfill again because the AUX section (and index metadata) comes after the key-space data. Is that right?

allenss-amazon: The save restore is different with V2 RDB format. With the V2 RDB format we actually capture/restore the current state w.r.t. backfill. Meaning that if a backfill was in progress when restoring, then a new backfill will be started. But if there's no backfill in progress, then no backfill will be triggered. This works because the V2 format has a list of the keys that are in the index. Rather than using backfill to locate the keys to ingest, the V2 format directly tells the system which keys to ingest.

allenss-amazon: Thus if SKIPINITIALSCAN is set on an index and a save is triggered after that index, there won't be a backfill in progress and hence on a reload of that save no backfill will be restarted.

KarthikSubbarao:
There are conflicts in COMMANDS.md file

After they are resolved, I can approve the PR

integration/test_skipinitialscan.py
```diff
@@ -0,0 +1,89 @@
...
+            "SCHEMA", "title", "TEXT", "tag", "TAG"
+        )
+
+
+        waiters.wait_for_true(lambda: IndexingTestHelper.is_backfill_complete_on_node(client, "idx_normal"))        
+        
+        for index, count in [["idx_normal", 2], ["idx_skip", 0]]:
+            results = client.execute_command("FT.SEARCH", index, "@tag:{red | blue}")
+            assert results[0] == count
+
```
KarthikSubbarao, line 125: Can we test the insertion of 4 keys here on both the indexes?

The expectation is that upon restore even the `idx_skip` will have these 4 keys present on it because they were added AFTER the index creation

src/index_schema.cc
KarthikSubbarao: We discussed this code for skipinitialscan should only be applied if RDB V2 is used. If RDB V1 is used, we should perform the backfill operation

allenss-amazon: Corrected. Test added for this behavior.

integration/test_skipinitialscan.py
allenss-amazon: done.

src/index_schema.cc
KarthikSubbarao, line 1581: The changes make sense, but adding comments on why we are doing certain things (marking as done) and why it matters for RDB V1 / V2 will be very helpful for readers

allenss-amazon:
The most recent fixes are due to the change in the way that the save of the V2 format "key-list" was done. Originally, prior to the availability of db_key_info, there wasn't a complete list of keys. The code looped over the attributes and picked the first non-vector index it encountered on the theory that by a key list could be generated by tracked_keys + untracked_keys. This fails in the presence of an index with invalid entries, where a key would be inserted into other fields, but not the selected field. By replacing tracked_keys + untracked_keys of some random index with the db_key_info. The missing field problem is bypassed. Another change is because if there's only a vector index present, there's no untracked_keys for a vector field, now the key-list is always complete and correct and the save code just uses it unconditionally. This in some cases to differing counts of the keys that we saved by this process, which required changing those values in the test.

PR #647, opened 2026-01-30
Move MINSTEMSIZE clause from per text field to per schema

Aksha1812:
```
.
.
.
.

5) attributes
 6) 1) 1) identifier
       2) desc
       3) attribute
       4) desc
       5) type
       6) TEXT
       7) WITH_SUFFIX_TRIE
       8) 0
       9) MIN_STEM_SIZE / NOSTEM
       10) 4 / 1
    2) 1) identifier
.
.
.
.
    32) will
    33) with
33) with_offsets
34) 1
35) language
36) english
````

changes to 

```
.
.
.
.
5) attributes
 6) 1) 1) identifier
       2) desc
       3) attribute
       4) desc
       5) type
       6) TEXT
       7) WITH_SUFFIX_TRIE
       8) 0
    2) 1) identifier
.
.
.
.
.
    32) will
    33) with
33) with_offsets
34) 1
35) min_stem_size
36) (integer) 5
37) language
38) english
.
.
.
.
```

### DIFF
--- a/src/commands/filter_parser.cc
+++ b/src/commands/filter_parser.cc
@@ -756,7 +756,7 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseUnquotedTextToken(
         SetupTextFieldConfiguration(field_mask, field_or_default, false));
     // Apply stemming if not exact match - use schema-level min_stem_size
     // directly
-    if (!exact && index_schema_.GetStemTextFieldMask() != 0) {
+    if (!exact && (index_schema_.GetStemTextFieldMask() & field_mask) != 0) {
       token = lexer.StemWord(token, true, index_schema_.GetMinStemSize(),
                              lexer.GetStemmer());
     }

### CODE REVIEW

File: src/index_schema.cc [hunk 1]
```diff
@@ -322,19 +325,9 @@ void IndexSchema::UpdateTextFieldMasksForIndex(const std::string &identifier,
...
-    // Update min stem sizes
+    // Track fields with stemming enabled (note: stemming not run for suffix)
     if (text_index->IsStemmingEnabled()) {
-      uint32_t stem_size = text_index->GetMinStemSize();
-      all_fields_min_stem_size_ =
-          all_fields_min_stem_size_.has_value()
-              ? std::min(*all_fields_min_stem_size_, stem_size)
-              : stem_size;
-      if (text_index->WithSuffixTrie()) {
-        suffix_fields_min_stem_size_ =
```
BCathcart, line 333: Do we not need to maintain a suffix stem field mask as well or are you just going to AND it with the `suffix_text_field_mask_` every time?

src/index_schema.proto
```diff
@@ -30,20 +30,19 @@ message IndexSchema {
   Stats stats = 7;
   optional uint32 db_num = 8;
 
-  // TODO: make these optional so that we can give them default values if loading an old version RDB
+  // TODO: make these optional so that we can give them default values if
+  // loading an old version RDB
```
BCathcart: nit: for things like this I wouldn't change the formatting of unless necessary. Now it assigns the TODO back to you in the git history

src/indexes/text.cc
```diff
@@ -72,18 +69,16 @@ int Text::RespondWithInfo(ValkeyModuleCtx* ctx) const {
   ValkeyModule_ReplyWithSimpleString(ctx, "WITH_SUFFIX_TRIE");
   ValkeyModule_ReplyWithSimpleString(ctx, with_suffix_trie_ ? "1" : "0");
 
-  // Show only one: if no_stem is specified, show no_stem, otherwise show
-  // min_stem_size
+  // Show NO_STEM if specified
```
BCathcart: Do we want to show NO_STEM with a "0" if it doesn't exist like we do for WITH_SUFFIX_TRIE? I'd ask @Nivesh-01 why he originally did it this way

testing/ft_create_parser_test.cc
```diff
@@ -1215,26 +1222,26 @@ INSTANTIATE_TEST_SUITE_P(
...
              .success = false,
-             .command_str = "idx1 on HASH SCHEMA text_field TEXT MINSTEMSIZE 0",
+             .command_str = "idx1 on HASH MINSTEMSIZE 0 SCHEMA text_field TEXT",
              .too_many_attributes = false,
              .hnsw_parameters = {},
              .flat_parameters = {},
              .tag_parameters = {},
              .text_parameters = {},
              .expected = {},
-             .expected_error_message = "Invalid field type for field `text_field`: Error parsing value for the parameter `MINSTEMSIZE` - MINSTEMSIZE must be positive",
```
BCathcart, line 1237: Can you add a test that keeps MINSTEMSIZE after the TEXT designator to validate it fails with an expected error message

testing/ft_info_test.cc
```diff
@@ -410,14 +410,14 @@ INSTANTIATE_TEST_SUITE_P(
                         )",
                      .expect_return_failure = false,
                      .expected_output =
-                         "*44\r\n+index_name\r\n+test_name\r\n+index_"
+                         "*46\r\n+index_name\r\n+test_name\r\n+index_"
```
BCathcart, line 413: I see we removed `MIN_STEM_SIZE` and added `min_stem_size`. Is there a reason that we changed the case and why did the string get longer (44 -> 46)? Are we now adding a value and before we just had the "MIN_STEM_SIZE" text present or not?

src/commands/filter_parser.cc
```diff
@@ -754,10 +752,13 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseUnquotedTextToken(
...
-    VMSDK_RETURN_IF_ERROR(SetupTextFieldConfiguration(field_mask, min_stem_size,
-                                                      field_or_default, false));
-    if (!exact && min_stem_size.has_value()) {
-      token = lexer.StemWord(token, true, *min_stem_size, lexer.GetStemmer());
+    VMSDK_RETURN_IF_ERROR(
+        SetupTextFieldConfiguration(field_mask, field_or_default, false));
+    // Apply stemming if not exact match - use schema-level min_stem_size
+    // directly
+    if (!exact && index_schema_.GetStemTextFieldMask() != 0) {
+      token = lexer.StemWord(token, true, index_schema_.GetMinStemSize(),
```
BCathcart, line 760: Should we only stem if the fields we're searching have stemming enabled?

Also by setting the token to the stem, we're not losing the original token are we? I'm guessing not since this is existing functionality but just want to double check

src/index_schema.cc
BCathcart: We don't apply stemming during a prefix/suffix search right? Was `suffix_fields_min_stem_size_` never needed?

Aksha1812: No not needed . Discussed with Karthik yesterday.

src/indexes/text.cc
Aksha1812: It was done because we either had a minstensize or a nostem . So if it was not nostem we showed minstemsize . Ill check redis behavious and change/keep accordingly

testing/ft_info_test.cc
Aksha1812: I am just aligning with other schema level text clauses. also the string getting longer is because we added min stem size to the schemal level info now, if you see field level info string would be shorter now 10 -> 8

src/commands/filter_parser.cc
Aksha1812: Yes are this is old functionality . This will be changed with stemming pr on top of it. Have kept this the same way it was before to avoid test failures etc.. to not show up . No search behaviour is changing with this PR . Only moving minstemsize

src/indexes/text.cc
Aksha1812: they don't display an integer at all, i remember we decided against that. and if not no stem they don't display it just like done here.

src/commands/filter_parser.cc
BCathcart: To my first point, aren't you changing the functionality if we're searching on a specific field? If the index schema has a text stem field mask (i.e. any fields have stemming), then we stem this token. Before we checked if the specific text index had stemming enabled at which point we set the min_stem_size and stemmed the token. I'm not too concerned if this is about to be replaced, but let's make sure we're aware of the changes.

src/indexes/text.cc
BCathcart: I thought we were moving to a world where we always displayed a value. Since we decided we don't need to match the Redis format, it's odd too me we wouldn't have the NO_STEM flag behave the same way as the WITH_SUFFIX_TRIE flag in ft.info

testing/ft_info_test.cc
BCathcart: Ok makes sense

BCathcart (commented):
In the future let's try not to rebase and force push changes. I can't see what you changed since my last review.

src/indexes/text.cc
Aksha1812: Not sure. But what you are saying is correct. I’ll change it.

src/commands/filter_parser.cc
Aksha1812: No i replaced this min_stem_size.has_value() with GetStemTextFieldMask() != 0 . same thing actually.

BCathcart: Help me understand how it is the same. This is how we set it in the single field case before, based off whether stemming is set on that particular field:
```   
  if (text_index->IsStemmingEnabled()) {
      min_stem_size = text_index->GetMinStemSize();
    }
```

Aksha1812: See  @BCathcart 

```
if (text_index->IsStemmingEnabled()) {
      stem_text_field_mask_ |= field_bit;

src/indexes/text.cc
Nivesh-01: Based on our descision for FT.INFO we always display the value wether is true or false. What redis does is it displays the parameter when it is true and does not display it when it is false but we have moved away from it.

src/commands/filter_parser.cc
BCathcart: Right so the check that stem field mask is not zero evaluates to true if any of the fields have stemming enabled. This is different than checking if just the current text index has stemming enabled.

Aksha1812: changed to `(index_schema_.GetStemTextFieldMask() & field_mask) != 0`

PR #667, opened 2026-02-02
Raise limit on number of attributes. Correct misnamed max-vector-attribute

allenss-amazon:
1. Move accounting for IndexSchema::document_cnt_ from per-attribute to per-key code. This eliminates O(N^2) ingestion with number of attributes.
2. Raise default limit on number of attributes to 1000.
3. Raise maximum number of attributes to 10000.
4. Implement WasSet configuration API to detect if a configuration variable was changed from default value.
5. Add "max-attributes" configuration to supercede inappropriately named "max-vector-attributes". Backward compatible code allowed either to be set. But if both a reset, then the old legacy value is used.

### DIFF
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -780,6 +780,7 @@ MutationSequenceNumber IndexSchema::UpdateDbInfoKey(
 
   auto &dbkeyinfo = dbkeyinfo_map[interned_key];
   dbkeyinfo.mutation_sequence_number_ = this_mutation;
+  stats_.document_cnt = dbkeyinfo_map.size();
 
   auto &attr_info_vec = dbkeyinfo.GetAttributeInfoVec();
   // Clear the array, we will re-use it.


src/commands/ft_create_parser.cc
```diff
@@ -78,7 +78,8 @@ constexpr absl::string_view kMaxPrefixesConfig{"max-prefixes"};
 constexpr absl::string_view kMaxTagFieldLenConfig{"max-tag-field-length"};
 constexpr absl::string_view kMaxNumericFieldLenConfig{
     "max-numeric-field-length"};
-constexpr absl::string_view kMaxAttributesConfig{"max-vector-attributes"};
```
KarthikSubbarao, line 81: Was this released already? If so, will this be a breaking change?

allenss-amazon: It shouldn't be a breaking change. If either max-vector-attributes or max-attributes is set it will be honored.  If they are both set, then max-vector-attributes takes precedent.

allenss-amazon: eventually we can deprecate max-vector-attributes.
'

src/commands/ft_create_parser.cc
```diff
@@ -566,7 +580,12 @@ bool HasVectorIndex(const data_model::IndexSchema &index_schema_proto) {
 absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
     ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
   // Get configuration values
-  const auto max_attributes_value = options::GetMaxAttributes().GetValue();
+  // The incorrectly named max_vector_attributes is kept for backward
+  // compatibility.
+  const auto max_attributes_value =
+      options::GetMaxVectorAttributes().WasSet()
+          ? options::GetMaxVectorAttributes().GetValue()
+          : options::GetMaxAttributes().GetValue();
```
KarthikSubbarao, line 588: We are honoring the legacy config over the the new config even if both were set?

KarthikSubbarao: It works and is simple, but it is odd to know a user can set both configs to a value. It will be helpful to document this behavior. I am fine with doing this


But I am trying to think of alternatives. The only one I can think of is update the OnConfigSet/OnConfigChange callback and reject the `CONFIG SET` command if the other config has already been set before. Here, the advantage is the CONFIG SET can be rejected with a clear error message.

KarthikSubbarao (approved):

PR #668, opened 2026-02-02
Add Memory Tracking for Index Schema Attributes

eifrah-aws:
## Brief

This change introduces a memory tracking system for index schema attributes to monitor and report the total size of indexed data. The data is presented in the `FT.INFO` command.

## Example

Example commands:

```
FT.CREATE hash_idx1 ON HASH PREFIX 1 hash: SCHEMA vec AS VEC VECTOR HNSW 6 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2 colour TAG
HSET hash:2 vec "\x00\x00\x00\x00\x00\x00\x00\x00" colour "red"
HSET hash:3 vec "\x00\x00\x00\x00\x00\x00\x00\x00" colour "yellow"
FT.INFO hash_idx1
```

Will output:

```
5) attributes
 6) 1)  1) identifier
        2) vec
        3) attribute
        4) VEC
        5) user_indexed_memory
        6) (integer) 16
        7) type
        8) VECTOR
        9) index
       10)  1) capacity
            2) (integer) 10240
            3) dimensions
            4) (integer) 2
            5) distance_metric
            6) L2
            7) size
            8) "2"
            9) data_type
           10) FLOAT32
           11) algorithm
           12) 1) name
               2) HNSW
               3) m
               4) (integer) 16
               5) ef_construction
               6) (integer) 200
               7) ef_runtime
               8) (integer) 10
    2)  1) identifier
        2) colour
        3) attribute
        4) colour
        5) user_indexed_memory
        6) (integer) 9
        7) type
        8) TAG
        9) SEPARATOR
       10) ,
       11) CASESENSITIVE
       12) 0
       13) size
       14) "2"
```

Notice the new fields: `user_indexed_memory` which appears for every attribute in the index. The `vec` attributes indexes total of `16` bytes(2 vectors) while the `colour` indexes `9` bytes ("yellow" + "red").

## Details

The implementation uses a compact bit-packed structure to efficiently store field indices and sizes within a 64-bit integer, reducing memory overhead for metadata storage.

A new AttributeInfo struct packs a 16-bit field index and 48-bit size value into a single uint64_t. The DbKeyInfo struct now maintains a vector of AttributeInfo entries and provides methods to get, set, and compute total attribute sizes. The IndexSchema class tracks global index size in bytes and assigns unique indices to attributes using an internal sequencer.

The ProcessMutation method has been refactored to extract database key info updates into a separate UpdateDbInfoKey method, which maintains accurate index size accounting by computing deltas between old and new attribute sizes during mutations and deletions.

* AttributeInfo: new bit-packed structure for field metadata
* DbKeyInfo: attribute size tracking and aggregation methods
* IndexSchema: attribute index assignment and global size tracking
* ProcessMutation: refactored to delegate key info updates

** Generated by CodeLite. **

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -52,6 +52,7 @@ endif()
 
 set(SRCS_VALKEY_SEARCH
     ${CMAKE_CURRENT_LIST_DIR}/attribute.cc
+    ${CMAKE_CURRENT_LIST_DIR}/attribute.h
     ${CMAKE_CURRENT_LIST_DIR}/valkey_search.cc
     ${CMAKE_CURRENT_LIST_DIR}/valkey_search.h
     ${CMAKE_CURRENT_LIST_DIR}/valkey_search_options.cc
--- a/src/attribute.cc
+++ b/src/attribute.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, valkey-search contributors
+ * Copyright (c) 2026, valkey-search contributors
  * All rights reserved.
  * SPDX-License-Identifier: BSD 3-Clause
  *
--- a/src/attribute.h
+++ b/src/attribute.h
@@ -20,6 +20,7 @@
 
 namespace valkey_search {
 
+using AttributePosition = uint16_t;
 class IndexSchema;
 class Attribute {
  public:
@@ -49,13 +50,14 @@ class Attribute {
   }
 
   // Return the attribute position during creation of the index.
-  inline uint16_t GetPosition() const { return position_; }
+  inline AttributePosition GetPosition() const { return position_; }
 
  private:
   std::string alias_;
   std::string identifier_;
   std::shared_ptr<indexes::IndexBase> index_;
-  uint16_t position_{UINT16_MAX};  // The attribute position during creation.
+  AttributePosition position_{
+      UINT16_MAX};  // The attribute position during creation.
 };
 
 }  // namespace valkey_search
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -384,7 +384,7 @@ absl::StatusOr<std::string> IndexSchema::GetIdentifier(
   return itr->second.GetIdentifier();
 }
 
-absl::StatusOr<uint16_t> IndexSchema::GetAttributePositionByAlias(
+absl::StatusOr<AttributePosition> IndexSchema::GetAttributePositionByAlias(
     absl::string_view attribute_alias) const {
   auto itr = attributes_.find(std::string{attribute_alias});
   if (itr == attributes_.end()) {
@@ -394,7 +394,7 @@ absl::StatusOr<uint16_t> IndexSchema::GetAttributePositionByAlias(
   return itr->second.GetPosition();
 }
 
-absl::StatusOr<uint16_t> IndexSchema::GetAttributePositionByIdentifier(
+absl::StatusOr<AttributePosition> IndexSchema::GetAttributePositionByIdentifier(
     absl::string_view identifier) const {
   VMSDK_ASSIGN_OR_RETURN(std::string attribute_alias, GetAlias(identifier));
   return GetAttributePositionByAlias(attribute_alias);
@@ -426,13 +426,14 @@ absl::Status IndexSchema::AddIndex(absl::string_view attribute_alias,
   auto [_, res] = attributes_.insert(
       {std::string(attribute_alias),
        Attribute{attribute_alias, identifier, index,
-                 static_cast<uint16_t>(attributes_size_vec_.size())}});
+                 static_cast<AttributePosition>(
+                     attributes_indexed_data_size_.size())}});
   if (!res) {
     return absl::AlreadyExistsError(
         absl::StrCat("Index field `", attribute_alias, "` already exists"));
   }
 
-  attributes_size_vec_.emplace_back(0);
+  attributes_indexed_data_size_.emplace_back(0);
   identifier_to_alias_.insert(
       {std::string(identifier), std::string(attribute_alias)});
   // Update schema level Text information for default field searches
@@ -776,15 +777,19 @@ MutationSequenceNumber IndexSchema::UpdateDbInfoKey(
   vmsdk::VerifyMainThread();
   MutationSequenceNumber this_mutation = ++schema_mutation_sequence_number_;
   auto &dbkeyinfo_map = db_key_info_.Get();
-  if (is_delete) {
-    auto iter = dbkeyinfo_map.find(interned_key);
-    if (iter != dbkeyinfo_map.end()) {
-      // Remove this key size(s) from the index tracked size array.
-      for (const auto &attr_info : iter->second.GetAttributeInfoVecMut()) {
-        CHECK(attr_info.GetPosition() < attributes_size_vec_.size());
-        attributes_size_vec_[attr_info.GetPosition()] -= attr_info.GetSize();
-      }
+
+  // When mutating (does not matter
+  auto iter = dbkeyinfo_map.find(interned_key);
+  if (iter != dbkeyinfo_map.end()) {
+    // Remove this key size(s) from the index tracked size array.
+    for (const auto &attr_info : iter->second.GetAttributeInfoVec()) {
+      CHECK(attr_info.GetPosition() < attributes_indexed_data_size_.size());
+      attributes_indexed_data_size_[attr_info.GetPosition()] -=
+          attr_info.GetSize();
     }
+  }
+
+  if (is_delete) {
     dbkeyinfo_map.erase(interned_key);
     return this_mutation;
   }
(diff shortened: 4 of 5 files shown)


src/index_schema.h
```diff
@@ -52,6 +52,48 @@ using RDBLoadFunc = void *(*)(ValkeyModuleIO *, int);
...
+    return (id_and_size_ >> 16) & 0xFFFFFFFFFFFFULL;
+  }
+
+  inline void SetSize(uint64_t v) {
+    // Clear the upper 48 bits, then set them with the new value shifted left
+    id_and_size_ = (id_and_size_ & 0xFFFF) | ((v & 0xFFFFFFFFFFFFULL) << 16);
+  }
+
+ private:
+  uint64_t id_and_size_{UINT64_MAX};
```
allenss-amazon: Bit fields would be much cleaner here and pretty much self-documenting.

src/index_schema.h
```diff
@@ -354,7 +506,33 @@ class IndexSchema : public KeyspaceEventSubscription,
...
+   * @param from_backfill Boolean flag indicating whether this update originates
+   * from a backfill operation (currently unused in the implementation but may
+   * affect future behavior).
+   * @param is_delete Boolean flag indicating whether this is a delete
+   * operation; if true, the key entry is removed from the map.
+   *
+   * @return MutationSequenceNumber The sequence number assigned to this
+   * mutation, representing the order in which this update occurred within the
+   * schema.
+   */
```
allenss-amazon, line 455: This is exactly the kind of documentation that shouldn't be present. It doesn't really give you any information that you wouldn't get from a few seconds of looking at the actual code.

src/index_schema.h
```diff
@@ -254,6 +402,7 @@ class IndexSchema : public KeyspaceEventSubscription,
  private:
   vmsdk::UniqueValkeyDetachedThreadSafeContext detached_ctx_;
   absl::flat_hash_map<std::string, Attribute> attributes_;
+  absl::flat_hash_map<std::string, uint16_t> attribute_name_to_idx_;
```
allenss-amazon: Should the index be a field in Attribute? Isn't this map a duplication of the attributes_ map?

src/index_schema.cc
```diff
@@ -762,20 +766,68 @@ bool ShouldBlockClient(ValkeyModuleCtx *ctx, bool inside_multi_exec,
...
+  vmsdk::VerifyMainThread();
   MutationSequenceNumber this_mutation = ++schema_mutation_sequence_number_;
-  auto &dbkeyinfo = db_key_info_.Get();
+  auto &dbkeyinfo_map = db_key_info_.Get();
   if (is_delete) {
-    dbkeyinfo.erase(interned_key);
+    auto iter = dbkeyinfo_map.find(interned_key);
+    if (iter != dbkeyinfo_map.end()) {
+      // Remove the size of the removed key from the total index size
+      index_size_bytes_ -= iter->second.GetTotalSize();
```
allenss-amazon: I thought the idea was to track the sizes of each attribute? Shouldn't you have to loop over the individual attributes and reduce the per-index_schema version? Why wouldn't that size be stored in Attribute (or IndexBase) ?

src/index_schema.cc
```diff
@@ -762,20 +766,68 @@ bool ShouldBlockClient(ValkeyModuleCtx *ctx, bool inside_multi_exec,
...
+      VMSDK_LOG(WARNING, ctx)
+          << "Index: [" << GetName()
+          << "]: could not find identifier for alias: " << attr.first;
+      continue;
+    }
+    const auto &attr_identifier = res.value();
+    auto attr_idx_opt = GetAttributeIndex(attr_identifier);
+    uint16_t attr_idx;
+    if (!attr_idx_opt.has_value()) {
+      attr_idx = AssignAttributeIndex(attr_identifier);
```
allenss-amazon: This shouldn't be lazy, the index values should be setup at index creation. Here this should be a CHECK, i.e., if we have an attr_identifier we can identify then it's an internal programming failure.

src/index_schema.cc
```diff
@@ -762,20 +766,68 @@ bool ShouldBlockClient(ValkeyModuleCtx *ctx, bool inside_multi_exec,
...
-                                  MutatedAttributes &mutated_attributes,
-                                  const Key &interned_key, bool from_backfill,
-                                  bool is_delete) {
-  //
-  // Update DbKeyInfo
-  //
+MutationSequenceNumber IndexSchema::UpdateDbInfoKey(
+    ValkeyModuleCtx *ctx, const MutatedAttributes &mutated_attributes,
+    const Key &interned_key, [[maybe_unused]] bool from_backfill,
+    bool is_delete) {
```
allenss-amazon, line 776: I like breaking out UPdateDbInfoKey into a separate method.

This method effectively replaces the old per-key info with a new per-key info (old or new might be nothing if key is new or it is being deleted). We can generalize this into an update operation.

The update operation conceptually removes all of the old sizes and then adds in the new sizes and retains the new sizes so that a future mutations will know the "old" sizes.

Thus, this is really a two step sequence:

1. Remove the old sizes. This is easily done by simply iterating down the "old" size array (the info array that you're adding) and decrementing the per-attribute size. The "old" size array is no longer used and can be discarded.
2. Add the new sizes. As you have coded, you iterate over the mutated attributes (which are all of them ;-)) and for any that now have a non-zero size, then you update the schema size and record that into the new array.

Hence in step 1, you consume the old vector. In step 2 you construct the new vector. You never need to "scan" the vector nor do you want to be trying to in-place edit it. Just build a new one and throw the old one away. Much simpler.

src/index_schema.cc
eifrah-aws: `index_size_bytes_` is a property of `IndexSchema`. So when we call `FT.INFO` the data is available. I will fine grain it per field type.

src/index_schema.h
eifrah-aws: I disagree here. 
Modern IDEs (via LSPs) will present you this text without the need to open the function at all...
If you strongly against it - I will remove it.

src/index_schema.cc
eifrah-aws: It is assigned in the `AddIndex` method. This code is defensive in its nature. I will change it to `CHECK`

src/CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ if(UNIX AND NOT APPLE)
 endif()
 
 set(SRCS_VALKEY_SEARCH
+    ${CMAKE_CURRENT_LIST_DIR}/attribute.cc
```
allenss-amazon, line 54: Don't you need to add attribute.h also?

src/attribute.cc
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
```
allenss-amazon: 2026.

src/attribute.h
```diff
@@ -54,10 +48,14 @@ class Attribute {
     return vmsdk::MakeUniqueValkeyString(absl::StrCat("__", alias_, "_score"));
   }
 
+  // Return the attribute position during creation of the index.
+  inline uint16_t GetPosition() const { return position_; }
```
allenss-amazon: let's make a typedef for AttributePosition or something like that.

src/index_schema.cc
```diff
@@ -407,13 +423,16 @@ absl::StatusOr<vmsdk::UniqueValkeyString> IndexSchema::DefaultReplyScoreAs(
 absl::Status IndexSchema::AddIndex(absl::string_view attribute_alias,
                                    absl::string_view identifier,
                                    std::shared_ptr<indexes::IndexBase> index) {
-  auto [_, res] =
-      attributes_.insert({std::string(attribute_alias),
-                          Attribute{attribute_alias, identifier, index}});
+  auto [_, res] = attributes_.insert(
+      {std::string(attribute_alias),
+       Attribute{attribute_alias, identifier, index,
+                 static_cast<uint16_t>(attributes_size_vec_.size())}});
```
allenss-amazon: Again, casting to "AttributePosition" rather than uint16_t makes this much clearer about what's happening.

Also, attributs_size_vec is pretty ambiguous. How about something like AttributesIndexedDataSize ?

src/index_schema.cc
```diff
@@ -750,20 +769,75 @@ bool ShouldBlockClient(ValkeyModuleCtx *ctx, bool inside_multi_exec,
...
+    }
+    dbkeyinfo_map.erase(interned_key);
+    return this_mutation;
+  }
+
+  auto &dbkeyinfo = dbkeyinfo_map[interned_key];
+  dbkeyinfo.mutation_sequence_number_ = this_mutation;
+
+  // Step 1: decrement the size per field from the index global size array.
+  auto &attr_info_vec = dbkeyinfo.GetAttributeInfoVecMut();
```
allenss-amazon: Don't repeat yourself.....

Either make a method for the decrementing. Or , restructure the code so it's clear that you're always going to decrement if there's a size vector available. This would be my preferred option. Move the dbkeyinfo_map.find before the "if delete" and then do the decrement there if the vector exists....

src/index_schema.cc
```diff
@@ -750,20 +769,75 @@ bool ShouldBlockClient(ValkeyModuleCtx *ctx, bool inside_multi_exec,
...
+      // If data is present, the operation is either INSERT or UPDATE.
+      // Otherwise, the field has been deleted and is treated as having size
+      // 0.
+      ValkeyModule_StringPtrLen(mutated_attr.second.data.get(), &data_len);
+      attr_info_vec.emplace_back(attr_pos, data_len);
+    }
+    // update the global tracking array
+    attributes_size_vec_[attr_pos] += data_len;
+  }
+  return this_mutation;
```
allenss-amazon, line 833: Since you're reusing the array, you'll want to have a "shrink_to_fit" call here.

src/index_schema.cc
```diff
@@ -750,20 +769,75 @@ bool ShouldBlockClient(ValkeyModuleCtx *ctx, bool inside_multi_exec,
...
+    CHECK(res.value() < attributes_size_vec_.size())
+        << "Invalid attribute position found";
+
+    const auto &attr_pos = res.value();
+    size_t data_len{0};
+    if (mutated_attr.second.data) {
+      // If data is present, the operation is either INSERT or UPDATE.
+      // Otherwise, the field has been deleted and is treated as having size
+      // 0.
+      ValkeyModule_StringPtrLen(mutated_attr.second.data.get(), &data_len);
```
allenss-amazon: vmsdk::ToStringView

src/index_schema.h
```diff
@@ -216,6 +238,11 @@ class IndexSchema : public KeyspaceEventSubscription,
                                   vmsdk::ArgsIterator &itr);
   struct DbKeyInfo {
     MutationSequenceNumber mutation_sequence_number_{0};
+    std::vector<AttributeInfo> attr_info_vec_;
+
+    inline std::vector<AttributeInfo> &GetAttributeInfoVecMut() {
```
allenss-amazon: InfoVecMut ?? Is this Rust??

src/CMakeLists.txt
eifrah-aws: Not really, similar to the fact that you don't add target for header files in `Makefile` - they are not getting compiled (unless it is a PCH file). However, for completeness, added.

src/index_schema.h
eifrah-aws: One can only wish...

src/index_schema.cc
```diff
@@ -750,20 +770,75 @@ bool ShouldBlockClient(ValkeyModuleCtx *ctx, bool inside_multi_exec,
...
 
+MutationSequenceNumber IndexSchema::UpdateDbInfoKey(
+    ValkeyModuleCtx *ctx, const MutatedAttributes &mutated_attributes,
+    const Key &interned_key, [[maybe_unused]] bool from_backfill,
+    bool is_delete) {
+  vmsdk::VerifyMainThread();
+  MutationSequenceNumber this_mutation = ++schema_mutation_sequence_number_;
+  auto &dbkeyinfo_map = db_key_info_.Get();
+
+  // When mutating (does not matter
```
allenss-amazon, line 781: can we fix this comment?

allenss-amazon (approved):
There's a lot of reformatting of comments that seems needless. Perhaps your editor setting is causing this??

PR #671, opened 2026-02-03
Endurance Testing - adding more search functionality testing.

Nivesh-01:
Increasing the search scenarios and the HSET scenarios. We are adding more number of HSETS to make the data to search from more varied data. We are also testing two different search processes each of which covers one instance of each type of search resulting in 10 different searches. The reason for having two processes is to increase the amount of searches in one test and to differentiate between the searches that use random data with binaries and searches that use real data.


testing/integration/stability_runner.py
```diff
@@ -220,16 +221,62 @@ def run(self) -> StabilityRunResult:
...
+            insert_command = (
+                f"{self.config.memtier_path}"
+                " --cluster-mode"
+                " -s localhost"
+                f" -p {self.config.ports[0]}"
+                f" -t {self.config.num_memtier_threads}"
+                f" -c {self.config.num_memtier_clients}"
+                " --reconnect-on-error"
+                " --max-reconnect-attempts=3"
+                " --random-data"
```
KarthikSubbarao, line 313: This flag makes the HSET commands use non utf8 data. How does this get handled on the server side?

I remember you mentioned it does not crash, but how is it handled in ingestion? Is it treated as one word per document?

testing/integration/stability_runner.py
```diff
@@ -140,6 +140,7 @@ def run(self) -> StabilityRunResult:
                 "tag": utils.TagDefinition(),
                 "numeric": utils.NumericDefinition(),
                 "content": utils.TextDefinition(),
+                "title": utils.TextDefinition(nostem=True, with_suffix_trie=True),
```
KarthikSubbarao: Why do we use nostem?

```
                "title": utils.TextDefinition(nostem=True, with_suffix_trie=True),
```

testing/integration/stability_runner.py
```diff
@@ -220,16 +221,62 @@ def run(self) -> StabilityRunResult:
...
+                f"{hset_fields_3}'"
+                " --command-ratio=1"
+                " --command-key-pattern=P"
+                " --command='HSET __key__ "
+                f"{hset_fields_4}'"
+                " --command-ratio=1"
+                " --command-key-pattern=P"
+                " --command='HSET __key__ "
+                f"{hset_fields_5}'"
+                " --command-ratio=4"
```
KarthikSubbarao: Let's make this 1 since all these commands will be rejected because it is binary data

testing/integration/stability_runner.py
```diff
@@ -355,30 +402,109 @@ def run(self) -> StabilityRunResult:
...
+                " --pipeline=1"
+                f" --test-time={self.config.test_time_sec}"
+                " --json-out-file"
+                f" {memtier_output_dir}/{self.config.index_name}_memtier_search.json"
+            )
+            
+            # Second search process for random data - short query patterns
+            # These target the random binary data inserted with __data__
+            # Using simple short patterns that might match random content
+            search_random_command = (
```
KarthikSubbarao: We can remove this.

Memtier already lets us continue threads and X number of clients per thread, so we don't need a separate process

testing/integration/stability_runner.py
```diff
@@ -414,6 +540,8 @@ def run(self) -> StabilityRunResult:
         logging.debug("delete_command: %s", delete_command)
         logging.debug("expire_command: %s", expire_command)
         logging.debug("search_command: %s", search_command)
+        if self.config.index_type == "TEXT":
+            logging.debug("search_random_command: %s", search_random_command)
```
KarthikSubbarao: We can remove this

testing/integration/stability_runner.py
```diff
@@ -434,6 +562,15 @@ def run(self) -> StabilityRunResult:
...
         )
+        # Add second search process for TEXT indexes targeting random data
+        if self.config.index_type == "TEXT":
+            processes.append(
+                utils.MemtierProcess(
+                    command=search_random_command,
+                    name="FT.SEARCH_RANDOM",
+                    error_predicate=lambda err: f"-Index with name '{self.config.index_name}' not found" not in err,
+                )
+            )
```
KarthikSubbarao: We can remove this

testing/integration/stability_runner.py
KarthikSubbarao: We want to test the code paths that enter snowballstemming through a workload which driving a lot of concurrent traffic/operations

testing/integration/stability_runner.py
```diff
@@ -355,30 +402,109 @@ def run(self) -> StabilityRunResult:
...
+                f" --command='FT.SEARCH {self.config.index_name} \"@title:prefix*\"'"
+                " --command-ratio=1"
+                f" --command='FT.SEARCH {self.config.index_name} \"@content:%fuzzy%\"'"
+                " --command-ratio=1"
+                f" --command='FT.SEARCH {self.config.index_name} \"@title:*device\"'"
+                " --command-ratio=1"
+                f" --command='FT.SEARCH {self.config.index_name} \"@title:smartwatch\"'"
+                " --command-ratio=1"
+                f" --command='FT.SEARCH {self.config.index_name} \"@title:\\\"fitness smartwatch\\\"\" SLOP 0 INORDER'"
+                " --command-ratio=1"
```
KarthikSubbarao, line 430: Lets add another command testing SLOP without the order requirement.

Use `systems matching enable` as the query and use SLOP of 3 


Also, in this query, don't use encapsulating double quotes because that will force inorder and slop 0

testing/integration/stability_runner.py
Nivesh-01: done

testing/integration/stability_runner.py
Nivesh-01: done

testing/integration/stability_runner.py
Nivesh-01: done

testing/integration/stability_runner.py
Nivesh-01: Changing nostem = false

testing/integration/stability_runner.py
Nivesh-01: This does not get added to the index it is used here to increase traffic

testing/integration/stability_runner.py
```diff
@@ -404,29 +451,86 @@ def run(self) -> StabilityRunResult:
...
+                f" -p {self.config.ports[0]}"
+                f" -t {self.config.num_memtier_threads}"
+                f" -c {self.config.num_search_clients}"
+                " -"
+                f" --command='FT.SEARCH {self.config.index_name} {search_query}'"
+                f" --test-time={self.config.test_time_sec}"
+                f" -d {self.config.vector_dimensions*4}"
+                " --json-out-file"
+                f" {memtier_output_dir}/{self.config.index_name}_memtier_search.json"
+            )
```
KarthikSubbarao, line 461: Why is there a change here?

KarthikSubbarao: (Tag and Numeric have the same code)

KarthikSubbarao: Same with `# Vector KNN search`

Can we try to avoid all the duplication?

testing/integration/stability_runner.py
```diff
@@ -152,6 +152,7 @@ def run(self) -> StabilityRunResult:
                 "tag": utils.TagDefinition(),
                 "numeric": utils.NumericDefinition(),
                 "content": utils.TextDefinition(),
+                "title": utils.TextDefinition(nostem=False, with_suffix_trie=True),
                 "category": utils.TextDefinition(nostem=True),
```
KarthikSubbarao: What is `category` used for? Are we querying it?

Nivesh-01: It is a field but we are not querying it for text. It is an additional field just in which we could do additional querying in the future.

testing/integration/stability_runner.py
Nivesh-01: I tried to remove redundancies in all of the code not just here to make it more compact. It should be much better now

testing/integration/stability_runner.py
```diff
@@ -152,7 +152,8 @@ def run(self) -> StabilityRunResult:
                 "tag": utils.TagDefinition(),
                 "numeric": utils.NumericDefinition(),
                 "content": utils.TextDefinition(),
-                "category": utils.TextDefinition(nostem=True),
+                "title": utils.TextDefinition(nostem=False, with_suffix_trie=True),
+                "category": utils.TextDefinition(nostem=False),
```
KarthikSubbarao: 1. `category` seems more like a TAG index based on the content it holds. 
2. `category` is unused, so let's delete it from here and update the text documents starting from line 285. It is not being queried anywhere. Let's not add unused aspects to the test framework

testing/integration/stability_test.py
```diff
@@ -590,7 +590,7 @@ def test_valkeyquery_stability(self, config):
                 result.halted,
                 msg=(
                     f"Expected memtier run {result.name} to not be halted (didn't "
-                    "make progress for >20sec)"
+                    "make progress for >10sec)"
```
KarthikSubbarao, line 593: I see a change in the message here, but no corresponding config/flag change.

 Did we actually change anything from 20 to 10? If this is a typo, you can revert it

Nivesh-01: No this is meant to be changed.

testing/integration/stability_runner.py
Nivesh-01: done

KarthikSubbarao (approved):

PR #682, opened 2026-02-04
Revert "Integrate ICU library for multiple language support (#443)"

VoletiRam:
This reverts commit 0f0a6bd3c9adc6532c1f358e466bb86fea000fcb. Removing as we are shipping only English for v1.


src/indexes/text/lexer.cc
```diff
@@ -137,7 +136,7 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
     }
 
     if (!word_buffer.empty()) {
-      std::string word = UnicodeNormalizer::CaseFold(word_buffer);
+      std::string word = absl::AsciiStrToLower(word_buffer);
```
BCathcart, line 139: I'm now hesitating about this for compatibility reasons with Redis. It looks like their behaviour is more aligned with what we have currently where we normalize all text regardless of the language.

```
127.0.0.1:6379> FT.CREATE myIdx ON HASH PREFIX 1 doc: LANGUAGE english SCHEMA content TEXT
OK
127.0.0.1:6379> HSET doc:1 content "Écrire is a French verb"
(integer) 1
127.0.0.1:6379> FT.SEARCH myIdx "écrire"
1) (integer) 1
2) "doc:1"
3) 1) "content"
   2) "\xc3\x89crire is a French verb"
```

VoletiRam: Hmm, this is a downside for only non english languages that requires normalization. ICU has normalization in built to help with it but now we don't handle it. This is in a way expected if we are saying we support only English for V1.

KarthikSubbarao: Let's check [this](https://github.com/valkey-io/valkey-search/pull/682#discussion_r2765899509)

and this on valkey-search with this PR
```
FT.SEARCH myIdx "Écrire"
```

KarthikSubbarao: @VoletiRam , when the ICU package was added only ingestion was made to use UnicodeNormalizer's CaseFold. However, the query parsing was left continuing to use `absl::AsciiStrToLower`. 



Do we need a change in query parsing for case folding? Or is this fine?

(There is a performance cost to use this)

PR #688, opened 2026-02-04
Rax Integration

BCathcart:
Follow-up to #605. The Rax symbols were conflicting with those from the Valkey core. The regular integration test run built with LTO while the ASAN one didn't, which seemed to make the difference. 

Re-named Rax symbols in the search module so that they can continue to be exported globally since it's my understanding that's needed for crash report backtraces

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,50 +1,53 @@
 # Use Ubuntu 24.04 as the base image
 FROM ubuntu:noble
 
-# Choose different mirror for Ubuntu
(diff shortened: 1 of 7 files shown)


src/query/predicate.cc
```diff
@@ -167,8 +167,7 @@ EvaluationResult PrefixPredicate::Evaluate(
   uint32_t word_count = 0;
   while (!word_iter.Done() && word_count < max_words) {
     std::string_view word = word_iter.GetWord();
-    if (!word.starts_with(term_)) break;
```
Aksha1812, line 170: wait why is this removed?

BCathcart: The word iterator is initialized with the term and will only traverse the subtree under that prefix so this check isn't needed

PR #693, opened 2026-02-05
Draft - composite query for stemming

Aksha1812:


src/commands/filter_parser.cc
```diff
@@ -758,12 +758,42 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseUnquotedTextToken(
     }
     VMSDK_RETURN_IF_ERROR(
         SetupTextFieldConfiguration(field_mask, field_or_default, false));
-    query_operations_ |= QueryOperations::kContainsTextTerm;
```
KarthikSubbarao, line 761: We need to keep this

src/commands/filter_parser.cc
```diff
@@ -758,12 +758,42 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseUnquotedTextToken(
...
+      FieldMaskPredicate stem_field_mask = 
+          field_mask & index_schema_.GetStemTextFieldMask();
+      
+      VMSDK_LOG(WARNING, nullptr) << "ParseUnquotedTextToken: token=" << token 
+                                  << ", field_mask=" << field_mask 
+                                  << ", stem_field_mask=" << stem_field_mask;
+      
+      if (stem_field_mask != 0) {
+        // Create OR composite: exact search of original OR non-exact of original
+        std::vector<std::unique_ptr<query::Predicate>> children;
```
KarthikSubbarao, line 774: Reserve with 2 would be nice

PR #707, opened 2026-02-06
Implement WEIGHT text field option, require a 1.0 value (fixes 517)

allenss-amazon:
* Add parsing for WEIGHT attribute but reject all values other than 1.0.
* Add weight to protobuf.
* Update tests.
*

### DIFF
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -663,12 +663,68 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         assert result[0] == 1, f"Expected 1 result for 'happi', got {result[0]}"
         assert set(result[1::2]) == {b"product:3"}, f"Expected {{product:3}}, got {set(result[1::2])}"
         
-        # Searching for "happy" should match both product:4 and product:5 (exact match)
+        # Searching for "happy" matches product:3 (happiness stems to happi), product:4 and product:5 (exact match)
         result = client.execute_command("FT.SEARCH", "testindex2", "happy", "DIALECT", "2")
-        assert result[0] == 2, f"Expected 2 results for 'happy', got {result[0]}"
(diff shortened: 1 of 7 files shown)


src/index_schema.proto
```diff
@@ -41,6 +41,7 @@ message IndexSchema {
 message TextIndex {
   bool with_suffix_trie = 1;
   bool no_stem = 2;
+  double weight = 3;
```
KarthikSubbarao, line 44: I see we are adding WEIGHT to the index field level protobuf in anticipation of when scoring is supported.

Should we do the same on the query side on text predicates? This would not be a question if we parsed the raw query string instead of sending the already constructed the query syntax tree over the wire

src/commands/ft_create_parser.cc
```diff
@@ -97,6 +97,7 @@ constexpr absl::string_view kNoStopWordsParam{"NOSTOPWORDS"};
 constexpr absl::string_view kStopWordsParam{"STOPWORDS"};
 constexpr absl::string_view kNoStemParam{"NOSTEM"};
 constexpr absl::string_view kMinStemSizeParam{"MINSTEMSIZE"};
+constexpr absl::string_view kWeight("WEIGHT");
```
KarthikSubbarao, line 100: Do we need to add anything else in FT.CREATE?
```
SCORE
SCORE_FIELD
```


Also, how about FT.SEARCH?
```
WITHSCORES
SCORER
```

KarthikSubbarao: We already have this
```
constexpr absl::string_view kScoreFieldParam{"SCORE_FIELD"};
constexpr absl::string_view kPayloadFieldParam{"PAYLOAD_FIELD"};
const absl::string_view kAsParam{"AS"};
const absl::string_view kOnParam{"ON"};
const absl::string_view kSeparatorParam{"SEPARATOR"};
const absl::string_view kCaseSensitiveParam{"CASESENSITIVE"};
const absl::string_view kScoreParam{"SCORE"};
```

PR #713, opened 2026-02-07
Test Configs and fix a bug in config

boda26:
1. Fix a misspelled name in config
2. Add check for duplicate config names
3. Create a script for testing configs

### DIFF
--- a/integration/test_configs.py
+++ b/integration/test_configs.py
@@ -1,8 +1,5 @@
-"""
-Test file for validating all search module configs.
-Tests CONFIG GET and CONFIG SET for each config parameter.
-"""
 import random
+import pytest
 from valkey_search_test_case import (
     ValkeySearchTestCaseBase, ValkeySearchClusterTestCase,
     ValkeySearchTestCaseDebugMode, ValkeySearchClusterTestCaseDebugMode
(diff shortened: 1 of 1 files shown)


src/query/response_generator.cc
```diff
@@ -40,7 +40,7 @@ constexpr absl::string_view kMaxSearchResultRecordSizeConfig{
 constexpr int kMaxSearchResultRecordSize{10 * 1024 * 1024};  // 10MB
 constexpr int kMinSearchResultRecordSize{100};
 constexpr absl::string_view kMaxSearchResultFieldsCountConfig{
-    "max-search-result-record-size"};
+    "max-search-result-fields-count"};
```
KarthikSubbarao, line 46: Let's please also fix `RegisterConfig` so that this never happens again.

Can you add a check to ensure we do not allow using the same config name?

```
void ModuleConfigManager::RegisterConfig(Registerable *config_item) {
  entries_.insert({std::string{config_item->GetName()}, config_item});
}
```

boda26: Added

integration/test_configs.py
```diff
@@ -0,0 +1,215 @@
...
+    ("prefer-partial-results", "Boolean", "APP"),
+    ("prefer-consistent-results", "Boolean", "APP"),
+    ("search-result-background-cleanup", "Boolean", "APP"),
+    ("high-priority-weight", "Number", "APP"),
+    ("ft-info-timeout-ms", "Number", "APP"),
+    ("ft-info-rpc-timeout-ms", "Number", "APP"),
+    ("local-fanout-queue-wait-threshold", "Number", "APP"),
+    ("thread-pool-wait-time-samples", "Number", "APP"),
+    ("max-term-expansions", "Number", "APP"),
+    ("search-result-buffer-multiplier", "String", "APP"),
```
KarthikSubbarao: Is this covered?

integration/test_configs.py
```diff
@@ -0,0 +1,227 @@
...
+    def test_hidden_configs_not_accessible(self):
+        self.assert_config_not_accessible(HIDDEN_CONFIGS)
+
+
+class TestHiddenConfigsCluster(ConfigTestMixin, ValkeySearchClusterTestCaseDebugMode):
+    def get_client(self):
+        return self.client_for_primary(0)
+
+    def test_hidden_configs_not_accessible(self):
+        self.assert_config_not_accessible(HIDDEN_CONFIGS)
```
KarthikSubbarao: Let's add a new line

PR #720, opened 2026-02-08
Add latency histogram for search workers time.

ohadshacham:
Add latency histogram for search workers time. When fanout, it takes the maximal time among all shards.


src/query/search.cc
```diff
@@ -721,15 +721,26 @@ SerializationRange SearchResult::GetSerializationRange(
 
 absl::StatusOr<SearchResult> Search(const SearchParameters &parameters,
                                     SearchMode search_mode) {
+  auto start_time = std::chrono::steady_clock::now();
```
eifrah-aws: I am not sure this is the correct clock to use. I would use platform specific API for getting the actual thread clock time (which excludes interrupts).

Something like:

```c++
timespec current;
clock_gettime(CLOCK_THREAD_CPUTIME_ID, &current);
```

eifrah-aws (changes requested):
Missing tests (UT or Integ)

integration/test_search_histogram.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+        
+        # Add some documents
+        for i in range(10):
+            vector = [float(i) / 10.0] * vector_dim
+            vector_bytes = b''.join(struct.pack('f', x) for x in vector)
+            self.client.hset(f"vdoc:{i}", "vec", vector_bytes)
+        
+        # Wait for indexing
+        time.sleep(0.5)
```
allenss-amazon: this isn't needed please remove. Ingesting data into a pre-existing index is synchronous.

integration/test_search_histogram.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+        
+        # Add some documents
+        for i in range(10):
+            vector = [float(i) / 10.0] * vector_dim
+            vector_bytes = b''.join(struct.pack('f', x) for x in vector)
+            self.client.hset(f"sdoc:{i}", "vec", vector_bytes)
+        
+        # Wait for indexing
+        time.sleep(0.5)
```
allenss-amazon: same

integration/test_search_histogram.py
```diff
@@ -0,0 +1,215 @@
...
+        # Queue FT.SEARCH
+        self.client.execute_command(
+            "FT.SEARCH", index_name,
+            "*=>[KNN 3 @vec $query]",
+            "PARAMS", "2", "query", query_bytes,
+            "DIALECT", "2"
+        )
+        
+        # Execute
+        results = self.client.execute_command("EXEC")
```
allenss-amazon: Is the only difference that this is within a multi/exec? If so, can we make loop out of the previous test to avoid an extra test case? Or maybe just drop it entirely.

integration/test_search_histogram.py
```diff
@@ -0,0 +1,215 @@
...
+        
+        # Add documents across shards
+        cluster_client = self.new_cluster_client()
+        for i in range(20):
+            vector = [float(i % 10) / 10.0] * vector_dim
+            vector_bytes = b''.join(struct.pack('f', x) for x in vector)
+            cluster_client.hset(f"fdoc:{i}", "vec", vector_bytes)
+        
+        # Wait for indexing
+        time.sleep(1.0)
```
allenss-amazon: remove

integration/test_search_histogram.py
```diff
@@ -0,0 +1,215 @@
...
+            "DIALECT", "2"
+        )
+        
+        # Verify search succeeded
+        assert result[0] > 0, "Fanout search should return results"
+        
+        # Check INFO on primary node 0
+        info = client.execute_command("INFO", "search")
+        
+        # Verify histogram was updated on this shard
```
allenss-amazon: If we're going to have a cluster-mode test, at least let's verify that the histogram gets updated on all nodes of the cluster.

src/query/fanout.cc
```diff
@@ -107,6 +108,12 @@ struct SearchPartitionResultsTracker {
       return;
     }
 
+    uint64_t shard_time = response.search_execution_time_us();
+    uint64_t current_max = max_search_time_us.load();
+    while (shard_time > current_max &&
+           !max_search_time_us.compare_exchange_weak(current_max, shard_time)) {
+    }
+
     absl::MutexLock lock(&mutex);
```
allenss-amazon: Is there some reason that the "max" loop can't be moved under the mutex??

src/query/fanout.cc
```diff
@@ -284,6 +292,12 @@ absl::Status PerformSearchFanoutAsync(
           if (result.ok()) {
             tracker->AddResults(result->neighbors);
             tracker->AddTotalCount(result->total_count);
+            uint64_t local_time = result->search_execution_time_us;
+            uint64_t current_max = tracker->max_search_time_us.load();
+            while (local_time > current_max &&
+                   !tracker->max_search_time_us.compare_exchange_weak(
+                       current_max, local_time)) {
+            }
```
allenss-amazon, line 363: Again, this feels like it should be done under the tracker's lock.

integration/test_search_histogram.py
ohadshacham: I would like to keep it since it uses the sync case and check a different path. extracted the shared parts between the tests.

src/query/fanout.cc
```diff
@@ -113,6 +114,12 @@ struct SearchPartitionResultsTracker {
     }
 
     absl::MutexLock lock(&mutex);
+
+    uint64_t shard_time = response.search_execution_time_us();
+    if (shard_time > max_search_time_us) {
+      max_search_time_us = shard_time;
```
eifrah-aws:
```c++
max_search_time_us = std::max(max_search_time_us, shard_time);
```

src/query/search.cc
```diff
@@ -762,15 +763,24 @@ SerializationRange SearchResult::GetSerializationRange(
 
 absl::StatusOr<SearchResult> Search(const SearchParameters &parameters,
                                     SearchMode search_mode) {
+  // Measure thread CPU time (excludes interrupts, context switches)
+  vmsdk::StopWatch timer(vmsdk::TimeType::kThreadCpu);
+
   auto result =
       MaybeAddIndexedContent(DoSearch(parameters, search_mode), parameters);
```
eifrah-aws: We can basically use here:

```c++
VMSDK_ASSIGN_OR_RETURN(auto res, MaybeAddIndexedContent(DoSearch(parameters, search_mode), parameters);
```

src/query/search.cc
```diff
@@ -762,15 +763,24 @@ SerializationRange SearchResult::GetSerializationRange(
...
+
+  uint64_t search_execution_time_us =
+      absl::ToInt64Microseconds(timer.Duration());
+
   if (!result.ok()) {
     return result.status();
   }
```
eifrah-aws: And we can remove this

vmsdk/src/utils.h
```diff
@@ -22,15 +23,55 @@
...
   ~StopWatch() = default;
-  void Reset() { start_time_ = absl::Now(); }
-  absl::Duration Duration() const { return absl::Now() - start_time_; }
+
+  void Reset() {
+    if (time_type_ == TimeType::kWallClock) {
+      start_time_ = absl::Now();
+    } else {
+      start_thread_time_ns_ = GetThreadTimeNanos();
+    }
```
eifrah-aws: In general, use `switch` over `if .. else ...` this allows the compiler to detect unhandled enumerators from the enum `TimeType`

vmsdk/src/utils.h
```diff
@@ -22,15 +23,55 @@
...
+  }
+
+  absl::Duration Duration() const {
+    if (time_type_ == TimeType::kWallClock) {
+      return absl::Now() - start_time_;
+    } else {
+      uint64_t current_ns = GetThreadTimeNanos();
+      return absl::Nanoseconds(current_ns - start_thread_time_ns_);
```
eifrah-aws: Use `switch`

vmsdk/src/utils.h
```diff
@@ -22,15 +23,55 @@
...
+    }
+  }
 
  private:
-  absl::Time start_time_;
+  TimeType time_type_;
```
eifrah-aws: Initialise with `TimeType::kWallClock`:

```c++
TimeType time_type_{TimeType::kWallClock};
```

vmsdk/src/utils.h
```diff
@@ -22,15 +23,55 @@
...
+enum class TimeType {
+  kWallClock,  // Wall-clock time (includes interrupts, context switches)
+  kThreadCpu   // Thread CPU time (excludes interrupts, context switches)
+};
+
 class StopWatch {
  public:
-  StopWatch() { Reset(); }
+  // Default constructor uses wall-clock time
+  StopWatch() : StopWatch(TimeType::kWallClock) {}
```
eifrah-aws: Do we need this if we set a default value for the other constructor?

vmsdk/src/utils.h
```diff
@@ -22,15 +23,55 @@
...
+
+  // Constructor that allows specifying the time type
+  explicit StopWatch(TimeType type) : time_type_(type) { Reset(); }
```
eifrah-aws: We can pass a default value here: `TimeType type = TimeType::kWallClock`

eifrah-aws (approved):

allenss-amazon:
closing, inactive

PR #722, opened 2026-02-08
Implement SLOP and INORDER in FT.AGGREGATE

allenss-amazon:
Because QMA isn't yet implemented, FT.AGGREGATE needs a command level SLOP and INORDER clause added.


src/commands/ft_aggregate_parser.cc
```diff
@@ -270,6 +272,10 @@ vmsdk::KeyValueParser<AggregateParameters> CreateAggregateParser() {
                         GENERATE_VALUE_PARSER(AggregateParameters, timeout_ms));
   parser.AddParamParser(kAddScoresParam,
                         GENERATE_FLAG_PARSER(AggregateParameters, addscores_));
+  parser.AddParamParser(kSlopParam,
```
KarthikSubbarao, line 275: I am thinking about this in case of FT.AGGREGATE and FT.SEARCH both. Should SLOP be limited by a max-slop config?

KarthikSubbarao: Checked on other search modules. They don't have a limit.

```
127.0.0.1:6381> FT.SEARCH myIdx "Écrire" SLOP 2147483647
1) (integer) 1
2) "doc:1"
3) 1) "content"
   2) "\xc3\x89crire is a French verb"
127.0.0.1:6381> FT.SEARCH myIdx "Écrire" SLOP 2147483648
(error) Bad arguments for SLOP: Value is outside acceptable bounds
```

KarthikSubbarao: Also, we discussed how a higher slop value does not change the memory overhead / cpu overhead

PR #726, opened 2026-02-09
Multi DB tests in CME/CMD for FT.SEARCH

Aksha1812:
Added test to verify FT.SEARCH database isolation for Valkey 9's Multi DB feature. Tests run in both CMD/CME modes, with TEXT, NUMERIC, TAG, and VECTOR fields across multiple indexes.

### DIFF
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -688,13 +688,11 @@ absl::Status SchemaManager::LoadIndex(
   // Load the index schema into memory
   auto index_schema_pb = std::unique_ptr<data_model::IndexSchema>(
       section->release_index_schema_contents());
-
   VMSDK_ASSIGN_OR_RETURN(auto index_schema,
                          IndexSchema::LoadFromRDB(ctx, mutations_thread_pool_,
                                                   std::move(index_schema_pb),
                                                   std::move(supplemental_iter)),
                          _ << "Failed to load index schema from RDB!");
-
   uint32_t db_num = index_schema->GetDBNum();
   const std::string &name = index_schema->GetName();
 


integration/test_multidb_search.py
```diff
@@ -0,0 +1,190 @@
...
+            vec = [0.0, 0.0, 0.0, 0.0]
+            vec[db_num % 4] = 1.0
+            clients[db_num].hset('0', mapping={
+                'name': f'product{db_num}',
+                'price': str(db_num * 100),
+                'category': f'cat{db_num}',
+                'vec': float_to_bytes(vec)
+            })
+        
+        # Check text search isolation
```
KarthikSubbarao: For all the searches, can we please check that the document and content matches the search based on the DB?

integration/test_multidb_search.py
```diff
@@ -0,0 +1,213 @@
+"""
+Multi DB tests for FT.SEARCH - checks that different databases are isolated.
```
KarthikSubbarao, line 2: Searching and ingestion is very good to validate. Thank you for adding them.


A few other areas which would be good to check are:
1. RDB Save/Load
2. Slot Migration

integration/test_multidb_search.py
```diff
@@ -0,0 +1,213 @@
...
+        ])
+        
+        for db_num in range(num_dbs):
+            index.create(clients[db_num])
+        
+        # Add unique data to each DB
+        for db_num in range(num_dbs):
+            vec = [0.0, 0.0, 0.0, 0.0]
+            vec[db_num % 4] = 1.0
+            clients[db_num].hset('0', mapping={
```
KarthikSubbarao: What is `'0'` here?

integration/test_multidb_search.py
```diff
@@ -0,0 +1,213 @@
...
+            Numeric('price'),
+            Tag('category'),
+            Vector('vec', dim=4)
+        ])
+        
+        for db_num in range(num_dbs):
+            index.create(clients[db_num])
+        
+        # Add unique data to each DB
+        for db_num in range(num_dbs):
```
KarthikSubbarao: Can we add at least two variants here when it comes to keys?
1. We want to test insert the same X key names across the Y DBs
2. We want to to test Z unique key names across the Y DBs.


Searches should be able to return results correctly when it comes to different keys and common keys across DBs

integration/test_multidb_search.py
KarthikSubbarao: Are we testing on a single key across all dbs?
See: https://github.com/valkey-io/valkey-search/pull/726/changes#r2797043120

Aksha1812: '0' is just for me to know which slot it hashes to so i could use cluster which that slot is in. otherwise i guess would have to look in all clusters.

Aksha1812:
Added tests . Ready for rereview @KarthikSubbarao 

RDB Save/Load
Slot Migration
We want to test insert the same X key names across the Y DBs
We want to to test Z unique key names across the Y DBs.

integration/test_multidb_search.py
allenss-amazon: I would agree about RDB save/load. OSS Search doesn't support server-side atomic slot migration. Old-style non-atomic migration doesn't require module support.

integration/indexes.py
```diff
@@ -131,6 +131,31 @@ def create(self, data_type: KeyDataType):
     def make_value(self, row: int, column: int, type: KeyDataType) -> Union[str, bytes, float, list[float]]:
         return f"Tag:{row}:{column}"
 
+
+class Text(Field):
```
KarthikSubbarao, line 135: There are conflicts here. You will need a merge from main to update your branch

integration/test_multidb_search.py
```diff
@@ -0,0 +1,386 @@
...
+    """Create clients for multiple databases."""
+    clients = {}
+    for db_num in range(num_dbs):
+        client = get_client_func()
+        client.select(db_num)
+        clients[db_num] = client
+    return clients
+
+
+def create_index(clients, num_dbs, prefixes=None):
```
KarthikSubbarao:
```suggestion
def create_indexes(clients, num_dbs, prefixes=None):
```

integration/test_multidb_search.py
```diff
@@ -0,0 +1,386 @@
...
+                'vec': float_to_bytes(vec)
+            })
+    
+    return db_keys
+
+
+class TestMultiDBCMD(ValkeySearchTestCaseDebugMode):
+    """Standalone mode tests"""
+
+    def test_multidb_common_keys_isolation_CMD(self):
```
KarthikSubbarao: Can we test both common and unique key creation + search in the same test?

If there is a reason why we cannot, that is fine - let me know. I just was trying to reduce test code here

integration/test_multidb_search.py
```diff
@@ -0,0 +1,386 @@
...
+        for db_num in range(num_dbs):
+            waiters.wait_for_true(lambda db=db_num: index.backfill_complete(clients[db]))
+        
+        verify_search_results(clients, num_dbs, key_names, index)
+
+
+class TestMultiDBCME(ValkeySearchClusterTestCaseDebugMode):
+    """Cluster mode tests"""
+
+    def test_multidb_common_keys_isolation_CME(self):
```
KarthikSubbarao: Can we test both common and unique key creation + search in the same test?

If there is a reason why we cannot, that is fine - let me know. I just was trying to reduce test code here

integration/test_multidb_search.py
```diff
@@ -0,0 +1,386 @@
...
+from valkey_search_test_case import (
+    ValkeySearchTestCaseDebugMode,
+    ValkeySearchClusterTestCaseDebugMode
+)
+from valkeytestframework.conftest import resource_port_tracker
+from valkeytestframework.util import waiters
+from indexes import Index, Text, Tag, Numeric, Vector, float_to_bytes
+
+
+def verify_search_results(clients, num_dbs, key_names, index):
```
KarthikSubbarao:
```suggestion
def verify_common_keys_results(clients, num_dbs, key_names, index):
```
?

integration/test_multidb_search.py
```diff
@@ -0,0 +1,386 @@
...
+    """Add same key names to all DBs with different data values."""
+    key_names = [f'{key_prefix}{i}' for i in range(num_keys)]
+    
+    for db_num in range(num_dbs):
+        for key_idx, key_name in enumerate(key_names):
+            vec = [0.0, 0.0, 0.0, 0.0]
+            vec[db_num % 4] = 1.0
+            clients[db_num].hset(key_name, mapping={
+                'name': f'product{db_num}_key{key_idx}',
+                'price': str(db_num * 100 + key_idx),
```
KarthikSubbarao, line 127: Is there a way we can try to search using queries which only match certain DBs and not all?

Example:: Have a term in 1 DB or have a numeric range in just 1 DB, and search for validating that case

Aksha1812: isn't that what i am doing already ? I have unque terms in each db and then i am searching for them and i am verifying that they are in that db and not in other dbs. Didn't fully understand the request here.

integration/test_multidb_search.py
Aksha1812: done

integration/test_multidb_search.py
Aksha1812: done

src/schema_manager.cc
```diff
@@ -688,21 +688,25 @@ absl::Status SchemaManager::LoadIndex(
...
-                         _ << "Failed to load index schema from RDB!");
-  uint32_t db_num = index_schema->GetDBNum();
-  const std::string &name = index_schema->GetName();
 
-  // Select the DB number in the context for subsequent usage.
+  // Select the DB number in the context BEFORE loading the index schema.
+  // This is critical because LoadFromRDB (specifically LoadIndexExtension in
+  // RDB v2) calls ProcessKeyspaceNotification which opens keys using the
+  // context, so the context must be pointing to the correct DB.
+  uint32_t db_num = index_schema_pb->db_num();
```
allenss-amazon: There's a select db guard object recently created. Should that be used here?

Aksha1812: Here the select db code block returns an error if unable to select. I see the guard does not , perhaps i could get its boolean response and give the same error but that it would be an unnecessary extra step. Also the guard seems to be used when there is a db change , i don't think that is the case here. Also i am not touching selectdb code at all here

src/schema_manager.cc
```diff
@@ -688,21 +688,25 @@ absl::Status SchemaManager::LoadIndex(
   // Load the index schema into memory
   auto index_schema_pb = std::unique_ptr<data_model::IndexSchema>(
       section->release_index_schema_contents());
-  VMSDK_ASSIGN_OR_RETURN(auto index_schema,
-                         IndexSchema::LoadFromRDB(ctx, mutations_thread_pool_,
-                                                  std::move(index_schema_pb),
```
murphyjacob4, line 693: Since index_schema_pb has DB num in it, should we make IndexSchema::LoadFromRDB do the select?

Aksha1812: do you mean move the selectdb call to loadfromrdb? makes sense. lmk if i understood you wrong. Updated PR

murphyjacob4: Yeah exactly

PR #728, opened 2026-02-10
Sanitize module logging 

### LINKED ISSUE #704: [BUG] Sanitize module logging
Sanitize module logging.

Add a config that decides whether to include data or no and have a common utility function which can be used across logging sites.

Arunsarma07:
This PR aims in sanitizing the logging information for the search module. 

Implements a new `--hide-user-data-from-log` configuration option similar to Valkey core to control whether to include data in log messages. 

### Related Issue

Closes #704

### DIFF
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -1402,31 +1402,12 @@ absl::Status IndexSchema::SaveIndexExtension(RDBChunkOutputStream out) const {
   // this key list explicitly which will trivially enable the SKIPINITIALSCAN
   // option.
   //
-  std::shared_ptr<indexes::IndexBase> index;
-  for (const auto &attribute : attributes_) {
-    if (!IsVectorIndex(attribute.second.GetIndex())) {
-      index = attribute.second.GetIndex();
-      break;
-    }
-  }
-  if (!index) {
-    VMSDK_RETURN_IF_ERROR(out.SaveObject<size_t>(0));  // zero keys
-  } else {
-    size_t key_count =
-        index->GetTrackedKeyCount() + index->GetUnTrackedKeyCount();
-    VMSDK_RETURN_IF_ERROR(out.SaveObject(key_count));
-    rdb_save_keys.Increment(key_count);
-    VMSDK_LOG(NOTICE, nullptr)
-        << "Writing Index Extension, keys = " << key_count;
-
-    auto write_a_key = [&](const Key &key) {
-      key_count--;
-      return out.SaveString(key->Str());
-    };
-    VMSDK_RETURN_IF_ERROR(index->ForEachTrackedKey(write_a_key));
-    VMSDK_RETURN_IF_ERROR(index->ForEachUnTrackedKey(write_a_key));
-    CHECK(key_count == 0) << "Key count mismatch for index "
-                          << vmsdk::config::RedactIfNeeded(GetName());
+  size_t key_count = db_key_info_.Get().size();
+  VMSDK_RETURN_IF_ERROR(out.SaveObject(key_count));
+  rdb_save_keys.Increment(key_count);
+  VMSDK_LOG(NOTICE, nullptr) << "Writing Index Extension, keys = " << key_count;
+  for (auto &[key, _] : db_key_info_.Get()) {
+    VMSDK_RETURN_IF_ERROR(out.SaveString(key->Str()));
   }
   //
   // Write out the mutation queue entries. As an optimization we only write
@@ -1672,11 +1653,13 @@ void IndexSchema::DrainMutationQueue(ValkeyModuleCtx *ctx) const {
 void IndexSchema::OnLoadingEnded(ValkeyModuleCtx *ctx) {
   if (loaded_v2_) {
     loaded_v2_ = false;
-    VMSDK_LOG_EVERY_N_SEC(NOTICE, ctx, 1)
-        << "RDB load completed, Mutation Queue contains "
-        << tracked_mutated_records_.size() << " entries."
-        << (backfill_job_.Get().has_value() ? " Backfill still required."
-                                            : " Backfill not needed.");
+    VMSDK_LOG(NOTICE, ctx) << "RDB load completed, "
+                           << " Mutation Queue contains "
+                           << tracked_mutated_records_.size() << " entries."
+                           << (backfill_job_.Get().has_value() &&
+                                       !backfill_job_.Get()->IsScanDone()
+                                   ? " Backfill still required."
+                                   : " Backfill not needed.");
     if (options::GetDrainMutationQueueOnLoad().GetValue()) {
       DrainMutationQueue(ctx);
     }


src/schema_manager.cc
```diff
@@ -790,8 +798,9 @@ absl::Status SchemaManager::ShowIndexSchemas(ValkeyModuleCtx *ctx,
     ValkeyModule_ReplyWithArray(ctx, inner_map.size());
     for (const auto &[name, schema] : inner_map) {
       auto proto = schema->ToProto()->DebugString();
-      VMSDK_LOG(NOTICE, ctx)
-          << "Index Schema in DB " << db_num << ": " << name << " " << proto;
+      VMSDK_LOG(NOTICE, ctx) << "Index Schema in DB " << db_num << ": "
```
KarthikSubbarao: This function is only used from the FT._DEBUG command. Maybe this can be DEBUG log level

src/schema_manager.cc
```diff
@@ -790,8 +798,9 @@ absl::Status SchemaManager::ShowIndexSchemas(ValkeyModuleCtx *ctx,
...
+                             << vmsdk::config::RedactIfNeeded(name) << " "
+                             << vmsdk::config::RedactIfNeeded(proto);
```
KarthikSubbarao: Question: What is proto here?

src/schema_manager.cc
```diff
@@ -687,14 +693,15 @@ absl::Status SchemaManager::LoadIndex(
   if (ValkeyModule_SelectDb(ctx, db_num) != VALKEYMODULE_OK) {
     return absl::InternalError(
         absl::StrFormat("Unable to select DB %d for loading of index schema %s",
-                        db_num, name.c_str()));
+                        db_num, vmsdk::config::RedactIfNeeded(name).data()));
```
KarthikSubbarao, line 700: Is this an error reply to a client or does this go to a log?

src/coordinator/server.cc
```diff
@@ -302,8 +302,8 @@ Service::GenerateInfoResponse(
         "Index fingerprint/version or slot fingerprint mismatch");
     response.set_error_type(
         coordinator::FanoutErrorType::INCONSISTENT_STATE_ERROR);
-    VMSDK_LOG(NOTICE, nullptr) << "DEBUG: Fingerprint, version or slot "
-                                  "fingerprint mismatch at server.cc";
+    VMSDK_LOG(DEBUG, nullptr) << "DEBUG: Fingerprint, version or slot "
```
KarthikSubbarao: @boda26 - Can you describe in what cases this error is logged? Based on that we can evaluate often we expect this error case to happen. If this is unexpected in a provisioned cluster, NOTICE is good. Just don't want to spam logs

src/query/response_generator.cc
```diff
@@ -308,8 +308,8 @@ void ProcessNeighborsForReply(ValkeyModuleCtx *ctx,
       ++Metrics::GetStats().query_result_record_dropped_cnt;
       VMSDK_LOG(WARNING, ctx)
           << "Content field number exceeds configured limit of "
-          << max_content_fields
-          << " for neighbor with ID: " << (*neighbor.external_id).Str();
+          << max_content_fields << " for neighbor with ID: "
+          << vmsdk::config::RedactIfNeeded((*neighbor.external_id).Str());
```
KarthikSubbarao: Hmm.. This can easily be spammed by a customer if they have documents that are large in their fields. We get here from every FT.SEARCH/FT.AGGREGATE command. We should for sure change this to DEBUG (unless we have some form of log throttler)

src/query/response_generator.cc
```diff
@@ -320,7 +320,8 @@ void ProcessNeighborsForReply(ValkeyModuleCtx *ctx,
         ++Metrics::GetStats().query_result_record_dropped_cnt;
         VMSDK_LOG(WARNING, ctx)
             << "Content size exceeds configured limit of " << max_content_size
-            << " bytes for neighbor with ID: " << (*neighbor.external_id).Str();
+            << " bytes for neighbor with ID: "
+            << vmsdk::config::RedactIfNeeded((*neighbor.external_id).Str());
```
KarthikSubbarao: This can easily be spammed by a customer if they have documents that are large in their fields. We get here from every FT.SEARCH/FT.AGGREGATE command. We should for sure change this to DEBUG (unless we have some form of log throttler)

src/index_schema.cc
KarthikSubbarao, line 104: Adding a reminder for myself. I need to start reviewing starting from this file downwards

src/query/response_generator.cc
allenss-amazon: VMSDK_LOG_EVERY_N_SEC

KarthikSubbarao: Thanks

src/schema_manager.cc
Arunsarma07: We can make this to DEBUG level.

src/schema_manager.cc
Arunsarma07: The proto is the protobuf object that contains all the information like the index schema name, db num, attribute type, mutable stats and whole other information.

src/schema_manager.cc
Arunsarma07: This is being logged in the RDB serialization. When we are calling the `PerformRDBLoad` method, we are storing the result and logging the result when it fails to do so. 

https://github.com/valkey-io/valkey-search/blob/main/src/rdb_serialization.cc#L176
https://github.com/valkey-io/valkey-search/blob/main/src/rdb_serialization.cc#L219

src/query/response_generator.cc
Arunsarma07: Sure. I can change this to DEBUG or  VMSDK_LOG_EVERY_N_SEC.

src/index_schema.cc
KarthikSubbarao, line 1512: I am questioning why we are adding a log for every attribute of every index. 

Won't this spam the logs?

src/index_schema.cc
```diff
@@ -1550,7 +1561,7 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> IndexSchema::LoadFromRDB(
               supplemental_content->key_to_id_map_header().attribute();
           VMSDK_LOG(NOTICE, nullptr)
               << "Loading Key to ID Map Content for attribute: "
-              << attribute.alias();
```
KarthikSubbarao: Is there a benefit for this to be using NOTICE log level? 

This is for every attribute/field of every index, right?

src/schema_manager.cc
```diff
@@ -513,12 +515,14 @@ void SchemaManager::OnFlushDBEnded(ValkeyModuleCtx *ctx) {
 
   auto to_delete = GetIndexSchemasInDBInternal(selected_db);
   for (const auto &name : to_delete) {
-    VMSDK_LOG(NOTICE, ctx) << "Deleting index schema " << name
+    VMSDK_LOG(NOTICE, ctx) << "Deleting index schema "
```
KarthikSubbarao: Same comment here and applies everywhere. Why do we want to add a NOTICE log for every index deleted. We just updated the default number of allowed indexes to be 1000, with a max of 10k.


I think just a log at the end of `OnFlushDBEnded` saying how many indexes were deleted is better for a Notice level log. Or a general log in the function outside the loop saying Deleting index schemas on FLUSHDB of DB X

src/schema_manager.cc
```diff
@@ -513,12 +515,14 @@ void SchemaManager::OnFlushDBEnded(ValkeyModuleCtx *ctx) {
...
+                           << vmsdk::config::RedactIfNeeded(name)
                            << " on FLUSHDB of DB " << selected_db;
     auto old_schema = RemoveIndexSchemaInternal(selected_db, name);
     if (!old_schema.ok()) {
       VMSDK_LOG(WARNING, ctx)
```
KarthikSubbarao, line 528: This is unexpected so it is fine to leave as WARNING

src/schema_manager.cc
```diff
@@ -527,13 +531,15 @@ void SchemaManager::OnFlushDBEnded(ValkeyModuleCtx *ctx) {
       // cluster-level construct, not a node-level construct. To delete,
       // FT.DROPINDEX must be done explicitly.
       auto to_add = old_schema.value()->ToProto();
-      VMSDK_LOG(NOTICE, ctx) << "Recreating index schema " << name
-                             << " on FLUSHDB of DB " << selected_db;
+      VMSDK_LOG(NOTICE, ctx)
+          << "Recreating index schema " << vmsdk::config::RedactIfNeeded(name)
+          << " on FLUSHDB of DB " << selected_db;
```
KarthikSubbarao: Ideally we do this once (outside the for loop) and not for every index schema

src/schema_manager.cc
KarthikSubbarao, line 673: Is this called once per index?

Arunsarma07: This is called over each RDB section and this method internally calls `LoadFromRDB` that iterates over each index in that particular RDB section.

src/schema_manager.cc
Arunsarma07: That make sense. But do you think if we can leave some of the logs on DEBUG level and add few generic logs outside the loops or remove all the spammed logs and add only generic logs?

src/index_schema.cc
Arunsarma07: It does spam with large number indexes and attributes.

src/index_schema.cc
Arunsarma07: It might benefit if we switch it to DEBUG.

src/coordinator/server.cc
KarthikSubbarao: @boda26 - ping

boda26: This happens only when there is a index fingerprint/version mismatch or a slot fingerprint mismatch between two nodes. This is unexpected in a normal cluster, so I also think NOTICE/WARNING is good. I was using NOTICE but someone changed to DEBUG

KarthikSubbarao: @Arunsarma07 , let's change this to NOTICE. A better approach would be using an internal DEV Metric

Arunsarma07: Sure, will change this.

allenss-amazon: Perhaps VMSDK_LOG_EVERY_N_SEC is more appropriate. I'm concerned about spamming the log for transient errors.

src/index_schema.cc
```diff
@@ -1243,11 +1247,14 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
 
   auto rdb_section_string = rdb_section->SerializeAsString();
   VMSDK_RETURN_IF_ERROR(rdb->SaveStringBuffer(rdb_section_string))
-      << "IO error while saving IndexSchema name: " << this->name_
+      << "IO error while saving IndexSchema name: "
+      << vmsdk::config::RedactIfNeeded(this->name_)
       << " in DB: " << this->db_num_ << " to RDB";
 
+  VMSDK_LOG(NOTICE, nullptr) << "Starting to save attributes";
```
KarthikSubbarao: Nit: We have the count, right? Can we include that?

Arunsarma07: Sure. Will add the count.

PR #732, opened 2026-02-10
Tests save and restore functionality at max memory.

Nivesh-01:
Tests save and restore functionality at max memory. We are going to use only one node and test for all indexes.


testing/integration/stability_runner.py
```diff
@@ -101,16 +104,25 @@ def run(self) -> StabilityRunResult:
...
-                require_full_coverage=True,
-                socket_timeout=10,
-            )
+            if self.is_standalone:
+                # Use standalone client for single node
+                client = valkey.Valkey(
+                    host="localhost",
+                    port=self.config.ports[0],
+                    socket_timeout=10,
+                )
```
BCathcart, line 113: You could keep the single node CME and not worry about a separate client configuration

testing/integration/stability_runner.py
```diff
@@ -466,15 +481,34 @@ def run(self) -> StabilityRunResult:
...
-        )
+        
+        # For OOM testing, HSET errors are expected when memory limit is reached
+        if self.is_standalone and self.config.maxmemory:
+            # OOM errors are expected - don't count them as failures
+            processes.append(
+                utils.MemtierProcess(
+                    command=insert_command, 
+                    name="HSET",
+                    # error_predicate=lambda err: "OOM command not allowed" in err
```
BCathcart, line 492: With this commented out it's the same as the else case

testing/integration/stability_runner.py
```diff
@@ -466,15 +481,34 @@ def run(self) -> StabilityRunResult:
...
+                )
+            )
+        else:
+            processes.append(
+                utils.MemtierProcess(command=insert_command, name="HSET")
+            )
+        
+        # Only add DEL and EXPIRE if we have replicas (normal stability test)
+        # For OOM testing with single node, we only want HSET to maximize memory usage
```
BCathcart, line 501: I don't love that we have test-specific logic in what is supposed to be a generic test runner. Can we add flags to enable/disable DEL and EXPIRE that any test can specify? They can be enabled by default

testing/integration/stability_runner.py
```diff
@@ -508,10 +542,41 @@ def run(self) -> StabilityRunResult:
         timeout_start = time.time()
         processes_killed_for_failover = False
         time_when_killed = 0
+        oom_detected = False
+        oom_recovery_triggered = False
         
         while time.time() - timeout_start < self.config.test_timeout:
             elapsed = time.time() - test_start_time
             
+            # Check for OOM in standalone mode
```
BCathcart, line 551: This is more test-specific code in the common runner. I'm getting the sense that this test doesn't fit very well into the endurance test framework, and I'm worried we're hurting the maintainability here.

testing/integration/stability_test.py
BCathcart, line 614: This is a common function as well right? A big portion of this code is now specific to the one-off save/restore testing.

I'm concluding at this point that we should move this testing to our integration testing framework. I was fine having the test as part of the endurance test framework if it fit into it nicely. I think the endurance test cadence with the daily or weekly github job is more suited for this test. I don't think we need to run it on every PR revision like we do with other integration tests. However, it seems like it's too fundamentally different to integrate into the endurance test framework. After we move it to the integration test framework, I'd suggest we have a special github to run it occasionally and exclude it from the regular integration test target.

testing/integration/stability_runner.py
Nivesh-01: Oh yes I will get that out, I was trying to mitigate teh error where we get repeated errors when we reach max memory but this way caused problems to another code logic in the testing

testing/integration/stability_runner.py
Nivesh-01: Yes we should be able to do that

PR #735, opened 2026-02-11
Solving Text Negation by using a Universal Set + Filtering

KarthikSubbarao:
Solving Text Negation by using a Universal Set + Filtering. Our plan is to optimize this in possible two in the future:
1. Have AND blocks containing Proximity properties (which is always the case when global inorder/slop is used) return a Universal Text Iterator - instead of applying demorgan's transformation on it.
2. Solve AND completely if it is containing Proximity properties (which is always the case when global inorder/slop is used). Solving AND completely needs a ComposedAND Iterator that takes all the children iterators (including numeric/tag/text/nested composed) and ensures that the NextKey API returns keys in asc order. If hybrid, it should not return that it has positions. The NextKey needs to performing an intersection, therefore requiring us to hold the keys of numeric and tag as ephemeral memory while the keys intersection is performed by the ComposedAND iterator. Similar iterator will be needed for ComposedOR potentially. But when the AND block does not have proximity properties, we can just use demorgan's transformation.

Right now, this PR detects if a query has Text + Negation, and skips the entries fetcher evaluator of predicates and simply returns an iterator that goes over the universal set. It also skips duplication checks (since there cannot be any), and goes into prefiltering where unmatching keys are removed. There is no memory overhead because the universal_text_fetcher is only iterating over the existing global indexmap's keys.

### DIFF
--- a/src/query/response_generator.cc
+++ b/src/query/response_generator.cc
@@ -81,13 +81,18 @@ namespace valkey_search::query {
 
 class PredicateEvaluator : public query::Evaluator {
  public:
-  explicit PredicateEvaluator(const RecordsMap &records)
-      : records_(records), text_index_(nullptr) {}
+  PredicateEvaluator(const RecordsMap &records,
+                     QueryOperations query_operations)
+      : Evaluator(query_operations), records_(records), text_index_(nullptr) {}
 
   PredicateEvaluator(const RecordsMap &records,
                      const valkey_search::indexes::text::TextIndex *text_index,
-                     InternedStringPtr target_key)
-      : records_(records), text_index_(text_index), target_key_(target_key) {}
+                     InternedStringPtr target_key,
+                     QueryOperations query_operations)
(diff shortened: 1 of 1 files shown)


integration/test_fulltext.py
```diff
@@ -2026,6 +1988,31 @@ def test_hybrid_non_vector_query(self):
         assert result[0] == 5
         assert set(result[1::2]) == {b"hash:00", b"hash:01", b"hash:02", b"hash:03", b"hash:04"}
 
+    def test_text_negation(self):
```
daddaman-amz, line 1991: we can leverage the  more comprehensive test from this PR: https://github.com/valkey-io/valkey-search/pull/733/changes

KarthikSubbarao: Yes, you can add them in the follow up

daddaman-amz (approved):
We need a more comprehensive test for this either in this or followup

PR #742, opened 2026-02-12
Restructure Content Fetching to match key blocking semantics

allenss-amazon:
Architecture Restructuring: Search Completion Model

  The core change is replacing main's callback-based search completion with a virtual method pattern:

  1. SearchResult moved into SearchParameters

  - SearchResult is now a member of SearchParameters (stored in-place), instead of being returned separately via absl::StatusOr<SearchResult> or passed through callbacks
  - Search() now returns absl::Status instead of absl::StatusOr<SearchResult>

  2. Callback pattern removed

  - Deleted SearchResponseCallback parameter from SearchAsync() and PerformSearchFanoutAsync()
  - Deleted the on_done_callback lambda in Execute() and the InitiatorSearch subclass
  - Deleted src/query/inflight_retry.h and src/query/inflight_retry.cc entirely

  3. Virtual method completion replaces callbacks

  - SearchParameters gains two pure virtual methods: QueryCompleteBackground() and QueryCompleteMainThread()
  - QueryCommand implements these — storing blocked_client as a member and unblocking the client on completion
  - New ContentProcessing enum classifies queries by their content needs (no content, content available, content required, contention required)
  - New src/query/content_resolution.cc/.h replaces the inflight retry logic

  4. Other changes

  - ComputeSearchTargets() extracted as a helper (with single-slot query support via hash tag parsing)
  - sortby_parameter moved from SearchCommand up to SearchParameters (available to all query types)
  - Coordinator server (server.cc) significantly simplified (~220 lines reduced)
  - SearchParameters constructor simplified (default constructor added)
  - Various test updates to match the new patterns

### DIFF
--- a/src/query/fanout.cc
+++ b/src/query/fanout.cc
@@ -68,6 +68,18 @@ struct NeighborComparator {
 // the top k results to the callback.
 struct SearchPartitionResultsTracker {
   absl::Mutex mutex;
+  // Holds the LocalResponderSearch after it completes, keeping its
+  // SearchParameters fields (return_attributes, sortby_parameter, etc.) alive.
+  // Neighbors moved from the local search into `results` contain RecordsMap
+  // entries whose string_view keys point into those fields. Without this, the
+  // LocalResponderSearch would be destroyed immediately after adding its
+  // results to the tracker, leaving dangling string_view keys that are read
+  // when the priority queue reallocates (triggering absl::flat_hash_map rehash
+  // on move).
+  //
+  // Since there can only be a single LocalResponder, this doesn't need a lock.
+  //
+  std::unique_ptr<SearchParameters> local_responder_;
   std::priority_queue<indexes::Neighbor, std::vector<indexes::Neighbor>,
                       NeighborComparator>
       results ABSL_GUARDED_BY(mutex);
@@ -218,6 +230,18 @@ class LocalResponderSearch : public query::SearchParameters {
           << "Error during local handling of FT.SEARCH: "
           << search_result.status.message();
     }
+    // Stash `self` (the LocalResponderSearch) in the tracker so that its
+    // SearchParameters fields outlive the Neighbor entries moved into
+    // `results` above. Those entries' RecordsMaps contain string_view keys
+    // that reference data owned by this SearchParameters object.
+    //
+    // We must break the circular reference first: this LocalResponderSearch
+    // holds a shared_ptr to the tracker, so storing `self` in the tracker
+    // would create a cycle (tracker -> self -> tracker). Copy the shared_ptr
+    // to a local, clear the member, then stash.
+    auto tracker_copy = tracker;
+    tracker.reset();
+    tracker_copy->parameters->local_responder_ = std::move(self);
   }
 };
 
--- a/src/query/search.h
+++ b/src/query/search.h
@@ -237,6 +237,13 @@ struct SearchParameters {
   // resolution due to contention with in-flight mutations.
   unsigned int content_resolution_blocked_{0};
 
+  // In CME, when a LocalResponderSearch is used, the neighbors of that
+  // operation get moved into this operation. But the neighbors has string_view
+  // references into the return_attributes of the owning operation. So in order
+  // to avoid a use-after-free we retain the LocalResponderSearch until this
+  // operation is destroyed.
+  std::unique_ptr<SearchParameters> local_responder_;
+
   SearchParameters() = default;
   SearchParameters(uint64_t timeout_ms, cancel::Token token, uint32_t db_num)
       : timeout_ms(timeout_ms), cancellation_token(token), db_num_(db_num) {}


allenss-amazon:
Architecture Restructuring: Search Completion Model

  The core change is replacing main's callback-based search completion with a virtual method pattern:

  1. SearchResult moved into SearchParameters

  - SearchResult is now a member of SearchParameters (stored in-place), instead of being returned separately via absl::StatusOr<SearchResult> or passed through callbacks
  - Search() now returns absl::Status instead of absl::StatusOr<SearchResult>

  2. Callback pattern removed

  - Deleted SearchResponseCallback parameter from SearchAsync() and PerformSearchFanoutAsync()
  - Deleted the on_done_callback lambda in Execute() and the InitiatorSearch subclass
  - Deleted src/query/inflight_retry.h and src/query/inflight_retry.cc entirely

  3. Virtual method completion replaces callbacks

  - SearchParameters gains two pure virtual methods: QueryCompleteBackground() and QueryCompleteMainThread()
  - QueryCommand implements these — storing blocked_client as a member and unblocking the client on completion
  - New ContentProcessing enum classifies queries by their content needs (no content, content available, content required, contention required)
  - New src/query/content_resolution.cc/.h replaces the inflight retry logic

  4. Other changes

  - ComputeSearchTargets() extracted as a helper (with single-slot query support via hash tag parsing)
  - sortby_parameter moved from SearchCommand up to SearchParameters (available to all query types)
  - Coordinator server (server.cc) significantly simplified (~220 lines reduced)
  - SearchParameters constructor simplified (default constructor added)
  - Various test updates to match the new patterns

src/query/search.cc
```diff
@@ -1062,4 +1073,15 @@ absl::Status query::SearchParameters::PostParseQueryString() {
   return absl::OkStatus();
 }
 
+ContentProcessing SearchParameters::GetContentProcessing() const {
+  if (no_content) {
```
KarthikSubbarao, line 1076: Is this also going to be the case when `RETURN 0` is true?

src/query/content_resolution.cc
```diff
@@ -0,0 +1,71 @@
...
+#include "vmsdk/src/managed_pointers.h"
+
+namespace valkey_search::query {
+
+void ResolveContent(std::unique_ptr<SearchParameters> params) {
+  // 1. Check cancellation
+  if (params->cancellation_token->IsCancelled()) {
+    params->QueryCompleteMainThread(std::move(params));
+    return;
+  }
```
KarthikSubbarao, line 27: Can we add a test for this where we enter this code path specifically (maybe through a pause point)?

src/index_schema.cc
```diff
@@ -1690,14 +1696,26 @@ vmsdk::BlockedClientCategory IndexSchema::GetBlockedCategoryFromProto() const {
...
+    }
+    // Sequence number mismatch — check the mutation queue for this key.
+    absl::MutexLock lock(&mutated_records_mutex_);
     auto itr = tracked_mutated_records_.find(neighbor.external_id);
     if (itr != tracked_mutated_records_.end()) {
-      itr->second.waiting_queries.insert(std::move(query_ctx));
+      if (params->content_resolution_blocked_++ == 0) {
+        ++Metrics::GetStats().fulltext_query_blocked_cnt;
+      }
+      ++Metrics::GetStats().fulltext_query_retry_cnt;
```
KarthikSubbarao: I thought we planned on renaming these to `text`. not fulltext

src/index_schema.h
```diff
@@ -232,8 +232,7 @@ class IndexSchema : public KeyspaceEventSubscription,
     std::optional<absl::flat_hash_map<std::string, AttributeData>> attributes;
     std::vector<vmsdk::BlockedClient> blocked_clients;
     // Queries waiting for this mutation to complete
-    absl::flat_hash_set<std::shared_ptr<query::InFlightRetryContext>>
-        waiting_queries;
+    std::vector<std::unique_ptr<query::SearchParameters>> waiting_queries;
```
KarthikSubbarao, line 235: Is this only used by text queries currently since they are the only one needing contention checks?

src/commands/commands.cc
```diff
@@ -242,26 +182,31 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
+void QueryCommand::QueryCompleteBackground(
+    std::unique_ptr<SearchParameters> parameters) {
+  blocked_client->SetReplyPrivateData(parameters.release());
+  blocked_client->UnblockClient();
+}
+
+void QueryCommand::QueryCompleteMainThread(
+    std::unique_ptr<SearchParameters> parameters) {
+  QueryCompleteBackground(std::move(parameters));
+}
```
KarthikSubbarao, line 217: This is called from ResolveContent. 


For a text query, this means we are unblocking the client from the main thread. 

There will be a time interval between this and the Replycallback from the blocked client.

My question (at this point in the review) is whether we are using a pre-formed reply / cached neighbors content from this RunByMain iterator to use as the response? If we don't do this, the index can be out of date if we try again to fetch and  verify the filter (unless the mutations are blocked on the text index)

src/query/content_resolution.cc
```diff
@@ -0,0 +1,71 @@
...
+      // ConsumeTrackedMutatedAttribute will call ResolveContent again
+      // via RunByMain when the mutation completes.
+      return;
+    }
+    // If PerformKeyContentionCheck returns false, params is still valid (not
+    // moved). Fall through to content fetch.
+  }
+
+  // 3. Content fetch + filter via ProcessNeighborsForReply
+  auto ctx = vmsdk::MakeUniqueValkeyThreadSafeContext(nullptr);
```
KarthikSubbarao, line 43: Is this needed if we are already on the Mainthread? or are we not guaranteed to be on the main thread at this time of code?

src/commands/commands.cc
```diff
@@ -242,26 +182,31 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
 
+void QueryCommand::QueryCompleteBackground(
+    std::unique_ptr<SearchParameters> parameters) {
+  blocked_client->SetReplyPrivateData(parameters.release());
+  blocked_client->UnblockClient();
+}
+
+void QueryCommand::QueryCompleteMainThread(
+    std::unique_ptr<SearchParameters> parameters) {
+  QueryCompleteBackground(std::move(parameters));
```
KarthikSubbarao: It feels odd to have mainthread call background thread. can we move the code into a helper fn?
This is thread independent 
```
  blocked_client->SetReplyPrivateData(parameters.release());
  blocked_client->UnblockClient();
```

src/query/content_resolution.cc
```diff
@@ -0,0 +1,71 @@
...
+
+  std::optional<std::string> vector_identifier = std::nullopt;
+  if (!params->attribute_alias.empty()) {
+    auto id = params->index_schema->GetIdentifier(params->attribute_alias);
+    if (id.ok()) {
+      vector_identifier = *id;
+    }
+  }
+
+  query::ProcessNeighborsForReply(ctx.get(), attribute_data_type,
```
KarthikSubbarao, line 56: Question: Does this store the content on the neighbors? so that when we are on the Replycallback, we skip fetching the content?

src/query/content_resolution.cc
```diff
@@ -0,0 +1,71 @@
...
+#include <utility>
+
+#include "src/index_schema.h"
+#include "src/query/response_generator.h"
+#include "src/query/search.h"
+#include "vmsdk/src/managed_pointers.h"
+
+namespace valkey_search::query {
+
+void ResolveContent(std::unique_ptr<SearchParameters> params) {
```
KarthikSubbarao, line 21: Can we add a verification that we are on the main thread?

src/coordinator/server.cc
```diff
@@ -125,58 +119,34 @@ class RemoteResponderSearch : public query::SearchParameters {
...
-        (total_count > removed) ? (total_count - removed) : 0;
-    SerializeNeighbors(response, neighbors);
-    response->set_total_count(adjusted_total_count);
+  void QueryCompleteBackground(
+      std::unique_ptr<SearchParameters> self) override {
+    CHECK(no_content);
+    SerializeNeighbors(response, search_result.neighbors);
+    response->set_total_count(search_result.total_count);
     reactor->Finish(grpc::Status::OK);
     RecordSearchMetrics(false, std::move(latency_sample));
```
KarthikSubbarao: This part of the code seems common with QueryCompleteMainThread

src/query/fanout.cc
```diff
@@ -193,28 +186,21 @@ struct SearchPartitionResultsTracker {
...
-    tracker->AddTotalCount(total_count);
+  void QueryCompleteMainThread(
+      std::unique_ptr<SearchParameters> self) override {
+    QueryCompleteBackground(std::move(self));
   }
 
-  void OnCancelled() override {
-    if (enable_partial_results) {
-      OnComplete(neighbors);
+  void QueryCompleteBackground(
```
KarthikSubbarao, line 217: Question: This is only called by `QueryCompleteMainThread` above, right?

KarthikSubbarao: IMO, if one of the two (QueryCompleteMainThread or QueryCompleteBackground) are not expected to be called in any class that implements `query::SearchParameters`, it would be clearer to add CHECK to ensure no one enters it. 
Also, doing this will help for readers (I am trying to see which ones are actively being called)

(Same comment for all other classes implementing these 2 APIs)

src/query/content_resolution.cc
```diff
@@ -0,0 +1,71 @@
...
+  // 1. Check cancellation
+  if (params->cancellation_token->IsCancelled()) {
+    params->QueryCompleteMainThread(std::move(params));
+    return;
+  }
+
+  // 2. If kContentionRequired, check for in-flight mutations
```
KarthikSubbarao: Is a better name `kContentionCheckRequired`? We actually want to avoid contention (so, kContentionRequired seems odd as a name)

src/query/content_resolution.h
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#ifndef VALKEY_SEARCH_SRC_QUERY_CONTENT_RESOLUTION_H_
+#define VALKEY_SEARCH_SRC_QUERY_CONTENT_RESOLUTION_H_
+
+#include <memory>
```
KarthikSubbarao, line 11: is this used?

src/query/search.cc
allenss-amazon: RETURN 0 doesn't set no_content. The parser should probably reject having RETURN and NO_CONTENT specified. I will open a bug for this case. https://github.com/valkey-io/valkey-search/issues/770

This PR treats no_content the same as it was before.

src/query/content_resolution.cc
allenss-amazon: yes. I'll add a test issue.

allenss-amazon: https://github.com/valkey-io/valkey-search/issues/771

src/index_schema.cc
allenss-amazon: I don't think I changed the name.

src/index_schema.h
allenss-amazon: It's used by any  search where GetContentProcessing returns kContentionRequired. Currently that's only text.

src/commands/commands.cc
allenss-amazon: Yes, at this point, the documentation is clear that all data has been fetched and no implemention of QueryCompletedMainThread or QueryCompletedBackground is allowed to access the database.

ResolveContent should have fully fetched all of the content BEFORE invoking these functions.

src/query/content_resolution.cc
allenss-amazon: It's needed deep down inside of GetContext, we have to set the DBNum in order to access the data and that requires a context.

allenss-amazon: I've opened an issue in the core PR to Fix the blocking logic so that we can get rid of the extra RunByMain which I think means we can use the blocked client context for this, right?

src/commands/commands.cc
allenss-amazon: It's not the mainthread calling the background thread. It's saying that regardless of which thread context the query has completed, I'm going to do the same work to complete the command.

src/query/content_resolution.cc
allenss-amazon: Sure, there used to be one, Apparently claude killed it.

src/query/fanout.cc
allenss-amazon: Adding checks are fine.

src/query/content_resolution.cc
allenss-amazon: Sure.

src/query/content_resolution.cc
allenss-amazon: Yes, this function will both call GetContent as needed as well as perform filtering on mutated keys (and update total_count!)

src/coordinator/server.cc
allenss-amazon: Refactored to share a common helper.

src/query/content_resolution.h
allenss-amazon: memory? Yes, for shared_ptr.

src/index_schema.cc
KarthikSubbarao: it was added here: https://github.com/valkey-io/valkey-search/pull/541

"Block Search Response on Conflicting In-Flight Keys for fulltext queries#541"


Since it is related to this PR, Can we fix the name? And also check the App/Dev status

KarthikSubbarao:
@allenss-amazon , can we check the clang and integration test failures?

PR #744, opened 2026-02-12
Add attribute data to FT.INFO PRIMARY and CLUSTER mode

boda26:
1. Provide a simple implementation of getting tracked keys in text index
2. Add per-attribute information, including user_memory_bytes and num_records, in the FT.INFO PRIMARY/CLUSTER response
3. Update integration tests to align with new command output

### DIFF
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -36,10 +36,18 @@ Text::Text(const data_model::TextIndex &text_index_proto,
 
 absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr &key,
                                      absl::string_view data) {
+  absl::MutexLock lock(&index_mutex_);
   auto result = text_index_schema_->StageAttributeData(
       key, data, text_field_number_, !no_stem_, with_suffix_trie_);
   if (result.ok() && *result) {
-    tracked_keys_.insert(key);
+    auto [_, succ] = tracked_keys_.insert(key);
+    if (!succ) {
+      return absl::AlreadyExistsError(
+          absl::StrCat("Key `", key->Str(), "` already exists"));
+    }
+    untracked_keys_.erase(key);
+  } else {
+    untracked_keys_.insert(key);
   }
   return result;
 }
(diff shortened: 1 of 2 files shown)


src/indexes/text.cc
```diff
@@ -48,14 +47,14 @@ absl::StatusOr<bool> Text::RemoveRecord(const InternedStringPtr &key,
...
-  // TODO: key tracking
+  tracked_keys_.erase(key);
 
   return true;
 }
 
 absl::StatusOr<bool> Text::ModifyRecord(const InternedStringPtr &key,
                                         absl::string_view data) {
-  // TODO: key tracking
+  tracked_keys_.insert(key);
```
KarthikSubbarao: tracked_keys is `using InternedStringSet`

Doesn't this lead to duplicate key tracking?

KarthikSubbarao: See `ModifyRecord` in other Numeric/Tag

daddaman-amz: Since its flat_hash_set, it wont add duplicate entries.

But maybe it would be better if we check the result of  staging in shared index (the return statement here and in Add record)  and then based on it  insert it into the set

boda26: Agreed, added the check before inserting into set

allenss-amazon:
I'd like to avoid anything that depends on tracked/untracked keys as I want to get rid of them, they are space-inefficient and I suspect provide no actual value. Counters are are different story...

src/indexes/text.h
```diff
@@ -139,6 +141,7 @@ class Text : public IndexBase {
   std::shared_ptr<text::TextIndexSchema> text_index_schema_;
 
   InternedStringSet untracked_keys_;
```
KarthikSubbarao, line 142: Can we delete this if it is unused?

src/indexes/text.h
```diff
@@ -139,6 +141,7 @@ class Text : public IndexBase {
...
+  InternedStringSet tracked_keys_;
```
KarthikSubbarao, line 145: Let's add a comment explaining that it is only used currently for metrics

KarthikSubbarao (approved):

PR #756, opened 2026-02-14
FT.AGGREGATE complex query test

boda26:
Added FT.AGGREGATE complex query test in CMD and CME.

### DIFF
--- a/integration/test_non_vector.py
+++ b/integration/test_non_vector.py
@@ -293,6 +293,10 @@ def validate_aggregate_complex_queries(client: Valkey):
     )
     assert result[0] == 5
     assert result[1][1] == b'1000'
+    assert result[2][1] == b'999'
+    assert result[3][1] == b'998'
+    assert result[4][1] == b'997'
+    assert result[5][1] == b'996'
 
     # 4. APPLY with arithmetic expression
     result = client.execute_command(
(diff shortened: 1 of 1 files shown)


integration/test_non_vector.py
```diff
@@ -243,6 +257,203 @@ def validate_aggregate_queries(client: Valkey):
...
+    assert result[3][1] == b'3'
+
+    # 3. SORTBY with MAX
+    result = client.execute_command(
+        "FT.AGGREGATE", "products", "@price:[0 1000]",
+        "LOAD", "1", "price",
+        "SORTBY", "2", "@price", "DESC", "MAX", "5"
+    )
+    assert result[0] == 5
+    assert result[1][1] == b'1000'
```
KarthikSubbarao, line 295: What does MAX do here? Why do we only check the first key?

boda26: https://redis.io/docs/latest/commands/ft.aggregate/
MAX is used to sort only for top n elements. Using it here is just checking if the arg works, the result is the same. Checking the first value is the largest should be fine

integration/test_non_vector.py
```diff
@@ -243,6 +257,203 @@ def validate_aggregate_queries(client: Valkey):
...
+    )
+    assert result[0] == 5
+    assert result[1][1] == b'990'
+    assert result[-1][1] == b'994'
+
+    # 16. multiple LIMIT stages
+    result = client.execute_command(
+        "FT.AGGREGATE", "products", "@price:[0 1000]",
+        "LOAD", "1", "price",
+        "SORTBY", "2", "@price", "ASC", "MAX", "1000",
```
KarthikSubbarao: Can we use a value other than 1000? Otherwise, we cannot see the effect of MAX

integration/test_non_vector.py
KarthikSubbarao: Let's check the other 4

PR #763, opened 2026-02-16
Add fanout-data-uniformity config

KarthikSubbarao:
Add fanout-data-uniformity config

The idea in the PR is to assume that users will have even data distribution and search results across shards, so we can fetch K / N (num shards) by default from each fanout request.
There is a config added in the PR to increase the amount of data fetched optionally. 
U defines how uniform the data and search result distribution across the shards are. U=100 means 100% uniformly distributed, U=0 means the data is all on one shard.

This will help reduce the amount of data sent over the network and processed across nodes (remote and coordinator)

```
/// Register the "fanout-data-uniformity" flag
/// U = uniformity (0-100): 100 = uniform distribution, 0 = all data in one shard
/// Formula: limit_per_shard = (K/N) + ((100-U) * (K - K/N) / 100)
```

### DIFF
--- a/src/valkey_search_options.cc
+++ b/src/valkey_search_options.cc
@@ -455,6 +455,28 @@ static auto async_fanout_threshold =
         kMaximumAsyncFanoutThreshold)  // max threshold (10k)
         .Build();
 
+/// Register the "--max-nonvector-search-results-fetched" flag. Controls the
+/// maximum number of results to fetch in background threads before content
+/// fetching on non-vector (numeric/tag/text) query paths. This controls
+/// potential OOM by limiting the result set size before expensive content
+/// fetching from the keyspace.
+/// Note: If queries use LIMIT with a large offset (e.g., LIMIT offset count
+/// where offset + count exceeds this value), consider increasing this config
+/// to ensure all paginated results are accessible.
+constexpr absl::string_view kMaxNonVectorSearchResultsFetchedConfig{
+    "max-nonvector-search-results-fetched"};
+constexpr uint32_t kDefaultMaxNonVectorSearchResultsFetched{
+    100000};  // 100K keys default
+constexpr uint32_t kMinimumMaxNonVectorSearchResultsFetched{0};
+constexpr uint32_t kMaximumMaxNonVectorSearchResultsFetched{UINT32_MAX};
+static auto max_nonvector_search_results_fetched =
+    vmsdk::config::NumberBuilder(
+        kMaxNonVectorSearchResultsFetchedConfig,   // name
+        kDefaultMaxNonVectorSearchResultsFetched,  // default limit (100K)
+        kMinimumMaxNonVectorSearchResultsFetched,  // min limit (0)
+        kMaximumMaxNonVectorSearchResultsFetched)  // UINT32_MAX
+        .Build();
+
 uint32_t GetQueryStringBytes() { return query_string_bytes->GetValue(); }
 
 vmsdk::config::Number& GetHNSWBlockSize() {
@@ -562,6 +584,11 @@ vmsdk::config::Number& GetFanoutUniformityMinIndexSize() {
 
 vmsdk::config::Number& GetAsyncFanoutThreshold() {
   return dynamic_cast<vmsdk::config::Number&>(*async_fanout_threshold);
+}
+
+vmsdk::config::Number& GetMaxNonVectorSearchResultsFetched() {
+  return dynamic_cast<vmsdk::config::Number&>(
+      *max_nonvector_search_results_fetched);
 }
 
 }  // namespace options
--- a/integration/test_non_vector.py
+++ b/integration/test_non_vector.py
@@ -7,6 +7,7 @@ import random
 from valkey.cluster import ValkeyCluster
 from valkey_search_test_case import ValkeySearchClusterTestCase
 import time
+import pytest
 
 """
 This file contains tests for non vector (numeric and tag) queries on Hash/JSON documents in Valkey Search - in CME / CMD.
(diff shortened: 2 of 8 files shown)


allenss-amazon (commented):
I think this is VERY important optimization, as will it substantially improve the scalability for network-heavy query operations.

I dislike having the fanout-limit-multiplier to change based on the number of shards. I don't think that's really usable. Surely we can find some numbering scheme that's shard-count independent. In other words, can we tie the setting of this configuration to the data distribution profile.

src/query/search.h
```diff
@@ -118,6 +118,7 @@ struct SearchParameters {
   bool inorder{false};
   std::optional<uint32_t> slop;
   bool verbatim{false};
+  std::optional<query::SortByParameter> sortby;
```
AlexFilipImproving: If we are adding sortby into SearchParameters, then we would also want to remove it from SearchCommand, so it's not duplicated. This is also the reason I made Pre/PostParseQueryCommand virtual, so those wouldn't need to be virtual anymore.
@allenss-amazon can you remind me why you asked me to put sortby into the SortCommand class?

allenss-amazon: The suggestion to put it in SortCommand made sense when I believed it didn't need to get transmitted across the wire. But that didn't hold true, because it's needed for remote searches, moving it into SearchParameters simplified everything.

AlexFilipImproving: Ok, in that case there is a bit to clean up, since I had to thread the sortby parameter through some function arguments and that's a bit redundant now, but we can do that at a later time.

allenss-amazon:
> > This is an interesting idea. The only downside is it requires client side changes to adopt the usage of this parameter
> 
> I’m concerned that a global configuration for fanout-data-uniformity will only be effective for a narrow subset of use cases. In practice, data distribution can vary significantly across different indexes, and even for the same index under different filter permutations.
> 
> As an alternative, we could dynamically determine the number of entries requested from each node using the following approach:
> 
> 1. Coordinator Estimation: The client-facing node estimates, in some cases calculates, the number of qualifying entries matching the filter within its own local store.
> 2. Dynamic Fanout: This value is communicated to all shards during the fanout process.
> 3. Adaptive Shard Responses: Each shard compares its local qualifying count against the coordinator’s base value to decide how many results to return:
> 
> * Higher local density: If the shard has significantly more matching entries than the coordinator, it returns more results to ensure global accuracy.
> * Lower local density: If the shard has fewer matching entries, it returns fewer results to minimize network overhead.
> * Uniform distribution: If the counts are similar, the node fetches about the same entries as the client facing node.
> 
> This approach allows the system to adapt to data skew in real-time without relying on a static uniformity assumption

I would agree that a global constant value is inferior. I think there are two directions we could move forward on. One would be to allow per-command override of the global value. This is still inferior, but is a simple short-term solution for wizard-level users. The second direction is more along the lines that @yairgott describes above. The second solution is clearly something that will require substantial research before it could be productized and hence is out of the 1.2 timeframe. It's conceivable that we would re-open 1.2 for the first direction -- but I'm skeptical that this is really worth it.

allenss-amazon (commented):
The need to change several tests strongly suggests that this change is a problem. It's clear that for small corpus sizes that this algorithm is causing problems.

I would recommend that we create a second configurable value, which the minimum index size (the local size of index_db_info works fine) before we apply this value. If the index is smaller than we just fetch everything. This should address the fact that small index sizes appear to have pathologically uneven data distribution.

KarthikSubbarao:
@allenss-amazon , I updated the PR with the config based on the index schema size to decide when to skip the optimization

KarthikSubbarao:
@allenss-amazon / @yairgott  , I changed the config to have U=0, meaning we assume that all data can reside in one shard , so we fetch K from every shard. Also, I made these configs Dev only

src/commands/ft_aggregate.cc
```diff
@@ -376,12 +376,12 @@ absl::Status SendReplyInner(ValkeyModuleCtx *ctx,
...
-// TODO: Implement the correct logic to detect if the FT.AGGREGATE query has a
-// clause (e.g. sorting) that requires all neighbors to be returned for the
-// correct search result.
+// SortBy and GroupBy stages need all results from shards.
+// Return true in these cases to indicate results must not be trimmed.
 bool AggregateParameters::RequiresCompleteResults() const {
   for (const auto &stage : stages_) {
-    if (dynamic_cast<const SortBy *>(stage.get()) != nullptr) {
+    if (dynamic_cast<const SortBy *>(stage.get()) != nullptr ||
+        dynamic_cast<const GroupBy *>(stage.get()) != nullptr) {
```
allenss-amazon: 1. I hate dynamic casts it almost always means that there's a better way to do it, i.e., Virtual functions here.
2. Doesn't the GetLimit function returning [0..MAX_POS] provide the same answer? Can't we reuse that?

src/query/fanout.cc
```diff
@@ -283,25 +285,47 @@ absl::Status PerformSearchFanoutAsync(
...
+    // calculate a limit based on the distribution profile to cover (offset +
+    // limit) results across the cluster.
+
+    // For queries requiring complete results (e.g., with SORTBY), we must
+    // fetch K results from each shard to guarantee global correctness.
+    // Also, for small indices (below the configured threshold), we skip the
+    // optimization.
+    if (index_size < min_index_size || parameters->RequiresCompleteResults()) {
+      request->mutable_limit()->set_first_index(0);
+      request->mutable_limit()->set_number(K);
```
allenss-amazon, line 303: K is wrong for non-vector searches.

KarthikSubbarao:
```
  uint64_t K = parameters->limit.first_index + parameters->limit.number;
```

src/commands/ft_aggregate.cc
KarthikSubbarao: > Doesn't the GetLimit function returning [0..MAX_POS] provide the same answer? Can't we reuse that?

What is the function name? I could not find `GetLimit`. 

Here, we want to indicate if the Search command needs complete results based on whether it has sorting stages

src/query/fanout.cc
allenss-amazon: The diff didn't show that. Sorry.

src/commands/ft_aggregate.cc
allenss-amazon: We had discussed a function that would return nullopt or an actual range. with 0..max_pos as indicating "all".

KarthikSubbarao: That is what `GetSerializationRange` does. Is that what you are referring to?

It tells us what range of keys to return. But it is built on `SearchResult`. This depends on the `RequiresCompleteResults` to be implemented correctly.

`RequiresCompleteResults` needs to return whether it needs complete results (due to sorting) or not

src/valkey_search_options.cc
```diff
@@ -407,6 +407,39 @@ constexpr absl::string_view kDrainMutationQueueOnSaveConfig{
...
+constexpr absl::string_view kFanoutDataUniformityConfig{
+    "fanout-data-uniformity"};
+constexpr uint32_t kDefaultFanoutDataUniformity{0};
+constexpr uint32_t kMinimumFanoutDataUniformity{0};
+constexpr uint32_t kMaximumFanoutDataUniformity{100};
+static auto fanout_data_uniformity =
+    config::NumberBuilder(
+        kFanoutDataUniformityConfig, kDefaultFanoutDataUniformity,
+        kMinimumFanoutDataUniformity, kMaximumFanoutDataUniformity)
+        .Dev()  // can only be set in debug mode
```
allenss-amazon, line 425: Hmm, if we believe in this, why isn't it an App item?

KarthikSubbarao: Yair mentioned we can use a Dev Config for now. If we see a strong benefit in this approach, we can change it to App

PR #764, opened 2026-02-16
Fix for handling trailing punctuations in query

daddaman-amz:
Fix to not throw syntax error when there is a punctuation towards end of the query

Old PR which was closed: https://github.com/valkey-io/valkey-search/pull/588

### DIFF
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -1874,8 +1874,8 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         result = client.execute_command("FT.SEARCH", "idx1", '%%drive%%')
         assert (result[0], set(result[1::2])) == (2, {b"doc:4", b"doc:11"})
         # stemming enabled: should give doc:3 (with word 'driving')
-        result = client.execute_command("FT.SEARCH", "idx2", '%%drive%%')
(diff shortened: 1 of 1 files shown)


src/commands/filter_parser.cc
```diff
@@ -878,7 +879,12 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::ParseTextTokens(
     node_count_ += terms.size() + 1;
   } else {
     if (terms.empty()) {
-      return absl::InvalidArgumentError("Invalid Query Syntax");
+      if (query_operations_ == QueryOperations::kNone) {
+        return absl::InvalidArgumentError("Invalid Query Syntax");
+      }
+      // Already processed valid operations, trailing punctuation already
+      // consumed by loop. Return std::nullopt to indicate "no more terms"
+      return std::nullopt;
```
KarthikSubbarao, line 887: Question: Is it ever possible for there to be any predicate beyond this? (I would imagine not)

daddaman-amz: Let me double check (since its been a while) if this path hits if there is remaining predicate in which case comment needs to be corrected

daddaman-amz: yes we won't have any more predicates. We hit this block when there are no more tokens which is true if punctuations are at the end

PR #774, opened 2026-02-17
Quick fix for unwanted concurrency on index_key_info_

BCathcart:
I suggest we move the fetching of the mutation sequence number into the time slice mutex reader phase which is currently within `DoSearch`, but as a very quick fix we can just acquire the mutex.


src/index_schema.h
```diff
@@ -270,6 +270,7 @@ class IndexSchema : public KeyspaceEventSubscription,
   };
 
   MutationSequenceNumber GetIndexMutationSequenceNumber(const Key &key) const {
```
KarthikSubbarao, line 272: This function should not used by the main thread, right? We could check for that

BCathcart: Looks like it can be called from the main thread `if (ABSL_PREDICT_FALSE(!ValkeySearch::Instance().SupportParallelQueries() ||
                           inside_multi_exec)) {`
                           
Since we have the mutex on it for now, it doesn't really matter where it's called from. If we get rid of the mutex, then we need to ensure it's within the read phase of the time-slice mutex and we need to move this call that can modify `index_key_info_` into the write phase of the time-slice mutex on the ingestion side: `auto mutation_record = ConsumeTrackedMutatedAttribute(key, first_time);`

PR #775, opened 2026-02-17
Change default RDB version to 2

boda26:
1. Change default RDB version to 2
2. Fix unit tests to use RDB version 1

### DIFF
--- a/testing/index_schema_test.cc
+++ b/testing/index_schema_test.cc
@@ -1172,7 +1172,7 @@ class IndexSchemaRDBTest : public ValkeySearchTest {
  protected:
   // Intercepting the setup to change RDB version to 1
   // Currently these tests only work with RDB version 1
-  // TODO: Will be fixed to work with RDB version 2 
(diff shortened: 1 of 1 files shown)


testing/index_schema_test.cc
```diff
@@ -1168,7 +1168,28 @@ INSTANTIATE_TEST_SUITE_P(
              (std::get<0>(info.param) ? "WithThreadPool" : "WithoutThreadPool");
     });
 
-class IndexSchemaRDBTest : public ValkeySearchTest {};
+class IndexSchemaRDBTest : public ValkeySearchTest {
+ protected:
+  void SetUp() override {
+    ValkeySearchTest::SetUp();
```
KarthikSubbarao, line 1177: Are you intercepting the setup to change RDB V2 to false? Can we add a comment that explains why and also that we plan to address this to also add unit testing capability for RDB V2 shortly

boda26: I added a comment

PR #777, opened 2026-02-17
Fix build script for ASAN integration tests

baswanth09:


integration/run.sh
```diff
@@ -197,4 +200,6 @@ if [[ "${SAN_BUILD}" != "no" ]]; then
   # And now we can check the logs
   logfiles=$(find ${LOGS_DIR} -name "*.log")
   check_for_san_errors "${logfiles}"
+  # Check pytest output for sanitizer errors
+  check_for_san_errors "${PYTEST_OUTPUT_LOG}"
```
KarthikSubbarao, line 204: We might need to suppress the leaks such as LUA from the core, but all others from the module should be tracked 

Usually a suppress pattern file can help here

baswanth09: Looks like we already have one: https://github.com/valkey-io/valkey-search/blob/main/ci/asan.supp

BCathcart:
Let's see if it passes once #929 is merged

BCathcart (approved):

BCathcart:
Failing test is a shutdown false positive in a new test I introduced. It should be fixed by #961.

PR #778, opened 2026-02-18
Draft - Reduce Stemming calls to improve performance

Aksha1812:
When running ingestion perf tests saw lot of time spent on running Lexer::sb_stemmer_stem and UnicodeNormalizer::CaseFold (which i think should be handled by #746 ). I saw while staging data for each word we try to stem . We filter on if stemming is enabled , and if word length is greater than  stem size. but that is not enough . While staging we don't need to do stemming if we have already stemmed that word once, since that has already been added to stem mappings, 

Currently adding a simple code change to temporarily track unique words within a document and skip stemming if already stemmed in past. Performance might be a lot better if we keep track across documents. But have to do this in a way that doesn't increase memory. So maybe a temp data structure while Staging data should work. Will check if performance improved with below code change or not, will follow up accordingly. Current change will only improve performance if we have high num of tokens in a doc , around >1000 tokens per doc and if there are repeated tokens in the doc.

### DIFF
--- a/src/utils/bloom_filter.h
+++ b/src/utils/bloom_filter.h
@@ -38,7 +38,8 @@ class BloomFilter {
     Clear();
   }
 
-  // Creates an optimally-sized bloom filter for expected capacity and false positive rate.
+  // Creates an optimally-sized bloom filter for expected capacity and false
+  // positive rate.
   static BloomFilter CreateOptimal(
       size_t expected_capacity,
       double false_positive_rate = kDefaultFalsePositiveRate) {
--- a/testing/lexer_test.cc
+++ b/testing/lexer_test.cc
@@ -199,21 +199,22 @@ TEST_F(LexerTest, StemMappingsBasic) {
       stem_mappings;
 
   auto result =
-      lexer_->Tokenize("running jumps happily", true, 3, &stem_mappings);
+      lexer_->Tokenize("swimming dances crazily", true, 3, &stem_mappings);
 
   ASSERT_TRUE(result.ok());
   // Original words (case-folded, not stemmed)
-  EXPECT_EQ(*result, std::vector<std::string>({"running", "jumps", "happily"}));
+  EXPECT_EQ(*result,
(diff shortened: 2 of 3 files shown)


src/indexes/text/lexer.cc
```diff
@@ -91,6 +91,11 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
 
   // Get or create the thread-local stemmer for this lexer's language
   sb_stemmer* stemmer = stemming_enabled ? GetStemmer() : nullptr;
+  
+  // Local cache to track words that have already been stemmed in this document.
+  // This avoids redundant stemming calls for repeated words.
+  absl::flat_hash_set<std::string> stemmed_words;
```
KarthikSubbarao: Can this grow to be the size of the field? We can have large fields, right? (in order of GBs). 

Should we consider having this tracking the most recent X variants to avoid OOMing?

src/indexes/text/lexer.cc
```diff
@@ -143,12 +148,15 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
...
-      if (stemming_enabled) {
+      if (stemming_enabled && !stemmed_words.contains(word)) {
+        // Only stem if we haven't already stemmed this word in this document
         std::string stemmed_word = StemWord(word, stemmer, min_stem_size);
         if (word != stemmed_word) {
           CHECK(stem_mappings) << "stem_mappings must not be null";
           (*stem_mappings)[stemmed_word].insert(word);
         }
+        // Mark this word as stemmed to avoid redundant stemming calls
+        stemmed_words.insert(word);
```
KarthikSubbarao: set creation and insertion (causing expansion) has some cost. We can inline instead?

Aksha1812:
Yes @KarthikSubbarao the prefix tree lookup replaced the multiple stemming calls but added more overhead and made it less efficient. will think of alternatives.

PR #781, opened 2026-02-18
Save and Restore at max memory for text, tag, and numeric

Nivesh-01:
Save and Restore at max memory for text, tag, and numeric. Currently max memory is set to 1GB can be easily changed. We can also have multiple cycles of the same run.


integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+    """
+    Endurance test for save/restore operations at maximum memory.
+    
+    This test:
+    1. Sets maxmemory with noeviction policy
+    2. Loads data until OOM
+    3. Saves RDB
+    4. Restarts and restores from RDB
+    5. Verifies data integrity and functionality
+    6. Repeats for multiple cycles to detect memory issues
```
BCathcart: What benefit does repeating give us? I think one save restore is enough

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+    def _print_memory_info(self, client: Valkey, label: str) -> dict:
+        """Print memory info and return the info dict"""
+        info = client.info("memory")
+        print(f"\n{'='*60}")
+        print(f"INFO MEMORY - {label}")
+        print('='*60)
+        
+        keys = [
+            'used_memory',
+            'used_memory_human',
```
BCathcart: Let's be consistent and print the human version for all cases. In general though, we can discuss what metrics we want to track. Maybe `used_memory_peak` relative to `maxmemory` will give us enough insight, or maybe we need to limit the process memory to the max memory value and monitor for swapping.

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+            "FT.SEARCH",
+            index_name,
+            query,
+            "LIMIT", "0", "10"
+        )
+        count = result[0]
+        logging.info(f"Search returned {count} results")
+        return count
+    
+    # currently not being used but added in for increasing functionality for later iterations.
```
BCathcart: What's the use case for this that is worth doing later down the line but not right now?

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+        # Track memory across cycles
+        memory_history = []
+        
+        for cycle in range(self.NUM_CYCLES):
+            logging.info(f"\n{'='*60}")
+            logging.info(f"STARTING CYCLE {cycle + 1}/{self.NUM_CYCLES}")
+            logging.info(f"{'='*60}")
+            print("\n" + "="*60)
+            print(f"CYCLE {cycle + 1}/{self.NUM_CYCLES}")
+            print("="*60)
```
BCathcart: What's the reason for having both logging and prints?

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+            self.server.restart(remove_rdb=False)
+            
+            # Reconnect and wait for server to be ready
+            client = self.server.get_new_client()
+            self.client = client
+            
+            # Wait for server to be responsive after restart
+            print("Waiting for server to be ready...")
+            logging.info("Waiting for server to be responsive...")
+            max_wait = 60
```
BCathcart: I wouldn't add a limit. A large dataset will take longer than this to restore. We should have a separate github job for this rather than having it as part of the regular integration test run, so no need to worry about being too slow.

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+                    time.sleep(1)
+                    waited += 1
+            else:
+                raise TimeoutError("Server did not become responsive after restart")
+            
+            # Wait for re-indexing to complete after restore
+            print("Waiting for re-indexing to complete after restore...")
+            logging.info("Waiting for re-indexing to complete...")
+            IndexingTestHelper.wait_for_backfill_complete_on_node(client, index_name)
+            logging.info("Re-indexing completed")
```
BCathcart: I believe with the new save/restore, the index will be rebuilt within the loading period and there's no follow-up backfill. Right now the default save/restore scheme may be the old one. @boda26 was going to make the new one the default, so I'd check if that's done now. Otherwise you can config set it to the new version specifically in this test.

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+            try:
+                ft_info_after = client.execute_command("FT.INFO", index_name)
+                logging.info(f"FT.INFO after RESTORE: {ft_info_after}")
+                # Log key metrics
+                for i, item in enumerate(ft_info_after):
+                    if item in [b"indexing", b"num_docs", b"hash_indexing_failures", b"total_indexing_time"]:
+                        logging.info(f"  {item}: {ft_info_after[i+1]}")
+                        print(f"  {item}: {ft_info_after[i+1]}")
+            except Exception as e:
+                logging.warning(f"Could not get FT.INFO after restore: {e}")
```
BCathcart: I wouldn't catch the exception. I think it's worth failing the test if we can't fetch basic info

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+            
+            # Check total keys in database
+            try:
+                dbsize = client.dbsize()
+                print(f"  Total keys in database: {dbsize}")
+                logging.info(f"Total keys in database after restore: {dbsize}")
+            except Exception as e:
+                logging.warning(f"Could not get DBSIZE: {e}")
```
BCathcart: I also wouldn't catch this

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+                for i, item in enumerate(ft_info_after):
+                    if item in [b"indexing", b"num_docs", b"hash_indexing_failures", b"total_indexing_time"]:
+                        logging.info(f"  {item}: {ft_info_after[i+1]}")
+                        print(f"  {item}: {ft_info_after[i+1]}")
+            except Exception as e:
+                logging.warning(f"Could not get FT.INFO after restore: {e}")
+            
+            # Check total keys in database
+            try:
+                dbsize = client.dbsize()
```
BCathcart: This doesn't seem useful to collect unless we assert it's the same as before the save

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+        elapsed = time.time() - start_time
+        logging.error(f"Write verification FAILED after {elapsed:.1f}s and {attempt} attempts")
+        return False
+    
+    def _get_index_doc_count(self, client: Valkey, index_name: str) -> int:
+        """Get the number of documents in the index"""
+        info_result = client.execute_command("FT.INFO", index_name)
+        
+        for i, item in enumerate(info_result):
+            if item == b"num_docs" or item == "num_docs":
```
BCathcart: I expect we might have a utility function that converts and FT.INFO result into a dictionary for easy lookup. If not, we should add one although it doesn't necessarily have to be part of this PR

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+                return int(info_result[i + 1])
+        
+        return 0
```
BCathcart: I'd assert rather than return 0. We don't know how many docs there are.

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,512 @@
...
+            # Step 6: Verify data integrity Verify index exists and has documents (may be re-indexing)
+            print("\nVerifying data integrity...")
+            restored_docs = self._get_index_doc_count(client, index_name)
+            print(f"  Index restored with {restored_docs} documents (initially)")
+            logging.info(f"Index restored with {restored_docs} documents (expected: {doc_count_before_save})")
+            
+            if restored_docs != doc_count_before_save:
+                pct_indexed = (restored_docs / doc_count_before_save * 100) if doc_count_before_save > 0 else 0
+                logging.warning(f"Re-indexing in progress: {restored_docs:,}/{doc_count_before_save:,} documents ({pct_indexed:.1f}%)")
+                print(f"Re-indexing in progress: {pct_indexed:.1f}% complete")
```
BCathcart: We should fail if the number of docs is different before and after. Did you check with @boda26 that the discrepancy you're seeing has now been resolved by his fix?

boda26: Yes, I just tested, the discrepancy is resolved

integration/test_endurance_save_restore.py
Nivesh-01: Done

integration/test_endurance_save_restore.py
Nivesh-01: Done

integration/test_endurance_save_restore.py
Nivesh-01: In case we want to check later if we can do new writes. But in hindsight if we need it later we can always add it in then so removing it now

integration/test_endurance_save_restore.py
Nivesh-01: The RDB version is default to version 2 for integration test and default to version 1 for unittests, so we are already testing with V2

integration/test_endurance_save_restore.py
Nivesh-01: Removed logging. I was trying to get logs while the test was running. With --capture flag we only need prints so removing logs.

integration/test_endurance_save_restore.py
Nivesh-01: changed the max_wait to 8 hours. Did not remove it completely because while true seems very dangerous

integration/test_endurance_save_restore.py
Nivesh-01: Removed backfill wait

integration/test_endurance_save_restore.py
Nivesh-01: done

integration/test_endurance_save_restore.py
Nivesh-01: done

integration/test_endurance_save_restore.py
Nivesh-01: adding in the assert. I initially added in for the logs but an assert before and after the save would also be good to test

integration/test_endurance_save_restore.py
Nivesh-01: Shifting to utility function.

integration/test_endurance_save_restore.py
Nivesh-01: Done

integration/test_endurance_save_restore.py
Nivesh-01: Done

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,351 @@
...
+from valkey import ResponseError, OutOfMemoryError
+from valkey.client import Valkey
+
+from valkey_search_test_case import ValkeySearchTestCaseBase
+from valkeytestframework.conftest import resource_port_tracker
+from utils import IndexingTestHelper
+from ft_info_parser import FTInfoParser
+
+
+class TestEnduranceSaveRestore(ValkeySearchTestCaseBase):
```
BCathcart, line 21: Let's skip this test for now by default since it'll be too expensive to run on every PR. For 1.2, let's evaluate it locally using a cgroup to limit the memory and make sure we don't go into swap. Later I think we should run it daily in a separate github job that can include the more expensive regression tests like this one and the one being introduced in #831.

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,351 @@
...
+        self.server.restart(remove_rdb=False)
+        
+        # Reconnect and wait for server to be ready
+        client = self.server.get_new_client()
+        self.client = client
+        
+        # Wait for server to be responsive after restart
+        # Large datasets can take significant time to restore, using generous timeout
+        print("Waiting for server to be ready...")
+        max_wait = 28800  # 8 hours;
```
BCathcart, line 273: There might be a timeout on integration test runtimes at a test infrastructure much shorter than this. Something to keep in mind if you see it time out

BCathcart (approved):

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,351 @@
...
+        
+        # Step 6: Check index state after restore
+        print("\nChecking index state after restore...")
+        
+        # Get detailed FT.INFO after restore
+        ft_info_after = client.execute_command("FT.INFO", index_name)
+        parser_after = FTInfoParser(ft_info_after)
+        print(f"FT.INFO after RESTORE:")
+        print(f"  num_docs: {parser_after.num_docs}")
+        print(f"  hash_indexing_failures: {parser_after.hash_indexing_failures}")
```
KarthikSubbarao, line 296: Can you copy and paste these results?

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,351 @@
...
+        
+        # Wait for server to be responsive after restart
+        # Large datasets can take significant time to restore, using generous timeout
+        print("Waiting for server to be ready...")
+        max_wait = 28800  # 8 hours;
+        waited = 0
+        while waited < max_wait:
+            try:
+                client.ping()
+                print(f"  Server is responsive after {waited}s")
```
KarthikSubbarao, line 278: Is this command rejected during the RDB Load? (some commands are permitted even in the middle of the load)

integration/test_endurance_save_restore.py
```diff
@@ -0,0 +1,351 @@
...
+                break
+            except Exception as e:
+                time.sleep(1)
```
KarthikSubbarao, line 281: Can you add a print here? I just wanted to check the exception message

integration/test_endurance_save_restore.py
Nivesh-01: done

integration/test_endurance_save_restore.py
Nivesh-01: currently for the amount of data being ingested this should not be an issue but I will keep it in mind when I manually test for higher memory

integration/test_endurance_save_restore.py
Nivesh-01:
```
Restarting server...
Monitoring restore progress from T+0s...
[T+79s] Restore status: in_progress=0, indexes=N/A, current_index_progress=0.0%
  Server is responsive and restore complete after 79s

============================================================
INFO MEMORY - AFTER RESTORE
============================================================
  used_memory: 5,658,959,624 bytes (5.27G)
  used_memory_rss: 5,328,584,704 bytes (4.96G)
  used_memory_peak: 5,658,959,624 bytes (5.27G)
  maxmemory: 5,368,709,120 bytes (5.00G)
  mem_fragmentation_ratio: 0.94
  allocator_frag_ratio: 1.0
  allocator_frag_bytes: 6,090,384 bytes (5.81 MB)
============================================================


Checking index state after restore...
FT.INFO after RESTORE:
  num_docs: 7340033
  hash_indexing_failures: 0
  Total keys in database after restore: 7340033
```

integration/test_endurance_save_restore.py
Nivesh-01: Yes this command is rejected during the RDB load.

integration/test_endurance_save_restore.py
Nivesh-01: LOADING Valkey is loading the dataset in memory

PR #785, opened 2026-02-18
Fix unexpected vk-search flag in RDB after dropped all indexs

boda26:
When an index is dropped, MetadataManager still keeps it in the inner map with a higher version as tombstone. This PR fix this by checking the content of each key (index) in the inner map.

### DIFF
--- a/integration/test_rdb_load_without_search_module.py
+++ b/integration/test_rdb_load_without_search_module.py
@@ -13,17 +13,15 @@ index = Index(
     [Numeric("n"), Tag("t")],
 )
 
-def _start_server_without_module(test_case, port, testdir, dbfilename):
+def _start_server_without_module(test_case, testdir, dbfilename):
     """Start a valkey server WITHOUT the valkey-search module, loading an existing RDB."""
     server_path = os.getenv("VALKEY_SERVER_PATH")
 
     # Use create_server from valkeytestframework, customize args to exclude module
     server, client = test_case.create_server(
         testdir=testdir,
-        port=port,
         server_path=server_path,
         args={
-            "enable-debug-command": "yes",
             "appendonly": "no",
             "dir": testdir,
             "dbfilename": dbfilename,
@@ -31,6 +29,11 @@ def _start_server_without_module(test_case, port, testdir, dbfilename):
     )
     logfile = os.path.join(server.cwd, server.args["logfile"])
     return server, client, logfile
+
+def _cleanup_data_files(rdb_path):
+    """Remove RDB files left behind by server.exit(cleanup=False)."""
+    if os.path.exists(rdb_path):
+        os.remove(rdb_path)
 
 def do_rdb_load_without_module_test_cmd(test_case):
     """
@@ -58,9 +61,8 @@ def do_rdb_load_without_module_test_cmd(test_case):
     server.exit(cleanup=False)
 
     # Phase 2: Start a new server WITHOUT valkey-search module
-    port = test_case.get_bind_port()
     new_server, new_client, logfile = _start_server_without_module(
-        test_case, port, testdir, dbfilename
+        test_case, testdir, dbfilename
     )
 
     assert new_client.ping()
@@ -76,6 +78,8 @@ def do_rdb_load_without_module_test_cmd(test_case):
     assert "search" not in module_names, "valkey-search should NOT be loaded"
 
     print("CMD: Server loaded RDB without module successfully, no crash.")
+
+    _cleanup_data_files(rdb_path)
 
 def do_rdb_load_without_module_test_cme(test_case):
     """
@@ -113,9 +117,8 @@ def do_rdb_load_without_module_test_cme(test_case):
     primary0.server.exit(cleanup=False)
 
     # Phase 2: Start first node WITHOUT valkey-search module
-    port = test_case.get_bind_port()
     new_server, new_client, logfile = _start_server_without_module(
-        test_case, port, testdir, dbfilename
+        test_case, testdir, dbfilename
     )
 
     assert new_client.ping()
@@ -131,6 +134,8 @@ def do_rdb_load_without_module_test_cme(test_case):
     assert "search" not in module_names
 
     print("CME: Node loaded RDB without module successfully, no crash.")
+
+    _cleanup_data_files(rdb_path)
 
 class TestRDBLoadWithoutModuleCMD(ValkeySearchTestCaseBase):
     """CMD mode: Verify server doesn't crash loading RDB from a module-enabled server."""


src/coordinator/metadata_manager.cc
```diff
@@ -621,8 +621,10 @@ bool DoesGlobalMetadataContainEntry(const GlobalMetadata &metadata) {
     return false;
   }
   for (const auto &[type_name, inner_map] : metadata.type_namespace_map()) {
-    if (!inner_map.entries().empty()) {
-      return true;
+    for (const auto &[id, entry] : inner_map.entries()) {
```
KarthikSubbarao, line 626: Can we add an integration test for this in both CME and CMD?

src/coordinator/metadata_manager.cc
```diff
@@ -621,8 +621,10 @@ bool DoesGlobalMetadataContainEntry(const GlobalMetadata &metadata) {
...
+      if (entry.has_content()) {
```
KarthikSubbarao, line 627: Is content guaranteed to be deleted immediately (sync) when we drop indexes? If not, we are going to end up in a timing related problem where a drop can index and without waiting use save and will see this storing the aux metadata.

Now that we are fixing it, it would be good to understand how reliable this is and when does it actually get deleted

boda26: I checked the code and I think yes. The DROPINDEX flow is FTDropIndexCmd -> SchemaManager::RemoveIndexSchema -> MetadataManager::DeleteEntry, and in DeleteEntry 
```
  (*(*metadata.mutable_type_namespace_map())[type_name]
        .mutable_entries())[encoded_id] = new_entry;
```
It creates a new value(new_entry) on the key(encoded_id), which just fully replaces the old entry. Then it just sets the top level version and fingerprint. Since it does not set content, the has_content will always return false. The whole flow runs on main thread, so when DoesGlobalMetadataContainEntry is called, the index should be dropped locally and has_content should return false

src/coordinator/metadata_manager.cc
KarthikSubbarao, line 626: Let's explain why we are checking this

integration/test_rdb_load_without_search_module.py
```diff
@@ -0,0 +1,212 @@
...
+    TEST_MAX_WAIT_TIME_SECONDS,
+)
+from indexes import Index, Numeric, Tag, Vector
+
+index = Index(
+    "test_rdb_no_module_idx",
+    [Numeric("n"), Tag("t")],
+)
+
+def _start_server_without_module(test_case, port, testdir, dbfilename):
```
KarthikSubbarao: This is a lot of code for something pretty standard. Why can't we use any existing function for this?

KarthikSubbarao: Let's please check the code in the test framework or base class functions and not duplicate test code so much

integration/test_rdb_load_without_search_module.py
```diff
@@ -0,0 +1,212 @@
...
+        f"dir {testdir}",
+        f"dbfilename {dbfilename}",
+        f"port {port}",
+        f"logfile logfile_{port}",
+    ]
+
+    # Only load JSON module if available, but NOT valkey-search
+    json_module = os.getenv("JSON_MODULE_PATH")
+    if json_module:
+        lines.append(f"loadmodule {json_module}")
```
KarthikSubbarao: Not relevant for this test

integration/test_rdb_load_without_search_module.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Grab first primary's RDB info before stopping
+    primary0 = test_case.replication_groups[0].primary
+    testdir = primary0.server.cwd
+    dbfilename = primary0.server.args["dbfilename"]
+    rdb_path = os.path.join(testdir, dbfilename)
+    assert os.path.exists(rdb_path), f"RDB file not found at {rdb_path}"
+
+    # Stop first primary, keep RDB
+    os.environ["SKIPLOGCLEAN"] = "1"
```
KarthikSubbarao: We should not change os envi variables in a test

integration/test_rdb_load_without_search_module.py
```diff
@@ -0,0 +1,212 @@
...
+    primary0.server.exit(cleanup=False)
```
KarthikSubbarao, line 117: Please review all the tests here. We don't want to create resources (e.g. logs, folders, rdb, other files) and not clean them up.

integration/test_rdb_load_without_search_module.py
boda26: I checked valkeytestframework, and there is a function for starting the server. Will reuse this one

integration/test_rdb_load_without_search_module.py
KarthikSubbarao: This comment is not resolved. Can you check this? @boda26

boda26: I already cleared all the changes to os.env variables

KarthikSubbarao: You are still passing cleanup=False here. See what happens as a result of it:
https://github.com/valkey-io/valkey-test-framework/blob/9170c1c09ac3d403a06f1e1585e5a210d29c51f0/src/valkey_test_case.py#L103

integration/test_rdb_load_without_search_module.py
```diff
@@ -0,0 +1,143 @@
...
+)
+
+def _start_server_without_module(test_case, port, testdir, dbfilename):
+    """Start a valkey server WITHOUT the valkey-search module, loading an existing RDB."""
+    server_path = os.getenv("VALKEY_SERVER_PATH")
+
+    # Use create_server from valkeytestframework, customize args to exclude module
+    server, client = test_case.create_server(
+        testdir=testdir,
+        port=port,
```
KarthikSubbarao: Question: Why do we need to specify a port?

Can you check if this is needed? (Does it not find and use one automatically?)

https://github.com/valkey-io/valkey-test-framework/blob/9170c1c09ac3d403a06f1e1585e5a210d29c51f0/tests/test_specific_setup_per_test.py#L16

integration/test_rdb_load_without_search_module.py
```diff
@@ -0,0 +1,143 @@
...
+        server_path=server_path,
+        args={
+            "enable-debug-command": "yes",
+            "appendonly": "no",
```
KarthikSubbarao: Are these needed?

integration/test_rdb_load_without_search_module.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Phase 1: Create and drop index, then save
+    index.create(client, wait_for_backfill=True)
+    index.load_data(client, 10)
+    index.drop(client)
+    client.execute_command("SAVE")
+
+    # Grab the RDB location info before stopping
+    testdir = server.cwd
+    dbfilename = server.args["dbfilename"]
```
KarthikSubbarao, line 56: Question: What is this value going to be? Is it is a default / fixed value?

if yes, why do we need to check and provide this as a startup config?


If no and each server has its own db name, I guess it makes sense to know what file to look for

integration/test_rdb_load_without_search_module.py
boda26: "appendonly" is required, otherwise server will load AOF instead of RDB.
"enable-debug-command" can be removed

integration/test_rdb_load_without_search_module.py
boda26: server start with appendonly=no, which will only produce RDB files. I added a cleanup at the end to delete all rdb files

integration/test_rdb_load_without_search_module.py
boda26: Port is not needed, removed port from here

integration/test_rdb_load_without_search_module.py
boda26: No, it is not fixed. In ValkeyServerHandle.__init__, 
self.args["dbfilename"] = f"testrdb-{port}.rdb"
So the RDB filename would be different because of different port number, so we should explicitly grab and pass the dbfilename

PR #791, opened 2026-02-19
update README.md

KarthikSubbarao:
update README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,13 @@
 # valkey-search
 
-**valkey-search** (BSD-3-Clause), provided as a Valkey module, is a high-performance Vector Similarity Search engine optimized for AI-driven workloads. It delivers single-digit millisecond latency and high QPS, capable of handling billions of vectors with over 99% recall.
+**Valkey-Search** (BSD-3-Clause), provided as a Valkey module, is a high-performance Search engine
+optimized for AI-driven / Search / Analytics / Recommendation System related workloads. It delivers single-digit millisecond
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,13 @@
 # valkey-search
 
-**valkey-search** (BSD-3-Clause), provided as a Valkey module, is a high-performance Vector Similarity Search engine optimized for AI-driven workloads. It delivers single-digit millisecond latency and high QPS, capable of handling billions of vectors with over 99% recall.
+**Valkey-Search** (BSD-3-Clause), provided as a Valkey module, is a high-performance Search engine
+optimized for AI-driven / Search / Analytics / Recommendation System related workloads. It delivers single-digit millisecond
```
rlunar, line 4: Nitpick _Optimized_

README.md
```diff
@@ -1,10 +1,13 @@
...
+latency and high QPS, capable of handling billions of vectors with over 99% recall as part of vector searches. It also provides
```
rlunar, line 5: What does high QPS means? Do we have a metric like 10K QPS per node on _X_ GB of data?

README.md
KarthikSubbarao: It is continuing on the same sentence from above

README.md
KarthikSubbarao: We do have some data on this for non vector searches (we have mainly been benchmarking fulltext and some combinational queries of numeric / tag). We should be able to have a clearer statement on this soon (ie - we will take our perf tests and those scenarios and convert that into a summary explaining when we can expect X QPS)

PR #793, opened 2026-02-20
Fix mutation sequence number timing bug

BCathcart:
Follow up to #774. Simply providing thread-safe access isn't enough. We want to get the sequence number of the key found in the search within the same read phase, or else a write that deletes the key has the opportunity to remove that key from `index_key_info_` before we've looked up the sequence number. Maybe we could drop the key from the results at this point since we know it's changed value form when the background search ran, but I'm not certain this won't break other assumptions.


src/index_schema.h
```diff
@@ -36,6 +36,7 @@
 #include "src/utils/string_interning.h"
 #include "vmsdk/src/blocked_client.h"
 #include "vmsdk/src/command_parser.h"
+#include "vmsdk/src/log.h"
```
Nivesh-01: Maybe we can remove this if we are not logging anything

src/query/search.cc
```diff
@@ -765,16 +764,26 @@ SerializationRange SearchResult::GetSerializationRange(
...
+        n.sequence_number =
+            parameters.index_schema->GetIndexMutationSequenceNumber(
+                n.external_id);
+      }
+    }
+  }
+
   VMSDK_ASSIGN_OR_RETURN(
-      auto result,
-      MaybeAddIndexedContent(DoSearch(parameters, search_mode), parameters));
```
KarthikSubbarao, line 771: Ideally, we could generate the sequence numbers inside `DoSearch`. I think that already has a time slice mutex and that is the fn which returns the vector of neighbors

src/query/search.cc
```diff
@@ -765,16 +764,26 @@ SerializationRange SearchResult::GetSerializationRange(
...
+      auto result, MaybeAddIndexedContent(std::move(neighbors), parameters));
   size_t total_count = result.size();
   parameters.search_result =
       SearchResult(total_count, std::move(result), parameters);
-  for (auto &n : parameters.search_result.neighbors) {
-    n.sequence_number =
-        parameters.index_schema->GetIndexMutationSequenceNumber(n.external_id);
-  }
```
KarthikSubbarao, line 778: Generating the sequence number after the SearchResult is created was helpful because the creation of `SearchResult` itself trims and removes keys (neighbors) which are not essential (based on LIMIT and Offset) to the response.

So, if we move this above (before SearchResult is constructed), we will spend more time checking and assigning the sequence number even though some of those keys will be dropped

src/query/search.cc
Nivesh-01: We are adding this lock over here to mitigate for one of the errors. An example of which would be like while we are doing a search a writer thread modifies one of the key (like deleting it) that is a part of the search. This is causing a race condition leading to itr != index_key_info_.end() Key not found error. This way of locking I believe should mitigate that

KarthikSubbarao: Yes, I understand the lock which is being added here and the reason. What I said in my comment above is that `DoSearch` already was acquiring lock on `vmsdk::ReaderMutexLock lock(&time_sliced_mutex);`, so the sequence numbers for the keys could be assigned in `DoSearch` function itself if that is a common function used across the code

KarthikSubbarao: Just checked, I think this if fine now. `DoSearch` is only used from `Search`. So this should be good.

But if we are doing this, we might want to consider having `SearchResult` result from `DoSearch`.

Otherwise, we will have the issue mentioned here: https://github.com/valkey-io/valkey-search/pull/793/changes#r2830966997

src/query/search.cc
allenss-amazon: I agree that it's better if we move the capture of the sequence numbers to after where the SearchResult is generated. Right after the pre-trimming logic seems like a good place.

BCathcart: How many neighbours do you think would typically be dropped in workloads that use these features? And what's our upper limit on the number of neighbours returned in a result in the first place? I'm guessing on average extending the read phase locking scope across the call to `MaybeAddIndexedContent` and `SearchResult` will have a bigger performance impact than fetching some unnecessary sequence numbers

KarthikSubbarao: > And what's our upper limit on the number of neighbours returned in a result in the first place? 

There is no limit right now for non-vector queries. This is an area I am looking into.

> I'm guessing on average extending the read phase locking scope across the call to MaybeAddIndexedContent and SearchResult will have a bigger performance impact than fetching some unnecessary sequence numbers

I am looking at your recent change. if it is possible, we can  just create this before calling `PopulateIndexMutationSequenceNumbers`. 
```
  parameters.search_result =
      SearchResult(total_count, std::move(result), parameters);
```

Then, `PopulateIndexMutationSequenceNumbers` can be provided with the neighbors from `parameters.search_result`

BCathcart: `PopulateIndexMutationSequenceNumbers` needs to happen within the read phase of the time slice mutex, so that means we're extending the read phase across all of this and likely starving writes longer. That's what I mean by I'm not sure doing this is a clear performance improvement. I think it behaves differently in different situations.

BCathcart: Turns out trimming is almost always used. Extending the read lock scope only really matters for the last query in the read phase and as long as there are a bunch of queries in the read phase it'll be proportionally insignificant. I'll make the change.

KarthikSubbarao (approved):

PR #794, opened 2026-02-20
FlatPosiitonMap::ReadCounts Optimization

Aksha1812:
Optimize ReadCounts by replacing variable-length memcpy with loop-based byte reading to reduce hot path overhead during search


Aksha1812:
CC: @VoletiRam

src/indexes/text/flat_position_map.cc
```diff
@@ -15,6 +15,19 @@
 #include "absl/log/check.h"
 #include "src/indexes/text/posting.h"
 
+namespace {
+
+// Read fixed-length unsigned int (little-endian) helper
+inline uint32_t ReadVarUint(const char* ptr, uint8_t num_bytes) {
```
KarthikSubbarao, line 21: Is it possible to capture a before and after perf test results (pick any scenario where we spend good time on ContainsFields)? Also, capturing the flamegraph for both will help

Alternatively, using micro benchmarking (BENCHMARK) fns help to easy test a single function without the whole e2e set up

Aksha1812: Seeing better performance by micro benchmarking
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_ReadCounts_Memcpy/0/0       6.53 ns         6.53 ns    108034946
BM_ReadCounts_Memcpy/1/0       6.82 ns         6.82 ns    103727962
BM_ReadCounts_Memcpy/1/1       6.83 ns         6.83 ns    102336774
BM_ReadCounts_Memcpy/3/0       6.83 ns         6.83 ns    103195278
BM_ReadCounts_Memcpy/3/3       1.55 ns         1.55 ns    444884977
BM_ReadCounts_Loop/0/0         3.02 ns         3.02 ns    233621996
BM_ReadCounts_Loop/1/0         3.04 ns         3.04 ns    229533826
BM_ReadCounts_Loop/1/1         3.49 ns         3.49 ns    202523435
BM_ReadCounts_Loop/3/0         3.52 ns         3.52 ns    198434506
BM_ReadCounts_Loop/3/3         5.22 ns         5.22 ns    134008009
```

Also seeing the similar proportional decrease in ReadCounts CPU time when i compare flamegraphs.

This could be optimized further since while reading num positions, and num partitions, the number of bytes are between 1 to 4 so maybe we can do do a single operation for each switch case 0 to 4 , and avoid loop completely. 

something like 
```
```cpp
inline uint32_t ReadVarUint_Switch(const char* ptr, uint8_t num_bytes) {
  switch (num_bytes) {
    case 1: return U8(ptr[0]);
    case 2: return U8(ptr[0]) | (U32(U8(ptr[1])) << 8);
    case 3: return U8(ptr[0]) | (U32(U8(ptr[1])) << 8) | (U32(U8(ptr[2])) << 16);
    case 4: return U8(ptr[0]) | (U32(U8(ptr[1])) << 8) | (U32(U8(ptr[2])) << 16) |
                   (U32(U8(ptr[3])) << 24);
    default: return 0;
  }
}
```

Stylistically this does not look good and conventional but micro benchmark showed 
```
BM_ReadCounts_Switch/0/0       1.09 ns         1.09 ns    631450185
BM_ReadCounts_Switch/1/0       1.16 ns         1.16 ns    603891178
BM_ReadCounts_Switch/1/1       1.46 ns         1.46 ns    479396839
BM_ReadCounts_Switch/3/0       1.36 ns         1.36 ns    506205629
BM_ReadCounts_Switch/3/3       1.70 ns         1.70 ns    409513959

Aksha1812: memcpy only works better when we have to read 4 bytes at a time. But it is very rare that pos_bytes and part bytes are upto 4 bytes ever. it is a very extreme case. Same goes with reading the partiitons while going through the partition map. These are the two places where we read variable length integers in the position map. Encoding and Decoding Positions and Field Mask is separate since it requires special encoding/ decoding. @KarthikSubbarao i have updated the code . Now it would handled most cases better than before except very very large position maps.

allenss-amazon: memcpy has a high startup overhead because it first needs to check the input arguments for overlap. In the overlap case sometimes the block move has to happen in the reverse direction. On some CPU architectures it will then check the alignment of the source & dest in order to enable fast operations like movement of whole words, etc.

This particular trick of using a fall through switch is statement is well-known exception to the usual formatting and and style guidelines.

PR #799, opened 2026-02-20
Address unsolved text path in PreFilter Evaluators

KarthikSubbarao:
Address unsolved text path in PreFilter Evaluators. There was an unsolved path (since we are trying to optimize text queries) in the Entries Fetcher where an AND contains an OR which has some other non text children and an AND child.

Here the AND child of the OR is not solved, so the prefilter evaluator needs to handle this.

```
    # Test OR with mixed predicates on new documents: AND(N1 T1 OR(N2 | T2 | AND(T3 T4))) where N2 and T2 don't match
    "@price:[40 60] alpha (@price:[100 200] | beta | (gamma delta))"
```

I added a test for this to validate it is handled by the prefilter evaluator (but updating the optimization code)

### DIFF
--- a/src/query/predicate.cc
+++ b/src/query/predicate.cc
@@ -441,8 +441,8 @@ EvaluationResult ComposedPredicate::EvaluateWithContext(Evaluator &evaluator,
       // AND is fully (recursively) resolved in the entries fetcher layer
       // already. The only cases where this is not true are:
       // 1) when an AND predicate contains an OR which has some other non text
-      // children and an AND child containing text This is not solved in entries
-      // fetcher yet.
+      // children and an AND child containing text. This is not solved in
+      // entries fetcher yet.
       // 2) when the query has negation on text. Currently, a universal set is
       // used in the entries fetcher layer for text+negate queries.
       if (evaluator.IsPrefilterEvaluator() &&


KarthikSubbarao:
@allenss-amazon , this morning I was thinking through what cases are not currently optimized (ie - not fully solved for Text in the Entries Fetcher), and there are only two such cases.
```
      // 1) when an AND predicate contains an OR which has some other non text
      // children and an AND child containing text. This is not solved in
      // entries fetcher yet.
      // 2) when the query has negation on text. Currently, a universal set is
      // used in the entries fetcher layer for text+negate queries.
```

I wanted to make sure because they are not solved in Entries Fetcher, the optimization logic in the Prefilter Evaluator should not skip text evaluation in these portions of the query in scenario 1. (We already test and handle scenario 2 well since it gets fully evaluated in Prefilter Evaluator)

After adding a more complex case of scenario 1, I saw we were matching more results than what we should due to prefilter evaluator optimization. So, I added a fix for it.

I think the compatibility tests generated by Miles is not creating very deeply tested queries. But even if it does, we should not just look at what documents match (this aspect was already correct). We should also look at what documents should not match (this needed a fix, which is what my PR covers).

PR #803, opened 2026-02-23
Remove Unused Variables And Add `UnusedVar` Helper Template

eifrah-aws:
Introduce a generic UnusedVar template function in utils.h to explicitly suppress unused variable warnings.

Remove several unused local variables across the codebase that are no longer referenced. Wrap `EXPECT_DEATH` return values in `UnusedVar` to suppress warnings about ignoring the return value which is marked with `nodiscard`.

**Generated by CodeLite**


src/index_schema.cc
```diff
@@ -1474,6 +1474,7 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
       VMSDK_ASSIGN_OR_RETURN(auto keyname_str, input.LoadString());
       VMSDK_ASSIGN_OR_RETURN(auto from_backfill, input.LoadObject<bool>());
       VMSDK_ASSIGN_OR_RETURN(auto from_multi, input.LoadObject<bool>());
+      vmsdk::UnusedVar(from_multi);
```
eifrah-aws: Note that I kept the call here, since I assumed that we do want to make sure that the call to `input.LoadObject<bool>()` succeeds.

testing/ft_aggregate_exec_test.cc
```diff
@@ -31,6 +31,7 @@ struct FakeIndexInterface : public IndexInterface {
       absl::string_view alias) const override {
     std::cout << "Fake get identifier for " << alias << "\n";
     VMSDK_ASSIGN_OR_RETURN(auto type, GetFieldType(alias));
+    vmsdk::UnusedVar(type);
```
eifrah-aws: Similar to above: I kept the call here, since I assumed that we do want to make sure that the call to `GetFieldType(alias)` succeeds.

testing/ft_aggregate_parser_test.cc
```diff
@@ -36,6 +36,7 @@ struct FakeIndexInterface : public IndexInterface {
       absl::string_view alias) const override {
     std::cout << "Fake get identifier for " << alias << "\n";
     VMSDK_ASSIGN_OR_RETURN(auto type, GetFieldType(alias));
+    vmsdk::UnusedVar(type);
```
eifrah-aws: Ditto.

Issue #761, opened 2026-02-15
[BUG] FT.SEARCH/FT.AGGREGATE can deadlock in cluster mode when executed from LUA/multi/exec


allenss-amazon:
**Describe the bug**

If two or more nodes in a cluster attempt a query command from a synchronous context (LUA/multi/exec), then it's possible to deadlock those nodes.

The failure happens because each makes a cross-shard RPC request. Upon receiving the RPC the mainthread must be used to access the database to provide current values of located keys. However, the mainthread can't be accessed because the LUA/multi/exec is sitting on the mainthread waiting for the result.

**To Reproduce**

As described above.

**Expected behavior**

Potential Solutions:

1. Disallow query options in CME form a LUA/multi context.
2. Enable the RPC "wait loop" to service remote RPC requests. The downside of this approach is that it exposes the state of the database in the middle of what's typically considered to be an atomic operation.

**Environment**

Additional information to reproduce the failure

- Valkey Search Module Version
- Valkey Core Version 

**Additional context**

PR #806, opened 2026-02-23
Add multi/lua integration tests with search commands

### LINKED ISSUE #718: [TEST] Need to write a test to ensure that multi/exec/LUA scripts and search are well behaved.
** Item not tested **

Need a test for the interactions between multi/exec/lua and all search operations.

** Testing Method **

FT._LIST, FT.SEARCH LOCALONLY and FT.INFO LOCAL need to be tested in a multi/exec and a lua context. Both CMD and CME. These commands should execute just fine, i.e., no error and give meaningful results.

The other search commands: FT.SEARCH (not localonly), FT.AGGREGATE, FT.CREATE, FT.DROPINDEX and FT.INFO PRIMARY / LOCAL. Should be fine in CMD but fail in CME. This needs to be tested in both multi/LUA context.

Ingestion needs to be tested, i.e., we need to ingestion keys into an index within a multi/LUA context.

Consistency needs to be tested also, i.e., an ingestion followed by a query must include the newly ingested data in the query (assuming it satisfies the query). Again must be tested both multi/LUA contexts.

baswanth09:
Add multi/lua integration tests for CMD and CME modes. 
Resolves: #718 
The potential deadlock scenario described in #761 is not possible as we skip fanout if the command is in multi/lua context.
https://github.com/valkey-io/valkey-search/blob/5fe1676e8ad03345e6d9818744097fbef4a10005/src/commands/commands.cc#L134-L138

Added an error message to reject the multi/lua queries in CME mode.

### DIFF
--- a/integration/test_multi_lua.py
+++ b/integration/test_multi_lua.py
@@ -17,10 +17,7 @@ from indexes import *
 INDEX = "idx"
 OK = b"OK"
 QUEUED = b"QUEUED"
-
-
-def _create_index(client: Valkey, index: str = INDEX):
-    assert client.execute_command("FT.CREATE", index, "SCHEMA", "price", "NUMERIC", "title", "TEXT") == OK
+FANOUT_NOT_SUPPORTED_ERR = "MULTI/EXEC or Lua script are not supported in CME mode"
 
 
 def _lua_call(cmd: str, *args: str) -> str:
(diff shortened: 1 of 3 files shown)


integration/test_multi_lua.py
```diff
@@ -1,79 +1,376 @@
...
+
+# ---------------------------------------------------------------------------
+# CMD (standalone) tests
+# ---------------------------------------------------------------------------
+
+class TestMultiLua(ValkeySearchTestCaseBase):
 
     def test_multi_exec_case1(self):
         """
         Test that HSET done outside a MULTI/EXEC on a key that was modified
```
allenss-amazon: This test doesn't accomplish what the text says it does. Without a second client nothing useful is being tested. Let's either nuke the test or fix it.

integration/test_multi_lua.py
```diff
@@ -1,79 +1,376 @@
...
         # for the cpp_book, followed by a search query.
-        assert client.execute_command("MULTI") == self.OK
-        assert client.hset("cpp_book", "price", 65) == self.QUEUED
-        assert client.hset("rust_book", "price", 50) == self.QUEUED
+        assert client.execute_command("MULTI") == OK
+        assert client.hset("cpp_book", "price", 65) == QUEUED
+        assert client.hset("rust_book", "price", 50) == QUEUED
         client.execute_command("EXEC")
         # This call should not be blocked.
         assert client.hset("cpp_book", "price", 70) == 0
```
allenss-amazon: Again, this isn't testing anything interesting.

integration/test_multi_lua.py
```diff
@@ -1,79 +1,376 @@
...
+        _create_books_price_index(client)
+        assert client.execute_command("MULTI") == OK
+        assert client.execute_command("FT.INFO", INDEX, *extra_args) == QUEUED
+        results = client.execute_command("EXEC")
+        assert results is not None
+        info = FTInfoParser(results[0])
+        assert info.index_name == INDEX
+
+    @pytest.mark.parametrize("extra_args", [[], ["LOCALONLY"]], ids=["search", "search localonly"])
+    def test_multi_exec_ft_search(self, extra_args):
```
KarthikSubbarao: Can we validate text search?

allenss-amazon (commented):
Let's reduce the number of distinct "tests" by combining these into a reduced number of tests (maybe 1 each for CMD and CME). There is high overhead for starting up a cluster which would happen for each of these trivial tests and thus this would add substantial wall-time to this test file as written.

integration/test_multi_lua.py
baswanth09: Updated in the revision

integration/test_multi_lua.py
baswanth09: These are existing tests. Removed them in the revision

allenss-amazon (approved):

Issue #792, opened 2026-02-19
[BUG] Memory Leak Triage


Aksha1812:
**Describe the bug**
Multiple tests showing memory leaks while running integration asan tests

**To Reproduce**
Vibe coded script to generate summary

you can run below like `./run_asan_tests.sh`  OR `./run_asan_tests.sh test_fulltext` , if you want to run specific tests. You would need to have everything setup before this this script just runs the tests and creates summary. also you might need to make small tweaks based on your setup. i just created this temporarily to generate summary of all leaks.  

run_asan_tests.sh
```
#!/bin/bash

# ASAN Integration Test Runner and Analyzer
# This script runs integration tests with ASAN, captures output, and generates a summary

# Don't exit on error - we want to capture test output even if tests fail
set +e

# Configuration
TIMESTAMP=$(date +%Y%m%d_%H%M%S)
OUTPUT_DIR="asan_results"
RAW_OUTPUT="${OUTPUT_DIR}/asan_raw_output_${TIMESTAMP}.txt"
SUMMARY_FILE="${OUTPUT_DIR}/asan_summary_${TIMESTAMP}.md"
TEST_FILTER="${1:-}"  # Optional test filter, runs all by default

# Colors for terminal output
RED='\033[0;31m'
GREEN='\033[0;32m'
YELLOW='\033[1;33m'
NC='\033[0m' # No Color

echo -e "${GREEN}========================================${NC}"
echo -e "${GREEN}ASAN Integration Test Runner${NC}"
echo -e "${GREEN}========================================${NC}"
echo ""

# Create output directory
mkdir -p "${OUTPUT_DIR}"

if [ -n "${TEST_FILTER}" ]; then
    echo -e "${YELLOW}Running ASAN tests for: ${TEST_FILTER}${NC}"
else
    echo -e "${YELLOW}Running all ASAN tests${NC}"
fi
echo "Output will be saved to: ${RAW_OUTPUT}"
echo ""

# Run the integration tests with ASAN
echo -e "${YELLOW}Starting tests...${NC}"

# Set up environment for ASAN integration tests
unset VALKEY_SERVER_PATH
unset VALKEY_CLI_PATH
export INTEG_RETRIES=1
export SAN_BUILD=address

# Set TEST_PATTERN if a filter was provided
if [ -n "${TEST_FILTER}" ]; then
    export TEST_PATTERN="${TEST_FILTER}"
fi

# Run tests and capture output (both stdout and stderr)
[...]

PR #811, opened 2026-02-24
Drop indexes on test teardown

BCathcart:
Addresses a chunk of the memory leaks being reported in #792 

> I’m not seeing any memory leaks in a simple test with a single ingestion and search if I drop the index at the end. Ownership of the shared pointers (InvasivePtr) containing the postings object is passed as an opaque pointer to the Rax tree. They’re called out as leaks by ASAN at process exit if ownership hasn’t been passed back to the C++ construct. So luckily there doesn’t seem to be an actual leak. We need to instrument all our integration tests to drop the indexes in teardown unless anyone has a better strategy to suppress this

### DIFF
--- a/integration/valkey_search_test_case.py
+++ b/integration/valkey_search_test_case.py
@@ -259,7 +259,6 @@ class ValkeySearchTestCaseBase(ValkeySearchTestCaseCommon):
         yield
 
         # Cleanup
-        self.client.execute_command("FLUSHALL", "SYNC") # Prevent ASAN false reporting
         ReplicationGroup.cleanup(self.rg)
 
     def get_config_file_lines(self, testdir, port) -> List[str]:
@@ -472,8 +471,6 @@ class ValkeySearchClusterTestCase(ValkeySearchTestCaseCommon):
         yield
 
         # Cleanup
-        for rg in self.replication_groups:
-            rg.primary.client.execute_command("FLUSHALL", "SYNC") # Prevent ASAN false reporting
         for rg in self.replication_groups:
             ReplicationGroup.cleanup(rg)
 


integration/valkey_search_test_case.py
```diff
@@ -148,6 +148,18 @@ def cleanup(rg):
...
+        """Drop all indexes to ensure proper cleanup before server shutdown.
+        
+        This helps avoid false positive memory leak reports from ASAN when
+        InvasivePtr objects stored in Rax nodes aren't explicitly freed.
+        """
+        indexes = client.execute_command("FT._LIST")
+        for index_name in indexes:
+            if isinstance(index_name, bytes):
+                index_name = index_name.decode('utf-8')
+            client.execute_command("FT.DROPINDEX", index_name)
```
KarthikSubbarao: This works for now and an alternative is to flushall


But, we can also update the module code to clean up all the indexes from the shutdown server event callback. 



See: https://github.com/valkey-io/valkey-search/issues/812

And See also: https://github.com/valkey-io/valkey-search/issues/792#issuecomment-3953443967

BCathcart: I switched to FLUSHALL which makes more sense. There are tests that restart the server part way through and ASAN still reports the shutdown leaks there. I gave the shutdown callback a quick try but ran into some issues. It needs a bit more attention to do right. I'd say let's merge this as a step in the right direction and we can follow-up with #812.

BCathcart:
I removed the integration test flush since it needs to be solve in the shutdown callback eventually anyways, which is now being worked on in #883. Let's merge the one small GRPC fix here and continue afterwards in the other PR.

PR #813, opened 2026-02-24
Using "Min Heap" approach for TermIterator

KarthikSubbarao:
Using "Min Heap" approach for TermIterator

This is an optimization on the Term Iterator which is used as the core iterator in text all searches. It performs a K way merge across different Key iterators (for different words) in prefix/suffix/fuzzy/stemmed term search and a K==1 search when exact term searches are used.
The expansion search (e.g. prefix/suffix) operations were showing in perf test runs that it was effected by the linear scan. So, this PR re-writes this using a newly added min heap structure using the inlinedvector as its storage. This min-heap is used when finding the next key and next position. Other APIs (DoneKeys/DonePositions) are optimized too

### DIFF
--- a/src/indexes/text/term.cc
+++ b/src/indexes/text/term.cc
@@ -21,8 +21,12 @@ TermIterator::TermIterator(
       current_field_mask_(0ULL),
       require_positions_(require_positions),
       has_original_(has_original) {
+  // Populate the key_set_ heap.
+  for (size_t i = 0; i < key_iterators_.size(); ++i) {
+    InsertValidKeyIterator(i);
+  }
   // Prime the first key and position if they exist.
-  if (!key_iterators_.empty()) {
+  if (!key_set_.empty()) {
     TermIterator::NextKey();
   }
 }
@@ -61,37 +65,37 @@ void TermIterator::InsertValidKeyIterator(size_t idx) {
 }
 
 bool TermIterator::FindMinimumValidKey() {
-  // 1. If the heap is empty, it means all iterators from the previous match
-  // were extracted. Attempt to re-populate the heap with any iterators
-  // that still have valid data.
-  if (key_set_.empty()) {
-    for (size_t i = 0; i < key_iterators_.size(); ++i) {
-      InsertValidKeyIterator(i);
-    }
-  }
-  // 2. If the heap is still empty after the re-population attempt, it means
-  // all underlying expansion iterators are exhausted.
+  // 1. If the heap is empty, all underlying iterators are exhausted.
+  // Note: This can be due to three cases:
+  //   a) Initialization: No valid keys were found across any
+  //      iterators during the first scan in the constructor.
+  //   b) Natural Exhaustion: All iterators were at some point the 'current_key'
+  //      (active indices) and were popped, advanced, and found to be
+  //      Done/Invalid.
+  //   c) Seek Exhaustion: During SeekForwardKey, laggards were popped and
+  //      skipped, but found to be invalid/done.
   if (key_set_.empty()) {
     ClearKeyState();
     return false;
   }
-  // 3. Restore the min-heap property. O(K) is efficient here because
-  // multiple expansions often advance simultaneously.
+  // 2. Restore the min-heap property. O(K).
   key_set_.heapify();
   current_key_ = key_set_.min().first;
   current_key_indices_.clear();
-  // 4. Extract all iterators that share this minimum key.
+  // 3. Extract all iterators that share this minimum key.
   // This physically removes them from the heap (making it "empty" if all
   // match).
   while (!key_set_.empty() && key_set_.min().first == current_key_) {
     current_key_indices_.push_back(key_set_.min().second);
     key_set_.pop_min();  // O(log K)
   }
-  // 5. Initialize position iteration for the specific new key if required.
+  // 4. Initialize position iteration for the specific new key if required.
   if (require_positions_) {
     ClearPositionState();
     for (size_t idx : current_key_indices_) {
       pos_iterators_.emplace_back(key_iterators_[idx].GetPositionIterator());
+      // Populate the position heap.
+      InsertValidPositionIterator(pos_iterators_.size() - 1);
     }
     TermIterator::NextPosition();
   }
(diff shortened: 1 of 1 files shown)


src/indexes/text/term.cc
```diff
@@ -32,156 +32,207 @@ FieldMaskPredicate TermIterator::QueryFieldMask() const {
...
 
 const InternedStringPtr& TermIterator::CurrentKey() const {
   CHECK(current_key_);
   return current_key_;
 }
 
+// Helper function to advance key iterators and populate the heap with valid
+// iterators for the new key.
+void TermIterator::InsertValidKeyIterator(
+    size_t idx, std::optional<InternedStringPtr> min_key) {
```
allenss-amazon: maybe change optional<InternedStringPtr> to optional<InternedStringPtr *>, avoid the extra increment/decrement for the copy of the string. Unfortunately optional<InternedStringPtr &> doesn't work.

src/indexes/text/term.cc
```diff
@@ -32,156 +32,207 @@ FieldMaskPredicate TermIterator::QueryFieldMask() const {
...
+  // Use query_field_mask for the original word (idx 0) or if no stem mask is
+  // provided.
+  const auto field_mask = ((idx == 0 && has_original_) || stem_field_mask_ == 0)
+                              ? query_field_mask_
+                              : stem_field_mask_;
+  // Advance the iterator until it matches the required fields and is >=
+  // min_key.
+  while (key_iter.IsValid() && (!key_iter.ContainsFields(field_mask) ||
+                                (min_key && key_iter.GetKey() < *min_key))) {
+    key_iter.NextKey();
```
allenss-amazon, line 60: Shouldn't this be a "SkipForward" to the key and then increment until you get a valid field?

src/indexes/text/term.cc
```diff
@@ -32,156 +32,207 @@ FieldMaskPredicate TermIterator::QueryFieldMask() const {
...
-    for (auto& pos_iter : pos_iterators_) {
-      if (pos_iter.IsValid() &&
-          pos_iter.GetPosition() == current_position_.value().start) {
-        pos_iter.NextPosition();
-      }
-    }
+// Helper function to advance position iterators and populate the heap with
+// valid iterators for the new position.
+void TermIterator::InsertValidPositionIterator(
+    size_t idx, std::optional<uint32_t> min_position) {
```
allenss-amazon: Seems like instead of optional<> min_position, you could just default min_position to max_positive and avoid the conditional, it'll speed up the loop compare because the min_position would always be valid...

src/indexes/text/term.cc
```diff
@@ -32,156 +32,207 @@ FieldMaskPredicate TermIterator::QueryFieldMask() const {
...
-          pos_iterators_.clear();
-          pos_iterators_.emplace_back(key_iter.GetPositionIterator());
-          current_key_ = key;
-        } else if (key == current_key_) {
-          pos_iterators_.emplace_back(key_iter.GetPositionIterator());
-        }
-      } else if (!current_key_ || key < current_key_) {
-        current_key_ = key;
-      }
+  // 1. Refill the heap if it was fully drained by the previous key move.
```
allenss-amazon: What circumstances cause this?

src/indexes/text/term.cc
```diff
@@ -32,156 +32,207 @@ FieldMaskPredicate TermIterator::QueryFieldMask() const {
...
+  if (key_set_.empty()) {
+    for (size_t i = 0; i < key_iterators_.size(); ++i) {
+      InsertValidKeyIterator(i);
     }
   }
-  if (!current_key_) {
+  // If still empty, we are done with all keys.
+  if (key_set_.empty()) {
```
allenss-amazon, line 77: This feels redundant with the test above.

src/indexes/text/term.cc
```diff
@@ -32,156 +32,207 @@ FieldMaskPredicate TermIterator::QueryFieldMask() const {
...
+  if (pos_set_.empty()) {
+    for (size_t i = 0; i < pos_iterators_.size(); ++i) {
+      InsertValidPositionIterator(i);
     }
   }
-  if (!found) {
-    current_position_ = std::nullopt;
-    current_field_mask_ = 0ULL;
+  // If still empty, we are done with all positions for the current key.
+  if (pos_set_.empty()) {
```
allenss-amazon, line 173: This feels redundant. Should it get moved up under the previous test for empty?

src/indexes/text/term.cc
KarthikSubbarao: this and `std::optional<InternedStringPtr> min_key` in the function declaration is actually just left over code from the earlier version. It should have been cleaned up and removed.  The seek/skip is actually already done in the SeekForward API using the `SkipForward`. 

I will remove it

src/indexes/text/term.cc
KarthikSubbarao: `std::optional<uint32_t> min_position` in the function declaration is actually just left over code from the earlier version. It should have been cleaned up and removed. The seek/skip is actually already done in the SeekForward API using the `SkipForward`. 


I will remove it

src/indexes/text/term.cc
KarthikSubbarao: There are 3 cases where this can happen:
1. First time constructing the heap (on Creation of the Term Iterator, during priming of the first key/position)
2. The heap becomes empty due to being popped since they were "current/active iterators" and they do not get added back later on since they are invalid/done
3. The heap becomes empty due to being popped while seeking forward and they do not get added back later on since they are invalid/done

src/indexes/text/term.cc
KarthikSubbarao: This was needed before because I used this same function for handling both the heap creation upon construction / new key handling as well as handling the case where the term iterator is done.

After a small refactor, this is no longer the case, so we just need one empty handling here.

src/indexes/text/term.cc
KarthikSubbarao: This was needed before because I used this same function for handling both the heap creation upon construction / new key handling as well as handling the case where the term iterator is done.

After a small refactor, this is no longer the case, so we just need one empty handling here.

PR #814, opened 2026-02-24
Add missing nocontent case in mixed scenario

VoletiRam:
Add missing nocontent case in mixed scenario.


Aksha1812 (approved):
Minor change, approved.

PR #815, opened 2026-02-24
tests for non-utf-8 and sanity multilang tests + negate comprehensive test

daddaman-amz:
Add coverage for non-utf8 data and other languages to ensure we don't see any undefined behavior

### DIFF
--- a/integration/test_fulltext.py
+++ b/integration/test_fulltext.py
@@ -2263,10 +2263,11 @@ class TestFullText(ValkeySearchTestCaseDebugMode):
         # Invalid UTF-8 queries should not crash (may return 0 or handle gracefully)
         assert client.execute_command("FT.SEARCH", "idx", b"invalid\xff")[0] == 0
         assert client.execute_command("FT.SEARCH", "idx", b"\xff\xfe")[0] == 0
-        assert client.execute_command("FT.SEARCH", "idx",  b"%invalid%")[0] == 0
+        assert client.execute_command("FT.SEARCH", "idx",  b"%invalid\xff%")[0] == 0
         assert client.execute_command("FT.SEARCH", "idx", b"@category:{invalid\xc3}")[0] == 1 #tag with non utf8 gives result
-        with pytest.raises(ResponseError):
+        with pytest.raises(ResponseError) as e:
             client.execute_command("FT.SEARCH", "idx", b"@price:invalid\xc3 invalid\xc3]")
+        assert "Invalid filter expression" in str(e.value)
 
     def test_multilanguage_text(self):
         """Test multi-language text handling - should not crash"""


integration/test_fulltext.py
```diff
@@ -2061,31 +2053,267 @@ def test_hybrid_non_vector_query(self):
...
+        client.execute_command("HSET", "doc:6", "content", "中文测试", "title", "测试", "price", "60")
+        # Arabic (RTL script)
+        client.execute_command("HSET", "doc:7", "content", "مرحبا بالعالم", "title", "عربي", "price", "70")
+        # Hindi
+        client.execute_command("HSET", "doc:8", "content", "नमस्ते दुनिया", "title", "हिन्दी", "price", "80")
+        # Emoji (4-byte UTF-8)
+        client.execute_command("HSET", "doc:9", "content", "Hello 👋 World 🌍 दुनिया 你好世界 مرحبا", "title", "emoji", "price", "90")
+        # ft.info
+        info_data = IndexingTestHelper.get_ft_info(client, "idx").parsed_data
+        assert info_data["num_docs"] == 9
```
KarthikSubbarao: 1. Do we see any indexing failures?
2. You tested mixed documents (one text, one numeric) which is good. What happens if it only contained a non-english / non-utf8 text field. Will the num_docs not go up since the doc ingestion fails?

integration/test_fulltext.py
```diff
@@ -2061,31 +2053,267 @@ def test_hybrid_non_vector_query(self):
...
+        # 1. Single term
```
KarthikSubbarao, line 2310: For all the search operations, what is the result? is it 0? Can we check this?

daddaman-amz: some have results, some don't but its not uniform. I can add asserts based on current behavior today.

integration/test_fulltext.py
```diff
@@ -2061,31 +2053,279 @@ def test_hybrid_non_vector_query(self):
...
+        client.execute_command("HSET", "doc:3", "content", "hello—world")  # U+2014 em-dash
+        client.execute_command("HSET", "doc:4", "content", "wait… done")  # U+2026 ellipsis
+        assert IndexingTestHelper.get_ft_info(client, "idx").parsed_data["hash_indexing_failures"] == 0
+        # Invalid UTF-8 sequences - should not crash
+        client.execute_command("HSET", "doc:5", "content", b"invalid\xff\xfe bytes")
+        client.execute_command("HSET", "doc:6", "content", b"truncated\xc3 char")
+        client.execute_command("HSET", "doc:7", "content", b"mixed valid\xffand\xfeinvalid")
+        # ft.info
+        info_data = IndexingTestHelper.get_ft_info(client, "idx").parsed_data
+        assert info_data["num_docs"] == 7
```
KarthikSubbarao: Hmm. Is this expected?

integration/test_fulltext.py
```diff
@@ -2061,31 +2053,279 @@ def test_hybrid_non_vector_query(self):
...
+        client.execute_command("HSET", "doc:10", "content", "纯中文文档", "title", "परीक्षण اختبار")
+        # check info
+        assert IndexingTestHelper.get_ft_info(client, "idx").parsed_data["hash_indexing_failures"] == 0
+        # Test different language in content + invalid UTF-8 in title (expect indexing failures)
+        client.execute_command("HSET", "doc:11", "content", "纯中文文档", "title", b"test\xff\xfe")
+        client.execute_command("HSET", "doc:12", "content", "مستند عربي فقط", "title", b"invalid\xc3")
+        # ft.info
+        info_data = IndexingTestHelper.get_ft_info(client, "idx").parsed_data
+        assert info_data["num_docs"] == 12
+        assert info_data["hash_indexing_failures"] == 2
```
KarthikSubbarao, line 2307: Does this mean we only increment this if all fields in the doc failed to be ingested, not just some?

integration/test_fulltext.py
```diff
@@ -2061,31 +2053,279 @@ def test_hybrid_non_vector_query(self):
...
+        assert info_data["num_records"] >= 27 # 9 docs with all fields
```
KarthikSubbarao: How do we get this? And why do we use >=27? Dont we know the exact number?

integration/test_fulltext.py
KarthikSubbarao: Ok, can you add them?

KarthikSubbarao: For certain things like simple diacritics (très), the search will still work. We can check. We can also check in what cases no matches are found

integration/test_fulltext.py
daddaman-amz: yes and added assert for that in new revision

integration/test_fulltext.py
daddaman-amz: num_docs increments if it matches the schema irrespective of utf8 or non utf8
hash_indexing_failures is incremented if the doc has atleast one non utf 8

num_records has aggregate of all successfully indexed fields (added comment in the test)

integration/test_fulltext.py
daddaman-amz: updated this to use accurate number with comment explaining count

integration/test_fulltext.py
```diff
@@ -2061,31 +2053,291 @@ def test_hybrid_non_vector_query(self):
...
+        # 8. Nested queries
+        assert client.execute_command("FT.SEARCH", "idx", "((Hola mundo) | (très bien) | (你好 世界) | (%بالعالم%) | (World 🌍*))")[0] == 3
+        # 9. Hybrid text + non-text
+        assert client.execute_command("FT.SEARCH", "idx", "@content:दुनिया @price:[80 100]")[0] == 2
+        assert client.execute_command("FT.SEARCH", "idx", "@content:très @price:[20 40]")[0] == 1
+        assert client.execute_command("FT.SEARCH", "idx", "@content:你好世界 @price:[40 60]")[0] == 1
+        # 10. NOCONTENT
+        assert client.execute_command("FT.SEARCH", "idx", "très", "NOCONTENT")[0] == 1 
+        assert client.execute_command("FT.SEARCH", "idx", "((Hola mundo) | (très bien) | (你好 世界) | (%بالعالم%) | (World 🌍*))", "NOCONTENT")[0] == 3
+        # 11. TAG fields with multi-language values - mostly return zero
```
KarthikSubbarao: How does it handle non utf 8?

daddaman-amz: Tags unlike numeric and Text  are indexing the non utf8 chars and returned results (refer test_non_utf8_english_chars)

daddaman-amz: added this issue for tags: https://github.com/valkey-io/valkey-search/issues/834 (will add more detail in it later)

integration/test_fulltext.py
```diff
@@ -2063,29 +2063,290 @@ def test_hybrid_non_vector_query(self):
...
+        assert info_data["num_docs"] == 9
+        # Current behavior: doc:9 tag is not rejected
+        assert info_data["hash_indexing_failures"] == 4  # doc: 5, doc:6, doc:7, doc:8
+        assert info_data["num_records"] == 5 # doc:1 to 4 , doc:9
+        # Query parsing: Verify tokenization handles non-ASCII
+        assert client.execute_command("FT.SEARCH", "idx", "“smart")[0] == 1
+        assert client.execute_command("FT.SEARCH", "idx", "café")[0] == 1
+        assert client.execute_command("FT.SEARCH", "idx", "hello—world")[0] == 1  
+        assert client.execute_command("FT.SEARCH", "idx", "wait…")[0] == 1
+        assert client.execute_command("FT.SEARCH", "idx", "done")[0] == 1
```
KarthikSubbarao, line 2259: Are these non-utf8 ?

integration/test_fulltext.py
```diff
@@ -2063,29 +2063,290 @@ def test_hybrid_non_vector_query(self):
...
+        # Prefix/fuzzy with non-ASCII UTF-8
+        assert client.execute_command("FT.SEARCH", "idx", "%café%")[0] == 1
+        assert client.execute_command("FT.SEARCH", "idx", "%naïve%")[0] == 1
+        # Invalid UTF-8 queries should not crash (may return 0 or handle gracefully)
+        assert client.execute_command("FT.SEARCH", "idx", b"invalid\xff")[0] == 0
+        assert client.execute_command("FT.SEARCH", "idx", b"\xff\xfe")[0] == 0
+        assert client.execute_command("FT.SEARCH", "idx",  b"%invalid%")[0] == 0
```
KarthikSubbarao: What part of this query string is non utf8?

integration/test_fulltext.py
```diff
@@ -2063,29 +2063,290 @@ def test_hybrid_non_vector_query(self):
...
+        assert client.execute_command("FT.SEARCH", "idx", b"@category:{invalid\xc3}")[0] == 1 #tag with non utf8 gives result
+        with pytest.raises(ResponseError):
+            client.execute_command("FT.SEARCH", "idx", b"@price:invalid\xc3 invalid\xc3]")
```
KarthikSubbarao: Is this numeric? Can you add an assert on the error message here? We can catch the exception as `e` and then assert the error message

integration/test_fulltext.py
daddaman-amz: yes this is typo. will correct this

integration/test_fulltext.py
daddaman-amz: No these are valid non-ASCII UTF-8 characters

daddaman-amz: The "done" query is more like sanity check for the doc
```
client.execute_command("HSET", "doc:4", "content", "wait… done")  # U+2026 ellipsis
```

integration/test_fulltext.py
daddaman-amz: Its a generic error with query when text is attempted in numeric fields.  Its more of a sanity check to ensure there is no undefined behavior.

```
valkey.exceptions.ResponseError: Invalid filter expression: `@price:invalid� invalid�]`. Index does not have any text field
```

PR #816, opened 2026-02-24
Refactor FT.AGGREGATE reducers to process records in batch

AlexFilipImproving:
Changed the ProcessRecords interface to accept all records at once instead of processing them one at a time. This enables more efficient implementations, particularly for COUNT which now simply returns the input size rather than incrementing a counter for each record.

Changes:
- Updated ReducerInstance::ProcessRecords to accept const std::vector<absl::InlinedVector<expr::Value, 4>>& all_values
- Modified GroupBy::Execute to collect all values per group before calling ProcessRecords once
- Updated all reducer implementations (Count, Min, Max, Sum, Avg, Stddev, CountDistinct) to process batched records


src/commands/ft_aggregate_exec.cc
```diff
@@ -183,6 +183,11 @@ absl::Status GroupBy::Execute(RecordSet& records) const {
   absl::flat_hash_map<GroupKey,
                       absl::InlinedVector<std::unique_ptr<ReducerInstance>, 4>>
       groups;
+  // Store all values for each group's reducers
```
allenss-amazon: I'd rather see these combined into a single map that accepts a new object.

src/commands/ft_aggregate_exec.cc
```diff
@@ -206,13 +211,14 @@ absl::Status GroupBy::Execute(RecordSet& records) const {
...
       }
+      group_values[group_it->first].resize(reducers_.size());
     }
     for (auto i = 0; i < reducers_.size(); ++i) {
       absl::InlinedVector<expr::Value, 4> args;
       for (auto& nargs : reducers_[i].args_) {
         args.emplace_back(nargs->Evaluate(ctx, *record));
       }
-      group_it->second[i]->ProcessRecord(args);
+      group_values[group_it->first][i].push_back(std::move(args));
```
allenss-amazon: Why std::move(args) ? I think it's declared to be passed by const reference.

src/commands/ft_aggregate_exec.cc
```diff
@@ -183,6 +183,11 @@ absl::Status GroupBy::Execute(RecordSet& records) const {
   absl::flat_hash_map<GroupKey,
                       absl::InlinedVector<std::unique_ptr<ReducerInstance>, 4>>
       groups;
+  // Store all values for each group's reducers
+  absl::flat_hash_map<
+      GroupKey,
+      absl::InlinedVector<std::vector<absl::InlinedVector<expr::Value, 4>>, 4>>
```
allenss-amazon: Maybe a typedef to replace `absl::InlinedVector<expr::Value, 4>` for an arg-vector would read better.

AlexFilipImproving:
I made the requested changes, but it looks like some of the compatibility tests are failing for reasons outside of this change. For example, these look like they have to do with the implementation of `DIALECT 2`

```
>>>>>>>>>>> ERROR FAILURE <<<<<<<<<<<<<<
>>>>>> Starting Test test_bad_numeric_data[json-2] So Far: Correct:0 Wrong:1 <<<<<<<<<
CMD: b'ft.search json_idx1 @n1:[-inf inf] DIALECT 2'
CMD:['ft.search', 'json_idx1', '@n1:[-inf inf]', 'DIALECT', '2']
Mismatched sizes RL:2 VK:3
--RL--
{'__key': b'json:0', '$': b'{"n1":0,"n2":0,"n3":0,"t1":"","t2":"","t3":"","v1":[0,0,0],"e1":1,"e2":"two"}'}
{'__key': b'json:4', '$': b'{"n1":0,"n2":0,"n3":0,"t2":"","t3":"","v1":[4,4,4],"e1":1,"e2":"two"}'}
--VK:--
{'__key': b'json:0', '$': b'{"n1":0,"n2":0,"n3":0,"t1":"","t2":"","t3":"","v1":[0,0,0],"e1":1,"e2":"two"}'}
{'__key': b'json:4', '$': b'{"n1":0,"n2":0,"n3":0,"t2":"","t3":"","v1":[4,4,4],"e1":1,"e2":"two"}'}
{'__key': b'json:5', '$': b'{"n1":0,"n2":0,"n3":0,"t2":"","t3":"","v1":[5,5,5,5]}'}
>>>>>>>>>>> ERROR FAILURE <<<<<<<<<<<<<<
>>>>>> Starting Test test_bad_numeric_data[json-2] So Far: Correct:0 Wrong:2 <<<<<<<<<
CMD: b'ft.search json_idx1 @n1:[-inf inf] DIALECT 2'
CMD:['ft.search', 'json_idx1', '@n1:[-inf inf]', 'DIALECT', '2']
Mismatched sizes RL:2 VK:3
--RL--
{'__key': b'json:0', '$': b'{"n1":0,"n2":0,"n3":0,"t1":"","t2":"","t3":"","v1":[0,0,0],"e1":1,"e2":"two"}'}
{'__key': b'json:4', '$': b'{"n1":0,"n2":0,"n3":0,"t2":"","t3":"","v1":[4,4,4],"e1":1,"e2":"two"}'}
--VK:--
{'__key': b'json:0', '$': b'{"n1":0,"n2":0,"n3":0,"t1":"","t2":"","t3":"","v1":[0,0,0],"e1":1,"e2":"two"}'}
{'__key': b'json:4', '$': b'{"n1":0,"n2":0,"n3":0,"t2":"","t3":"","v1":[4,4,4],"e1":1,"e2":"two"}'}
{'__key': b'json:5', '$': b'{"n1":0,"n2":0,"n3":0,"t2":"","t3":"","v1":[5,5,5,5]}'}
```

src/commands/ft_aggregate_exec.cc
```diff
@@ -204,15 +210,20 @@ absl::Status GroupBy::Execute(RecordSet& records) const {
     if (inserted) {
       DBG << "Was inserted, now have " << groups.size() << " groups\n";
       for (auto& reducer : reducers_) {
-        group_it->second.emplace_back(reducer.info_->make_instance());
+        ArgVector args;
+        for (auto& nargs : reducer.args_) {
+          args.emplace_back(nargs->Evaluate(ctx, *record));
+        }
+        group_it->second.emplace_back(std::move(reducer.info_->make_instance()),
+                                      std::vector<ArgVector>{});
```
allenss-amazon: This can't be correct, "args" isn't used. I think we can just eliminate the computation of args entirely.

src/commands/ft_aggregate_exec.cc
```diff
@@ -204,15 +210,20 @@ absl::Status GroupBy::Execute(RecordSet& records) const {
...
       }
     }
-    for (auto i = 0; i < reducers_.size(); ++i) {
-      absl::InlinedVector<expr::Value, 4> args;
+    for (int i = 0; i < reducers_.size(); ++i) {
```
allenss-amazon: I think this is un-optimizing the cases where the reducer doesn't need to process the entire vector in a single block. Maybe what we want to do is to move the accumulation of a vector of intermediate values into the reducers themselves and leave the interface unchanged. Seems like we could have a class derived from reducer than implements the accumulation capability and then the reducers that want to have the capability derive from the sub-class.

src/commands/ft_aggregate_exec.cc
```diff
@@ -336,19 +342,62 @@ class CountDistinct : public GroupBy::ReducerInstance {
 };
 
 template <typename T>
-std::unique_ptr<GroupBy::ReducerInstance> MakeReducer() {
-  return std::unique_ptr<GroupBy::ReducerInstance>(std::make_unique<T>());
+struct BasicReducer : GroupBy::Reducer {
+  // BasicReducer(std::string name) : GroupBy::Reducer(std::move(name)) {}
```
allenss-amazon, line 346: A better comment would be to describe when it's used, i.e., for Reducers that can compute their value with out preserving all of the values.

allenss-amazon (approved):

Issue #703, opened 2026-02-05
[FEATURE] Need to run TSAN on regression tests.


allenss-amazon:
**The problem/use-case that the feature addresses**

Need to re-enable TSAN for regression tests.

It was working and now it's broken. It needs to be fixed.

baswanth09:
The valkey-server itself is crashing even without modules. 
```
export TSAN_OPTIONS="suppressions=/workplace/baswanth/tsan/valkey-search/ci/tsan.supp"
```

```
/workplace/baswanth/tsan/valkey-search/.build-release-tsan/valkey-server/.build-release/bin/valkey-server --port 16379

zsh: segmentation fault   --port 16379
```
```
gdb -batch -ex "run" -ex "bt" --args /workplace/baswanth/tsan/valkey-search/.build-release-tsan/valkey-server/.build-release/bin/valkey-server --port 16379

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff6de0fb0 in ?? ()
#0  0x00007ffff6de0fb0 in ?? ()
#1  0x00000000004b89a7 in string2ll_resolver.lto_priv ()
#2  0x00007ffff7de4d94 in _dl_relocate_object () from /lib64/ld-linux-x86-64.so.2
#3  0x00007ffff7ddceaa in dl_main () from /lib64/ld-linux-x86-64.so.2
#4  0x00007ffff7df0b5f in _dl_sysdep_start () from /lib64/ld-linux-x86-64.so.2
#5  0x00007ffff7ddad2e in _dl_start () from /lib64/ld-linux-x86-64.so.2
#6  0x00007ffff7dd9ef8 in _start () from /lib64/ld-linux-x86-64.so.2
#7  0x0000000000000003 in ?? ()
#8  0x00007fffffffe04c in ?? ()
#9  0x00007fffffffe0b6 in ?? ()
#10 0x00007fffffffe0bd in ?? ()
#11 0x0000000000000000 in ?? ()
```

PR #818, opened 2026-02-25
Fix TSAN build for integration tests

baswanth09:
Resolves: #703 

Encountered the following series of crashes during engine startup. 
1. SIGSEGV in `string2ll_resolver` — IFUNC resolver called before TSAN runtime initializes
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff6de0fb0 in ?? ()
#0  0x00007ffff6de0fb0 in ?? ()
#1  0x00000000004b89a7 in string2ll_resolver.lto_priv ()
#2  0x00007ffff7de4d94 in _dl_relocate_object () from /lib64/ld-linux-x86-64.so.2
#3  0x00007ffff7ddceaa in dl_main () from /lib64/ld-linux-x86-64.so.2
#4  0x00007ffff7df0b5f in _dl_sysdep_start () from /lib64/ld-linux-x86-64.so.2
#5  0x00007ffff7ddad2e in _dl_start () from /lib64/ld-linux-x86-64.so.2
#6  0x00007ffff7dd9ef8 in _start () from /lib64/ld-linux-x86-64.so.2
#7  0x0000000000000003 in ?? ()
#8  0x00007fffffffe04c in ?? ()
#9  0x00007fffffffe0b6 in ?? ()
#10 0x00007fffffffe0bd in ?? ()
#11 0x0000000000000000 in ?? ()
```
2. TSAN abort on `RTLD_DEEPBIND` — module loader uses `RTLD_DEEPBIND` which is incompatible with TSAN
```
cat /workplace/baswanth/tsan/valkey-search/testing/integration/.build-release-tsan/output/6379_stdout.txt 2>/dev/null | tail -20
45443:M 25 Feb 2026 01:57:58.222 # WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled, it can also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
45443:M 25 Feb 2026 01:57:58.222 * oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo
45443:M 25 Feb 2026 01:57:58.222 * Valkey version=9.0.1, bits=64, commit=ab3c953b, modified=1, pid=45443, just started
45443:M 25 Feb 2026 01:57:58.222 * Configuration loaded
45443:M 25 Feb 2026 01:57:58.224 * monotonic clock: POSIX clock_gettime
45443:M 25 Feb 2026 01:57:58.226 * Running mode=cluster, port=6379.
45443:M 25 Feb 2026 01:57:58.226 * No cluster configuration found, I'm a13c403e8a798f9ad051c654ff540ce1b3833b9d
==45443==You are trying to dlopen a /workplace/baswanth/tsan/valkey-search/.build-release-tsan/valkey-json/build/src/libjson.so shared library with RTLD_DEEPBIND flag which is incompatible with sanitizer runtime (see https://github.com/google/sanitizers/issues/611 for details). If you want to run /workplace/baswanth/tsan/valkey-search/.build-release-tsan/valkey-json/build/src/libjson.so library under sanitizers please remove RTLD_DEEPBIND from dlopen flags.
```
3. `CHECK failed: sanitizer_deadlock_detector.h:67` — TSAN deadlock detector overflows during module load
```
7826:M 25 Feb 2026 02:39:45.284 * Valkey version=9.0.1, bits=64, commit=ab3c953b, modified=1, pid=17826, just started
17826:M 25 Feb 2026 02:39:45.284 * Configuration loaded
17826:M 25 Feb 2026 02:39:45.286 * monotonic clock: POSIX clock_gettime
17826:M 25 Feb 2026 02:39:45.288 * Running mode=cluster, port=6379.
17826:M 25 Feb 2026 02:39:45.289 * No cluster configuration found, I'm 134edf06efe1a2e084ce9c94a43dab67e8376dfc
17826:M 25 Feb 2026 02:39:45.298 . <module> Setting shards to 524287
ThreadSanitizer: CHECK failed: sanitizer_deadlock_detector.h:67 "((n_all_locks_)) < (((sizeof(all_locks_with_contexts_)/sizeof((all_locks_with_contexts_)[0]))))" (0x40, 0x40) (tid=17826)
```

core changes that needs to be added.
```
cd /workplace/baswanth/tsan/valkey-search/.build-release-tsan/valkey-server && git diff src/util.c src/module.c src/config.h
diff --git a/src/config.h b/src/config.h
index 0c6b111a9..f80fa8cbc 100644
--- a/src/config.h
+++ b/src/config.h
@@ -172,6 +172,10 @@
 #define VALKEY_NO_SANITIZE(sanitizer)
 #endif
 
+#if defined(__SANITIZE_THREAD__)
+#define VALKEY_THREAD_SANITIZER 1
+#endif
+
 #if defined(__SANITIZE_ADDRESS__)
 /* GCC */
 #define VALKEY_ADDRESS_SANITIZER 1
diff --git a/src/module.c b/src/module.c
index 8d1bad4fe..0435499a8 100644
--- a/src/module.c
+++ b/src/module.c
@@ -12545,7 +12545,7 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
     }
 
     int dlopen_flags = RTLD_NOW | RTLD_LOCAL;
-#if (defined(__GLIBC__) || defined(__FreeBSD__)) && !defined(VALKEY_ADDRESS_SANITIZER) && __has_include(<dlfcn.h>)
+#if (defined(__GLIBC__) || defined(__FreeBSD__)) && !defined(VALKEY_ADDRESS_SANITIZER) && !defined(VALKEY_THREAD_SANITIZER) && __has_include(<dlfcn.h>)
     /* RTLD_DEEPBIND, which is required for loading modules that contains the
      * same symbols, does not work with ASAN. Therefore, we exclude
      * RTLD_DEEPBIND when doing test builds with ASAN.
diff --git a/src/util.c b/src/util.c
index 0631736c8..cf5a073b1 100644
--- a/src/util.c
+++ b/src/util.c
@@ -622,7 +622,7 @@ static int string2llScalar(const char *s, size_t slen, long long *value) {
 }
 
 #if HAVE_IFUNC && HAVE_X86_SIMD
-__attribute__((no_sanitize_address, used)) static int (*string2ll_resolver(void))(const char *, size_t, long long *) {
+__attribute__((no_sanitize_address, no_sanitize("thread"), used)) static int (*string2ll_resolver(void))(const char *, size_t, long long *) {
     /* Ifunc resolvers run before ASan initialization and before CPU detection
      * is initialized, so disable ASan and init CPU detection here. */
     __builtin_cpu_init();
 ```

### DIFF
--- a/.github/benchmark_configs/fts-benchmarks-arm.json
+++ b/.github/benchmark_configs/fts-benchmarks-arm.json
@@ -2,6 +2,7 @@
   {
     "test_name": "FTS Performance Benchmarks",
     "seed": false,
+    "socket_timeout": 300,
     "dataset_generation": {
       "hybrid_data_100k.csv": {
         "doc_count": 100000,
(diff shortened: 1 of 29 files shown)

### CODE REVIEW

File: scripts/common.rc [hunk 1]
```diff
@@ -191,6 +193,22 @@ function setup_valkey_server() {
   if [ ! -z "${SAN_BUILD}" ] && [[ "${SAN_BUILD}" != "no" ]]; then
     VALKEY_CMAKE_EXTRA_ARGS="-DBUILD_SANITIZER=${SAN_BUILD}"
     LOG_INFO "Building Valkey with Sanitizer options: -DBUILD_SANITIZER=${SAN_BUILD}"
+    # Temporary workaround until these are merged into core upstream:
+    # 1. IFUNC resolvers run before TSAN runtime initializes, so no_sanitize("thread") is required.
```
allenss-amazon, line 197: Would it be better/easier to just build valkey-server with the ADDRESS_SANITIZER #define set instead of trying to create a new #define ? Don't we want the same alternate code path set for thread sanitizer as well as address sanitizer?

baswanth09: Yes, we want the same `RTLD_DEEPBIND` code path for TSAN as ASAN. Unfortunately we can't reuse VALKEY_ADDRESS_SANITIZER directly because it is set automatically by the compiler only when `-fsanitize=address` is active (via `__SANITIZE_ADDRESS__`). When building with `-fsanitize=thread`, the compiler defines `__SANITIZE_THREAD__` and `__SANITIZE_ADDRESS__` is never set, so `VALKEY_ADDRESS_SANITIZER` is never defined and the existing guard doesn't trigger for TSAN builds.
Fixes 2 and 3 introduce `VALKEY_THREAD_SANITIZER` mirroring the exact same pattern as `VALKEY_ADDRESS_SANITIZER` in config.h

baswanth09: PR in core: https://github.com/valkey-io/valkey/pull/3257

PR #821, opened 2026-02-25
Add config to limit keys from pre-content fetching phase

daddaman-amz:
Adding new config which will limit the number of keys fetched in the background threads which will help in controlling the ephemeral memory from going out of bounds

Config name: `search.max-nonvector-search-results-fetched`

Applies to only non vector queries.  The vector paths are already controlled by K.

Dev metric added: `nonvector_results_fetched_limited_count`

### DIFF
--- a/src/query/search.cc
+++ b/src/query/search.cc
@@ -371,7 +371,7 @@ void EvaluatePrefilteredKeys(
     absl::AnyInvocable<bool(const InternedStringPtr &,
                             absl::flat_hash_set<const char *> &)>
         appender,
-    size_t max_keys, bool stop_on_fetch_limit = false) {
+    size_t max_keys, bool stop_on_fetch_limit) {
   // If there was a union operation, we need to handle deduplication.
   // This implementation skips deduplication (flat_hash_set usage) if not needed
   // for performance.
@@ -442,7 +442,8 @@ CalcBestMatchingPrefilteredKeys(
                                            results, top_keys);
   };
   EvaluatePrefilteredKeys(parameters, entries_fetchers,
-                          std::move(results_appender), qualified_entries);
+                          std::move(results_appender), qualified_entries,
+                          /*stop_on_fetch_limit=*/false);
   return results;
 }
 
--- a/src/valkey_search_options.cc
+++ b/src/valkey_search_options.cc
@@ -427,6 +427,9 @@ static auto async_fanout_threshold =
 /// fetching on non-vector (numeric/tag/text) query paths. This controls
 /// potential OOM by limiting the result set size before expensive content
 /// fetching from the keyspace.
+/// Note: If queries use LIMIT with a large offset (e.g., LIMIT offset count
+/// where offset + count exceeds this value), consider increasing this config
+/// to ensure all paginated results are accessible.
 constexpr absl::string_view kMaxNonVectorSearchResultsFetchedConfig{
     "max-nonvector-search-results-fetched"};
 constexpr uint32_t kDefaultMaxNonVectorSearchResultsFetched{


src/valkey_search_options.cc
```diff
@@ -422,6 +422,24 @@ static auto async_fanout_threshold =
...
+/// number of keys to accumulate in background threads before content fetching.
+/// This controls potential OOM by limiting the result set size before expensive
+/// content fetching from the keyspace.
+constexpr absl::string_view kMaxSearchKeysAccumulatedConfig{
+    "max-search-keys-accumulated"};
+constexpr uint32_t kDefaultMaxSearchKeysAccumulated{
+    100000};  // 100K keys default
+constexpr uint32_t kMinimumMaxSearchKeysAccumulated{1};
+constexpr uint32_t kMaximumMaxSearchKeysAccumulated{10000000};  // Max 10M keys
+static auto max_search_keys_accumulated =
```
KarthikSubbarao: I don't think this should have a limit

src/valkey_search_options.cc
```diff
@@ -422,6 +422,24 @@ static auto async_fanout_threshold =
...
         .Build();
 
+/// Register the "--max-search-keys-accumulated" flag. Controls the maximum
+/// number of keys to accumulate in background threads before content fetching.
+/// This controls potential OOM by limiting the result set size before expensive
+/// content fetching from the keyspace.
+constexpr absl::string_view kMaxSearchKeysAccumulatedConfig{
+    "max-search-keys-accumulated"};
+constexpr uint32_t kDefaultMaxSearchKeysAccumulated{
+    100000};  // 100K keys default
```
KarthikSubbarao: What is the overhead of 100k search results per query?

src/query/search.cc
```diff
@@ -605,17 +619,33 @@ absl::StatusOr<std::vector<indexes::Neighbor>> SearchNonVectorQuery(
...
         }
       }
+      // If truncated or cancelled, stop processing remaining fetchers
+      if (truncated || parameters.cancellation_token->IsCancelled()) {
+        break;
+      }
     }
-    return neighbors;
+  } else {
+    EvaluatePrefilteredKeys(parameters, entries_fetchers,
```
KarthikSubbarao: Should we not handle `EvaluatePrefilteredKeys`?

src/valkey_search_options.cc
```diff
@@ -422,6 +422,24 @@ static auto async_fanout_threshold =
         kMaximumAsyncFanoutThreshold)  // max threshold (10k)
         .Build();
 
+/// Register the "--max-search-keys-accumulated" flag. Controls the maximum
+/// number of keys to accumulate in background threads before content fetching.
+/// This controls potential OOM by limiting the result set size before expensive
+/// content fetching from the keyspace.
+constexpr absl::string_view kMaxSearchKeysAccumulatedConfig{
+    "max-search-keys-accumulated"};
```
KarthikSubbarao: > "max-search-keys-accumulated"};


Let's think of names:

max-search-keys-fetched
max-search-results-fetched
max-search-results

src/query/search.cc
```diff
@@ -64,6 +64,7 @@ DEV_INTEGER_COUNTER(query_stats, query_text_fuzzy_count);
 DEV_INTEGER_COUNTER(query_stats, query_text_proximity_count);
 DEV_INTEGER_COUNTER(query_stats, query_numeric_count);
 DEV_INTEGER_COUNTER(query_stats, query_tag_count);
+DEV_INTEGER_COUNTER(query_stats, query_keys_truncated_cnt);
```
KarthikSubbarao: 1. What is the metric's name?
2. Let's not use truncated. Let's use "limited" and make the metric be identifiable / similar to the config which limits it

src/valkey_search_options.cc
daddaman-amz: yeah I agree, we can make it `UINT32_MAX`

src/valkey_search_options.cc
```diff
@@ -422,6 +422,24 @@ static auto async_fanout_threshold =
...
+constexpr uint32_t kDefaultMaxSearchKeysAccumulated{
+    100000};  // 100K keys default
+constexpr uint32_t kMinimumMaxSearchKeysAccumulated{1};
```
KarthikSubbarao: 0?

src/valkey_search_options.cc
daddaman-amz: `max-search-results-fetched` as config 

`results_fetched_limited_count`   as metric

src/query/search.cc
```diff
@@ -573,13 +574,26 @@ absl::StatusOr<std::vector<indexes::Neighbor>> SearchNonVectorQuery(
...
       parameters.index_schema.get());
+
+  // Get the config for maximum number of keys to accumulate before content
+  // fetching
+  const size_t max_keys =
+      static_cast<size_t>(options::GetMaxSearchKeysAccumulated().GetValue());
   std::vector<indexes::Neighbor> neighbors;
   neighbors.reserve(std::min(qualified_entries, static_cast<size_t>(5000)));
+  // Track accumulated count and truncation status (shared across both paths)
+  size_t accumulated = 0;
```
KarthikSubbarao: fetched

KarthikSubbarao: Can we use the len of the vector instead?

src/valkey_search_options.cc
daddaman-amz: As discussed since config is visible to all,   I will rename the config to `max-nonvector-search-results-fetched` which signifies it applies to nonvector path

Dev metric also to be `nonvector_results_fetched_limited_count` for consistency

We earlier noted that if vector is involved the cuurent code is already limiting results by "K" which has its own config range

src/query/search.cc
daddaman-amz: we update the lambda prior which is fed to this functions.
Also as we disucssed this restricting this config and restriction to nonvector paths only.  Vector paths are limited by the K

src/valkey_search_options.cc
```diff
@@ -422,6 +422,25 @@ static auto async_fanout_threshold =
         kMaximumAsyncFanoutThreshold)  // max threshold (10k)
         .Build();
 
+/// Register the "--max-nonvector-search-results-fetched" flag. Controls the
```
KarthikSubbarao, line 425: Document the pagination behavior and what is recommended for cases where LIMIT or OFFSET for non vector queries goes higher than this config

src/query/search.cc
```diff
@@ -370,7 +371,7 @@ void EvaluatePrefilteredKeys(
     absl::AnyInvocable<bool(const InternedStringPtr &,
                             absl::flat_hash_set<const char *> &)>
         appender,
-    size_t max_keys) {
+    size_t max_keys, bool stop_on_fetch_limit = false) {
```
KarthikSubbarao: Nit: Using defaults is not generally safe since caller sites can unknowingly use a default value when it might not be correct

PR #822, opened 2026-02-26
test: add comprehensive timeout coverage for all search execution paths

### LINKED ISSUE #686: [TEST] Add test using FT.DEBUG to trigger inf loop in all search paths
** Item not tested **

Add test using ft.debug to trigger inf loop inside a background search code path which is part of evaluating a particular key (1) in the entries fetcher, (2) in the prefilter evaluators and (3) in the predicate evaluators. 
For the background thread operations, test if reader thread remains at `100% cpu or if it is all good and it gracefully times out the cmd within the appropriate duration.

** Testing Method **

Add integration tests

### LINKED ISSUE #769: [BUG] FT.AGGREGATE timeout handling is inferior.
**Describe the bug**

FT.AGGREGATE should check for OOM and timeout before each stage.

**To Reproduce**
Steps to reproduce the behavior. Ideally, a script to reproduce the problem.

**Expected behavior**

**Environment**

- Valkey Core Version 

**Additional context**

Arunsarma07:
## Summary
This PR adds comprehensive timeout handling and test coverage for both `FT.SEARCH` and `FT.AGGREGATE` commands.

### FT.SEARCH Testing (#686)

Added pausepoint-based integration tests to verify timeout behavior across all search execution paths.

**Added 9 pausepoints in search loops:**
- Entries fetcher loop
- Prefilter evaluation loop
- Inline vector filter
- Text predicates (term, prefix, suffix, fuzzy)
- Composed predicate children iteration

**Added 9 integration tests that:**
- Set a pausepoint to simulate an infinite loop
- Run FT.SEARCH with a 5-second timeout in a background thread
- Verify the server stays responsive (not spinning at 100% CPU)
- Verify the command times out and returns an error

**Coverage includes all query types:**
- Non-vector queries (Tag, Numeric, Text)
- Text searches (term, prefix, suffix, fuzzy, composed)
- Hybrid vector queries (prefiltering and inline filtering)

### FT.AGGREGATE Timeout Fix (#769)

FT.AGGREGATE did not have timeout checks during aggregation stage execution. Added timeout checks before processing stages in both standalone and cluster modes.

---

Closes #686  
Fixes #769

### DIFF
--- a/.github/benchmark_configs/fts-benchmarks-arm.json
+++ b/.github/benchmark_configs/fts-benchmarks-arm.json
@@ -1537,7 +1537,7 @@
               "NOCONTENT": "_nocontent"
             },
             "warmup": 60,
-            "command": "FT.SEARCH rd0 \"(@title:__field:search_term__)=>[KNN 10 @embedding $vec AS score]\" PARAMS 2 vec \"__field:query_vector__\" RETURN 2 title score"
+            "command": "FT.SEARCH rd0 \"(@title:__field:search_term__)=>[KNN 10 @embedding $vec AS score]\" PARAMS 2 vec \"__field:query_vector__\" RETURN 2 title score DIALECT 2"
           },
           {
             "id": "c",
@@ -1552,7 +1552,7 @@
               "NOCONTENT": "_nocontent"
             },
             "warmup": 60,
-            "command": "FT.SEARCH rd0 \"(@price:[100 500])=>[KNN 10 @embedding $vec AS score]\" PARAMS 2 vec \"__field:query_vector__\" RETURN 2 price score"
+            "command": "FT.SEARCH rd0 \"(@price:[100 500])=>[KNN 10 @embedding $vec AS score]\" PARAMS 2 vec \"__field:query_vector__\" RETURN 2 price score DIALECT 2"
           }
         ]
       }
--- a/integration/test_filter_expressions.py
+++ b/integration/test_filter_expressions.py
@@ -901,3 +901,32 @@ class TestFilterExpressions(ValkeySearchTestCaseBase):
         assert result[0] >= 1
         keys = set(result[i].decode('utf-8') for i in range(1, len(result)))
         assert "item:9" in keys
+
+    def test_10k_tag_filters(self):
+        """
+        Test with 10,000 tag filters to verify query complexity limits.
+        """
+        client: Valkey = self.server.get_new_client()
+        # Set query-string-bytes to 100k to handle large query string size
+        assert client.execute_command("CONFIG", "SET", "search.query-string-bytes", "100000") == b"OK"
+        # Create index with tag field
+        assert client.execute_command(
+            "FT.CREATE", "large_idx",
+            "ON", "HASH",
+            "PREFIX", "1", "item:",
+            "SCHEMA", "tags", "TAG"
+        ) == b"OK"
+        # Add a few test documents
+        assert client.execute_command("HSET", "item:1", "tags", "tag_1") == 1
+        assert client.execute_command("HSET", "item:2", "tags", "tag_5000") == 1
+        assert client.execute_command("HSET", "item:3", "tags", "tag_9999") == 1
+        # Generate 10k tag filters using OR syntax: @tags:{tag_1|tag_2|...|tag_10000}
+        tag_list = [f"tag_{i}" for i in range(1, 10001)]
+        tags_string = '|'.join(tag_list)
+        query = f"@tags:{{{tags_string}}}"
+        # This should work within the increased 100k query terms limit
+        result = client.execute_command("FT.SEARCH", "large_idx", query, "NOCONTENT")
+        # Should find the 3 documents that match some of the tags
+        assert result[0] == 3
+        keys = set(result[i].decode('utf-8') for i in range(1, len(result)))
+        assert keys == {"item:1", "item:2", "item:3"}
--- a/src/indexes/text/lexer.cc
+++ b/src/indexes/text/lexer.cc
@@ -84,8 +84,7 @@ Lexer::Lexer(data_model::Language language, const std::string& punctuation,
 
 absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
     absl::string_view text, bool stemming_enabled, uint32_t min_stem_size,
-    absl::flat_hash_map<std::string, absl::flat_hash_set<std::string>>*
-        stem_mappings) const {
+    InProgressStemMap* stem_mappings) const {
   if (stemming_enabled) {
     CHECK(stem_mappings) << "stem_mappings must not be null";
   }
@@ -222,11 +221,9 @@ void Lexer::StemWordInPlace(std::string& word, sb_stemmer* stemmer,
   }
 }
 
-void Lexer::UpdateStemMap(
-    absl::string_view original_word, sb_stemmer* stemmer,
-    uint32_t min_stem_size,
-    absl::flat_hash_map<std::string, absl::flat_hash_set<std::string>>&
-        stem_mappings) const {
+void Lexer::UpdateStemMap(absl::string_view original_word, sb_stemmer* stemmer,
+                          uint32_t min_stem_size,
+                          InProgressStemMap& stem_mappings) const {
   std::string_view stemmed_view =
       DoStemming(original_word, stemmer, min_stem_size);
   if (stemmed_view != original_word) {
@@ -235,8 +232,12 @@ void Lexer::UpdateStemMap(
       // New Stem Root: create a new map entry
       it = stem_mappings.try_emplace(std::string(stemmed_view)).first;
     }
-    // Add original word to the set of variants for this stem root
-    it->second.emplace(original_word);
+    // Add original word to the list of variants for this stem root
+    auto& variants = it->second;
+    if (std::find(variants.begin(), variants.end(), original_word) ==
+        variants.end()) {
+      variants.emplace_back(original_word);
+    }
   }
 }
 }  // namespace valkey_search::indexes::text
--- a/src/indexes/text/lexer.h
+++ b/src/indexes/text/lexer.h
@@ -28,7 +28,9 @@ Tokenization Pipeline:
 #include <string>
 #include <vector>
 
+#include "absl/container/flat_hash_map.h"
 #include "absl/container/flat_hash_set.h"
+#include "absl/container/inlined_vector.h"
 #include "absl/status/statusor.h"
 #include "absl/strings/string_view.h"
 #include "src/index_schema.pb.h"
@@ -37,6 +39,15 @@ struct sb_stemmer;
 
 namespace valkey_search::indexes::text {
 
+// Inline capacity for per-document stem mapping
+constexpr size_t kInProgressStemVariantsInlineCapacity = 4;
+
+// Per-document stem mappings: stemmed_word -> list of original words that stem
+// to it
+using InProgressStemMap = absl::flat_hash_map<
+    std::string,
+    absl::InlinedVector<std::string, kInProgressStemVariantsInlineCapacity>>;
+
 struct Lexer {
   Lexer(data_model::Language language, const std::string& punctuation,
         const std::vector<std::string>& stop_words);
@@ -44,8 +55,7 @@ struct Lexer {
 
   absl::StatusOr<std::vector<std::string>> Tokenize(
       absl::string_view text, bool stemming_enabled, uint32_t min_stem_size,
-      absl::flat_hash_map<std::string, absl::flat_hash_set<std::string>>*
-          stem_mappings = nullptr) const;
+      InProgressStemMap* stem_mappings = nullptr) const;
 
   bool IsPunctuation(char c) const {
     return punct_bitmap_[static_cast<unsigned char>(c)];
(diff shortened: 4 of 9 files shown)


src/commands/ft_aggregate.cc
```diff
@@ -307,7 +307,10 @@ absl::Status ExecuteAggregationStages(AggregateParameters &parameters,
                                       RecordSet &records) {
   agg_input_records.Increment(records.size());
   for (auto &stage : parameters.stages_) {
-    // Todo Check for timeout
+    // Check for timeout
+    if (parameters.cancellation_token->IsCancelled()) {
+      return absl::CancelledError("Search operation cancelled due to timeout");
```
KarthikSubbarao: Search or Aggregate? What do we normally use for Aggregate command errors?

src/query/predicate.cc
```diff
@@ -127,6 +128,9 @@ EvaluationResult TermPredicate::Evaluate(
     }
     // Search for stem variants - these should all exist from ingestion
     for (const auto &variant : stem_variants) {
+      if (!vmsdk::IsMainThread()) {
```
KarthikSubbarao: This means we are not testing the timeout of the predicate evaluator which runs on the main thread, right?

src/commands/ft_aggregate.cc
```diff
@@ -390,6 +393,12 @@ bool AggregateParameters::RequiresCompleteResults() const {
 
 void AggregateParameters::SendReply(ValkeyModuleCtx *ctx,
                                     query::SearchResult &result) {
+  if (cancellation_token->IsCancelled()) {
```
KarthikSubbarao: What happens there when partial results are enabled? I was reading the FT Search command for comparison. From the code, it sounds like it just returns the results fetched (Which might not be the complete result set).
For Aggregate, this is probably fine

src/commands/ft_aggregate.cc
Arunsarma07: I'll make a correction. It should be Aggregate here as the check is in the execution stage of FT.AGGREGATE.

src/query/predicate.cc
Arunsarma07: Correct. I have added the tests for background thread executions for various search paths. But as far as the main thread execution there is a generic test `test_timeoutCMD` in the same `test_cancel.py` file which forces a timeout. I can add some more tests if needed.

src/commands/ft_aggregate.cc
Arunsarma07: I think there was a brief discussion on this that we do not return any kind of partial results for the timeout for FT.AGGREGATE command.

allenss-amazon: Correct, for aggregate, there really isn't interesting "partial results". So a timeout for this command should always be reported as a timeout.

integration/test_cancel.py
```diff
@@ -9,6 +9,8 @@
 import logging
 from typing import Any, Union
 from valkeytestframework.util import waiters
+import threading
```
KarthikSubbarao, line 12: This test is 700 lines long. Is there anything here which can be cleaned up before I review it? Try checking for existing helper methods and not creating one off functions for this test

src/indexes/text.cc
```diff
@@ -185,6 +186,9 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
     }
     // Search for stem variants - these should all exist from ingestion
     for (const auto &variant : stem_variants) {
+      if (!vmsdk::IsMainThread()) {
+        PAUSEPOINT("search_term_predicate");
+      }
```
KarthikSubbarao: This is not on the main thread

KarthikSubbarao: Also, what was the purpose of adding the pause point here and not on the other predicates? I am trying to understand what you are planning on testing with this

src/query/predicate.cc
```diff
@@ -165,6 +169,9 @@ EvaluationResult PrefixPredicate::Evaluate(
   uint32_t max_words = options::GetMaxTermExpansions().GetValue();
   uint32_t word_count = 0;
   while (!word_iter.Done() && word_count < max_words) {
+    if (!vmsdk::IsMainThread()) {
```
KarthikSubbarao: It would be nice to have a helper function which helps us avoid having to add this check everywhere:
```
if (!vmsdk::IsMainThread()) 
```

integration/test_cancel.py
Arunsarma07: I have added a dedicated test to hit the each of the pausepoints to see if the command times out. I'll see if I can reuse some existing methods but I think it is mostly the new tests.

src/indexes/text.cc
Arunsarma07: I have added the same pausepoint in the `predicate.cc` file in the `Evaluate` method for other predicates too. I thought I cleaned it up while fixing the integration tests but missed it. I will remove this.

vmsdk/src/debug.h
```diff
@@ -34,6 +34,12 @@ namespace debug {
   vmsdk::debug::PausePoint(name, std::source_location::current())
 void PausePoint(absl::string_view point, std::source_location location);
 
+#define BACKGROUND_PAUSEPOINT(name) \
+  do {                              \
```
KarthikSubbarao: Can we remove the while loop?

integration/test_cancel.py
```diff
@@ -62,6 +64,110 @@ def search(
             assert str(e) == "Search operation cancelled due to timeout"
         return []
 
+def wait_for_pausepoint(client: Valkey, pausepoint_name: str, timeout: int = 5) -> bool:
```
KarthikSubbarao: Can we move this to utilities?

integration/test_cancel.py
```diff
@@ -62,6 +64,110 @@ def search(
...
+    """
+    Wait for a pausepoint to be hit by at least one thread.
+    """
+    start = time.time()
+    while time.time() - start < timeout:
+        count = client.execute_command("FT._DEBUG", "PAUSEPOINT", "TEST", pausepoint_name)
+        if count > 0:
+            return True
+        time.sleep(0.1)
```
KarthikSubbarao: It is not ideal to use time.sleep in tests since it can be flaky based on where it is used

integration/test_cancel.py
```diff
@@ -62,6 +64,110 @@ def search(
...
+        return result
+    except ResponseError as e:
+        if not expect_timeout:
+            raise
+        error_msg = str(e)
+        assert "cancelled due to timeout" in error_msg, f"Expected timeout error, got: {error_msg}"
+        return []
+
+
+def run_pausepoint_timeout_test(self, pausepoint_name, setup_fn, search_cmd, timeout_ms=5000):
```
KarthikSubbarao: Is this a utility function as well?

integration/test_cancel.py
```diff
@@ -62,6 +64,110 @@ def search(
...
+    3. Timeout fires and command returns error
+
+    Args:
+        pausepoint_name: Name of the pausepoint to set
+        setup_fn: Callable(client) that creates indexes and loads data
+        search_cmd: List of command args to execute as the search.
+        timeout_ms: Timeout in milliseconds for the search command
+    """
+    client = self.server.get_new_client()
+    client.execute_command("FLUSHALL", "SYNC")
```
KarthikSubbarao: Why do we need to flush here?

integration/test_cancel.py
```diff
@@ -62,6 +64,110 @@ def search(
...
+
+    thread = threading.Thread(target=run_search)
+    thread.start()
+
+    assert wait_for_pausepoint(client, pausepoint_name, timeout=10), \
+        f"Pausepoint {pausepoint_name} was not hit"
+    assert client.ping() == True, "Server not responsive while pausepoint is held"
+
+    # Wait for timeout to fire (timeout_ms + 1s buffer)
+    time.sleep(timeout_ms / 1000 + 1)
```
KarthikSubbarao: Is this needed?

integration/test_cancel.py
Arunsarma07: Will move it to utilities.

vmsdk/src/debug.h
Arunsarma07: Will remove this.

integration/test_cancel.py
Arunsarma07: I can change this to `waiters.wait_for_true`.

integration/test_cancel.py
Arunsarma07: It helps in isolating each test so leftover state from a prior test doesn't interfere.

integration/test_cancel.py
Arunsarma07: The search command has timeout of 5 sec and I'm waiting till the timeout is occurred and the thread is stuck at the pausepoint. Without the sleep I will be resetting the pausepoint before it hits.

integration/test_cancel.py
KarthikSubbarao: Each test here is running on a new server, so it is not needed. if you combine the search tests into a single test fn, it is needed

integration/test_cancel.py
KarthikSubbarao: We want to test whether we are able to break out of inf loops (if any exist today due to bugs). We should be able to safely timeout from these operations without causing the module to be stuck / thread to be stuck running

integration/test_cancel.py
```diff
@@ -62,6 +64,110 @@ def search(
...
+
+    client.execute_command("FT._DEBUG", "PAUSEPOINT", "RESET", pausepoint_name)
```
KarthikSubbarao: Let's remove this

integration/test_cancel.py
KarthikSubbarao: We can wait until the blocked client count increments instead of a time.sleep

integration/test_cancel.py
```diff
@@ -62,6 +64,110 @@ def search(
...
+    try:
+        result = client.execute_command(*cmd)
+        if expect_timeout:
+            assert False, f"Expected timeout but got result: {result}"
+        return result
+    except ResponseError as e:
+        if not expect_timeout:
+            raise
+        error_msg = str(e)
+        assert "cancelled due to timeout" in error_msg, f"Expected timeout error, got: {error_msg}"
```
KarthikSubbarao: Let's add the full message here. Check that it is timing out from the aggregate path

integration/test_cancel.py
```diff
@@ -180,6 +286,192 @@ def test_timeoutCMD(self):
...
+        # Baseline - verify normal operation
+        nominal_hnsw = aggregate(client, "hnsw", False, stages=["LOAD", "1", "@n"])
+        nominal_flat = aggregate(client, "flat", False, stages=["LOAD", "1", "@n"])
+        assert nominal_hnsw[0] == 10
+        assert nominal_flat[0] == 10
+        assert client.info("SEARCH")["search_cancel-timeouts"] == 0
+        
+        # Enable forced timeouts
+        assert client.execute_command("FT._DEBUG CONTROLLED_VARIABLE SET ForceTimeout yes") == b"OK"
+        assert client.execute_command("ft._debug CONTROLLED_VARIABLE SET timeoutpollfrequency 1") == b"OK"
```
KarthikSubbarao: This is not testing the Aggregate Stage based timeout. It is going to force a timeout in the background search and later force the error message in the SendReply before entering the aggregate stage logic

integration/test_cancel.py
KarthikSubbarao: lets remove timeout_ms

integration/test_cancel.py
```diff
@@ -287,3 +562,37 @@ def test_timeoutCME(self):
         self.check_info_sum("search_query_prefiltering_requests_cnt", 6)
         self.check_info_sum("search_test-counter-ForceCancels", 9)
 
+    def test_aggregate_timeout_cluster(self):
+        """Test FT.AGGREGATE timeout handling in cluster mode."""
+        self.execute_primaries(["flushall sync"])
```
KarthikSubbarao, line 567: Not needed. This is on a new test and new server

integration/test_cancel.py
```diff
@@ -180,6 +270,191 @@ def test_timeoutCMD(self):
...
+
+        run_pausepoint_timeout_test(
+            self, "search_composed_predicate", setup,
+            ["FT.SEARCH", "idx", "(@text:word* | @tag:{tag1}) @num:[0 500]", "TIMEOUT", "5000"]
+        )
+
+    def test_aggregate_timeout(self):
+        """Test FT.AGGREGATE timeout handling across all aggregation stages."""
+        client: Valkey = self.server.get_new_client()
+        client.execute_command("FLUSHALL SYNC")
```
KarthikSubbarao, line 408: Not needed. This is on a new test and new server

integration/test_cancel.py
```diff
@@ -287,3 +562,37 @@ def test_timeoutCME(self):
         self.check_info_sum("search_query_prefiltering_requests_cnt", 6)
         self.check_info_sum("search_test-counter-ForceCancels", 9)
 
+    def test_aggregate_timeout_cluster(self):
+        """Test FT.AGGREGATE timeout handling in cluster mode."""
+        self.execute_primaries(["flushall sync"])
+        self.config_set("search.info-developer-visible", "yes")
+
+        client: Valkey = self.new_cluster_client()
+        self.check_info("search_cancel-timeouts", 0)
```
KarthikSubbarao, line 571: What is this check helping with?

integration/test_cancel.py
```diff
@@ -180,6 +270,191 @@ def test_timeoutCMD(self):
...
+        hnsw_index.create(client)
+        flat_index.create(client)
+        hnsw_index.load_data(client, 1000)
+        
+        # Baseline - verify normal operation
+        nominal_hnsw = aggregate(client, "hnsw", False, stages=["LOAD", "1", "@n"])
+        nominal_flat = aggregate(client, "flat", False, stages=["LOAD", "1", "@n"])
+        assert nominal_hnsw[0] == 10
+        assert nominal_flat[0] == 10
+        assert client.info("SEARCH")["search_cancel-timeouts"] == 0
```
KarthikSubbarao, line 425: What is this check helping with?

PR #823, opened 2026-02-26
Compatibility test

boda26:
Adding text compatibility tests with Redis
To run it:
1. in integration/compatibility, run pytest generate_text.py to generate Redis answers as a pickle.gz file
2. in integration, first run export ROOT_DIR=/home/ubuntu/oss/compatibility/valkey-search, then run TEST_PATTERN=test_answers[text-search-answers.pickle.gz] ./run.sh

### DIFF
--- a/integration/compatibility/unsupported_tests.md
+++ b/integration/compatibility/unsupported_tests.md
@@ -56,3 +56,26 @@ Some special cases of the group queries with INORDER and SLOP are not supported 
 
 ## 3. Fuzzy Search with Levenshtein > 1
 Some queries are excluded due to different behavior in fuzzy search. ValkeySearch uses Damerau-Levenshtein to calculate the distance in fuzzy search. The results produced by Valkey is consistent with the Damerau-Levenshtein rules.
+
+## 4. Escaped character search in JSON
+Some queries are excluded due to different behavior in escaped character search. Any punctuation followed by a backslash will be escaped. This is the standard format in both ingesting documents and performing search.
+   
+   Example:
+
+   ```
+   127.0.0.1:7001> ft.create idx on json prefix 1 json: schema $.body as body text
+   OK
+   127.0.0.1:7001> json.set json:1 $ '{"body":"black\\,white"}'
+   OK
+   127.0.0.1:7001> ft.search idx 'black\\,white'
+   1) (integer) 1
+   2) "json:1"
+   3) 1) "$"
+      2) "{\"body\":\"black\\\\,white\"}"
+   127.0.0.1:7001> ft.search idx "black\\\\,white"
+   1) (integer) 1
+   2) "json:1"
+   3) 1) "$"
+      2) "{\"body\":\"black\\\\,white\"}"
+   ```
+


integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
+            print(f"Got exception for Error: '{exc}', Cmd:{cmd}")
+            answer["result"] = {}
+            answer["exception"] = True
+        self.answers.append(answer)
 
-    def _run_test(self, builder_fn, data_set_name, key_type, dialect, inorder=False, slop=False):
+    # helper function to parse the result of Redis FT.EXPLAINCLI
+    # use it to detect parsing differences
+    def parse_explaincli_to_query(self, index_name: str, query: str, dialect: int = 2) -> tuple[str, bool]:
+        import re
```
VoletiRam: Can we move imports to top?

integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
+        except Exception as exc:
+            print(f"Got exception for Error: '{exc}', Cmd:{cmd}")
+            answer["result"] = {}
+            answer["exception"] = True
+        self.answers.append(answer)
 
-    def _run_test(self, builder_fn, data_set_name, key_type, dialect, inorder=False, slop=False):
+    # helper function to parse the result of Redis FT.EXPLAINCLI
+    # use it to detect parsing differences
+    def parse_explaincli_to_query(self, index_name: str, query: str, dialect: int = 2) -> tuple[str, bool]:
```
VoletiRam, line 114: Do you think we can break this function into small functions? We can also move reconstruct and validation parts as separate functions out of this function.

integration/compatibility/data_sets.py
```diff
@@ -110,6 +110,100 @@
             ],
             'price': (0, 50)
         }
+    },
+    # ,.<>{}[]"':;!@#$%^&*()-+=~
```
VoletiRam, line 114: Where is forward and backslash in this list?

If we are going with default list, lets remove this comment

integration/compatibility/generate_text.py
```diff
@@ -28,7 +348,18 @@ def _run_test(self, builder_fn, data_set_name, key_type, dialect, inorder=False,
             key_type: Type of key ("json" or "hash")
             dialect: Query dialect version
         """
-        self.setup_data(data_set_name, key_type)
+        try:
+            info = self.client.execute_command("INFO", "SERVER")
+            # Decode if bytes
+            if isinstance(info, bytes):
+                info = info.decode('utf-8')
+            is_redis = "redis_version" in info
```
VoletiRam: Don't we have `redis_version` in valkey info too for compatibility reason?

```
127.0.0.1:6379> info server
# Server
redis_version:7.2.4
server_name:valkey
valkey_version:255.255.255
valkey_release_stage:dev
```

Try looking for server_name instead

integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
-    def setup_data(self, data_set_name, key_type):
-        """Override to specify text data source."""
+    def setup_data(self, data_set_name, key_type, schema_type):
+        """Override to specify text data source."""        
         self.data_set_name = data_set_name
         self.key_type = key_type
-        load_data(self.client, data_set_name, key_type, data_source='text')
+        self.schema_type = schema_type
+        self.client.execute_command("FLUSHALL SYNC")
+        load_data(self.client, data_set_name, key_type, data_source='text', schema_type=schema_type)
```
VoletiRam, line 91: Why are we passing this down when we already defining them in the object? Should we define load_data in the same class and access necessary data via self.*?

integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
 class TestTextSearchCompatibility(BaseCompatibilityTest):
     TEXT_QUERY_TEST_SEED = 3948
-    MAX_QUERIES = 20
+    MAX_QUERIES = 1000
     ANSWER_FILE_NAME = "text-search-answers.pickle.gz"
 
-    def setup_data(self, data_set_name, key_type):
-        """Override to specify text data source."""
+    def setup_data(self, data_set_name, key_type, schema_type):
+        """Override to specify text data source."""
```
VoletiRam, line 86: Don't we use this method for non text related code?

integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
+        elif len(stack) == 1:
+            reconstructed = reconstruct(stack[0])
+        else:
+            reconstructed = ' '.join(reconstruct(node) for node in stack)
+        
+        # Normalize function that handles Redis query normalization rules
+        def normalize(q):
+            # Remove all whitespace first for easier processing
+            q = re.sub(r'\s+', ' ', q.strip())
+
```
VoletiRam: Try breaking this function. You can also move parse_to_tree out of this function.

integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
+        reconstructed_normalized = normalize(reconstructed)
+        
+        is_valid = original_normalized == reconstructed_normalized
+        
+        return reconstructed, is_valid
+
+    # extract only words from the result of FT.EXPLAINCLI
+    # return a list of words only
+    def extract_words_from_explaincli(self, index_name: str, query: str, dialect: int = 2) -> list:
+        import re
```
VoletiRam: Move imports to top so that we can avoid redundants.

integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
+            # If it's a simple word, add it
+            if re.match(r'^[a-zA-Z0-9_-]+$', cleaned):
+                terms.append(cleaned)
+        
+        return terms
+    
+    # extract only words from the result of FT.EXPLAINCLI
+    # return a list of words only
+    def extract_words_from_query(self, query: str) -> list:
+        import re
```
VoletiRam: Move imports to top so that we can avoid redundants.

integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
+        original_normalized = normalize(query)
+        reconstructed_normalized = normalize(reconstructed)
+        
+        is_valid = original_normalized == reconstructed_normalized
+        
+        return reconstructed, is_valid
+
+    # extract only words from the result of FT.EXPLAINCLI
+    # return a list of words only
+    def extract_words_from_explaincli(self, index_name: str, query: str, dialect: int = 2) -> list:
```
VoletiRam: Are we using this method? I don't see calling function.

integration/compatibility/generate_text.py
```diff
@@ -5,21 +5,341 @@
...
+            
+            # If it's a simple word, add it
+            if re.match(r'^[a-zA-Z0-9_-]+$', cleaned):
+                terms.append(cleaned)
+        
+        return terms
+    
+    # extract only words from the result of FT.EXPLAINCLI
+    # return a list of words only
+    def extract_words_from_query(self, query: str) -> list:
```
VoletiRam: Are we using this method? I don't see calling function.

integration/compatibility/data_sets.py
boda26: This is the Redis punctuation list. Since we are comparing with Redis, I am only including characters that are both in Redis and Valkey.

integration/compatibility/generate_text.py
boda26: Yes, but for text indexes, we also have schema_type (default, nostem), so we have to override it for text here.

integration/compatibility/generate_text.py
boda26: removed

integration/compatibility/generate_text.py
boda26: removed

integration/compatibility/generate_text.py
boda26: load_data is in data_sets.py, which is a different file. I am reusing it here to avoid duplication since the logic is mostly same

integration/compatibility/data_sets.py
```diff
@@ -406,7 +470,7 @@ def make_field_definition(key_type, name, typ, i):
 
 # TODO: stopwords, punctuation, numbers in text field (with periods, float, +/-)
 # TODO: special case with lexers (random lexical chars, prove lexical analyzer work correct) (isolation)
```
VoletiRam: Does this TODO needs to be updated?

integration/compatibility/text_query_builder.py
```diff
@@ -193,4 +209,70 @@ def gen_depth2(vocab: List[str], rng: random.Random) -> str:
...
+    count = rng.randint(1, 3)
+
+    # add extra pair of backslashes to make query work in Valkey
+    result = []
+    for _ in range(count):
+        word = rng.choice(vocab)
+        # Double the backslash: \, -> \\, to match server-side escaping
+        query = word.replace('\\', '\\\\')
+        result.append(EscapedTerm(query))
+    return result
```
VoletiRam: Do we need this replacement?  I would update the punctuation list instead to include two backslashes if the reason for adding this was to bypass preprocessing of single backslash by command line.

integration/compatibility/data_sets.py
```diff
@@ -443,8 +507,15 @@ def compute_text_data_sets(dataset_name, seed=123):
...
-        json_schema_parts.append(f"$.{field} AS {field} TEXT WITHSUFFIXTRIE NOSTEM")
+        match schema_type:
+            case "default":
+                hash_schema_parts.append(f"{field} TEXT WITHSUFFIXTRIE")
+                json_schema_parts.append(f"$.{field} AS {field} TEXT WITHSUFFIXTRIE")
+            case "nostem":
+                hash_schema_parts.append(f"{field} TEXT WITHSUFFIXTRIE NOSTEM")
+                json_schema_parts.append(f"$.{field} AS {field} TEXT WITHSUFFIXTRIE NOSTEM")
+            case _:
+                raise ValueError(f"Unknown index schema type: {schema_type}")
```
VoletiRam: We are essential duplicating both json and hash schema building here. 

I would suggest a refactor of this function and extract schema builder helper
```
def _build_field_schema(field: str, field_type: str, schema_type: str, for_json: bool = False) -> str:
```

I also see scheme related configs are hardcoded. Can we define them like:
```
SCHEMA_FLAGS = {
        "text": {
            "default": "WITHSUFFIXTRIE",
            "nostem": "WITHSUFFIXTRIE NOSTEM",
        },
        "tag": "TAG",
        "numeric": "NUMERIC",
    }
```
This way adding new configs will be easy and scalable

We can then simplify the json vs hash update as
```
 if field_type == "text":
        if schema_type not in SCHEMA_FLAGS["text"]:
            raise ValueError(f"Unknown index schema type: {schema_type}")
        flags = SCHEMA_FLAGS["text"][schema_type]
    else:
        flags = SCHEMA_FLAGS[field_type]
    
    field_def = f"{field} {field_type.upper()} {flags}".strip()
    
    if for_json:
        return f"$.{field} AS {field_def}"
    return field_def
```

Other attribute types can also be easily extended
```
hash_schema_parts = []
json_schema_parts = []

for field in text_fields:
    hash_schema_parts.append(_build_field_schema(field, "text", schema_type, for_json=False))
    json_schema_parts.append(_build_field_schema(field, "text", schema_type, for_json=True))

for field in tag_fields:
    hash_schema_parts.append(_build_field_schema(field, "tag", schema_type, for_json=False))
    json_schema_parts.append(_build_field_schema(field, "tag", schema_type, for_json=True))
```

integration/compatibility_test.py
```diff
@@ -381,14 +383,30 @@ def mark_as_failed(testname):
...
+    if (expected['data_set_name'], expected['key_type'], expected.get('schema_type')) != data_set:
         print("Loading data set:", expected['data_set_name'], "key type:", expected['key_type'])
         client.execute_command("FLUSHALL SYNC")
-        load_data(client, expected['data_set_name'], expected['key_type'])
-        data_set = (expected['data_set_name'], expected['key_type'])
+        schema_type = expected.get('schema_type')
+        if schema_type:
+            load_data(client, expected['data_set_name'], expected['key_type'], schema_type=schema_type)
+        else:
+            load_data(client, expected['data_set_name'], expected['key_type'])
```
VoletiRam: why branching for schema type here? We can just pass directly and let load data handle it inside
```
load_data(client, expected['data_set_name'], expected['key_type'], schema_type=schema_type)
```

integration/compatibility/data_sets.py
boda26: I deleted this since these are implemented

integration/compatibility/text_query_builder.py
boda26: I removed this and added in punctuation list.

integration/compatibility/data_sets.py
boda26: reformatted to simplify schema building

integration/compatibility_test.py
boda26: Combined the branch

integration/compatibility/unsupported_tests.md
```diff
@@ -0,0 +1,52 @@
+# Unsupported Tests
+
+This is the list of all the unsupported tests in text-search compatibility tests.
+
+## 1. Exact Phrase Query
+
+Exact phrase query tests are unsupported because of a bug in Redis exact phrase search.
+Redis sometimes returns a document when the query string is separated in two fields.
+
```
VoletiRam: Lets not point other engine bugs here. We can just say we follow linguistic processing or due to x feature we have, our behavior is different and hence not adding the tests for this.

integration/compatibility/unsupported_tests.md
```diff
@@ -0,0 +1,52 @@
...
+   3) "body"
+   4) "four five six"
+```
+
+## 2. Group Queries with INORDER and SLOP
+
+Some special cases of the group queries with INORDER and SLOP are not supported:
+
+1. **Parsing difference** — Valkey parses some queries differently than Redis. This leads to different search results. The compatibility tests check the parsing and filter out these queries.
+
```
VoletiRam: Do we want to list out what are those parsing differences?

integration/compatibility/unsupported_tests.md
```diff
@@ -0,0 +1,52 @@
...
+2. **Nested OR queries** — Redis has bugs in some of the nested OR queries. The bugs include specific orderings in the words, and the same word appearing in both sides of an OR clause. The queries with bugs are extracted and added into an excluded query set in `generate_text.py`.
```
VoletiRam: Same here

integration/compatibility/unsupported_tests.md
VoletiRam, line 1: I expect the list of tests that are not supported due to valkey way of handling is different and we can explain what is our behavior and we can simply say this may not align with other engines and that is expected. Lets not point any engine names or bugs there.

integration/compatibility/unsupported_tests.md
```diff
@@ -0,0 +1,58 @@
...
+   127.0.0.1:6379> hset hash:2 body "ocean carrot jump quiet build shark onion"
+   (integer) 1
+   127.0.0.1:6379> ft.search idx "(((carrot quiet) | (quiet desert)) ((shark | forest)))" SLOP 2
+   1) (integer) 0
+   ```
+
+   Valkey does not return the document, because "carrot", "shark" is picked from two groups, and the distance is 3, which is larger than 2 in the query
+
+## 3. Fuzzy Search with Levenshtein > 1
+Some queries are excluded due to different behavior in fuzzy search. ValkeySearch uses Damerau-Levenshtein to calculate the distance in fuzzy search. The results produced by Valkey is consistent with the Damerau-Levenshtein rules.
```
VoletiRam, line 58: Please leave a new line at the end for better readability

integration/compatibility/unsupported_tests.md
```diff
@@ -0,0 +1,58 @@
...
+   ```
+
+## 2. Group Queries with INORDER and SLOP
+
+Some special cases of the group queries with INORDER and SLOP are not supported in the test:
+
+1. **Parsing difference** — Some group queries are excluded due to parsing difference. In Valkey, all the group queries are directly parsed as it is, without flattening any groups.
+
+   Example:
+
```
VoletiRam, line 25: Curious what we do about incompatible part of multiple backslash characters? Do we need to mention that in file?

VoletiRam (approved):

integration/compatibility/unsupported_tests.md
boda26: Good catch, I will add this part

Issue #664, opened 2026-02-01
[FEATURE] Size Estimation in Prefix/Suffix Tree


allenss-amazon:
Prefix/Suffix tree implementation should provide sum of positions below each node in the tree. This is required for the query planner to make good choices for performance optimization.

PR #824, opened 2026-02-26
Text Size Estimation

BCathcart:
First stage of solving #664.

Unoptimized implementation of Text size estimation. Will degrade text ingestion performance marginally in the short term.

### DIFF
--- a/.github/benchmark_configs/fts-benchmarks-arm.json
+++ b/.github/benchmark_configs/fts-benchmarks-arm.json
@@ -191,7 +191,7 @@
         "fields": [
           {
             "name": "field1",
-            "size": 0,
+            "size": 25000000,
             "transforms": [
               {
                 "type": "repeated_token",
@@ -207,7 +207,7 @@
         "fields": [
           {
             "name": "field1",
-            "size": 0,
+            "size": 1000000,
             "transforms": [
               {
                 "type": "cyclic_pattern",
@@ -223,7 +223,7 @@
         "generate_fields": {
           "count": 50,
           "prefix": "field",
-          "size": 1000,
+          "size": 100000,
           "transforms": [
             {
               "type": "wikipedia"
@@ -236,7 +236,7 @@
         "fields": [
           {
             "name": "field1",
-            "size": 500,
+            "size": 5000,
             "transforms": [
               {
                 "type": "wikipedia"
@@ -250,7 +250,7 @@
         "generate_fields": {
           "count": 20,
           "prefix": "field",
-          "size": 0,
+          "size": 25,
           "transforms": [
             {
               "type": "random_from_set",
@@ -276,7 +276,7 @@
         "fields": [
           {
             "name": "field1",
-            "size": 0,
+            "size": 10000000,
             "transforms": [
               {
                 "type": "unique_tokens",
@@ -291,7 +291,7 @@
         "fields": [
           {
             "name": "field1",
-            "size": 0,
+            "size": 250000,
             "transforms": [
               {
                 "type": "uuid_tokens",
@@ -306,7 +306,7 @@
         "fields": [
           {
             "name": "field1",
-            "size": 0,
+            "size": 1000000,
             "transforms": [
               {
                 "type": "progressive_prefix",
@@ -322,7 +322,7 @@
         "fields": [
           {
             "name": "keyname",
-            "size": 0,
+            "size": 150,
             "transforms": [
               {
                 "type": "uuid_tokens",
@@ -333,7 +333,7 @@
           },
           {
             "name": "field1",
-            "size": 500,
+            "size": 5000,
             "transforms": [
               {
                 "type": "wikipedia"
@@ -347,7 +347,7 @@
         "generate_fields": {
           "count": 10,
           "prefix": "field",
-          "size": 500,
+          "size": 5000,
           "transforms": [
             {
               "type": "wikipedia"
@@ -360,7 +360,7 @@
         "fields": [
           {
             "name": "field1",
-            "size": 50000,
+            "size": 500000,
             "transforms": [
               {
                 "type": "wikipedia"
@@ -374,7 +374,7 @@
         "fields": [
           {
             "name": "field1",
-            "size": 50000,
+            "size": 500000,
             "transforms": [
               {
                 "type": "stemmable_words",


src/indexes/text.cc
```diff
@@ -261,29 +261,41 @@ std::unique_ptr<indexes::text::TextIterator> FuzzyPredicate::BuildTextIterator(
       std::move(key_iterators), field_mask, require_positions);
 }
 
-// Size apis for estimation
+/* Size APIs for pre-filter or inline filter planning */
+
 size_t TermPredicate::EstimateSize() const {
-  // TODO: Implementation
+  auto iter =
+      text_index_schema_->GetTextIndex()->GetPrefix().GetWordIterator(term_);
```
KarthikSubbarao, line 268: Note: GetWordIterator is seen to be expensive and visible as a portion of the time spent on flamegraphs in perf runs. So, this means the `EstimateSize` is not free. We can think about optimizing  GetWordIterator in the future

src/indexes/text.cc
```diff
@@ -261,29 +261,41 @@ std::unique_ptr<indexes::text::TextIterator> FuzzyPredicate::BuildTextIterator(
       std::move(key_iterators), field_mask, require_positions);
 }
 
-// Size apis for estimation
+/* Size APIs for pre-filter or inline filter planning */
+
 size_t TermPredicate::EstimateSize() const {
```
KarthikSubbarao, line 266: Note: We can now change the search.cc file to use the size instead of hard coded 5000
```
  // TODO: For now, we just reserve a fixed size because text search operators
  // return a size of 0 currently.
  neighbors.reserve(5000);
```

```
      // TODO: Use the qualified_entries size when text indexes return correct
      // size.
      seen_keys.reserve(5000);
```

We can use the `qualified_entries` now. But maybe worth using a min of 5000 and qualified_entries because we have cases where it can be the entire index size...

src/indexes/text.cc
BCathcart: Traversing the tree is going to have some cost of course. If it's wasting time allocating memory and maintaining state for the iteration capabilities, we can add a separate API to fetch a target or count without constructing an iterator if we only care about an exact match, which has crossed my mind as a potential optimization. There are also a couple small optimizations I want to make to the tree traversal that might yield some marginal improvement.

src/valkey_search_options.cc
```diff
@@ -318,13 +318,48 @@ static auto max_term_expansions =
...
+        .WithValidationCallback([](const std::string& value) -> absl::Status {
+          double parsed_value;
+          if (!absl::SimpleAtod(value, &parsed_value)) {
+            return absl::InvalidArgumentError(
+                "Prefiltering threshold ratio must be a valid number");
+          }
+          if (parsed_value < kMinimumPrefilteringThresholdRatio ||
+              parsed_value > kMaximumPrefilteringThresholdRatio) {
+            return absl::InvalidArgumentError(absl::StrFormat(
+                "Prefiltering threshold ratio must be between %.3f and %.1f",
```
BCathcart: No need for .3f here

src/indexes/text/rax/rax.h
```diff
@@ -165,6 +166,7 @@ typedef struct raxNode {
     uint32_t isnull : 1;  /* Associated value is NULL (don't store it). */
     uint32_t iscompr : 1; /* Node is compressed. */
     uint32_t size : 29;   /* Number of children, or compressed string len. */
+    uint32_t subtree_items; // SEARCH
```
KarthikSubbarao, line 169: Is this sufficient? This is the number of keys, right?

src/indexes/text/text_index.h
```diff
@@ -190,6 +192,10 @@ class TextIndexSchema {
   // IndexSchema::stem_text_field_mask_)
   uint64_t stem_text_field_mask_ = 0;
 
+  // We track subtree items if the index has a HNSW field to enable filtering
+  // planning decisions with prefix/suffix text filtering.
+  bool track_subtree_item_counts_ = false;
```
KarthikSubbarao, line 197: Does this mean we do not have size estimation when the index does not have a vector field?

KarthikSubbarao: What will the EstimateSize fn return for all the predicates?

src/valkey_search_options.cc
```diff
@@ -318,13 +318,48 @@ static auto max_term_expansions =
                           kMaximumMaxTermExpansions)  // max limit (100k)
         .Build();
 
+/// Register the "--prefiltering-threshold-ratio" flag
+/// Controls when pre-filtering is used vs inline-filtering for hybrid queries
+constexpr absl::string_view kPrefilteringThresholdRatioConfig{
+    "prefiltering-threshold-ratio"};
```
KarthikSubbarao, line 324: Ok, so you changed it from a static to a config

src/indexes/text/rax/rax.h
BCathcart: It's a rough upper bound on the number of keys that match the prefix. Keys can be double counted. We discussed a while back about this being "good enough" for a first implementation. Depends on the use case how well it'll perform.

src/valkey_search_options.cc
BCathcart: Yes to help with testing but also I think this is good to have operationally

src/indexes/text/text_index.h
BCathcart: This is something I meant to ask you about. Are they used outside of the decision for pre-filtering/filtering? In which case they wouldn't be used if there is no HNSW field. If they are used, right now I think it will return 0 for prefix/suffix estimation but I can make a quick change to make it return the total number of keys in the text index.

testing/rax_wrapper_test.cc
```diff
@@ -516,6 +530,46 @@ TEST_F(RaxTest, PathIteratorAPIs) {
   EXPECT_TRUE(ca_iter.Done());
 }
 
+TEST_F(RaxTest, SubtreeKeyCount) {
+  AddWords({{"c", 0},
```
KarthikSubbarao, line 534: What is 0? is that a unique ID?

BCathcart: It's just the value of that key in the tree which has no real meaning in this test

src/indexes/text/rax/rax.h
KarthikSubbarao: We can use u64 for keys

src/indexes/text/text_index.h
KarthikSubbarao: It is ideal to have a non 0 size returned which is a rough estimate. We use the size in the non vector queries in AND for choosing one of the terms and also in the vector reserve. You can fix this in a follow up

testing/rax_wrapper_test.cc
KarthikSubbarao: Let's document that this is duplicated across fields

PR #825, opened 2026-02-26
Fix for ingestion scenarios datasets getting truncated to 0

Aksha1812:
Fixing perf json, earlier had set size : 0 which was causing the data generated to truncate


VoletiRam (approved):
Small Change. Approving.

PR #827, opened 2026-02-26
adding the commented out tests

Nivesh-01:
Adding the tests that were commented out because of a crash


testing/integration/stability_test.py
```diff
@@ -377,30 +377,30 @@ def tearDown(self):
             ),
         ),
         # THIS TEST FAILS BECAUSE OF A KNOWN CRASH
-        # dict(
```
KarthikSubbarao: Was this the num ops crash?

If yes, where is the change to stop execution of EXPIRE commands? Or even better, use a very large expiration TTL (>1 day)

testing/integration/stability_test.py
```diff
@@ -280,30 +280,30 @@ def tearDown(self):
...
```
KarthikSubbarao: Was this the num ops crash?

1. If yes, where is the change to stop execution of EXPIRE commands? Or even better, use a very large expiration TTL (>1 day)

testing/integration/stability_test.py
```diff
@@ -474,30 +474,30 @@ def tearDown(self):
...
```
KarthikSubbarao: Was this the num ops crash?

If yes, where is the change to stop execution of EXPIRE commands? Or even better, use a very large expiration TTL (>1 day)

testing/integration/stability_test.py
KarthikSubbarao: 2. Please test this locally and share if there are any other crashes seen after addressing 1

testing/integration/stability_test.py
Nivesh-01: done

testing/integration/stability_test.py
Nivesh-01: ttl > 1 done. There do not seem to be any crashes seen currently, running more of the tests. There is the socket error that we are getting which is leading to a clusterdown error but it is a bug not a crash.

testing/integration/stability_test.py
Nivesh-01: done

testing/integration/stability_runner.py
```diff
@@ -367,6 +367,8 @@ def run(self) -> StabilityRunResult:
             " --json-out-file"
             f" {memtier_output_dir}/{self.config.index_name}_memtier_del.json"
         )
+        # Use 2-day TTL (172800 seconds) to avoid crash bug with short expiration times
+        # while still testing expiration functionality
```
KarthikSubbarao: You can remove the comment now

testing/integration/stability_test.py
```diff
@@ -279,31 +279,32 @@ def tearDown(self):
...
-        #         test_time_sec=60,
-        #         test_timeout=120,
-        #         keyspace_size=1000000,
-        #         use_coordinator=True,
-        #         replica_count=1,
-        #         repl_diskless_load="swapdb",
-        #     ),
-        # ),
+        dict(
+            testcase_name="text_with_backfill_coordinator_replica",
```
KarthikSubbarao, line 283: Can you share the logs for this run?

testing/integration/stability_runner.py
Nivesh-01: done

PR #831, opened 2026-02-27
Fix crash due to lazy expiration in the middle of Search commands

KarthikSubbarao:
Fix crash due to lazy expiration in the middle of Search commands.

I looked into the numops crash for some time this evening. I was able to reproduce it from an integration test with HSET, FT.SEARCH and EXPIRE (no other operations) commands being executed in parallel. The setup needed a primary and a replica.

From what I can tell, the crash only happens when there is an expiration during the workload that would lead to the core trying to propagate the UNLINK command to the replica node. Expiration happens on the primary and the command is sent as UNLINK.

When using a very high TTL in the EXPIRE command, this crash is not encountered.
Primary queues UNLINK command for replication to replica, but also_propagate queue isn't flushed before handleClientsBlockedOnKeys() is called (from beforeSleep) and this leads to the assert.

Understanding the state in the blocked.c will help figure out the next steps

Note: From my tests, at the time of the crash, I have seen at least one client with the blocked flag. e.g.:
```
id=332 addr=127.0.0.1:57310 laddr=127.0.0.1:19311 fd=187 name=*redacted* age=3 idle=0 flags=b capa= db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=28 multi-mem=0 rbs=1024 rbp=570 obl=0 oll=0 omem=0 tot-mem=1716 events=r cmd=FT.SEARCH user=*redacted* redir=-1 resp=2 lib-name=valkey-py lib-ver=6.1.1 tot-net-in=8090 tot-net-out=89416 tot-cmds=165
```

The crash was happening because there was lazy expiration in the middle of Search based commands.

The Search command would access the key and mark it for lazy expiration (this is the default mode). This is due to how the Module was opening the key when fetching content using `ValkeyModule_OpenKey`.

The key would be added to the also_propagate queue in the core and when we are done with the current command (example: Search), it would crash on the num_ops assert in handleClientsBlockedOnKeys called from blockedBeforeSleep because it already has the UNLINK command from expiration.

I updated how the module opens the key to include flags (e.g. `VALKEYMODULE_OPEN_KEY_NOEXPIRE` or simply `VALKEYMODULE_OPEN_KEY_NOEFFECTS`) to avoid having the searched keys marked for lazy expiration. This prevents the crash: https://github.com/valkey-io/valkey-search/pull/831

In all other sites in the Module we were using `VALKEYMODULE_OPEN_KEY_NOEFFECTS | VALKEYMODULE_READ`, except for the `FetchAllRecords` function.

If we use this `VALKEYMODULE_OPEN_KEY_NOEFFECTS `, we do not encounter the crash. But also, we miss out on the following:
```
VALKEYMODULE_OPEN_KEY_NOTOUCH - Avoid touching the LRU/LFU of the key when opened.
VALKEYMODULE_OPEN_KEY_NONOTIFY - Don't trigger keyspace event on key misses.
VALKEYMODULE_OPEN_KEY_NOSTATS - Don't update keyspace hits/misses counters.
VALKEYMODULE_OPEN_KEY_NOEXPIRE - Avoid deleting lazy expired keys.
VALKEYMODULE_OPEN_KEY_NOEFFECTS - Avoid any effects from fetching the key.
```

So, instead, if we use `VALKEYMODULE_OPEN_KEY_NOEXPIRE`, we will get all other capabilities and avoid crash due to lazy expire. It will have to be expired by other core operations.

I also updated the PR to perform an expiration check (ValkeyModule_GetExpire) and we skip opening the key in the FetchRecords fn

### DIFF
--- a/src/query/response_generator.cc
+++ b/src/query/response_generator.cc
@@ -211,8 +211,12 @@ absl::StatusOr<RecordsMap> GetContentNoReturnJson(
   auto key_str = vmsdk::MakeUniqueValkeyString(key);
   auto key_obj = vmsdk::MakeUniqueValkeyOpenKey(
       ctx, key_str.get(), VALKEYMODULE_OPEN_KEY_NOEFFECTS | VALKEYMODULE_READ);
-  if (!key_obj || ValkeyModule_GetExpire(key_obj.get()) == 0) {
-    return absl::NotFoundError("Key not found or expired");
+  if (!key_obj) {
+    return absl::NotFoundError("Key not found");
+  }
+  mstime_t expire = ValkeyModule_GetExpire(key_obj.get());
+  if (expire != VALKEYMODULE_NO_EXPIRE && expire <= 0) {
+    return absl::NotFoundError("Key expired");
   }
   VMSDK_ASSIGN_OR_RETURN(auto content, attribute_data_type.FetchAllRecords(
                                            ctx, vector_identifier,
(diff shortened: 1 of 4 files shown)


integration/test_fulltext.py
```diff
@@ -2311,6 +2230,113 @@ def test_ft_info_text_index_fields(self):
 
 class TestFullTextCluster(ValkeySearchClusterTestCaseDebugMode):
 
+    @pytest.mark.parametrize("setup_test", [{"replica_count": 1}], indirect=True)
+    def test_concurrent_workload(self, setup_test):
```
BCathcart, line 2235: There are a class of integration tests developing that I don't think should be run on every PR like this one and the one being introduced in #781. I think it's fine for now, but I'll create a follow-up issue to consider marking some to run separately in a daily job.

BCathcart:
This seems like a good fix for the moment but is there generally a problem with the assert or the blocking logic? If a lazy expire happens during the execution of a blocking command, why not flush that to the replication buffer and continue on? I've seen this assert be hit outside of search as well

integration/test_fulltext.py
KarthikSubbarao: Oh yeah, I dont even know if this is reliable on Github runners.  I was going to remove this once we are done with the fix.

Or at least find a less flaky way to repro this without having hundreds of threads to create a timed sequence of operations

PR #832, opened 2026-02-27
Replace blocks with token field in json files

adam-fowler:
This change reduces the number of blocks in the json schema for the ft.aggregate, ft.create and ft.search commands. It replaces blocks that consist of just a pure-token and one other field, with just other field with a `token` field added.

eg 
```json
{
  "name": "SLOP",
  "type": "block",
  "optional": true,
  "arguments": [
    {
      "name": "slop_token",
      "type": "pure-token",
      "token": "SLOP"
    },
    {
      "name": "slop",
      "type": "integer"
    }
  ]
}
```
can be replaced with
```json
{
  "name": "SLOP",
  "token": "SLOP",
  "type": "integer",
  "optional": true
}
```
Both of these generate the same command ie `SLOP value`
It is a good thing to reduce the number of blocks as it reduces the amount of generated code. ie the original json would create a type `Slop` that held an integer value `slop`. The new version just uses an integer.


allenss-amazon:
Can you update the DCO so we can merge this?

allenss-amazon (approved):

PR #836, opened 2026-02-28
fix to be able to query with both alias and identifier

### LINKED ISSUE #482: [BUG] Incorrect syntax scanning for JSON vector indexes
**Describe the bug**
If a vector field is declared on JSON., i.e.,

```FT.CREATE index on json schema $.path as vector HNSW .....```

Then the following search command is rejected.

```FT.SEARCH index "*=[KNN 1 @vector $ref]" ....

Instead the following is accepted:

```FT.SEARCH index "*=[KNN 1 @$.path $ref]" ....

**To Reproduce**

As described above.

**Expected behavior**

It should accept as described above.

**Environment**

- Valkey Search Module Version 1.0
- Valkey Core Version  8.x

**Additional context**
Add any other context about the problem here.

### PR DESCRIPTION
**Pull request by daddaman-amz**
If the fields has an alias and user tries to query with original identifier we throw an error.    

RL behavior silently gives zero results for same case:
```
127.0.0.1:6379> FT.CREATE idx2 ON HASH SCHEMA content as con TEXT NOSTEM category as cat TAG price as pr NUMERIC
OK
127.0.0.1:6379> ft.search idx2 @con:"smart" dialect 2
1) (integer) 1
127.0.0.1:6379> ft.search idx2 @content:"smart" dialect 2
1) (integer) 0
127.0.0.1:6379> ft.search idx2 @cat:{test} dialect 2
1) (integer) 1
127.0.0.1:6379> ft.search idx2 @category:{test} dialect 2
1) (integer) 0
```

Ideal behavior would be to handle both cases from a query which this PR does

### CODE REVIEW

File: src/index_schema.cc [hunk 1]
```diff
@@ -445,6 +458,14 @@ absl::StatusOr<vmsdk::UniqueValkeyString> IndexSchema::DefaultReplyScoreAs(
...
   auto itr = attributes_.find(std::string{attribute_alias});
   if (ABSL_PREDICT_FALSE(itr == attributes_.end())) {
+    // Fallback: try input as an identifier and resolve to alias
+    auto alias_result = GetAlias(attribute_alias);
+    if (alias_result.ok()) {
+      itr = attributes_.find(alias_result.value());
+      if (itr != attributes_.end()) {
+        return itr->second.DefaultReplyScoreAs();
+      }
+    }
```
KarthikSubbarao: Can this be a helped fn? 

Should `GetAttributePositionByAlias` be updated?

daddaman-amz: `GetAttributePositionByAlias` is used only by internal  ingestion function  `UpdateDbInfoKey` and is not user facing. They are using this function with an other function call in case `GetAttributePositionByIdentifier` incase alias is not found.

src/index_schema.cc
```diff
@@ -408,7 +408,10 @@ absl::flat_hash_set<std::string> IndexSchema::GetTextIdentifiersByFieldMask(
 absl::StatusOr<std::string> IndexSchema::GetIdentifier(
```
daddaman-amz, line 408: Used by filter_parser.cc for numeric, tags

ParseNumericPredicate
ParseTagPredicate

src/index_schema.cc
```diff
@@ -441,10 +444,23 @@ absl::StatusOr<std::string> IndexSchema::GetAlias(
...
+// Returns nullptr if the identifier is not found.
+const Attribute *IndexSchema::FindAttributeByIdentifier(
+    absl::string_view identifier) const {
+  auto alias = GetAlias(identifier);
+  if (!alias.ok()) return nullptr;
+  auto itr = attributes_.find(alias.value());
+  return itr != attributes_.end() ? &itr->second : nullptr;
+}
+
 absl::StatusOr<vmsdk::UniqueValkeyString> IndexSchema::DefaultReplyScoreAs(
```
daddaman-amz, line 457: Used in PreParseQueryString (search.cc)

allenss-amazon:
I think the original behavior is correct.

Issue #655, opened 2026-01-31
[TEST] Fix Text Index memory usage sub-module statistics 


allenss-amazon:
** Item not tested **

Fix Text Index memory usage sub-module statistics 

** Testing Method **

TBD

PR #837, opened 2026-02-28
Serialize position map before lock context

Aksha1812:
Moving serialization of position map outside of the text index lock context

### DIFF
--- a/src/indexes/text/text_index.cc
+++ b/src/indexes/text/text_index.cc
@@ -44,7 +44,6 @@ InvasivePtr<Postings> AddKeyToPostings(InvasivePtr<Postings> existing_postings,
     postings = InvasivePtr<Postings>::Make();
   }
 
-  // FlatPositionMap is pre-created and metadata is already updated by caller
   postings->InsertKey(key, flat_map);
   return postings;
 }
@@ -247,15 +246,13 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
                      std::string(token.rbegin(), token.rend()))
                : std::nullopt;
 
-    // Update metadata from PositionMap BEFORE creating FlatPositionMap
-    // This is more efficient than computing counts from FlatPositionMap later
+    // Update metadata from PositionMap
     metadata_.total_positions += pos_map.size();
     for (const auto &[_, field_mask] : pos_map) {
       metadata_.total_term_frequency += field_mask->CountSetFields();
     }
 
-    // Create FlatPositionMap BEFORE the lock context
-    // This serialization work is done outside the critical section
+    // Create FlatPositionMap from PositionMap
     FlatPositionMap *flat_map =
         FlatPositionMap::Create(pos_map, num_text_fields_);
 
@@ -263,7 +260,7 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
     // target_set_fn, so that all trees point to the same postings object
     InvasivePtr<Postings> updated_target;
 
-    // Pass the pre-created FlatPositionMap to AddKeyToPostings
+    // Pass the FlatPositionMap to AddKeyToPostings
     auto target_add_fn = CreateTargetMutateFn(
         metadata_.posting_memory_pool_,
         [&](InvasivePtr<Postings> existing) {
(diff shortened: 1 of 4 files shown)


src/indexes/text/text_index.cc
```diff
@@ -247,26 +247,34 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
                      std::string(token.rbegin(), token.rend()))
                : std::nullopt;
 
+    // Update metadata from PositionMap BEFORE creating FlatPositionMap
+    // This is more efficient than computing counts from FlatPositionMap later
```
BCathcart: Just a nitpick but we're overexplaining the change we're making. I'm not a fan when AI rehashes the prompt in the comments since it doesn't really matter what came before the current code. I'd just leave a short comment here mentioning we're doing these two things outside the locked context and not re-iterate it at every level. The APIs and code speak for themselves.

src/indexes/text/text_index.cc
```diff
@@ -247,26 +247,34 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
...
+
     // The updated target gets set in target_add_fn and later used in
     // target_set_fn, so that all trees point to the same postings object
     InvasivePtr<Postings> updated_target;
 
-    // Note: Right now the memory tracking won't include the position map memory
-    // since it's already allocated and moved into the postings object. Once we
-    // start creating a serialized version instead then it will be tracked. At
-    // that point stop moving the pos_map and just pass a reference so that it
-    // doesn't get cleaned up in the memory scope.
```
BCathcart, line 258: Can we just comment out these text index memory tracking metrics in this PR since we're here and leave a TODO with a link to the follow-up SIM to fix them?

src/indexes/text/text_index.cc
```diff
@@ -247,26 +247,34 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
...
-                                  &metadata_, num_text_fields_);
+          return AddKeyToPostings(std::move(existing), key, flat_map,
+                                  &metadata_);
         },
         updated_target);
     auto target_set_fn = CreateTargetSetFn(updated_target);
 
     // Update the postings object for the token in the schema-level index with
-    // the key and position map
+    // the pre-created FlatPositionMap
```
BCathcart:
```suggestion
    // the key and its serialized position map
```

BCathcart (approved):

src/indexes/text/text_index.cc
```diff
@@ -247,26 +247,34 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
...
+    // This is more efficient than computing counts from FlatPositionMap later
+    metadata_.total_positions += pos_map.size();
+    for (const auto &[_, field_mask] : pos_map) {
+      metadata_.total_term_frequency += field_mask->CountSetFields();
+    }
+
+    // Create FlatPositionMap BEFORE the lock context
+    // This serialization work is done outside the critical section
+    FlatPositionMap *flat_map =
+        FlatPositionMap::Create(pos_map, num_text_fields_);
```
BCathcart, line 257: Btw I don't think we need to do it in this PR, but I feel like we might as well move to a unique_ptr for the ownership semantics as a best practice

src/indexes/text/text_index.cc
Aksha1812: Ok, sorry missed that.

src/indexes/text/text_index.cc
Aksha1812: I don't think I should do that here. I can do a quick separate change if needed. because this will require changing a lot of info related function and some tests. So just in case we need to revert or modify/ update either of these changes, it would be a little hard to bifurcate. I can raise a quick separate PR if needed. Also let's see if we have time before RC3 to fix the sub module statistics, I think they could be fixed enough to atleast used in debug mode. if not removing this code will take few minutes whenever we need, also it is not affecting any other code yet. Will keep a note of this and make sure I do it before RC3 if https://github.com/valkey-io/valkey-search/issues/655 this issue not closed

BCathcart: Yes we'll need to update the info tests but I think this is a 15 minute change. I don't think it's high priority to fix these before the release, but if you have time they would be helpful of course. I'm OK waiting but I don't want this to slip through and have issues being raised from people that tested it out, looked at the memory statistics, and are now thinking we have a broken product all because we didn't take 15 minutes to comment out broken metrics.

PR #839, opened 2026-02-28
Quick fix/ perf improvement for Field Mask Create

Aksha1812:


src/indexes/text/posting.cc
```diff
@@ -37,131 +34,42 @@ class FieldMaskImpl : public FieldMask {
   size_t MaxFields() const override { return num_fields_; }
 
  private:
-  MaskType mask_;
+  uint64_t mask_{0};
```
BCathcart: Are we trading space performance for potential time performance by always using an 8-byte field mask now? We're already lagging on our space goals so I'm not sure this is the direction we want to go

Aksha1812: So when i asserted the size of each field mask template each of them were 24 bytes. 8 bytes for the field mask . (even for smaller field masks it took 8 bytes due to alignment) + 8 bytes for storing num_text_fields (1 byte + 7 byte padding)  + 8 bytes for vtable ptr used because we determined FieldMask type at runtime. If we want to actually just consume as many minimal bytes as the fieldmask requires we would have to allocate memory and variable length encode it or have different structs and bit pack it as far as i understand. So I'll be removing the Impl class as well and bit packing the uint64_t fieldmask and num_text_fields  so it takes only 9 bytes always (though i read that CPU might need extra cycles for this and might be slower on some CPUs , so need suggestion on this) .  And all of this would add onto the stack and not worth the effort to minimize the memory consumption on this any further than i would be reducing with this PR. also currently lot of setfield(), asuint64() , countsetfields() function calls while ingestion which would be simplified if no multiple templates. Also I think me and @allenss-amazon had discussed that this may need an update since current implementation not the most efficient.

src/indexes/text/posting.cc
```diff
@@ -10,158 +10,49 @@
...
-}
+// FieldMask Implementation
 
-// Initialize field mask with specified field count
-template <typename MaskType, size_t MAX_FIELDS>
-FieldMaskImpl<MaskType, MAX_FIELDS>::FieldMaskImpl(size_t num_fields)
-    : num_fields_(num_fields) {
-  CHECK(num_fields <= MAX_FIELDS)
-      << "Field count exceeds maximum for this mask type";
+FieldMask::FieldMask() : mask_(0), num_fields_(1) {}
```
BCathcart: Maybe default to 0 fields?

src/indexes/text/posting.cc
```diff
@@ -10,158 +10,49 @@
...
-  }
-
-  if constexpr (std::is_same_v<MaskType, EmptyFieldMask>) {
-    return true;  // Single field case: presence of object implies field is set
-  } else {
-    return (mask_ & (MaskType(1) << field_index)) != 0;
-  }
+bool FieldMask::HasField(size_t field_index) const {
+  if (field_index >= num_fields_) return false;
+  return (mask_ & (1ULL << field_index)) != 0;
```
BCathcart: nit: `!= 0` is redundant here and could be dropped

src/indexes/text/posting.cc
```diff
@@ -10,158 +10,49 @@
...
-void FieldMaskImpl<MaskType, MAX_FIELDS>::SetAllFields() {
-  if constexpr (std::is_same_v<MaskType, EmptyFieldMask>) {
-    // No-op: field is already implicitly set by object presence
-  } else if (num_fields_ == MAX_FIELDS && MAX_FIELDS == 64) {
-    mask_ = ~MaskType(
-        0);  // Special case: avoid undefined behavior for 64-bit shift
-  } else {
-    mask_ = (MaskType(1) << num_fields_) - 1;  // Set num_fields_ bits
-  }
+void FieldMask::SetAllFields() {
```
BCathcart: Where is this used?

src/indexes/text/posting.cc
```diff
@@ -10,158 +10,49 @@
...
 }
 
-// Clear all field bits
-template <typename MaskType, size_t MAX_FIELDS>
-void FieldMaskImpl<MaskType, MAX_FIELDS>::ClearAllFields() {
-  if constexpr (!std::is_same_v<MaskType, EmptyFieldMask>) {
-    mask_ = MaskType{};  // Zero-initialize
-  }
-}
+void FieldMask::ClearAllFields() { mask_ = 0; }
```
BCathcart: Same question here

src/indexes/text/posting.cc
```diff
@@ -10,158 +10,49 @@
...
-  } else if constexpr (std::is_same_v<MaskType, uint8_t>) {
-    return __builtin_popcount(
-        static_cast<unsigned int>(mask_));  // Count bits in uint8_t
-  } else if constexpr (std::is_same_v<MaskType, uint64_t>) {
-    return __builtin_popcountll(mask_);  // Count bits in uint64_t
-  } else {
-    CHECK(false) << "Unsupported mask type for CountSetFields";
-  }
-}
+size_t FieldMask::CountSetFields() const { return __builtin_popcountll(mask_); }
```
BCathcart, line 33: How about storing this value in the FieldMask so we don't have to recompute? We have 7 bytes of memory we could be using. Although I guess since we are deconstructing this object after the "staging" phase so we'd only call this once. It probably doesn't matter either way much from a performance perspective then

src/indexes/text/posting.cc
BCathcart: I forgot we are serializing the field masks into a different form in the flat position map, so I'm not very worried about the temporary space usage while we stage them. This change LGTM

src/indexes/text/posting.cc
Aksha1812: yes either way it is computed only once so doesn't matter.

PR #843, opened 2026-03-03
Index locking optimization during ingestion.

VoletiRam:
Previously, all concurrent writes to the text index were serialized under a single mutex that covered the entire token processing loop including Postings object updates and radix tree structural changes. This meant multiple writer threads indexing completely independent words would still block each other, providing no concurrency benefit even when there was no actual data contention.

This PR replaces that coarse lock with a two-level locking scheme. A sharded pool of per-word bucket mutexes (`PostingsMutexPool`) now protects Postings object updates, so threads working on different words proceed in parallel. The radix tree mutex is upgraded to an `absl::Mutex` reader-writer lock, allowing concurrent search reads while write-exclusive access is reserved for inserting a new word or removing the last occurrence of a word. For the common case of updating an existing word's Postings object, no tree lock is needed at all. Reading the tree to check whether a word exists is done under the shared reader lock, which permits concurrent access from all other threads.

A `text-posting-mutex-pool-size` configuration parameter is added to control the pool size. The `Rax::GetTarget` method is added to support non-mutating existence checks. Unit tests cover `GetTarget` correctness and concurrent commit correctness under ASAN. Integration tests verify functional correctness and a high-concurrency stress test validates the implementation under production-like load.

### DIFF
--- a/src/indexes/text/text_index.cc
+++ b/src/indexes/text/text_index.cc
@@ -250,11 +250,8 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
       }
       bool is_new_word = !existing;
 
-      {
-        NestedMemoryScope posting_scope{metadata_.posting_memory_pool_};
-        updated_target =
-            AddKeyToPostings(std::move(existing), key, flat_map, &metadata_);
-      }
+      updated_target =
+          AddKeyToPostings(std::move(existing), key, flat_map, &metadata_);
 
       if (is_new_word) {
         absl::WriterMutexLock tree_lock(&text_index_mutex_);
@@ -316,11 +313,8 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
       }
 
       InvasivePtr<Postings> updated_target;
-      {
-        NestedMemoryScope posting_scope{metadata_.posting_memory_pool_};
-        updated_target =
-            RemoveKeyFromPostings(std::move(existing), key, &metadata_);
-      }
+      updated_target =
+          RemoveKeyFromPostings(std::move(existing), key, &metadata_);
 
       if (!updated_target) {
         absl::WriterMutexLock tree_lock(&text_index_mutex_);


src/indexes/text/text_index.cc
```diff
@@ -329,63 +318,71 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
...
+
+      InvasivePtr<Postings> updated_target;
+      {
+        NestedMemoryScope posting_scope{metadata_.posting_memory_pool_};
+        updated_target =
+            RemoveKeyFromPostings(std::move(existing), key, &metadata_);
+      }
+
+      if (!updated_target) {
+        auto drop_fn = [](void *old_val) -> void * {
```
BCathcart: You can use `CreateTargetSetFn` here. I want to limit the places where InvasivePtr ownership is changing. Also since you added the `MutateWordInIndex` helper, I'd probably move the `CreateTargetSetFn` call into that helper.

src/indexes/text/rax_wrapper.cc
```diff
@@ -75,6 +75,15 @@ void Rax::MutateTarget(absl::string_view word,
              << " (" << strerror(errno) << ")";
 }
 
+void *Rax::GetTarget(absl::string_view word) const {
```
BCathcart: Can we name this `FindTarget`? I want to differentiate it from the `GetTarget` APIs for the iterators, where we're already at the node and simply fetching its target. I think this is a good API to add and is one I already had in mind. Right now the word iterator is always used to find targets, even if we only want a single target. I'd expect it won't make much difference from a performance perspective, but there is a chance it does in which case we should use it elsewhere too and/or optimize the iterator creation.

To match the pattern in the iterators, you could add a `FindPostingsTarget` API so the caller doesn't have to do the CopyRaw. It's not a great pattern to have a separate API for every target type, but I do think limiting the number of sites we translate between `void *` and `InvasivePtr` is important to limit the chances of it being accidentally broken.

src/indexes/text/postings_mutex_pool.h
```diff
@@ -0,0 +1,57 @@
...
+// This pool provides N mutexes shared across all words via
+// hashing. The mutex for a given word is chosen by:
+//
+//   index = absl::HashOf(word) % pool_size
+//
+//
+// Pool size is configured at construction time and does not change. absl::Mutex
+// is neither copyable nor movable, so the pool cannot be resized after
+// construction.
+class PostingsMutexPool {
```
BCathcart: You may want to make this name more generic like `RaxTargetMutexPool`, since there is nothing specifically tying this to the Postings target type. We could also use this for the stem tree if we think it'll be a worthwhile improvement.

src/indexes/text/text_index.cc
```diff
@@ -122,6 +97,18 @@ std::function<void *(void *)> CreateSimpleTargetMutateFn(MutateFn mutate_fn) {
   };
 }
 
+// Applies |fn| to the prefix tree for |word| and, if the index has a suffix
+// tree, to the suffix tree for reverse(word).
+void MutateWordInIndex(TextIndex &index, absl::string_view word,
```
BCathcart: What do you think about defining this as `MutateTarget` on the `TextIndex` object itself?

src/indexes/text.cc
```diff
@@ -205,6 +214,7 @@ std::unique_ptr<indexes::text::TextIterator> TermPredicate::BuildTextIterator(
 std::unique_ptr<indexes::text::TextIterator> PrefixPredicate::BuildTextIterator(
     const std::shared_ptr<indexes::text::TextIndex> &text_index,
     FieldMaskPredicate field_mask, bool require_positions) const {
+  absl::ReaderMutexLock tree_lock(&GetTextIndexSchema()->GetTreeMutex());
```
BCathcart: Do these `BuildTextIterator` and `EstimateSize` functions ever get called outside of the read phase of the reader/writer time slice mutex? I didn't think they did which is why not having locks here previously wasn't a concern. It's good to double check this though

src/indexes/text/text_index.cc
```diff
@@ -259,49 +247,50 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
     // The updated target gets set in target_add_fn and later used in
     // target_set_fn, so that all trees point to the same postings object
     InvasivePtr<Postings> updated_target;
+    {
+      absl::MutexLock word_lock(&postings_mutex_pool_.Get(token));
```
BCathcart: The pattern you have here is a bit different than the original design, but I don't see any issues. It seems good at first glance especially as a first implementation!

BCathcart (commented):
I don't see any major issues after the first pass. My comments are mostly about having clean APIs and naming to make sure it's easily maintainable. There is danger around the translation between our C++ code and the Rax C APIs, so I think keeping things as clean as possible here is important.

src/indexes/text.cc
KarthikSubbarao: +1, It is only used from the DoSearch code path

src/indexes/text/text_index.cc
VoletiRam: Trying to understand the invasive pointers functionality. Here, I believe I am just duplicating the pointer. Instead, I can use CreateTargetSetFn which just adopts, uses, and release back to raw. Is my understanding right?

src/indexes/text/postings_mutex_pool.h
VoletiRam: Got it, will rename the config name etc.

src/indexes/text.cc
VoletiRam: Previously, it is protected at high-level by coarse std::mutex text_index_mutex_. Not sure if it is design or just accidental protection. Now with fine-grained bucket locks, we need explicit reader locks on the search path to prevent rax node reallocation during writes from racing with rax traversal during search.

src/indexes/text/text_index.cc
BCathcart: Yeah I'm really just trying to limit the number of places we call the dangerous InvasivePtr APIs to pass ownership around because I'm worried about refcounting bugs leading to memory leaks/corruption.

This code should simply return a nullptr if the `updated_target` passed in is a nullptr:

```
    // Grab a new reference to the new shared target and pass ownership
    // of it to the tree.
    InvasivePtr<Target> copy = updated_target;
    return static_cast<void *>(std::move(copy).ReleaseRaw());
```

We could make it more clear by returning nullptr explicitly if `updated_target` is a nullptr, like how we check if the `old_val` is not null even though `InvasivePtr<Target>::AdoptRaw(static_cast<InvasivePtrRaw<Target>>(old_val));` should be a no-op if `old_val` is a nullptr.

src/indexes/text.cc
BCathcart: Where was `text_index_mutex_` being acquired on the search path before? The time slice mutex provides read/write phases, so the `text_index_mutex_` was really only to prevent conflicting writes during the write phase. The main thread search evaluation complicates things because that can run during the write phase on the background threads, but that is only evaluating against the per-key indexes and the blocking response logic from Allen and Baswanth should prevent conflicting key updates. It shouldn't need to acquire the `text_index_mutex_`.

src/indexes/text/text_index.cc
VoletiRam: I initially implemented the design doc's caching approach. Some Integration tests failed with data loss, documents were indexed but some words were missing from search results. The failure was deterministic when two documents with overlapping words were indexed concurrently. When a thread checks whether a word is new and caches an insertion, it releases the bucket lock. A second thread then checks the same word and also sees it as new because the first thread's cached insertion hasn't been applied yet. Both threads cache insertions for the same word pointing to different Postings objects. When both cached insertions are applied, one overwrites the other, silently losing the first thread's data. That's why had to hold main tree lock while still having bucket lock for atomicity purpose.

VoletiRam: The problem I described was due to local-cache thread i built. I should've built the shared cached as per design doc.

src/indexes/text.cc
VoletiRam: Okay, I misread about time slice mutex logic with text index mutex and thought reader also acquired it. I will remove reader related mutex as we already have protection at time mutex.

src/indexes/text/text_index.cc
VoletiRam: Do you see value in exploring shared cache as per design peformance wise.  

I wll address the rest of the comments first and explore shared cache to a map as per design doc.

BCathcart: There are three aspects to the coordination in my mind:
1.	Mutually exclusive modifications to the postings object
2.	Mutually exclusive modifications to the trees
3.	Atomic updates to the prefix and suffix tree targets (i.e. the updates to the prefix tree targets need to happen in the same order in the suffix tree)

The design here achieves them as follows:
1.	We lock on the word across the lookup to the target, applying the mutation to it, and putting its new value (if needed) back into the tree.
2.	We use a read/write lock on the text index.
3.	We hold the write lock across the update to the prefix and suffix tree.

The original design achieved them as follows:
1.	A sharded mutex like you have today. However, because we’re leaving all tree modification until the end, we could hash on the target pointer since it won’t change. This may be relevant if hashing on the words is expensive.
2.	We apply all the modifications to each tree at the end sequentially.
3.	Multiple updates to the target of the same word will be cached and not actually applied to the trees until the end. We’re only applying a single target update per word to the trees so a bad interleaving isn’t possible.

My instinct is that these approaches won’t be substantially different. If a word is being added and removed several times, then only updating the tree once at the end in the original design would be beneficial, but that seems like an uncommon edge case. The current approach has potentially many transitions between the read and write phase. There’s a cost to the transition, but also other threads can do some work during the write phase in other parts of the ingestion flow like StageAttributeData, so I don’t know how the thread utilization shapes up in practice.

I think this current approach is simpler to implement and it’s not worth exploring the other one at the moment. We can always spend a day doing a rough implementation of the other approach and test it. That seems like a worthwhile exercise but probably not a high priority right now unless we see evidence in the perf tests that there are some limitations with the current approach that could be solved by it.

src/indexes/text/text_index.h
```diff
@@ -206,6 +219,14 @@ class TextIndexSchema {
   uint64_t GetPositionMemoryUsage() const;
   uint64_t GetTotalTextIndexMemoryUsage() const;
 
+  absl::Mutex &GetTreeMutex() ABSL_LOCK_RETURNED(text_index_mutex_) {
+    return text_index_mutex_;
+  }
+  const absl::Mutex &GetTreeMutex() const
+      ABSL_LOCK_RETURNED(text_index_mutex_) {
+    return text_index_mutex_;
+  }
```
BCathcart: Are these used?

src/indexes/text/text_index.cc
```diff
@@ -259,49 +243,41 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
...
+      InvasivePtr<Postings> existing;
+      {
+        // Tree read lock prevents rax node reallocation racing with FindTarget.
+        absl::ReaderMutexLock tree_read(&text_index_mutex_);
+        existing = text_index_->GetPrefix().FindPostingsTarget(token);
+      }
+      bool is_new_word = !existing;
+
+      {
+        NestedMemoryScope posting_scope{metadata_.posting_memory_pool_};
```
BCathcart: @KarthikSubbarao has a PR to remove this memory tracking which I fully agree with doing for now. You might need to rebase on top.

src/indexes/text/text_index.h
VoletiRam: Will remove it. Thanks.

src/indexes/text/text_index.cc
KarthikSubbarao: Yes, it is here: https://github.com/valkey-io/valkey-search/pull/856

src/indexes/text/text_index.cc
```diff
@@ -329,63 +305,63 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
...
-      suffix_opt.value().get().MutateTarget(
-          reverse_word, target_set_fn,
-          ITEM_COUNT_TRACKING_ENABLED(item_count_op::SUBTRACT));
+    std::string word_str(iter.GetWord());
+    {
+      absl::MutexLock word_lock(&rax_target_mutex_pool_.Get(word_str));
+
+      InvasivePtr<Postings> existing;
+      {
+        absl::ReaderMutexLock tree_read(&text_index_mutex_);
```
KarthikSubbarao, line 311: I am thinking through whether this is needed.

If we are accessing the postings object, the word lock is good enough here

src/valkey_search_options.cc
```diff
@@ -422,6 +422,17 @@ static auto async_fanout_threshold =
         kMaximumAsyncFanoutThreshold)  // max threshold (10k)
         .Build();
 
+constexpr absl::string_view kRaxTargetMutexPoolSizeConfig{
+    "text-rax-target-mutex-pool-size"};
+constexpr uint32_t kDefaultRaxTargetMutexPoolSize{256};
```
KarthikSubbarao, line 460: Should there be a check in the on_update_callback to check that write thread count? 

Edit - worst case threads will just be waiting on the word lock

src/indexes/text.cc
```diff
@@ -304,4 +309,4 @@ size_t FuzzyPredicate::EstimateSize() const {
   // Right now we return the upper bound
   return text_index_schema_->GetTrackedKeyCount();
 }
-}  // namespace valkey_search::query
+}  // namespace valkey_search::query
```
KarthikSubbarao: New line Mising

src/indexes/text/text_index.cc
BCathcart: Other words can share a portion of the same path

KarthikSubbarao: But we are not mutating here, right?

There is a block below which obtains the write lock on the text_index_mutex and then performs the mutation.


Did you mean that even when accessing the postings object, we need to have a lock on the text index?

(my comment here was for both Add and Delete paths)

VoletiRam: Okay, evaluating example "apple" and "application"

Thread A could try walking A ->P-> P but thread B could reshuffle rax to add application so we could hit realloc issue on thread A

KarthikSubbarao: Ack, when I was reading the "Approach 1", I thought we planned on requesting a lock the part of the tree relevant to the token rather than the entire tree. I am wondering if this is feasible for us to implement

KarthikSubbarao (approved):

src/indexes/text/text_index.cc
```diff
@@ -319,62 +296,60 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
...
+            ITEM_COUNT_TRACKING_ENABLED(item_count_op::SUBTRACT));
+        if (stem_text_field_mask_) {
+          empty_words.push_back(word_str);
+        }
+      }
     }
+    iter.Next();
+  }
 
-    // If the postings are now empty, remove from stem tree if it was a parent
```
KarthikSubbarao, line 351: There is a lot of valuable documentation which is deleted. Let's please add them back in the next update here. Check all the files

PR #845, opened 2026-03-04
Improve text.cc locking

BCathcart:
Improving the locking domain for tracked and untracked keys which was added in #744.

### DIFF
--- a/src/indexes/text.cc
+++ b/src/indexes/text.cc
@@ -34,9 +34,10 @@ Text::Text(const data_model::TextIndex &text_index_proto,
 
 absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr &key,
                                      absl::string_view data) {
-  absl::MutexLock lock(&index_mutex_);
   auto result = text_index_schema_->StageAttributeData(
       key, data, text_field_number_, !no_stem_, with_suffix_trie_);
+
+  absl::MutexLock lock(&index_mutex_);
   if (result.ok() && *result) {
     auto [_, succ] = tracked_keys_.insert(key);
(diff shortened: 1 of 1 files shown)


src/indexes/text.cc
```diff
@@ -34,9 +34,10 @@ Text::Text(const data_model::TextIndex &text_index_proto,
 
 absl::StatusOr<bool> Text::AddRecord(const InternedStringPtr &key,
                                      absl::string_view data) {
-  absl::MutexLock lock(&index_mutex_);
   auto result = text_index_schema_->StageAttributeData(
       key, data, text_field_number_, !no_stem_, with_suffix_trie_);
+
+  absl::MutexLock lock(&index_mutex_);
```
KarthikSubbarao, line 40: Makes sense

src/indexes/text.cc
```diff
@@ -74,25 +75,20 @@ absl::StatusOr<bool> Text::ModifyRecord(const InternedStringPtr &key,
...
-    }
-    auto result = text_index_schema_->StageAttributeData(
-        key, data, text_field_number_, !no_stem_, with_suffix_trie_);
-    if (!result.ok() || !*result) {
-      need_remove = true;
-    }
-  }
-  if (need_remove) {
-    [[maybe_unused]] auto res =
-        RemoveRecord(key, indexes::DeletionType::kIdentifier);
```
KarthikSubbarao, line 93: Is this call needed for some tracking?

KarthikSubbarao: Ok, I see you have used the sets as though one made a call to this API

Issue #841, opened 2026-03-02
[Bug] Provide correct serialization range in FT Aggregate based on stage


KarthikSubbarao:
**The problem/use-case that the feature addresses**

Provide correct serialization range in FT Aggregate based on stage

PR #846, opened 2026-03-04
Support GetSerializationRange in Aggregation

KarthikSubbarao:
Support GetSerializationRange in Aggregation https://github.com/valkey-io/valkey-search/issues/841

### DIFF
--- a/integration/test_aggregate_metrics.py
+++ b/integration/test_aggregate_metrics.py
@@ -48,10 +48,11 @@ class TestAggregateMetrics(ValkeySearchTestCaseBase):
         client.execute_command("FT.AGGREGATE", "products", "@rating:[-inf inf]", "LIMIT", "0", "5")
         info = client.info("SEARCH")
         assert int(info["search_agg_limit_stages"]) == initial_limit + 1
-        assert int(info["search_agg_limit_input_records"]) == initial_limit_input + 20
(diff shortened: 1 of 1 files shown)


src/commands/ft_aggregate_parser.h
```diff
@@ -139,6 +145,11 @@ class Stage {
   virtual ~Stage() = default;
   virtual absl::Status Execute(RecordSet& records) const = 0;
   virtual void Dump(std::ostream& os) const = 0;
+  // std::nullopt means this stage doesn't require a specific number of input
```
daddaman-amz, line 148: Does any stage return  `std::nullopt` today?

src/commands/ft_aggregate_parser.h
```diff
@@ -68,6 +68,12 @@ struct AggregateParameters : public expr::Expression::CompileContext,
   // Determine if we need full results or if we can optimize with trimming via
   // LIMIT offset & count.
   bool RequiresCompleteResults() const override;
+  //
+  // Number of records required as output of the query phase.
```
daddaman-amz, line 72: Nit: should be Range of records

src/commands/ft_aggregate_parser.h
KarthikSubbarao: Not currently. In the future, with more stages supported, or with improvements done to existing stages in terms of detecting cases, we can likely use this.

PR #853, opened 2026-03-05
Refactor Lexer::Tokenize / Stem / Normalize - reduce string allocations

KarthikSubbarao:
Refactor Lexer::Tokenize / Stem / Normalize -  reduce string allocations

### DIFF
--- a/src/indexes/text/lexer.cc
+++ b/src/indexes/text/lexer.cc
@@ -6,8 +6,8 @@
 
 #include "src/indexes/text/lexer.h"
 
-#include <deque>
 #include <memory>
+#include <vector>
 
 #include "absl/container/flat_hash_map.h"
 #include "absl/log/check.h"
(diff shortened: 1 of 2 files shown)


src/indexes/text/lexer.cc
```diff
@@ -81,17 +82,24 @@ Lexer::Lexer(data_model::Language language, const std::string& punctuation,
...
   }
 
   // Get or create the thread-local stemmer for this lexer's language
   sb_stemmer* stemmer = stemming_enabled ? GetStemmer() : nullptr;
-  std::vector<std::string> tokens;
+  // Deque grows by adding new blocks—avoids the cost of copying
+  // existing elements during reallocation.
+  std::deque<std::string> tokens;
+  std::string word;
+  word.reserve(64);
```
Aksha1812: How do we got to this number. Can you add constant at top so that easy to adjust later?

PR #855, opened 2026-03-05
raxMutate() optimization

BCathcart:
Maintained the same behaviour with a single tree traversal. Previously there would be a traversal to get the target, a traversal to change the target if needed, and a traversal to update the item tracking at all the nodes along the path to said target.

### DIFF
--- a/src/indexes/text/rax/rax.c
+++ b/src/indexes/text/rax/rax.c
@@ -1273,67 +1273,7 @@ uint32_t raxGetSubtreeItemCount(rax *rax, unsigned char *s, size_t len) {
     return h->subtree_items;
 }
 
-// /* Atomically mutates the value at the given key by calling the provided
-//  * callback function. The callback receives the current value (NULL if the
-//  * key doesn't exist) and caller context, and returns the new value (NULL
-//  * to delete the key).
-//  * 
-//  * OWNERSHIP: The callback takes ownership of the old value. If returning a
-//  * new value, the callback MUST clean up the old value.
-//  * 
-//  * Returns 1 on success, 0 on error (errno will be set to ENOMEM
-//  * on out of memory). */
-// // TODO: optimize to a single-pass version that doesn't traverse the tree multiple times.
-// int raxMutate(rax *rax, unsigned char *s, size_t len, raxMutateCallback callback, void *caller_context, item_count_op op) {
-//     void *current_value = NULL;
-
-//     /* Find the current value */
-//     int found = raxFind(rax, s, len, &current_value);
-
-//     /* Call the callback to get the new value */
-//     void *new_value = callback(current_value, caller_context);
-
-//     /* Apply the mutation */
-//     int success = 1;
-//     if (new_value == NULL) {
-//         /* Delete the key if it exists */
-//         if (found) {
-//             success = raxRemove(rax, s, len, NULL);
-//         }
-//         /* Key doesn't exist and callback returned NULL - nothing to do */
-//     } else if (new_value != current_value) {
-//         /* Callback created a new target. Insert or update the key. */
-//         int res = raxInsert(rax, s, len, new_value, NULL);
-//         if (res == 0 && errno != 0) {
-//             /* Actual failure (OOM) */
-//             success = 0;
-//         }
-//     }
-
-//     /* Update subtree_items along the path after tree structure finalized */
-//     if (success && op != NONE) {
-//         raxStack ts;
-//         raxStackInit(&ts);
-//         raxNode *h;
-//         raxLowWalk(rax, s, len, &h, NULL, NULL, &ts);
-
-//         int delta = (op == ADD) ? 1 : -1;
-//         h->subtree_items += delta;
-
-//         raxNode *ancestor;
-//         while ((ancestor = raxStackPop(&ts)) != NULL) {
-//             ancestor->subtree_items += delta;
-//         }
-//         raxStackFree(&ts);
-//     }
-
-//     return success;
-// }
-
 int raxMutate(rax *rax, unsigned char *s, size_t len, raxMutateCallback callback, void *caller_context, item_count_op op) {
-    // Rax low walk to find the node if it exists or where the insertion needs to begin
-    // if it doesn't
-
     raxNode *h, **parentlink;
     int splitpos = 0;
     raxStack ts;
@@ -1343,9 +1283,8 @@ int raxMutate(rax *rax, unsigned char *s, size_t len, raxMutateCallback callback
         raxStackInit(ts_ptr);
     }
 
-    // Find the current value
+    // Search for an existing target
     size_t match_len = raxLowWalk(rax, s, len, &h, &parentlink, &splitpos, ts_ptr, &op);
-    // int exists = !(match_len != len || (h->iscompr && splitpos != 0) || !h->iskey);
     int exists = (match_len == len && (!h->iscompr || splitpos == 0) && h->iskey);
     void *current_value = exists ? raxGetData(h) : NULL;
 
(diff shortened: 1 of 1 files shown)


src/indexes/text/rax/rax.c
```diff
@@ -716,7 +717,7 @@ int raxGenericInsert(rax *rax, unsigned char *s, size_t len, void *data, void **
          * We replace the "h" node with the trimmed node if there is one.
          * If the split position is the first character of the compressed path from h,
          * then there is no trimmed node and we replace h with the split node itself. */
-        splitnode->subtree_items = trimmedlen ? next->subtree_items : h->subtree_items;
+        splitnode->subtree_items = (trimmedlen ? next->subtree_items : h->subtree_items) + 1;
```
Aksha1812: Can you add a comment here as well .
e.g 
```
// SEARCH - inherits existing subtree count plus the newly inserted item
```

src/indexes/text/rax/rax.c
```diff
@@ -412,6 +412,16 @@ raxNode *raxCompressNode(raxNode *n, unsigned char *s, size_t len, raxNode **chi
     return n;
 }
 
+/* BEGIN SEARCH
```
Aksha1812, line 415: Now that we are adding/ modifying/ refactoring code here , unit testing would be useful

src/indexes/text/rax/rax.c
```diff
@@ -922,82 +949,12 @@ int raxFind(rax *rax, unsigned char *s, size_t len, void **value) {
...
- * An empty prefix (len == 0) returns the total key count of the tree. */
-uint32_t raxGetSubtreeItemCount(rax *rax, unsigned char *s, size_t len) {
-    raxNode *h;
-    int splitpos = 0;
-    size_t i = raxLowWalk(rax, s, len, &h, NULL, &splitpos, NULL);
-    if (i < len) return 0;
-    return h->subtree_items;
-}
-
-/* Atomically mutates the value at the given key by calling the provided
```
Aksha1812, line 943: Can we add this comment back

src/indexes/text/rax/rax.c
```diff
@@ -1289,10 +1232,96 @@ int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
...
     }
-    raxStackFree(&ts);
+    // raxStackFree(&ts); - SEARCH moved to caller
     return 1;
 }
 
+/* Remove the specified item. Returns 1 if the item was found and
+ * deleted, 0 otherwise. */
+int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
+    raxNode *h;
```
Aksha1812, line 1246: If you move raxRemove() above raxRemoveHelper() i bet it will not show this code block 
```
 raxNode *h;
    raxStack ts;

    debugf("### Delete: %.*s\n", (int)len, s);
    raxStackInit(&ts);
    int splitpos = 0;
    size_t i = raxLowWalk(rax, s, len, &h, NULL, &splitpos, &ts);
    if (i != len || (h->iscompr && splitpos != 0) || !h->iskey) {
        raxStackFree(&ts);
        return 0;
    }
    if (old) *old = raxGetData(h);
```
    
in the diff. currently looks like lot of deletions and additions in the PR, whereas this very small change. you can rearrange to make the diff and history cleaner. Same goes above with raxGenericInsert() and raxGenericInsertHelper() also raxMutate() for example raxGetSubtreeItemCount has no changes still shows up in the diff

src/indexes/text/rax/rax.c
```diff
@@ -1289,10 +1232,96 @@ int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
...
+    // Search for an existing target
+    size_t match_len = raxLowWalk(rax, s, len, &h, &parentlink, &splitpos, ts_ptr, &op);
+    int exists = (match_len == len && (!h->iscompr || splitpos == 0) && h->iskey);
+    void *current_value = exists ? raxGetData(h) : NULL;
+
+    // The tree became out of sync with the user data if there is no
+    // target and the op is a delete.
+    assert(current_value || op != SUBTRACT);
+
+    // Apply the mutation
```
Aksha1812:
```suggestion
    /* Call the callback to get the new value */
```

src/indexes/text/rax/rax.c
```diff
@@ -1289,10 +1232,96 @@ int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
...
+    void *new_value = callback(current_value, caller_context);
+
```
Aksha1812, line 1311:
```suggestion
/* Apply the mutation */
```

src/indexes/text/rax/rax.c
```diff
@@ -1289,10 +1232,96 @@ int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
...
+    int res;
+    if (new_value && new_value == current_value) {
```
Aksha1812, line 1313: could you structure the if else statements here close to the original. I went through them , the logic seems fine

src/indexes/text/rax/rax.c
```diff
@@ -1289,10 +1232,96 @@ int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
...
+      // Existing target updated. No tree change.
+      res = 1;
+    } else if (new_value) {
+      // New target added.
+      res = raxGenericInsertHelper(rax, s, len, new_value, NULL, 1, h, parentlink, match_len, splitpos);
+      // Check if it's an actual error (OOM)
```
Aksha1812: Earlier we used to update the counts after insert was successful, if we hit this error , would the counts already have been updated before , do we want to reset them?

src/indexes/text/rax/rax.c
```diff
@@ -1289,10 +1232,96 @@ int raxRemove(rax *rax, unsigned char *s, size_t len, void **old) {
...
+      if (errno == 0) {
+        res = 1;
+      }
+    } else if (current_value) {
+      // Old target removed.
+      res = raxRemoveHelper(rax, s, len, NULL, h, ts, match_len, splitpos);
+    } else {
+      // Tree untouched.
+      assert(0);
```
Aksha1812: Not sure what happens here, what happened earlier when current_value and new_value both ==NULL

src/indexes/text/rax/rax.c
BCathcart: I have a `SubtreeKeyCount` test in `rax_wrapper_test.cc` along with a bunch of mutation tests.  Anything particular you think is missing?

src/indexes/text/rax/rax.c
BCathcart: It's common to have the helper above the function that calls it so that it doesn't need to be forward declared. I could be wrong, but I think git blame is smart enough to detect that I simply moved the lines and attribute the code back to the original author.

src/indexes/text/rax/rax.c
BCathcart: Do you have a specific suggestion about how the current one can be approved?

src/indexes/text/rax/rax.c
BCathcart: Good point. The code will already assert higher up if the insertion fails and it is never expected to. I added an assert here to make it more clear.

src/indexes/text/rax/rax.c
BCathcart: This is never expected to happen the way the code is currently used, but from a API perspective we could just return a success response which is what we had before.

src/indexes/text/rax/rax.c
Aksha1812: i was looking at the Update subtree_items block IGNORE

Aksha1812 (approved):

PR #858, opened 2026-03-06
Introduce support for Double configuration variables

allenss-amazon:
A good simulation of having double configuration values in Valkey. They are registered as strings, but the conversion between string a double is handled by the class automatically.


allenss-amazon:
superceded by #1063

PR #859, opened 2026-03-06
WIP - use sorted tokens for cache locality in StageAttributeData

KarthikSubbarao:
In this PR, we change the Tokenize function signature from returning a `std::vector<std::string>` to instead return a vector of a Token struct.

This vector is also sorted by the token's text content in the end of the Tokenize function. The benefit here is the user of this function, StageAttributeData, will see higher cache locality.

The StageAttributeData function previously would loop across a vector of strings where each element could be a different token. Then it would perform a find on a hash map, to obtain the `PositionMap`. These random jumps for every element of the vector of tokens are what I believe is the cause for large amount of page faults in the ingestion perf test scenario 12A (and other similar scenarios). By having the same tokens grouped token and by holding the current PositionMap of the current token in StageAttributeData, I suspect the page faults will reduce significantly. 

We are yet to perf test this.

I also believe the Tokenize function can be improved further based on an approach I describe here: https://github.com/valkey-io/valkey-search/pull/859#discussion_r2898036747

### DIFF
--- a/src/indexes/text/lexer.cc
+++ b/src/indexes/text/lexer.cc
@@ -151,12 +151,11 @@ absl::StatusOr<std::deque<std::string>> Lexer::Tokenize(
       }
 
       // Save original before stemming
-      std::string token_to_store = scratch;  // Save original before stemming
+      std::string token_to_store = scratch;
       if (stemming_enabled) {
-        StemWord(scratch, stemmer, min_stem_size, stem_mappings);
+        UpdateStemMap(scratch, stemmer, min_stem_size, *stem_mappings);
       }
-      tokens.push_back(
-          token_to_store);  // Store original word like original code
+      tokens.push_back(token_to_store);
       scratch.clear();
     }
   }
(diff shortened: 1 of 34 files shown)


src/indexes/text/lexer.cc
```diff
@@ -95,12 +95,11 @@ absl::StatusOr<std::vector<std::string>> Lexer::Tokenize(
 
   // Get or create the thread-local stemmer for this lexer's language
   sb_stemmer* stemmer = stemming_enabled ? GetStemmer() : nullptr;
-  // Deque grows by adding new blocks—avoids the cost of copying
-  // existing elements during reallocation.
-  std::vector<std::string> tokens;
```
KarthikSubbarao, line 99: Note: It should be possible for us to move away from using a string here and instead use a tagged pointer which can operate in two modes:

1) Original - No transformations needed (normalization, escaped char, etc). In this case, we point to the portion of the `absl::string_view text` data received for the entire field's content.

2) Transformed - Due to normalization/escaped char handling, we have a new string allocated for this token. Therefore, we hold the content of this string rather than the original content in the field. The token here is marked using LSB for freeing later on.


This needs some testing and review of the tagged pointed Token struct. I have prototyped it in a separate branch, but don't plan on including this for 1.2 unless we have high confidence in it. See [here](https://github.com/valkey-io/valkey-search/compare/main...KarthikSubbarao:valkey-search:tagptr_tokenization?expand=1) for details.

So, we can return a Vector of Tokens as the result.

The benefits of this will be:
1. Less string allocations when transformations were not done on the token
2. Smaller Token Struct size (less overhead for the overall tokenization process)
3. Faster processing of tokens / sorting at the end of the tokenization

PR #860, opened 2026-03-06
Some tsan fixes

daddaman-amz:
TSAN fixes for:

- zmalloc_used_memory:                    Suppression
- zmalloc_usable:                         Suppression
- sdsFree (related to above) : Fixed by above
- decrRefCount related to Free String :                        Fix 
- malloc_mutex_trylock_final (jemalloc):  Suppression
- data race absl::Mutex::Dtor() - gRPC:   Fix (NoDestructor singleton )
- data race VM_BlockedClientMeasureTimeEnd:   -- Fix (Move the MeasureTimend call when the refer count for underlying blocked count = 0)
- data race ServerCron replicaReceiveRDBFromPrimaryToDisk :  valkey core unrelated to module Suppression

### DIFF
--- a/src/commands/commands.cc
+++ b/src/commands/commands.cc
@@ -67,10 +67,10 @@ void Free([[maybe_unused]] ValkeyModuleCtx *ctx, void *privdata) {
   // Some things can only be cleaned up on the main thread.
   // We need to do this here.
   parameters->index_schema = nullptr;
-  // return_attributes and score_as holds ValkeyModuleStrings retained
-  // from client argv. Must be freed here (main thread) to avoid
-  // racing with freeClientArgv().
+  // return_attributes holds ValkeyModuleStrings retained from client argv.
+  // Must be freed here (main thread) to avoid racing with freeClientArgv().
   parameters->return_attributes.clear();
+  // Cleanup of score_as
   parameters->score_as = nullptr;
   ValkeySearch::Instance().ScheduleSearchResultCleanup(
       [parameters]() { delete parameters; });


src/coordinator/grpc_suspender.h
```diff
@@ -23,8 +24,12 @@ namespace valkey_search::coordinator {
 class GRPCSuspender {
  public:
   static GRPCSuspender& Instance() {
-    static GRPCSuspender instance;
-    return instance;
+    // Use NoDestructor to avoid atexit destruction racing with gRPC
+    // event_engine threads that may still be executing GRPCSuspensionGuard
+    // destructors (which call Decrement() -> MutexLock). The GRPCSuspender
+    // singleton outlives all threads; the OS reclaims memory at process exit.
```
KarthikSubbarao, line 30: Would this be considered a leak by ASAN tests then? @BCathcart  FYA, since you were also looking at this part of the code

vmsdk/src/managed_pointers.h
```diff
@@ -19,7 +19,7 @@ namespace vmsdk {
 
 struct ValkeyStringDeleter {
   void operator()(ValkeyModuleString *str) {
-    ValkeyModule_FreeString(nullptr, str);
+    vmsdk::RunByMain([str]() { ValkeyModule_FreeString(nullptr, str); });
   }
```
KarthikSubbarao, line 23: Triggering a RunByMain call for every ValkeyModuleString seems excessive. Is there any perf impact this causes?

daddaman-amz: Let me recheck the traces which deal with ValkeyModuleString and see if we can reduce the scope.
will check with @VoletiRam how i can run a perf test with and without this

allenss-amazon: Well, the StringFree call isn't thread-safe, it needs to happen from the mainthread. The performance cost of the RunByMain is two malloc/free calls. We could certainly optimize that to remove the malloc/free calls. For example, we could create a deque/vector (with a lock) that background threads could insert a string to free, then during the CRON job do the actual free.

src/coordinator/grpc_suspender.h
daddaman-amz: I didn't see any new memory leak due to this.   Ran some CME integ tests in asan mode

vmsdk/src/managed_pointers.h
daddaman-amz: I have reverted this fix to now more focused fix by clearing `return_attributes`   in `async::Free`

Root cause: `return_attributes` holds `ValkeyModuleString` objects retained from the client's argv (via `RetainUniqueValkeyString`). The race was specifically between the worker thread's `ScheduleSearchResultCleanup` lambda calling `VM_FreeString` on those strings, and the main thread's `freeClientArgv` freeing the same argv strings concurrently after the client unblocks.


We may need to do "RunByMain" in future if we see more signatures like this with `ValkeyModuleString`

src/commands/commands.cc
```diff
@@ -67,6 +67,9 @@ void Free([[maybe_unused]] ValkeyModuleCtx *ctx, void *privdata) {
   // Some things can only be cleaned up on the main thread.
   // We need to do this here.
   parameters->index_schema = nullptr;
+  // return_attributes holds ValkeyModuleStrings retained from client argv.
+  // Must be freed here (main thread) to avoid racing with freeClientArgv().
+  parameters->return_attributes.clear();
   ValkeySearch::Instance().ScheduleSearchResultCleanup(
       [parameters]() { delete parameters; });
 }
```
KarthikSubbarao, line 77: This does not solve the other ValkeyModuleStrings in the SearchResults. I thought we discussed handling this by the following:
1. Short Term: Disable Scheduling
2. Implement a queue system of pending frees of ValkeyModuleStrings (frmo bg threads) to be done on the RunByMain loop

src/coordinator/grpc_suspender.h
BCathcart: The memory leaks reported in the ASAN integ test runs is cut down drastically now. It will be easier to see if this introduces anything new now.

BCathcart: ASAN test results on this PR look good. ASAN shouldn't report it as a leak since it will still be reachable memory at shutdown.

src/commands/commands.cc
daddaman-amz: Disabled scheduling

PR #863, opened 2026-03-07
Fix flakey mrmwmutex test

allenss-amazon:
Test relied on timing, it no longer does that.


vmsdk/testing/thread_pool_test.cc
```diff
@@ -242,67 +242,46 @@ TEST_P(ThreadPoolTest, ConcurrentWorkers) {
...
-#ifdef BROKEN_UNIT_TEST
 TEST_F(ThreadPoolTest, priority) {
-  // Test that high priority tasks are executed before low priority tasks
-  const size_t thread_count = 5;
-  ThreadPool thread_pool("test-pool", thread_count);
-  const size_t tasks = thread_count * 2;
+  // Test that high priority tasks are dequeued (and executed) before low
+  // priority tasks. A single-threaded pool ensures dequeue order matches
+  // execution order, eliminating the race that existed with multiple threads.
+  ThreadPool thread_pool("test-pool", 1);
```
VoletiRam, line 249: Here we are miss testing if priority ordering is respected or not on concurrent threads. Wondering if a high-priority task that arrives while a low-priority task is already running will preempt/precede the next slot or not. Do you see value testing this case?

VoletiRam (approved):
Approved with a small comment

vmsdk/testing/thread_pool_test.cc
VoletiRam: The other way is to record walltime of overall high and low priority tasks and compare them to see if high priority ones done faster.

allenss-amazon: I don't see a need for low prio tasks to get interrupted for high-prio at this time.

Aksha1812:
This should be backported if i am not wrong. FYI : valkey-search would be using valkey-ci-agent for backporting changes . Actively working with Sarthak on it. After that we would just need to PRs to respective project boards with status to be backported and agent will auto raise backport PR. That said we should close all other backporting changes for now if that sounds fine

allenss-amazon:
I've assigned it to the project 1.2 with a to-be-backported status. I believe that's what's required.

sarthakaggarwal97:
> I've assigned it to the project 1.2 with a to-be-backported status. I believe that's what's required.

Yes! Thank you. Example - https://github.com/sarthakaggarwal97/valkey-search/pull/3
I will merge my changes today, hoping it will look better for search tom!

Aksha1812:
@allenss-amazon I think we might need to backport this in 1.0 and 1.1 as well if i am not wrong. Respective boards below

- 1.0 - https://github.com/orgs/valkey-io/projects/36/views/1
- 1.1 - https://github.com/orgs/valkey-io/projects/34/views/1

Issue #606, opened 2026-01-13
[FEATURE] Add Regex to FT._LIST command


allenss-amazon:
**The problem/use-case that the feature addresses**

When the number of indexes gets large, it's hard to deal with it on the CLI.

**Description of the feature**

Allow optional regex to filter the returned list of indexes.

**Alternatives you've considered**

doing nothing.

**Additional information**

PR #864, opened 2026-03-08
Implement Regex to FT._LIST command

bandalgomsu:
Implement Regex to FT._LIST command to visibility

issue : #606

### DIFF
--- a/src/commands/ft_list.cc
+++ b/src/commands/ft_list.cc
@@ -5,13 +5,17 @@
  *
  */
 
+#include <memory>
 #include <string>
 
 #include "absl/container/flat_hash_set.h"
 #include "absl/status/status.h"
-#include "absl/strings/string_view.h"
+#include "absl/strings/ascii.h"
+#include "absl/strings/str_cat.h"
+#include "re2/re2.h"
 #include "src/commands/commands.h"
 #include "src/schema_manager.h"
+#include "vmsdk/src/type_conversions.h"
 #include "vmsdk/src/utils.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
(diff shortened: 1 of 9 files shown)


allenss-amazon:
Love the idea. It's too late for 1.2 release, but this is definitely something to include in the next release.

KarthikSubbarao:
I tagged the 1.3 label, this is a very nice improvement to the FT._LIST command

src/commands/ft_list.cc
```diff
@@ -5,28 +5,58 @@
...
     return absl::InvalidArgumentError(vmsdk::WrongArity(kListCommand));
   }
+
+  std::optional<std::regex> filter_regex;
+  if (argc == 3) {
+    if (absl::AsciiStrToUpper(vmsdk::ToStringView(argv[1])) != "REGEX") {
+      return absl::InvalidArgumentError(vmsdk::WrongArity(kListCommand));
+    }
+    try {
+      filter_regex.emplace(std::string(vmsdk::ToStringView(argv[2])));
```
allenss-amazon: std::regex suffers from "Catastrophic backtracing", meaning that certain patterns can cause exponential explosion in run-times. Could we replace std::regex with an alternative library that doesn't have this issue?

https://github.com/hanickadot/compile-time-regular-expressions was suggested and seems easy to integrate (header only) and license compatible.

bandalgomsu: `Compile Time Regular Expression` seems like it may require the regex pattern to be fixed at compile time, so it might not be the best fit for our case.

would [re2](https://github.com/google/re2) be a better alternative here?

allenss-amazon: Yes, the description of re2 addresses my concerns head-on.

Aksha1812:
**Reviewers for this PR**

src/commands/ft_list.cc
```diff
@@ -5,28 +5,60 @@
...
   absl::flat_hash_set<std::string> names =
       SchemaManager::Instance().GetIndexSchemasInDB(
           ValkeyModule_GetSelectedDb(ctx));
-  ValkeyModule_ReplyWithArray(ctx, names.size());
-  for (const auto &name : names) {
+
+  absl::flat_hash_set<std::string> filtered_names;
+  if (filter_regex != nullptr) {
+    for (const auto &name : names) {
+      if (RE2::PartialMatch(name, *filter_regex)) {
```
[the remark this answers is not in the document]
mohanrajendran: This is matching the documented behavior.

> - `REGEX <pattern>` (optional): Returns only indexes whose name partially matches the regular expression pattern. Use anchors such as `^` and `$` for exact-name matching.

mohanrajendran (approved):

Issue #798, opened 2026-02-20
[FEATURE] Support for lua/multi on single slot indexes when properly routed.


allenss-amazon:
**The problem/use-case that the feature addresses**

Currently, in CME, FT.SEARCH and FT.AGGREGATE are blocked from multi/lua context due to deadlock issues. However, if the search is known to be local, i.e., it's a single slot index and the current node owns that slot -- then the query can be executed safely.

**Description of the feature**

A description of what you want to happen.

**Alternatives you've considered**

Any alternative solutions or features you've considered, including references to existing open and closed feature requests in this repository.

**Additional information**

Any additional information that is relevant to the feature request.

PR #866, opened 2026-03-09
Refactor support for lua/multi on single slot indexes

bandalgomsu:
Enabled FT.SEARCH/FT.AGGREGATE in MULTI/EXEC and Lua for CME only when the index is single-slot (hash-tagged) and the current node owns that slot

issue : #798

### DIFF
--- a/src/commands/commands.cc
+++ b/src/commands/commands.cc
@@ -81,6 +81,22 @@ void Free([[maybe_unused]] ValkeyModuleCtx *ctx, void *privdata) {
 CONTROLLED_BOOLEAN(ForceReplicasOnly, false);
 DEV_INTEGER_COUNTER(stats, single_slot_queries);
 
+bool IsSingleSlotQueryRoutedToLocalNode(ValkeyModuleCtx *ctx,
+                                        const QueryCommand &parameters) {
+  if (!parameters.index_schema) {
+    return false;
+  }
+
+  const auto &single_slot_number =
+      parameters.index_schema->GetSingleSlotNumber();
+  if (!single_slot_number.has_value()) {
+    return false;
+  }
+
+  const auto cluster_map = ValkeySearch::Instance().GetOrRefreshClusterMap(ctx);
+  return cluster_map && cluster_map->IOwnSlot(*single_slot_number);
+}
+
 std::vector<vmsdk::cluster_map::NodeInfo> ComputeSearchTargets(
     ValkeyModuleCtx *ctx, const QueryCommand &parameters) {
   auto mode = /* !vmsdk::IsReadOnly(ctx) ? query::fanout::kPrimaries ? */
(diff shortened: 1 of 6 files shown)


src/commands/commands.cc
```diff
@@ -76,6 +76,18 @@ void Free([[maybe_unused]] ValkeyModuleCtx *ctx, void *privdata) {
 CONTROLLED_BOOLEAN(ForceReplicasOnly, false);
 DEV_INTEGER_COUNTER(stats, single_slot_queries);
 
+bool IsSingleSlotQueryRoutedToLocalNode(ValkeyModuleCtx *ctx,
```
allenss-amazon, line 84: Let's not recompute whether this Index is single-slot on every query access. Rather, IndexSchema should have an "std::optional<uint16_t> single_slot_number" that is setup when the index is created (or recreated from an rdb load). If that optional is populated, then it will be the slot number for the single-slot index.

bandalgomsu: thanks for good suggestion :+1:  
I will apply it.

src/commands/commands.cc
```diff
@@ -136,7 +153,8 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
                            ValkeySearch::Instance().IsCluster() &&
                            !parameters->local_only;
 
-    if (ABSL_PREDICT_FALSE(inside_multi_exec && do_fanout)) {
+    if (ABSL_PREDICT_FALSE(inside_multi_exec && do_fanout) &&
```
yairgott, line 161: Can you please remind me why MULTI/EXEC or Lua script are not supported in CME mode?

allenss-amazon: Deadlock/Timeout. If node A fans out a request to node B and simultaneously node B sends a request to node A. then both A & B have mainthreads that are stuck in LUA and can't service the content requests from remote. 

There is a solution to this problem, which I think we should consider, which is that when executing a fanout from within a LUA or multi that instead of doing the traditional blocking (which causes a busy loop in the core) that we actually do our own busy-waiting, and explicitly allow remote content fetching queries while in that state. 

This eliminates the deadlock, but exposes the local database in the middle of a LUA/multi that isn't normally visible. But since you're issuing the query, you're making it visible at that point. So I think the query problem is simply a documentation issue.

I think this should be put on the queue as a feature for the next release.

allenss-amazon (commented):
If you can resolve the conflicts and update the documentation we can merge this.

src/commands/commands.cc
```diff
@@ -141,7 +158,8 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
+        !IsSingleSlotQueryRoutedToLocalNode(ctx, *parameters)) {
       return absl::InvalidArgumentError(
           "MULTI/EXEC or Lua script are not supported in CME mode.");
```
allenss-amazon, line 164: Need to update this message.

allenss-amazon (approved):
Can we fix the message in a separate PR ?

Issue #870, opened 2026-03-09
[BUG] Numeric Index having more memory after stabalisation than during save.


Nivesh-01:
**Describe the bug**
Numeric Index unlike other index untilises 10% extra memory of memory at save.

**To Reproduce**
Run the integration test test_endurance_save_restore. This is being seen in all test runs in the statistics that are provided at the end of each test run. You can also create a cgroup with OOM crash disabled and then use the watch command to see this happening in action.

**Expected behavior**
The memory consumption after restore should be about the same as after save (before restore).

**Environment**
Just the normal testing environment nothing extra needed

These are the statistics for memory observed when the test is ran with cgroups and we are watching the RAM memory consumption.
```
before save - 2.3 
after save - 2.3
after restore - 3.5
after stabalisation (Once enough time has passed for the mutation queue to release backlog stress) - 2.8

```

Issue #869, opened 2026-03-09
[BUG] - Extremely high overhead memory during restore of indexes


Nivesh-01:
**Describe the bug**
We have an extremely high memory overhead during restore. Tag and Text having as high as 25% overhead of the max memory and Numeric having about 33% overhead of max memory. This is bad behavior which will lead the machine to go to swap when we have a restore at max memory.

**To Reproduce**
Run the integration test test_endurance_save_restore. This is being seen in all test runs in the statistics that are provided at the end of each test run. You can also create a cgroup with OOM crash disabled and then use the watch command to see this happening in action.

**Expected behavior**
We should not have such a high overhead of memory during restore. It should be close to the memory usage we had during save before the restore was initiated.

**Environment**
Just the normal testing environment nothing extra needed

These are the statistics for memory observed when the test is ran with cgroups and we are watching the RAM memory consumption.
```
before save - 2.3 
after save - 2.3
after restore - 3.5
after stabalisation (Once enough time has passed for the mutation queue to release backlog stress) - 2.8

```

PR #871, opened 2026-03-09
Reducing mutation queue growth during restore

Nivesh-01:
Reducing mutation queue growth during restore

This "relieves pressure" on the mutation queue by limiting the number of keys added to it during restore. We are seeing an overhead memory upward of ~25% during restore of Text, tag, and numeric Index. To solve this we are batching the numer of keys we add to the mutation queue during restore and check if the mutation queue is above a threshold limit we have set we sleep for one milliseconds. This keeps the mutation queue from growing to large and increasing the overhead memory. This saves the memory from going to swap.
This also mitigates Numeric index extra memory consumption after restore as compared to save.

Solves the bugs:
https://github.com/valkey-io/valkey-search/issues/869
https://github.com/valkey-io/valkey-search/issues/870

### DIFF
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -1485,13 +1485,12 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
       // Apply backpressure if mutation queue is too large
       while (current_queue_size >= max_queue_size) {
         // Use ValkeyModule_Yield to cooperatively yield during loading
-        ValkeyModule_Yield(ctx, VALKEYMODULE_YIELD_FLAG_CLIENTS,
-                           "slow module operation");
+        ValkeyModule_Yield(ctx, VALKEYMODULE_YIELD_FLAG_CLIENTS, nullptr);
         Metrics::GetStats().rdb_restore_backpressure_wait_cycles++;
         current_queue_size = GetMutatedRecordsSize();
 
         // Log periodically to show progress
-        VMSDK_LOG_EVERY_N_SEC(NOTICE, ctx, 2)
+        VMSDK_LOG_EVERY_N_SEC(NOTICE, ctx, 30)
             << "RDB restore backpressure: waiting for mutation queue to drain. "
             << "Queue size: " << current_queue_size << ", Progress: " << i
             << "/" << key_count << " keys loaded";
@@ -1501,8 +1500,7 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
     VMSDK_ASSIGN_OR_RETURN(auto keyname_str, input.LoadString());
     auto keyname = vmsdk::MakeUniqueValkeyString(keyname_str);
     ProcessKeyspaceNotification(ctx, keyname.get(), false);
-    ValkeyModule_Yield(ctx, VALKEYMODULE_YIELD_FLAG_CLIENTS,
-                       "Slow module operation");
+    ValkeyModule_Yield(ctx, VALKEYMODULE_YIELD_FLAG_CLIENTS, nullptr);
     keys_since_last_check++;
 
     // Update restore progress counter


src/index_schema.cc
```diff
@@ -1442,10 +1442,52 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
...
+    // Batch queue size checks to reduce mutex lock overhead
+    // Only check every kQueueCheckBatchSize keys, or when queue was recently full
+    if (keys_since_last_check >= kQueueCheckBatchSize || 
+        current_queue_size >= max_queue_size) {
+      current_queue_size = GetMutatedRecordsSize();
+      keys_since_last_check = 0;
+      
+      // Apply backpressure if mutation queue is too large
+      while (current_queue_size >= max_queue_size) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
KarthikSubbarao: This is going to block the current thread until mutation queue is drained. The better approach would be to yield back to let it continue with other operations until we are ready to continue.

See: https://valkey.io/topics/modules-api-ref/#ValkeyModule_Yield

If we are blocking the main thread, commands / clients will not be processed in this time and can lead to timeouts / unresponsiveness of the server.

So, it would be better to replace this blocking sleep with a call to ValkeyModule_Yield and maintain a timer in this function to ensure we have yielded to the valkey engine's event loop for X duration

src/index_schema.cc
```diff
@@ -1442,10 +1442,52 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
...
+        backpressure_wait_count++;
+        current_queue_size = GetMutatedRecordsSize();
+        
+        // Log periodically to show progress
+        if (backpressure_wait_count % 1000 == 0) {
+          VMSDK_LOG(NOTICE, ctx)
```
KarthikSubbarao: Can we use `VMSDK_LOG_EVERY_N_SEC` instead to avoid spamming the logs?

src/index_schema.cc
KarthikSubbarao: Let us also have a metric regarding this mutation queue related back pressure that is made visible from the INFO SEARCH.

With the YIELD logic, clients will be able to see the metric during this phase

src/index_schema.cc
Nivesh-01: I will log every two seconds.

src/index_schema.cc

src/index_schema.cc
```diff
@@ -1442,10 +1442,52 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
...
+        current_queue_size >= max_queue_size) {
+      current_queue_size = GetMutatedRecordsSize();
+      keys_since_last_check = 0;
+      
+      // Apply backpressure if mutation queue is too large
+      while (current_queue_size >= max_queue_size) {
+        // Use ValkeyModule_Yield to cooperatively yield during loading
+        ValkeyModule_Yield(ctx, VALKEYMODULE_YIELD_FLAG_NONE, nullptr);
+        backpressure_wait_count++;
+        current_queue_size = GetMutatedRecordsSize();
```
KarthikSubbarao, line 1490: Note: `GetMutatedRecordsSize` requests the lock on `mutated_records_mutex_` and this can cycle pretty quickly

src/index_schema.cc
```diff
@@ -1442,10 +1442,52 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
...
+    if (keys_since_last_check >= kQueueCheckBatchSize || 
+        current_queue_size >= max_queue_size) {
+      current_queue_size = GetMutatedRecordsSize();
+      keys_since_last_check = 0;
+      
+      // Apply backpressure if mutation queue is too large
+      while (current_queue_size >= max_queue_size) {
+        // Use ValkeyModule_Yield to cooperatively yield during loading
+        ValkeyModule_Yield(ctx, VALKEYMODULE_YIELD_FLAG_NONE, nullptr);
+        backpressure_wait_count++;
```
KarthikSubbarao: Can this be shown in the INFO metrics? Also, a clearer metric name will be helpful

KarthikSubbarao:
Can we fix the formatting?

src/index_schema.cc
```diff
@@ -1454,11 +1441,61 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
...
+            << "RDB restore backpressure: waiting for mutation queue to drain. "
+            << "Queue size: " << current_queue_size
+            << ", Progress: " << i << "/" << key_count << " keys loaded";
+        
+      }
+    }
+    
     VMSDK_ASSIGN_OR_RETURN(auto keyname_str, input.LoadString());
     auto keyname = vmsdk::MakeUniqueValkeyString(keyname_str);
     ProcessKeyspaceNotification(ctx, keyname.get(), false);
```
KarthikSubbarao, line 1502: In the core, after the rioRead, we have a process events call invoked.

Let's also call the VM_Yield API here. Otherwise, if we are always under the `max_queue_size`, we will never call the yield, so no clients can be processed during this time

src/index_schema.cc
```diff
@@ -1457,11 +1457,61 @@ absl::Status IndexSchema::LoadIndexExtension(ValkeyModuleCtx *ctx,
...
```
BCathcart, line 1502:
```suggestion
    ProcessKeyspaceNotification(ctx, keyname.get(), false);
    ValkeyModule_Yield(ctx, VALKEYMODULE_YIELD_FLAG_CLIENTS, "slow module operation");
```

src/index_schema.cc
BCathcart: This doesn't appear to be an issue from our testing. The lock is only held briefly and the ingestion threads aren't going to hold the lock for a long time either, so there shouldn't be much contention

src/valkey_search_options.cc
```diff
@@ -414,6 +414,20 @@ static auto drain_mutation_queue_on_load =
         .Dev()  // can only be set in debug mode
         .Build();
 
+/// Register the "max-mutation-queue-size-on-restore" parameter
+/// Limit the mutation queue size during RDB restore to prevent memory spikes
+constexpr absl::string_view kMaxMutationQueueSizeOnRestoreConfig{
+    "max-mutation-queue-size-on-restore"};
+constexpr uint32_t kDefaultMaxMutationQueueSizeOnRestore{10000};
+constexpr uint32_t kMinimumMaxMutationQueueSizeOnRestore{100};
```
BCathcart: This seems like a reasonable lower limit and we wouldn't expect to need it to be lower, but operationally when we've hit surprises where my mitigations have been hindered by lower limits like this. I'd make it `1`

BCathcart:
```suggestion
constexpr uint32_t kMinimumMaxMutationQueueSizeOnRestore{1};
```

Nivesh-01: done

src/index_schema.cc
BCathcart: Maybe "key_loading_delay_cycles"? We'll keep this metric as dev so it can always be improved in the future

src/valkey_search.cc
```diff
@@ -325,6 +325,55 @@ static vmsdk::info_field::Integer rdb_save_failure_cnt(
...
+    "rdb", "rdb_restore_total_indexes",
+    vmsdk::info_field::IntegerBuilder().Dev().Computed([]() -> long long {
+      return Metrics::GetStats().rdb_restore_total_indexes;
+    }));
+
+static vmsdk::info_field::Integer rdb_restore_completed_indexes(
+    "rdb", "rdb_restore_completed_indexes",
+    vmsdk::info_field::IntegerBuilder().Dev().Computed([]() -> long long {
+      return Metrics::GetStats().rdb_restore_completed_indexes;
+    }));
```
BCathcart: Why do we need these if we already have the ratio?

BCathcart: Whatever they are dev metrics we can always remove later

src/index_schema.cc
Nivesh-01: done changed the name and added as a info metric - rdb_restore_backpressure_wait_cycles

src/index_schema.cc
Nivesh-01: added in both the places

src/valkey_search.cc
Nivesh-01: The ratio is calculated using this moreover this will give us a the exact number instead of a decimal if we might need it in some debugging case

PR #872, opened 2026-03-09
Removed multiple reverse_token string allocations from MutateTarget()

Aksha1812:
Removed multiple reverse_token string allocations from MutateTarget(). Passing optional string which is just declared once instead , which can be used by multiple MutateTarget calls.

### DIFF
--- a/src/indexes/text/text_index.cc
+++ b/src/indexes/text/text_index.cc
@@ -105,12 +105,12 @@ TextIndex::TextIndex(bool suffix)
 
 void TextIndex::MutateTarget(absl::string_view word,
                              const InvasivePtr<Postings> &target,
+                             const std::optional<std::string> &reverse_word,
                              item_count_op op) {
   auto target_set_fn = CreateTargetSetFn(target);
   prefix_tree_.MutateTarget(word, target_set_fn, op);
-  if (suffix_tree_) {
-    std::string rev(word.rbegin(), word.rend());
-    suffix_tree_->MutateTarget(rev, target_set_fn, op);
+  if (suffix_tree_ && reverse_word.has_value()) {
+    suffix_tree_->MutateTarget(*reverse_word, target_set_fn, op);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/indexes/text/text_index.cc
```diff
@@ -221,11 +221,6 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
     const std::string &token = entry.first;
     auto &[pos_map, suffix] = entry.second;
 
-    const std::optional<std::string> reverse_token =
-        suffix ? std::optional<std::string>(
-                     std::string(token.rbegin(), token.rend()))
-               : std::nullopt;
-
```
BCathcart, line 228: Good catch. What do you think about keeping it and passing the optional to the `MutateTarget()` APIs? We've seen performance impacts from unnecessary memory allocations, so might as well only create the reverse string once

Aksha1812: Currently only needed here 

```
void TextIndex::MutateTarget(absl::string_view word,
                             const InvasivePtr<Postings> &target,
                             item_count_op op) {
  auto target_set_fn = CreateTargetSetFn(target);
  prefix_tree_.MutateTarget(word, target_set_fn, op);
  if (suffix_tree_) {
    std::string rev(word.rbegin(), word.rend());
    suffix_tree_->MutateTarget(rev, target_set_fn, op);
  }
  ```
  
 The string after this change already would be created just once inside the MutateTarget function itself. We can avoid extra 8 bytes of optional pointer by keeping it the way it is in MutateTarget, rather than keeping in CommitKeyData and passing to it. That way it can be used while both adding and deleting. this variable was just being assigned and not being used anywhere. lmk your thoughts

BCathcart: We need to add it into the main text index and per-key tree index so I assume we're calling MutateTarget twice?

Aksha1812: yeah that's true. you are right. I understand what youare saying now. Thanks.

Issue #653, opened 2026-01-31
[TEST] Audit log messages


allenss-amazon:
Log messages need to be audited:

1. Need to ensure that only appropriate data is written to log files.
2. Need to ensure that proper levels are applied to each log message.
3. Need to ensure that log messages are throttled to avoid spamming/swamping the log.

BCathcart:
@Arunsarma07 will do a last audit before we cut RC3

PR #873, opened 2026-03-09
Search module - Audit log messages

### LINKED ISSUE #649: [BUG] Update warn/error logging to not include logs that do not need to logged at this level
Update warn/error logging to not include logs that do not need to logged at this level

Need to review the code base to check all of the warn/error level logs.

This is one of them:
```
{"pid":70992,"role":"replica","timestamp":"30 Jan 2026 12:31:21.001","level":"warning","message":"<search> [notice], tid: 8771510848, /src/index_schema.cc:289: Index schema idx12 dropped from DB 0"}
```

Arunsarma07:
## Summary

This PR aims in auditing the log messages for the Valkey Search Module that includes
- only appropriate data is written to log files.
- proper levels are applied to each log message.
- log messages are throttled to avoid spamming/swamping the log.

Closes #649 , #653

### DIFF
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -1237,7 +1237,7 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
     DrainMutationQueue(detached_ctx_.get());
   }
 
-  VMSDK_LOG(DEBUG, nullptr)
+  VMSDK_LOG(NOTICE, nullptr)
       << "Starting RDB save for index schema: "
       << vmsdk::config::RedactIfNeeded(name_) << " Saving in version "
       << (RDBWriteV2() ? "2" : "1") << " format";
@@ -1265,7 +1265,7 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
       << vmsdk::config::RedactIfNeeded(this->name_)
       << " in DB: " << this->db_num_ << " to RDB";
 
-  VMSDK_LOG(DEBUG, nullptr)
+  VMSDK_LOG(NOTICE, nullptr)
       << "Starting to save " << attributes_.size() << " attributes.";
 
   for (auto &attribute : attributes_) {
@@ -1406,7 +1406,7 @@ absl::Status IndexSchema::SaveIndexExtension(RDBChunkOutputStream out) const {
   size_t key_count = db_key_info_.Get().size();
   VMSDK_RETURN_IF_ERROR(out.SaveObject(key_count));
   rdb_save_keys.Increment(key_count);
-  VMSDK_LOG(DEBUG, nullptr) << "Writing Index Extension, keys = " << key_count;
+  VMSDK_LOG(NOTICE, nullptr) << "Writing Index Extension, keys = " << key_count;
   for (auto &[key, _] : db_key_info_.Get()) {
     VMSDK_RETURN_IF_ERROR(out.SaveString(key->Str()));
   }
@@ -1422,7 +1422,7 @@ absl::Status IndexSchema::SaveIndexExtension(RDBChunkOutputStream out) const {
                                            [](const auto &entry) {
                                              return !entry.second.from_backfill;
                                            });
-  VMSDK_LOG(DEBUG, nullptr)
+  VMSDK_LOG(NOTICE, nullptr)
       << "Writing mutation queue records = " << count
       << " Total queue:" << tracked_mutated_records_.size();
   VMSDK_RETURN_IF_ERROR(out.SaveObject(count));


src/coordinator/metadata_manager.cc
```diff
@@ -457,8 +457,7 @@ void MetadataManager::HandleBroadcastedMetadata(
       [address, this](grpc::Status s, GetGlobalMetadataResponse &response) {
         if (!s.ok()) {
           VMSDK_LOG_EVERY_N_SEC(WARNING, detached_ctx_.get(), 1)
-              << "Failed to get GlobalMetadata from " << address << ": "
-              << s.error_message();
```
KarthikSubbarao, line 461: Is there no way to print a redacted error message here?

src/index_schema.cc
```diff
@@ -1406,7 +1405,7 @@ absl::Status IndexSchema::SaveIndexExtension(RDBChunkOutputStream out) const {
                                            [](const auto &entry) {
                                              return !entry.second.from_backfill;
                                            });
-  VMSDK_LOG(NOTICE, nullptr)
+  VMSDK_LOG(DEBUG, nullptr)
```
KarthikSubbarao: Question: This log was was not including customer data, why was it changed it DEBUG? Is it not worth having as NOTICE

src/index_schema.cc
```diff
@@ -1427,8 +1426,8 @@ absl::Status IndexSchema::SaveIndexExtension(RDBChunkOutputStream out) const {
   VMSDK_RETURN_IF_ERROR(
       out.SaveObject<size_t>(multi_mutations_keys_.Get().size()));
   rdb_save_multi_exec_entries.Increment(multi_mutations_keys_.Get().size());
-  VMSDK_LOG(NOTICE, nullptr) << "Writing Multi/Exec Queue, records = "
-                             << multi_mutations_keys_.Get().size();
+  VMSDK_LOG(DEBUG, nullptr) << "Writing Multi/Exec Queue, records = "
```
KarthikSubbarao, line 1446: Question: This log was was not including customer data, why was it changed it DEBUG? Is it not worth having as NOTICE

src/index_schema.cc
```diff
@@ -1204,8 +1203,8 @@ static absl::Status SaveSupplementalSection(
   rdb_save_sections.Increment();
   auto header = std::make_unique<data_model::SupplementalContentHeader>();
   header->set_type(type);
-  VMSDK_LOG(NOTICE, nullptr) << "Writing supplemental section type "
-                             << data_model::SupplementalContentType_Name(type);
+  VMSDK_LOG(DEBUG, nullptr) << "Writing supplemental section type "
```
KarthikSubbarao, line 1219: Question: Starting from here to the rest of the file,,,, This log was was not including customer data, why was it changed it DEBUG? Is it not worth having as NOTICE

src/query/fanout.cc
```diff
@@ -229,8 +228,7 @@ class LocalResponderSearch : public query::SearchParameters {
       tracker->AddTotalCount(search_result.total_count);
     } else {
       VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
-          << "Error during local handling of FT.SEARCH: "
-          << search_result.status.message();
+          << "Error during local handling of FT.SEARCH";
```
KarthikSubbarao: Is there no way to print a redacted error message here?

src/index_schema.cc
Arunsarma07: The problem is the logs which are marked to DEBUG in this file blow up when we have large number of attributes and indexes. They are printed per attribute, per index. I created a long running test with 200 indexes and 100 attributes and the logs blew each time when there is a `save` happening. I think it is best to keep it to DEBUG unless there is any specific reason for which we need to included them in NOTICE level.

src/query/fanout.cc
Arunsarma07: One way is to see the status error code of each message and filter out or redact based on the code. Initially I had the idea of redacting the message based on the error codes like `NOT_FOUND (5)` error code (specific to Index not found). But there are bunch of error codes and various error messages that may or may not contain sensitive data. To make it easier, for some of the logs I have included just the error code which can help in debugging later.

src/index_schema.cc
KarthikSubbarao: Thanks for the context

src/query/fanout.cc
```diff
@@ -109,8 +109,7 @@ struct SearchPartitionResultsTracker {
       if (status.error_code() != grpc::DEADLINE_EXCEEDED ||
           status.error_code() != grpc::FAILED_PRECONDITION) {
         VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
-            << "Error during handling of FT.SEARCH on node " << address << ": "
-            << status.error_message();
+            << "Error during handling of FT.SEARCH on node " << address;
       }
```
KarthikSubbarao: We can remove the IF statement and also change to DEBUG

src/coordinator/metadata_manager.cc
Arunsarma07: I added the error code here instead of the message.

src/index_schema.cc
```diff
@@ -1407,7 +1406,7 @@ absl::Status IndexSchema::SaveIndexExtension(RDBChunkOutputStream out) const {
   size_t key_count = db_key_info_.Get().size();
   VMSDK_RETURN_IF_ERROR(out.SaveObject(key_count));
   rdb_save_keys.Increment(key_count);
-  VMSDK_LOG(NOTICE, nullptr) << "Writing Index Extension, keys = " << key_count;
+  VMSDK_LOG(DEBUG, nullptr) << "Writing Index Extension, keys = " << key_count;
```
KarthikSubbarao: This is only during Save, why does it have to be DEBUG?

src/index_schema.cc
KarthikSubbarao: This is only during Save, why does it have to be DEBUG?

src/index_schema.cc
```diff
@@ -1266,7 +1265,7 @@ absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
       << vmsdk::config::RedactIfNeeded(this->name_)
       << " in DB: " << this->db_num_ << " to RDB";
 
-  VMSDK_LOG(NOTICE, nullptr)
+  VMSDK_LOG(DEBUG, nullptr)
```
KarthikSubbarao: Let's make this NOTICE for observability in logs

PR #874, opened 2026-03-09
Use vector for storing stem variants

Aksha1812:
The set containing stem variants in most cases is less than 10 elements. Vector is proven to be faster than hash set in these cases. Even though we would have to linearly add/ delete in the stem variant list. since maintaing hash set also involves hash computataion and vector provides better cache locality. Also vector consumes less memory than flat_hash_set.

Changes : 
```
using StemParents = absl::flat_hash_set<std::string>; 

// updated to

using StemParents = std::vector<std::string>;
```

```
absl::node_hash_map<
      Key, absl::flat_hash_map<std::string, absl::flat_hash_set<std::string>>>
      in_progress_stem_mappings_; 
      
// updated to 

absl::node_hash_map<Key, InProgressStemMap> in_progress_stem_mappings_;
 
// where InProgressStemMap is

constexpr size_t kInProgressStemVariantsInlineCapacity = 4;
using InProgressStemMap = absl::flat_hash_map<
    std::string,
    absl::InlinedVector<std::string, kInProgressStemVariantsInlineCapacity>>;       
```

### DIFF
--- a/src/indexes/text/text_index.cc
+++ b/src/indexes/text/text_index.cc
@@ -222,9 +222,9 @@ void TextIndexSchema::CommitKeyData(const InternedStringPtr &key) {
     auto &[pos_map, suffix] = entry.second;
 
     const std::optional<std::string> reverse_token =
-        suffix ? std::optional<std::string>(
-                     std::string(token.rbegin(), token.rend()))
-               : std::nullopt;
+        with_suffix_trie_ ? std::optional<std::string>(
+                                std::string(token.rbegin(), token.rend()))
+                          : std::nullopt;
 
     // Update metadata from PositionMap
     metadata_.total_positions += pos_map.size();


Aksha1812:
### Micro benchmark results for stem_mappings (flat_hash_set vs InlinedVector ) 

#### 3 stems (typical doc) (n=3, 1000000 iterations) 
StageAttributeData Operation               |    HashSet |    InlVec<4> | Speedup
------------------------------------------ | ---------- | ------------ | -------
StageAttributeData: UpdateStemMap          |    305.5 ns |      139.5 ns |   2.2x
StageAttributeData: create + extract map   |    300.8 ns |      143.2 ns |   2.1x
CommitKeyData: iterate stem mappings       |      4.9 ns |        3.6 ns |   1.4x

#### 5 stems (common doc) (n=5, 1000000 iterations) 
StageAttributeData Operation               |    HashSet |    InlVec<4> | Speedup
------------------------------------------ | ---------- | ------------ | -------
StageAttributeData: UpdateStemMap          |    498.4 ns |      231.6 ns |   2.2x
StageAttributeData: create + extract map   |    492.5 ns |      235.3 ns |   2.1x
CommitKeyData: iterate stem mappings       |      7.6 ns |        5.3 ns |   1.4x

#### 10 stems (large doc) (n=10, 1000000 iterations) 
StageAttributeData Operation               |    HashSet |    InlVec<4> | Speedup
------------------------------------------ | ---------- | ------------ | -------
StageAttributeData: UpdateStemMap          |   1057.9 ns |      468.4 ns |   2.3x
StageAttributeData: create + extract map   |   1057.9 ns |      472.2 ns |   2.2x
CommitKeyData: iterate stem mappings       |     12.0 ns |        6.2 ns |   1.9x

#### 26 stems (extreme doc) (n=26, 1000000 iterations) 
StageAttributeData Operation               |    HashSet |    InlVec<4> | Speedup
------------------------------------------ | ---------- | ------------ | -------
StageAttributeData: UpdateStemMap          |   2737.6 ns |     1253.7 ns |   2.2x
StageAttributeData: create + extract map   |   2750.5 ns |     1258.1 ns |   2.2x
CommitKeyData: iterate stem mappings       |     34.8 ns |       10.3 ns |   3.4x

Aksha1812:
Went through the PR again to reverify , merging.

src/indexes/text/rax_wrapper.h
```diff
@@ -47,9 +47,9 @@ query planning.
 
 namespace valkey_search::indexes::text {
 
-// Stem tree target: maps stem root to set of parent words that stem to it
+// Stem tree target: maps stem root to list of parent words that stem to it.
 // Example: "happi" → {"happy", "happiness", "happily"}
-using StemParents = absl::flat_hash_set<std::string>;
+using StemParents = std::vector<std::string>;
```
BCathcart, line 52: @Aksha1812 were you able to measure any improvement for this particular StemParents change? The InProgressStemMap inline vector change definitely makes sense, but my guess is that this StemParents change wouldn't have much impact. It's a contiguous allocation in either case. There are usually only a few small strings per StemParents so lookup using a vector probably isn't noticeably more expensive than lookup in the flat_hash_set. Anyways just curious

PR #875, opened 2026-03-09
Add 10k tag filter test and Change prefiltering-threshold-ratio and text-rax-target-mutex-pool-size to Dev configs

KarthikSubbarao:
Add 10k tag filter test and Change prefiltering-threshold-ratio and text-rax-target-mutex-pool-size to Dev configs

### DIFF
--- a/integration/test_filter_expressions.py
+++ b/integration/test_filter_expressions.py
@@ -905,7 +905,6 @@ class TestFilterExpressions(ValkeySearchTestCaseBase):
     def test_10k_tag_filters(self):
         """
         Test with 10,000 tag filters to verify query complexity limits.
-        This tests the query-string-terms-count configuration limit.
         """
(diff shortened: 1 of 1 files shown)


integration/test_filter_expressions.py
```diff
@@ -901,3 +901,32 @@ def test_deep_or_operator_nesting_depth_4(self):
         assert result[0] >= 1
         keys = set(result[i].decode('utf-8') for i in range(1, len(result)))
         assert "item:9" in keys
+
+    def test_10k_tag_filters(self):
```
BCathcart, line 905: If we're verifying a limit set by the config, should this test also validate that the query is rejected if it's above the limit? It's not ensured by just reading this test that the config does anything

KarthikSubbarao: There is no limit. We wanted to check that there is no limit

PR #876, opened 2026-03-09
Improve Evaluator Locking

BCathcart:
The main change here is to prevent the response generator from holding the lock on the `per_key_text_indexes_` for the entire evaluation. This was useful before the text blocking response logic was implemented between #541 and #742, to prevent the per-key text index from being deconstructed while we were evaluating against it. Now that we wait for any in-flight mutations to be indexed for a key before we evaluate it, we don't need to worry about background ingestion deleting the text index. All we need to lock is the access to the `per_key_text_indexes_` to safely fetch the correct text index.

I also added locking on `per_key_text_indexes_` when its size is fetched for INFO since that can happen during the ingestion phase and tried to simplify the flow a bit by passing around the `text_index_schema` rather than the `per_key_text_indexes_`.

### DIFF
--- a/src/indexes/text/text_index.h
+++ b/src/indexes/text/text_index.h
@@ -235,7 +235,8 @@ class TextIndexSchema {
   }
 
   // Helper function to lookup text index for a key.
-  // Locking needs to be true if called outside of read phase of time sliced mutex.
+  // Locking needs to be true if called outside of read phase of time sliced
+  // mutex.
   const TextIndex *GetPerKeyTextIndex(const Key &key, bool lock);
 
   // TODO: remove this because we'll always track the counts once it's optimized
--- a/testing/filter_test.cc
+++ b/testing/filter_test.cc
@@ -149,8 +149,8 @@ TEST_P(FilterTest, ParseParams) {
 
     // Set up text index for text predicate evaluation
     if (index_schema->GetTextIndexSchema()) {
-      auto text_index =
-          index_schema->GetTextIndexSchema()->GetPerKeyTextIndex(interned_key, false);
+      auto text_index = index_schema->GetTextIndexSchema()->GetPerKeyTextIndex(
+          interned_key, false);
       indexes::PrefilterEvaluator evaluator(
           text_index, parse_results.value().query_operations);
       EXPECT_EQ(test_case.evaluate_success.value(),


src/indexes/text/text_index.h
```diff
@@ -223,40 +223,21 @@ class TextIndexSchema {
...
+  // Locking-enabled version of GetTrackedKeyCount.
+  size_t GetTrackedKeyCount(bool lock) {
+    std::optional<std::lock_guard<std::mutex>> per_key_guard;
+    if (lock) per_key_guard.emplace(per_key_text_indexes_mutex_);
+    return GetTrackedKeyCount();
   }
 
+  // Helper function to lookup text index for a key.
+  // Locking needs to be specified if called outside of read phase.
+  const TextIndex *GetPerKeyTextIndex(const Key &key, bool lock = false);
```
daddaman-amz: Should we make the lock explicit instead of default parameter and also document the reason at each call site whenever lock is not needed?

src/query/search.cc
```diff
@@ -382,13 +374,9 @@ void EvaluatePrefilteredKeys(
     result_keys.reserve(max_keys);
   }
   // Get per-key text indexes directly since we have reader lock
```
daddaman-amz: Nit: Move this comment to the callsite

src/index_schema.cc
```diff
@@ -1058,7 +1058,7 @@ int IndexSchema::GetTextItemCount() const {
     return 0;
   }
   // Count documents that actually have text content indexed
-  return text_index_schema->GetPerKeyTextIndexes().size();
+  return text_index_schema->GetTrackedKeyCount(true);
```
daddaman-amz, line 1061: oh good catch,  so old code used non-locked version of metric.

In worst case of crash with the lock held, the crash report may also hang due to deadlock. Maybe we should try lock mechanism for diagnostic path in future.  what do you think?

src/indexes/text/text_index.h
BCathcart: I think removing the default parameter is way better. This enforces us to explicitly document via code that we are calling it from within the read phase of the reader/writer lock or not. IMO we don't need additional comments after this change since the code is self-documenting, but I could be convinced otherwise

src/index_schema.cc
BCathcart: That's a good point. It's definitely a problem we ran into before. Do you know how common it is to need a lock in order to retrieve INFO fields? I think this is something we could follow-up with post 1.2 GA, assuming the newer code hasn't made this problem significantly more likely to occur

BCathcart: In any case if you could create an issue for it whenever you have time that'd be great

src/indexes/text/text_index.h
daddaman-amz: Nit: Yeah, the only place a comment adds value is when the lock is already acquired higher in the call stack — where it's not immediately obvious from the function alone

PR #878, opened 2026-03-10
Simplify text size estimation for non-vector queries

BCathcart:
Solves the last piece of #664:

> Even if we aren't doing a vector search filtered by TEXT, EstimateSize() is still being called to size some response collections. I expect the cost of traversing the tree will be greater than the benefit from accurately reserving the response collection size and we should instead do a O(1) rough estimation.

- I mention the HNSW specifically, but a proper query planning decision is also relevant for FLAT.

`test_text_size_estimation_prefilter_decision` validates that the query planning still works as expected in the hybrid case. I haven't validated that we avoid the expensive size estimation in the non-vector case since it's only used to size an internal buffer, but we should validate that our perf tests should show no degradation in the non-vector case from what we have now.

### DIFF
--- a/testing/search_test.cc
+++ b/testing/search_test.cc
@@ -248,7 +248,7 @@ TEST_P(EvaluateFilterAsPrimaryTest, ParseParams) {
   UnitTestSearchParameters params;
   params.index_schema = index_schema;
   params.filter_parse_results = std::move(filter_parse_results.value());
-  params.attribute_alias = ""; // Makes is_vec_query = false
+  params.attribute_alias = "";  // Makes is_vec_query = false
   EXPECT_EQ(EvaluateFilterAsPrimary(
                 params, params.filter_parse_results.root_predicate.get(),
                 entries_fetchers, false),


src/indexes/text.cc
```diff
@@ -268,28 +268,38 @@ std::unique_ptr<indexes::text::TextIterator> FuzzyPredicate::BuildTextIterator(
...
+ * bound using the total number of tracked keys since the estimation will only
+ * be used to reserve space in a collection.
  */
 
-size_t TermPredicate::EstimateSize() const {
-  auto iter =
-      text_index_schema_->GetTextIndex()->GetPrefix().GetWordIterator(term_);
-  if (!iter.Done() && iter.GetWord() == term_) {
-    return iter.GetPostingsTarget()->GetKeyCount();
+size_t TermPredicate::EstimateSize(bool is_vec_query) const {
```
KarthikSubbarao, line 279: This approach seems fine assuming the cost of estimating the size is going to be a major impact to performance (we will find out after the perf test run)

src/query/search.cc
```diff
@@ -668,7 +668,7 @@ absl::StatusOr<std::vector<indexes::Neighbor>> DoSearch(
   size_t qualified_entries = EvaluateFilterAsPrimary(
       parameters.filter_parse_results.root_predicate.get(), entries_fetchers,
       false, parameters.filter_parse_results.query_operations,
-      parameters.index_schema.get());
+      parameters.index_schema.get(), parameters.IsVectorQuery());
```
KarthikSubbarao: Observation: EvaluateFilterAsPrimary is being loaded with variables from parameters which will not change during the execution of it. 

These are the same through out the execution, so maybe passing SearchParameters is cleaner
```
parameters.filter_parse_results.query_operations,
      parameters.index_schema.get(), parameters.IsVectorQuery()
```

KarthikSubbarao:
Approved, but would suggest this if possible: https://github.com/valkey-io/valkey-search/pull/878#discussion_r2909126525

Also, clang needs to be fixed

PR #879, opened 2026-03-10
Add dialect 2 for vss hybrid test

VoletiRam:
Add dialect 2 for vss hybrid tests as redis uses it for hybrid test scenarios


Aksha1812 (approved):
Simple change

PR #880, opened 2026-03-10
Update to oom flags on search commands and fix error propagation in coordinator / iscancelled cases

KarthikSubbarao:
The PR does the following:

1. Update to oom flags on search commands. We are marking search commands with the denyoom flags following what was done since 1.0.
2. Updated Error Tracking / Propagation to ensure we do not always return the "timeout" message even we cancel a fanout operation.
3. Updated the normal code path of search commands (commands.cc) to first attempt replying with an error message based on the status. If no error, then check if canceled and return the standard timeout error.

Note: The logic for early cancellation when partial results are disabled is still retained in this PR. If there is a timeout when partial results are disabled, we cancel without waiting and return the timed out error message.

The parts which are new are the following:
We have added handling for the case (with partial results disabled) and one of the fanouts fails with OOM. We now have correct with error propagation for it
It also adds handling for the case (with partial results enabled) where we have not timed out, and all nodes have failed. Since it tracks the first error message, we are able to reply with the error now

### DIFF
--- a/src/query/fanout.cc
+++ b/src/query/fanout.cc
@@ -181,19 +181,14 @@ struct SearchPartitionResultsTracker {
     if (consistency_failed) {
       // Consistency failures always take precedence
       status = absl::FailedPreconditionError(kFailedPreconditionMsg);
-    } else if (!parameters->enable_partial_results && has_node_error.load()) {
-      // When partial results are disabled, use the first error we encountered
-      status = first_node_error;
-    } else if (parameters->enable_partial_results && has_node_error.load() &&
-               !has_successful_node.load()) {
-      // When partial results are enabled but all nodes failed (no successful
-      // responses)
+    } else if (has_node_error.load() && (!parameters->enable_partial_results ||
+                                         !has_successful_node.load())) {
+      // Use first error when:
+      // - Partial results disabled (any error fails the operation), OR
+      // - Partial results enabled but no nodes succeeded (all failed)
       status = first_node_error;
     } else {
       // No errors detected - success case
-      status = absl::OkStatus();
-    }
-    if (status.ok()) {
       std::vector<indexes::Neighbor> neighbors;
       neighbors.resize(results.size());
       size_t i = neighbors.size();
@@ -214,7 +209,6 @@ struct SearchPartitionResultsTracker {
           accumulated_total_count, std::move(neighbors), *parameters, true);
       status = absl::OkStatus();
     }
-
     parameters->search_result.status = status;
     // The destructor runs on whichever thread drops the last shared_ptr
     // reference. If remote shards complete first and the local shard (which


integration/test_oom_handling.py
```diff
@@ -72,11 +72,20 @@ def test_search_oom_cme(self):
...
+        # Run search query using third primary, and expect to fail on OOM.
+        result = run_search_query(client_primary_2)
+        assert result[0] == 2
+        # Disable partial results.
+        for node in self.nodes:
+            node.client.execute_command("CONFIG", "SET", "search.enable-partial-results", "no")
+        # The OOM error during fanout results in the coordinator cancelling the command, and we
+        # expect to see timeout error on the client side.
+        with pytest.raises(ResponseError, match="Search operation cancelled due to timeout"):
+            run_search_query(client_primary_2)
```
KarthikSubbarao: This is odd to me.


It appears that In 1.0 and 1.1, when enable partial results were diabled, when the coordinator received the OOM error from the other times, it cancels it and returns a timeout error

It was "simpler" when we considered removing the DENYOOM flag for 1.2. Now, I am thinking this needs to be handled better and we should fix this error handing

KarthikSubbarao: Will open an issue to track this. AFAIK, this behavior existed since 1.0

KarthikSubbarao: Issue: https://github.com/valkey-io/valkey-search/issues/888

integration/test_valkey_search_acl.py
```diff
@@ -199,12 +199,7 @@ def test_valkey_search_cmds_categories(self):
             ),
             (
                 "FT.SEARCH",
-                [b"readonly", b"module"],
-                [b"@read", b"@slow", b"@search"],
-            ),
-            (
-                "FT.AGGREGATE",
```
Aksha1812, line 206: are we removing this check entirely

src/query/fanout.cc
```diff
@@ -224,10 +231,13 @@ class LocalResponderSearch : public query::SearchParameters {
 
  private:
   void QueryCompleteImpl(std::unique_ptr<SearchParameters> self) {
-    if (search_result.status.ok() || enable_partial_results) {
+    if (search_result.status.ok()) {
```
Aksha1812, line 245: is the goal to not enable partial results during OOM only? if so we can add that to the if statement or are we changing behaviour here

Aksha1812 (approved):
Approving. high level LGTM. Left 2 minor comments

integration/test_oom_handling.py
KarthikSubbarao: I decided it is better to fix it now. It was hard to develop a PR with the error logic not fully functional. I updated the PR

PR #881, opened 2026-03-10
Shard the radix into small trees

VoletiRam:
Shard the radix into small trees. Previously, we just had one big tree for prefix/suffix each at schema. We had to acquire lock at global level on entire tree for any writes. Break into small configurable trees and lock the tree at shard level. Use first byte of the word to hash the tree. Allows parallel writes on different sharded tree

### DIFF
--- a/src/indexes/text/text_index.h
+++ b/src/indexes/text/text_index.h
@@ -8,6 +8,8 @@
 #ifndef VALKEY_SEARCH_INDEXES_TEXT_INDEX_H_
 #define VALKEY_SEARCH_INDEXES_TEXT_INDEX_H_
 
+#include <algorithm>
+#include <array>
 #include <atomic>
 #include <bitset>
 #include <cctype>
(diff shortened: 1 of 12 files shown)


src/indexes/text/text_index.h
```diff
@@ -56,34 +56,77 @@ struct TextIndexMetadata {
...
+
+    explicit Shard(void (*free_callback)(void *), bool with_suffix)
+        : prefix_tree(free_callback),
+          suffix_tree(with_suffix ? std::make_unique<Rax>(free_callback)
+                                  : nullptr) {}
+  };
+
+  // Use unique_ptr for indirection - Shard contains non-movable Mutex
+  std::vector<std::unique_ptr<Shard>> shards_;
+  size_t num_shards_;
```
allenss-amazon: Sharding in the per-key TextIndex isn't helpful and will dramatically hurt our per-key space performance numbers (which are already awful). How hard would it be to make TextIndex be a template and pass into it the # of shards are compile time, then use a hard-coded array here. This means that for the per-key index we have close to zero space overhead.

BCathcart:
My question here is similar to yesterday: what's the benefit of this over having one tree and still locking on the first byte? Essentially locking the paths branching from the root node. We'd need to share a lock on the root node, but the number of times the root node is updated is probably very small relative to the life of the main text index. My initial though here is that this adds complexity for minimal benefit.

allenss-amazon:
Can we resolve this or close it?

Aksha1812:
**Reviewers for this PR**

VoletiRam:
Closing the PR for now and will get back explore other suggested options like single tree with lock on first few bytes vs sharded rax etc.

Issue #812, opened 2026-02-24
[FEATURE] Delete all index schema objects upon SHUTDOWN Server Event


KarthikSubbarao:
**The problem/use-case that the feature addresses**

Because the search module creates a data type, it is not allowed to be unloaded.

The search module will not receive an unload callback, due to core logic, because it has created a module data type

```
static int moduleUnloadInternal(struct ValkeyModule *module, const char **errmsg) {
    if (listLength(module->types)) {
        *errmsg = "the module exports one or more module-side data "
                  "types, can't unload";
        return C_ERR;
    }
```

It does not reach the unload block of code:
```
    /* Give module a chance to clean up. */
    const char *onUnloadNames[] = {"ValkeyModule_OnUnload", "RedisModule_OnUnload"};
```

So, alternatively, we can register a server event on the shutdown server event and perform the clean up there.
```
VALKEYMODULE_EVENT_SHUTDOWN
```

The clean up can involve deleting all the index related objects in the schema. This will also prevent integration ASAN tests from treating the index data as a leak

Frank-Gu-81:
Fix is in #811 and #883. Closing this issue.

PR #883, opened 2026-03-10
Implement clean up index schemas on SHUTDOWN event

bandalgomsu:
Implement cleaning up all index schema objects on `VALKEYMODULE_EVENT_SHUTDOWN`.
This avoids relying on unload callbacks, which are not invoked for modules exporting module data types.
Includes unit test updates for shutdown event subscription and shutdown cleanup behavior.

issue : #812

### DIFF
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -748,6 +748,23 @@ void SchemaManager::OnServerCronCallback(ValkeyModuleCtx *ctx,
       ctx, options::GetBackfillBatchSize().GetValue());
 }
 
+void SchemaManager::OnShutdownCallback(ValkeyModuleCtx *ctx,
+                                       [[maybe_unused]] ValkeyModuleEvent eid,
+                                       [[maybe_unused]] uint64_t subevent,
+                                       [[maybe_unused]] void *data) {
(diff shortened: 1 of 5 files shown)


src/schema_manager.cc
```diff
@@ -748,6 +748,23 @@ void SchemaManager::OnServerCronCallback(ValkeyModuleCtx *ctx,
       ctx, options::GetBackfillBatchSize().GetValue());
 }
 
+void SchemaManager::OnShutdownCallback(ValkeyModuleCtx *ctx,
```
KarthikSubbarao, line 751: Is there any other GRPC related clean up / shutdown needed here?

allenss-amazon: There is a shutdown routine for grpc

KarthikSubbarao: We were seeing leaks related to GRPC in the Cluster mode ASAN integration tests, so just wanted to check if we are hooking into it correctly and if this `OnShutdownCallback` is


Also, how about threadpools? Do they need explicit clean up?

BCathcart: I'll follow-up on this

PR #884, opened 2026-03-10
Implement memory reservation on ingestion

allenss-amazon:


src/valkey_search_options.cc
```diff
@@ -608,5 +608,65 @@ vmsdk::config::Number& GetMaxNonVectorSearchResultsFetched() {
       *max_nonvector_search_results_fetched);
 }
 
+/// Register the "--mutation-weight-vector" flag. Controls the weight multiplier
```
KarthikSubbarao, line 611: Do we want these to be App or Dev configs?

src/index_schema.cc
```diff
@@ -1771,6 +1807,8 @@ bool IndexSchema::TrackMutatedRecord(ValkeyModuleCtx *ctx, const Key &key,
     itr->second.from_backfill = from_backfill;
     itr->second.from_multi = from_multi;
     itr->second.sequence_number = sequence_number;
+    itr->second.weighted_buffer.resize(
```
KarthikSubbarao, line 1810: Can we add documentation on `weighted_buffer` explaining what this helps us solve?

allenss-amazon: Let's see if this resolves the problem. If it does we can clean this up.

src/index_schema.cc
```diff
@@ -1768,6 +1756,42 @@ bool IndexSchema::InTrackedMutationRecords(
...
+  size_t total = 0;
+  for (const auto &[alias, attr_data] : attributes) {
+    size_t data_size = 0;
+    if (attr_data.data.get() != nullptr) {
+      data_size = vmsdk::ToStringView(attr_data.data.get()).length();
+    }
+    uint32_t weight = 0;
+    auto attr_itr = attributes_.find(alias);
+    if (attr_itr != attributes_.end()) {
+      switch (attr_itr->second.GetIndex()->GetIndexerType()) {
```
KarthikSubbarao, line 1770: Can each Attribute type instead implement a Weight function where we just return the weight of the config? It will help avoid this switch statement / case handling here

src/valkey_search_options.cc
KarthikSubbarao: IMO, we can mark all of these as Dev configs for now

src/index_schema.cc
allenss-amazon: Yes. A virtual function would have been a better choice.

PR #894, opened 2026-03-11
fix: prevent infinite drain loop in forked BGSave child process

eliastam:
DrainMutationQueue spins forever in the BGSave forked child process because the mutation queue is a frozen snapshot that will never be drained by writer threads. Check VALKEYMODULE_CTX_FLAGS_IS_CHILD and return an error instead of attempting to drain.

Also adds a configurable timeout (drain-mutation-queue-timeout-ms, default 5s) to DrainMutationQueue as a safety net, and an integration test covering the BGSAVE with non-empty mutation queue scenario.

### DIFF
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -1226,20 +1226,20 @@ static absl::Status SaveSupplementalSection(
 }
 
 absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
-  // Drain mutation queue before save if configured.
-  // Skip draining in forked child (BGSave) — the queue is a frozen snapshot
-  // and will never drain, causing an infinite loop.
-  if (options::GetDrainMutationQueueOnSave().GetValue()) {
-    int flags = ValkeyModule_GetContextFlags(nullptr);
-    bool is_bgsave = (flags & VALKEYMODULE_CTX_FLAGS_IS_CHILD) != 0;
-    if (is_bgsave) {
-      return absl::InternalError(
-          "Cannot drain mutation queue in forked child process");
-    }
+  // Drain mutation queue before save if configured and queue is non-empty.
+  // In forked child (BGSave), the queue is a frozen snapshot that will never
+  // drain, so skip it instead.
+  int flags = ValkeyModule_GetContextFlags(nullptr);
+  bool is_bgsave = (flags & VALKEYMODULE_CTX_FLAGS_IS_CHILD) != 0;
+  if (options::GetDrainMutationQueueOnSave().GetValue() &&
+      !tracked_mutated_records_.empty() && !is_bgsave) {
     VMSDK_LOG(NOTICE, nullptr)
         << "Draining mutation queue before RDB save for index "
         << vmsdk::config::RedactIfNeeded(name_);
-    VMSDK_RETURN_IF_ERROR(DrainMutationQueue(detached_ctx_.get()));
+
+    VMSDK_RETURN_IF_ERROR(DrainMutationQueue(
+        detached_ctx_.get(),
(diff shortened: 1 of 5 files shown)


src/index_schema.cc
```diff
@@ -1635,19 +1643,29 @@ void IndexSchema::OnSwapDB(ValkeyModuleSwapDbInfo *swap_db_info) {
   }
 }
 
-void IndexSchema::DrainMutationQueue(ValkeyModuleCtx *ctx) const {
+absl::Status IndexSchema::DrainMutationQueue(ValkeyModuleCtx *ctx) const {
   static const auto max_sleep = std::chrono::milliseconds(100);
   auto sleep_duration = std::chrono::milliseconds(1);
+  auto timeout = std::chrono::milliseconds(
+      options::GetDrainMutationQueueTimeoutMs().GetValue());
```
otherscase: GetDrainMutationQueueOnSaveTimeoutMs?

src/index_schema.cc
```diff
@@ -1226,12 +1226,20 @@ static absl::Status SaveSupplementalSection(
...
 
 absl::Status IndexSchema::RDBSave(SafeRDB *rdb) const {
-  // Drain mutation queue before save if configured
+  // Drain mutation queue before save if configured.
+  // Skip draining in forked child (BGSave) — the queue is a frozen snapshot
+  // and will never drain, causing an infinite loop.
   if (options::GetDrainMutationQueueOnSave().GetValue()) {
+    int flags = ValkeyModule_GetContextFlags(nullptr);
+    bool is_bgsave = (flags & VALKEYMODULE_CTX_FLAGS_IS_CHILD) != 0;
+    if (is_bgsave) {
```
otherscase: We should still continue if the queue is empty at this point.

src/index_schema.cc
```diff
@@ -1669,7 +1687,12 @@ void IndexSchema::OnLoadingEnded(ValkeyModuleCtx *ctx) {
                                    ? " Backfill still required."
                                    : " Backfill not needed.");
     if (options::GetDrainMutationQueueOnLoad().GetValue()) {
-      DrainMutationQueue(ctx);
+      auto status = DrainMutationQueue(ctx);
+      if (!status.ok()) {
+        VMSDK_LOG(WARNING, ctx)
+            << "Failed to drain mutation queue on load for index "
+            << vmsdk::config::RedactIfNeeded(name_) << ": " << status;
+      }
```
otherscase: Timeout should only apply for save. We should keep the load behaviour unchanged.

otherscase (changes requested):
Few comments

src/index_schema.cc
eliastam: reverted this change

src/index_schema.cc
eliastam: yes

src/index_schema.cc
eliastam: done

allenss-amazon:
Is this still needed? If so can we update and resolve the merge problems?

Aksha1812:
**Reviewers for this PR**

zackcam:
I think this has mostly been fixed in https://github.com/valkey-io/valkey-search/pull/904/changes so this most likely can be closed

zackcam:
Only thing this at that does differently is that this contains:
>Also adds a configurable timeout (drain-mutation-queue-timeout-ms, default 5s) to DrainMutationQueue as a safety net, and an integration test covering the BGSAVE with non-empty mutation queue scenario.

Even so if we want to add we should do this in a separate PR not here most likely. I would recommend either requester or maintainer to close

Frank-Gu-81:
Checked with @eliastam offline. Similar change was merged in another PR. So I will cancel this.

Issue #895, opened 2026-03-12
[FEATURE] Add configuration for TAG wildcard minimum prefix length


bandalgomsu:
**The problem/use-case that the feature addresses**
`kDefaultMinPrefixLength` is annotated as `configurable`, but it is currently hardcoded and cannot be changed at runtime.

**Description of the feature**

Add a module config (e.g., search.tag-min-prefix-length) to control the minimum prefix length for TAG wildcard queries (@field:{prefix*}).

**Alternatives you've considered**

Any alternative solutions or features you've considered, including references to existing open and closed feature requests in this repository.

**Additional information**

Any additional information that is relevant to the feature request.

PR #896, opened 2026-03-12
Implement configuration for TAG wildcard minimum prefix length

bandalgomsu:
kDefaultMinPrefixLength is annotated as “configurable”, but it is currently hardcoded and cannot be changed at runtime.
This PR adds a module config (search.tag-min-prefix-length) to control the minimum prefix length for TAG wildcard queries (e.g., @field:{prefix*}).

issue : #895


allenss-amazon (approved):
assuming that the regression tests pass.

PR #914, opened 2026-03-17
Allow FT._debug to use when client is admin

### LINKED ISSUE #779: [BUG] FT._debug is usable only when engine is started in debug-mode
**Describe the bug**
Today FT._DEBUG can be used only if the engine is started with debug-mode which is a startup only config.  Instead the command can be an admin command where a developer can run it on running cluster.   

**To Reproduce**
Steps to reproduce the behavior. Ideally, a script to reproduce the problem.

**Expected behavior**

**Environment**

- Valkey Core Version 

**Additional context**

bandalgomsu:
Implement an admin check function and, through it, allow the use of FT._debug if the user is an admin.

issue : #779

### DIFF
--- a/integration/test_debug.py
+++ b/integration/test_debug.py
@@ -1,6 +1,7 @@
 from valkey import ResponseError
 from valkey.client import Valkey
 from valkey_search_test_case import (
+    ValkeySearchTestCaseBase,
     ValkeySearchTestCaseDebugMode,
     ValkeySearchClusterTestCaseDebugMode,
 )
@@ -79,3 +80,17 @@ class TestFtDebugCommand(ValkeySearchTestCaseDebugMode):
         assert byref[6] == {'Count': 10, 'Bytes': 140, 'AvgSize': b'14', 'Allocated': 320, 'AvgAllocated': b'32', 'Utilization': 43}
         assert bysize[-12] == {'Count': 10, 'Bytes': 120, 'AvgSize': b'12', 'Allocated': 280, 'AvgAllocated': b'28', 'Utilization': 42}
         assert bysize[14] == {'Count': 10, 'Bytes': 140, 'AvgSize': b'14', 'Allocated': 320, 'AvgAllocated': b'32', 'Utilization': 43}
+
+    def test_ft_debug_denied_without_admin_permissions(self):
+        self.client.execute_command(
+            "ACL", "SETUSER", "debug_user", "on", ">password",
+            "~*", "&*", "-@all", "+@read"
+        )
+
+        restricted_client = self.server.get_new_client()
+        restricted_client.execute_command("AUTH", "debug_user", "password")
+
+        with pytest.raises(ResponseError, match="has no permissions to run"):
+            restricted_client.execute_command("FT._DEBUG", "HELP")
+
+        self.client.execute_command("ACL", "DELUSER", "debug_user")
--- a/src/commands/commands.h
+++ b/src/commands/commands.h
@@ -32,6 +32,7 @@ constexpr absl::string_view kWriteCategory{"@write"};
 constexpr absl::string_view kFastCategory{"@fast"};
 constexpr absl::string_view kSlowCategory{"@slow"};
 constexpr absl::string_view kAdminCategory{"@admin"};
+constexpr absl::string_view kDangerousCategory{"@dangerous"};
 
 constexpr absl::string_view kCreateCommand{"FT.CREATE"};
 constexpr absl::string_view kDropIndexCommand{"FT.DROPINDEX"};
@@ -55,7 +56,7 @@ const absl::flat_hash_set<absl::string_view> kInfoCmdPermissions{
 const absl::flat_hash_set<absl::string_view> kListCmdPermissions{
     kSearchCategory, kReadCategory, kSlowCategory, kAdminCategory};
 const absl::flat_hash_set<absl::string_view> kDebugCmdPermissions{
-    kSearchCategory, kReadCategory, kSlowCategory, kAdminCategory};
+    kSearchCategory, kSlowCategory, kAdminCategory, kDangerousCategory};
 
 inline absl::flat_hash_set<absl::string_view> PrefixACLPermissions(
     const absl::flat_hash_set<absl::string_view> &cmd_permissions,
--- a/src/commands/ft_debug.cc
+++ b/src/commands/ft_debug.cc
@@ -347,19 +347,6 @@ absl::Status HelpCmd(ValkeyModuleCtx *ctx, vmsdk::ArgsIterator &itr) {
 absl::Status FTDebugCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
                         int argc) {
   std::string msg;
-  if (!vmsdk::config::IsDebugModeEnabled()) {
-    // Pretend like we don't exist
-    msg = "ERR unknown command '";
-    msg += vmsdk::ToStringView(argv[0]);
-    msg += "', with args beginning with:";
-    for (int i = 1; i < argc; ++i) {
-      msg += " '";
-      msg += vmsdk::ToStringView(argv[i]);
-      msg += "'";
-    }
-    ValkeyModule_ReplyWithError(ctx, msg.data());
-    return absl::OkStatus();
-  }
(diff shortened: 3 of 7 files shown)


src/commands/ft_debug.cc
```diff
@@ -347,7 +348,7 @@ absl::Status HelpCmd(ValkeyModuleCtx *ctx, vmsdk::ArgsIterator &itr) {
 absl::Status FTDebugCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
```
KarthikSubbarao, line 347: Can we add a comment here explaining when this command is allowed to be used?

src/acl.cc
```diff
@@ -403,4 +403,38 @@ absl::Status AclPrefixCheck(ValkeyModuleCtx *ctx, acl::KeyAccess access,
...
+      vmsdk::UniqueValkeyString(ValkeyModule_GetCurrentUserName(ctx));
+  if (!username) {
+    return absl::PermissionDeniedError(
+        "The user does not have admin permissions");
+  }
+
+  auto reply = vmsdk::UniquePtrValkeyCallReply(
+      ValkeyModule_Call(ctx, "ACL", "cs3", "GETUSER", username.get()));
+  VMSDK_ASSIGN_OR_RETURN(auto acl_views, GetAclViewFromCallReply(reply.get()));
+  for (const auto &acl_view : acl_views) {
```
KarthikSubbarao: Can we instead make use of this API?
```
ValkeyModule_GetModuleUserACLString
```
https://valkey.io/topics/modules-api-ref/#ValkeyModule_GetModuleUserACLString

We should not need to execute ACL commands since we have these Module APIs

src/commands/ft_debug.cc
```diff
@@ -347,7 +348,7 @@ absl::Status HelpCmd(ValkeyModuleCtx *ctx, vmsdk::ArgsIterator &itr) {
...
                         int argc) {
   std::string msg;
-  if (!vmsdk::config::IsDebugModeEnabled()) {
+  if (!vmsdk::config::IsDebugModeEnabled() && !AclAdminCheck(ctx).ok()) {
```
KarthikSubbarao: I can agree that this is a way to allow a non-breaking change. ie, we continue to allow the command to be used by all clients in DebugModeEnabled whether they are admin or not.


But I don't understand why the FT.DEBUG command was not flagged as "ADMIN" to begin with. @murphyjacob4 , do you have context on this?

The DEBUG command in the valkey core is flagged as admin-only. See: https://valkey.io/commands/debug/
Non-admin clients cannot use it.

If we flagged the FT.DEBUG command as "admin" in the command flags we use to [register it](https://valkey.io/topics/modules-api-ref/#ValkeyModule_CreateCommand), we would not need to add these custom checks.


If we don't want this to be a breaking change, we can continue with what you have in this PR

KarthikSubbarao (changes requested):

src/commands/ft_debug.cc
KarthikSubbarao: Update: Checked with Jacob, the admin command flag was likely just missed when FT.DEBUG was created.

@bandalgomsu , so let's just add the admin and dangerous command flags to the FT.DEBUG command registration. Then, we can skip the custom code for ACL checks because the valkey core will handle it

src/acl.cc
KarthikSubbarao: Also, all of this can be skipped if we follow: https://github.com/valkey-io/valkey-search/pull/914#discussion_r2962262426

src/commands/ft_debug.cc
bandalgomsu: It looks more consistent and simple. thus I agree with that 👍 
could you please confirm if my understanding is correct?

1. set the ACL categories of `FT._DEBUG` to `ADMIN`, `DANGEROUS`, and `SLOW` (Should `Search` be included as well ??) 
2. delete the debug mode check logic.

KarthikSubbarao: > 1. set the ACL categories of FT._DEBUG to ADMIN, DANGEROUS, and SLOW


Yes

> (Should Search be included as well ??)

Just FT._DEBUG

> delete the debug mode check logic.


Yes

bandalgomsu: Oh sorry for confusion

'Search' mean that is acl category 

https://github.com/valkey-io/valkey-search/blob/main/src/commands/ft._debug.json

It should be included in FT._DEBUG's acl categories ?

KarthikSubbarao: Oh Yes, the SEARCH category too

integration/test_debug.py
```diff
@@ -79,3 +80,17 @@ def test_StringPoolStats(self):
         assert byref[6] == {'Count': 10, 'Bytes': 140, 'AvgSize': b'14', 'Allocated': 320, 'AvgAllocated': b'32', 'Utilization': 43}
         assert bysize[-12] == {'Count': 10, 'Bytes': 120, 'AvgSize': b'12', 'Allocated': 280, 'AvgAllocated': b'28', 'Utilization': 42}
         assert bysize[14] == {'Count': 10, 'Bytes': 140, 'AvgSize': b'14', 'Allocated': 320, 'AvgAllocated': b'32', 'Utilization': 43}
+
+    def test_cannot_run_ft_debug_when_debug_mode_enabled(self):
```
KarthikSubbarao: We can rename the test now.

Also, do we have a good way to test admin vs non admin client usage from the integration tests (non admin clients should not be able to access this command)

integration/test_debug.py
```diff
@@ -79,3 +80,17 @@ def test_StringPoolStats(self):
...
+            "~*", "&*", "-@all", "+@read"
+        )
+
+        restricted_client = self.server.get_new_client()
+        restricted_client.execute_command("AUTH", "debug_user", "password")
+
+        with pytest.raises(ResponseError):
+            restricted_client.execute_command("FT._DEBUG", "HELP")
+
+        self.client.execute_command("ACL", "DELUSER", "debug_user")
```
KarthikSubbarao: Missing new line

src/commands/commands.h
```diff
@@ -55,7 +56,7 @@ const absl::flat_hash_set<absl::string_view> kInfoCmdPermissions{
 const absl::flat_hash_set<absl::string_view> kListCmdPermissions{
     kSearchCategory, kReadCategory, kSlowCategory, kAdminCategory};
 const absl::flat_hash_set<absl::string_view> kDebugCmdPermissions{
```
KarthikSubbarao, line 58: Valkey commands have both ACl flags and Command flags.

You updated the ACL flags.... This looks good.

The command flags are here: https://github.com/valkey-io/valkey-search/blob/main/src/module_loader.cc#L88. We need to update this next


Refer to the core Valkey `DEBUG ` cmd flags here: https://github.com/valkey-io/valkey/blob/unstable/src/commands/debug.json#L12

integration/test_debug.py
bandalgomsu: How about switching the inheritance from `ValkeySearchTestCaseDebugMode` to `ValkeySearchTestCaseBase` and adding a test similar to [test_acl_category_permissions](https://github.com/valkey-io/valkey-search/pull/914/changes/BASE..7a9c4bf81f5b7bc823767580903a7c8fca1199b9#diff-7e217b4c06ae87524b4729ae4802c3224c5441b307951c862634524f6aef7c9dL45) ? ?

src/commands/commands.h
bandalgomsu: `PROTECTED` command flag is only using a core command (only `DEBUG`, `MODULE`).  
and It is not supported as a module command flag in `ValkeyModule_CreateCommand`, so we cannot set it for FT._DEBUG and did not add it. 

https://github.com/valkey-io/valkey/blob/unstable/src/server.c#L4292
https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L1243C1-L1243C42

allenss-amazon:
Would like to merge this, can you respond to review items?

KarthikSubbarao (approved):

Aksha1812:
@KarthikSubbarao @bandalgomsu does this need to be backported to 1.2.1 ?

Issue #897, opened 2026-03-12
[BUG] Query results from un-owned slots are still reported.


allenss-amazon:
**Describe the bug**

In CME during slot migration there are periods of time when a shard has keys that belong to un-owned slots. These keys will be in the index and need to be filtered out of the results.

**To Reproduce**
Steps to reproduce the behavior. Ideally, a script to reproduce the problem.

**Expected behavior**

**Environment**

Additional information to reproduce the failure

- Valkey Search Module Version
- Valkey Core Version 

**Additional context**

PR #921, opened 2026-03-19
Slot Ownership check

KarthikSubbarao:
Slot Ownership check in the search path.

This PR adds a "Slot Ownership check" in the search path for FT.SEARCH and FT.AGGREGATE commands and ensures we are only include the keys which are owned by this shard. This helps avoid a double counted key (from source and target) during slot migration when the indexes are applying the mutations from the slot purge.

It uses the `GetOrRefreshClusterMap` functionality to obtain the slot map and checks if the node owns the slot containing the key. It is important to know that this is controlled by `search.cluster-map-expiration-ms` which has a default refresh rate of 0.25 seconds.
```
// configurable variable for cluster map expiration time
static auto cluster_map_expiration_ms =
    vmsdk::config::Number("cluster-map-expiration-ms",
                          250,       // default: 0.25 second
                          0,         // min: 0 (no cache)
                          3600000);  // max: 1 hour
```

Note: The change will not work without Atomic Slot Migration
If we migrated a slot, and we are still indexing keys on the target, we will not include them in the source (because it is not owned by the old source) and we will not include on the target node because it is not indexed yet. So, we get back partial search results. This will be addressed once we correctly hook into Atomic Slot Migration hooks

https://github.com/valkey-io/valkey-search/issues/897

### DIFF
--- a/integration/test_multidb_search.py
+++ b/integration/test_multidb_search.py
@@ -10,13 +10,19 @@ from valkey_search_test_case import (
 from valkeytestframework.conftest import resource_port_tracker
 from valkeytestframework.util import waiters
 from indexes import Index, Text, Tag, Numeric, Vector, float_to_bytes
-
+import time
 
 def verify_common_keys_results(clients, num_dbs, key_names, index):
     """Verify search isolation across all DBs for text, numeric, tag, and vector searches."""
     for db_num in range(num_dbs):
         # Text search
         for key_idx, key_name in enumerate(key_names):
+            print(clients[db_num].execute_command(
+                'INFO SEARCH'
+            ))
+            print(clients[db_num].execute_command(
+                'FT.INFO ' + str(index.name)
+            ))
             result = clients[db_num].execute_command(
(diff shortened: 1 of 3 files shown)


integration/test_multidb_search.py
```diff
@@ -382,5 +382,6 @@ def test_multidb_slot_migration_CME(self):
                 len(key_names),
                 timeout=10
             )
-
+        # Wait longer than cluster map expiration (250ms default) plus some buffer
+        time.sleep(1)
```
VoletiRam, line 386: Wondering if we can update the test to validate the results to not have keys belong to the slot that we don't own

VoletiRam (approved):

integration/test_multidb_search.py
KarthikSubbarao: Actually, we are already testing the intended behavior. When you look at either node, it fans out. So it includes results for both shards. What we are testing is that we are not double counting (since the verify_common_keys_results fn checks for count for each search command's results)

Issue #907, opened 2026-03-15
[BUG] Querying TAG from values from wildcard JSON paths


eric-musliner:
**Describe the bug**
There seems to be a bug when querying/extracting TAG values from wildcard JSON paths like `$.address[*].city`. This behavior works fine in RediSearch.

**To Reproduce**
Create index
```
FT.CREATE idx ON JSON PREFIX 1 user: SCHEMA $.address[*].city AS address__city TAG
```
Add test JSON document
```
JSON.SET user:1 $ '{"address":[{"city":"Pittsburg"},{"city":"New York"}]}'
```
Search on idx
```
FT.SEARCH idx '@address__city:{New York}'
```
Returns no results

**Expected behavior**
```
FT.SEARCH idx '@address__city:{New York}'
```
Should return the test JSON doc user:1

**Environment**

Running `valkey-bundle` docker image. Tag `unstable`. Which should include the latest Valkey Search [1.2.0-rc3](https://github.com/valkey-io/valkey-search/releases/tag/1.2.0-rc3)

https://github.com/valkey-io/valkey-bundle/pull/95

MODULE LIST
```
2) 1) "name"
   2) "ldap"
   3) "ver"
   4) "16777471"
   5) "path"
   6) "/usr/lib/valkey/libvalkey_ldap.so"
   7) "args"
   8) (empty list or set)
3) 1) "name"
   2) "search"
   3) "ver"
   4) "66048"
   5) "path"
   6) "/usr/lib/valkey/libsearch.so"
   7) "args"
   8) (empty list or set)
4) 1) "name"
   2) "bf"
   3) "ver"
   4) "10001"
   5) "path"
   6) "/usr/lib/valkey/libvalkey_bloom.so"
   7) "args"
   8) (empty list or set)
5) 1) "name"
   2) "lua"
   3) "ver"
   4) "1"
   5) "path"
   6) "libvalkeylua.so"
   7) "args"
   8) (empty list or set)
```

webdevelopersrinu:
Same problem while porting redis-om to Valkey. A tag field on a JSON array
doesn't match the values:

    FT.CREATE t ON JSON PREFIX 1 t: SCHEMA $["tags"][*] AS tags TAG SEPARATOR |
    JSON.SET t:1 $ '{"tags":["a","b"]}'
    FT.SEARCH t '@tags:{b}'   -> 0, I expected 1

Works fine on HASH. Tested on the latest bundle (search 1.2.1).

PR #924, opened 2026-03-23
Fix normalize json wildcard string-array values for TAG indexing

bandalgomsu:
Fix normalize wildcard string-array values for TAG indexing

For JSON wildcard paths, `JSON.GET` returns an array string (e.g. `["Seoul","New York"]`).
The previous normalization incorrectly transformed this into `Seoul","New York` instead of a TAG-splittable format, causing lookup mismatches for values such as `New York`.

`["Seoul","New York"]`
Before: `Seoul","New York`
After:  `Seoul,New York`

issue : #907

### DIFF
--- a/integration/test_non_vector.py
+++ b/integration/test_non_vector.py
@@ -517,6 +517,52 @@ class TestNonVector(ValkeySearchTestCaseBase):
         create_bulk_data_standalone(client)
         validate_bulk_limit_queries(client)
 
+    def test_json_tag_query_wildcard_path(self):
+        client: Valkey = self.server.get_new_client()
+
+        assert client.execute_command(
+            "FT.CREATE", "idx_wildcard_json_tag",
+            "ON", "JSON",
+            "PREFIX", "1", "user:",
+            "SCHEMA", "$.address[*].city", "AS", "address__city", "TAG"
+        ) == b"OK"
+
+        assert client.execute_command(
+            "JSON.SET", "user:1", "$",
+            '{"address":[{"city":"Seoul"},{"city":"New York"},{"city":"Seoul"},{"city":""}]}'
+        ) == b"OK"
+
+        assert client.execute_command(
+            "JSON.SET", "user:2", "$",
+            '{"address":[{"city":"Busan"}]}'
+        ) == b"OK"
+
+        time.sleep(1)
+
+        result = client.execute_command(
+            "FT.SEARCH", "idx_wildcard_json_tag",
+            "@address__city:{Seoul}",
+            "NOCONTENT"
+        )
+        assert result[0] == 1
+        assert result[1] == b"user:1"
+
+        result = client.execute_command(
+            "FT.SEARCH", "idx_wildcard_json_tag",
+            "@address__city:{New York}",
+            "NOCONTENT"
+        )
+        assert result[0] == 1
+        assert result[1] == b"user:1"
+
+        result = client.execute_command(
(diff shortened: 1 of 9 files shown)


integration/test_non_vector.py
```diff
@@ -517,6 +517,32 @@ def test_bulk_limit_background_changes(self):
...
+        assert client.execute_command(
+            "FT.CREATE", "idx_wildcard_json_tag",
+            "ON", "JSON",
+            "PREFIX", "1", "user:",
+            "SCHEMA", "$.address[*].city", "AS", "address__city", "TAG"
+        ) == b"OK"
+
+        assert client.execute_command(
+            "JSON.SET", "user:1", "$",
+            '{"address":[{"city":"Seoul"},{"city":"New York"}]}'
```
allenss-amazon: Let's add a few more data values. Especially the harder cases. Like an empty string and some duplicate values.

src/attribute_data_type.cc
```diff
@@ -115,9 +116,20 @@ absl::Status NormalizeJsonRecord(absl::string_view record,
     return absl::NotFoundError("Invalid record");
   }
   if (absl::ConsumePrefix(&record, "[")) {
-    absl::ConsumeSuffix(&record, "]");
```
allenss-amazon: This routine is shared with the other data types: numeric, string and vector. Either we need additional testing to make sure that these changes don't cause a problem for those data types OR perhaps we should pass in the data type that it's being used for -- so that the conversions are handled on a per-type basis.

bandalgomsu: on second thought, to correctly handle string/tag types, I think we should implement a JSON array parser that iterates through the input. 

if the data type is known in advance, we can apply a fast path (like previously method) for numeric/vector types.

alternatively, we can detect the type by checking the first character (e.g., '"' for string arrays) and choose the appropriate path.

what do you think about ? ?

allenss-amazon: The original GCP code only supported HASH, JSON was added later. That's why this routine has a single string output, because the ingestion engine is built around the idea of one attribute -> one string's worth of data. 

IMO, we should either pass into this routine the data type OR break this up into virtual functions. In either case, there's type specific parsing behavior.

For text, only a single string should be supported. I believe that some special character handling may be required.
For numeric, only a single string or a single number should be supported. (Arguably, an array of numbers could be supported in the future).
For Tags, only a single string or an array of strings should be supported. In the array case, it should reformat the data into the same format that a HASH value of multiple tags would have -- this may require accessing the index definition to get the separator character.
For Vector, only an array should be supported.

Aksha1812:
**Reviewers for this PR**

allenss-amazon:
Can we get these conflicts resolved so we can review this and get it into the 1.3 release? Also, since it's a breaking change, it should be conditioned by the COMPATIBILITY_FIX macro.

allenss-amazon (changes requested):
Let's change the release # from 1.2.1 to 1.3.0 and I'll merge it into the current release.

BTW, default version is now 1.3.0, so the places in the tests where you set 1.3.0 can be removed.

Frank-Gu-81:
Hi @bandalgomsu 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

PR #932, opened 2026-03-25
feat: add TOLIST reducer for FT.AGGREGATE GROUPBY stage

AlexFilipImproving:
## Summary

Adds the `TOLIST` reduction function for `FT.AGGREGATE`'s `GROUPBY` stage. TOLIST collects all distinct values of a specified field within each group into a vector (RESP array), providing RediSearch-compatible behavior.

## Changes

- Added `ToList` reducer class in `ft_aggregate_exec.cc` using `absl::flat_hash_set<expr::Value>` for deduplication
- Registered `TOLIST` in `GroupBy::reducerTable` with min/max nargs of 1
- Nil values are skipped; vector inputs are flattened one level
- `GetResult()` returns an `expr::Value::Vector` (serialized via existing `SerializeVectorToResp`)

## Testing

- Unit tests in `ft_aggregate_exec_test.cc` covering basic collection, deduplication, vector type, and size
- Parser tests in `ft_aggregate_parser_test.cc` for valid syntax, AS alias, and invalid argument counts
- Integration test covering end-to-end RESP output, deduplication, multi-reducer GROUPBY, and AS alias

## Example

```
FT.AGGREGATE idx "*" GROUPBY 1 @category REDUCE TOLIST 1 @name AS names
```

Returns a nested RESP array of distinct `@name` values per `@category` group.


integration/compatibility_test.py
```diff
@@ -490,6 +490,20 @@ def do_answer_cluster(cluster_client, expected, data_set, test_case):
 
     return data_set
 
+def _check_and_regenerate_answers(answer_file, generator_file):
```
allenss-amazon: I think this code belongs in the build scripts, not here.

integration/compatibility_test.py
```diff
@@ -490,6 +490,20 @@ def do_answer_cluster(cluster_client, expected, data_set, test_case):
...
+    
+    answer_time = subprocess.run(["git", "log", "-1", "--format=%ct", answer_path], 
+                                  capture_output=True, text=True, cwd=root_dir).stdout.strip().split('\n')[-1]
+    generator_time = subprocess.run(["git", "log", "-1", "--format=%ct", generator_path], 
+                                     capture_output=True, text=True, cwd=root_dir).stdout.strip().split('\n')[-1]
+    
+    if not answer_time or (generator_time and int(generator_time) > int(answer_time)):
+        print(f"Regenerating {answer_file}...")
+        os.system(f"cd {root_dir}/integration/compatibility && pytest {generator_file}")
+
```
allenss-amazon: I love the idea of having the answer files being re-generated. But I think the cmake/build scripts are the right place for that code.

src/commands/ft_aggregate_exec.cc
```diff
@@ -325,16 +348,47 @@ class Stddev : public GroupBy::ReducerInstance {
...
+      if (values[0].IsNil()) {
+        continue;
+      }
+      if (values[0].IsVector()) {
+        auto vec = values[0].GetVector();
+        for (const auto& elem : *vec) {
+          if (!elem.IsNil()) {
+            values_.insert(elem);
+          }
+        }
```
allenss-amazon: Can this happen?

src/commands/ft_aggregate_exec.cc
```diff
@@ -325,16 +348,47 @@ class Stddev : public GroupBy::ReducerInstance {
...
+      }
     }
   }
   expr::Value GetResult() const override {
     return expr::Value(double(values_.size()));
   }
 };
 
+class ToList : public GroupBy::ReducerInstance {
+  absl::flat_hash_set<expr::Value> values_;
```
allenss-amazon: Is a hash_set correct or should it be a vector? Does the compatibility testing cover the cases:

1. Duplicate values.
2. Ordering of values.

src/expr/value.h
```diff
@@ -38,17 +45,29 @@ class Value {
   explicit Value(const char* s) : value_(absl::string_view(s)) {}
   explicit Value(std::string&& s) : value_(std::move(s)) {}
 
+  // Vector constructors
+  explicit Value(Vector vec) : value_(vec) {}
```
allenss-amazon: const & ?

src/expr/value.h
```diff
@@ -22,12 +28,13 @@ class Value {
...
-    Nil(const char* reason) : reason_(reason) {}
-    const char* GetReason() const { return reason_; }
+    explicit Nil(std::string reason) : reason_(reason) {}
+    const char* GetReason() const { return reason_.c_str(); }
 
    private:
-    const char* reason_;
+    std::string reason_;
   };
+  using Vector = std::shared_ptr<std::vector<Value>>;
```
allenss-amazon: I'd like to see us differentiate between a vector, in the vector-search sense and a collection of Values -- which I think could be called an Array.

A vector is guaranteed to be a collection of numbers (no nils, or strings, etc.). This will allow future functions to efficiently compute values from vectors (like vector distance, etc.)

An Array is a heterogeneous collection of values.  It's less efficient because you have to test each element as you go -- and like end up with Nils in some cases.

We can also think about a Tensor type, where a "Vector" is a rank 1 Tensor.

allenss-amazon:
Can we update the conflicts?

allenss-amazon (approved):

allenss-amazon:
Additional compatibilty tests associated with ARRAY data type are forth coming in a different PR from allen.

PR #947, opened 2026-04-01
feat: add Quantile reducer to FT.Aggregate groupby stage

rileydes-improving:
## Summary
Adds a `QUANTILE `reducer to `FT.AGGREGATE` using the Greenwald-Khanna approximate quantile algorithm (±1% error bound), and allows `GROUPBY 0` to aggregate all records into a single group. Provides RediSearch-compatible behavior.

## Changes
- `Quantile` class in `ft_aggregate_exec.cc` implements GK approximate quantile estimation with buffered insertion, sorted merge, and backward-pass compression; registered in `GroupBy::reducerTable` as `QUANTILE` with 2 arguments (@field, quantile_value).
- `GROUPBY 0` guard removed in `ft_aggregate_parser.cc`, zero-group count now aggregates all records into one group, matching RediSearch behavior.
- `ft.aggregate.json` and `ft.aggregate.md` updated to include `QUANTILE` as a valid reducer.

## Testing
- Unit: `QUANTILE` correctness verified across percentiles, group sizes, nil values, duplicates, and negative numbers, property-based tests validate GK accuracy and reject out-of-range quantile values. Parser tests also added to test quantile usage.
- Integration: `test_non_vector.py` covers median, p99, multiple reducers in one query, per-group quantile with `GROUPBY 1`, and `GROUPBY 0` global aggregation
- Compatibility tests against RediSearch

## Example
```
FT.AGGREGATE products "@price:[1 100]"
  LOAD 1 price
  GROUPBY 0
  REDUCE QUANTILE 2 @price 0.5 AS median_price
  REDUCE QUANTILE 2 @price 0.99 AS p99
```
Returns a single record with median_price ≈ 50.5 and p99 ≈ 99 for a uniform price distribution


rileydes-improving:
The pattern of using cerr matches the rest of the file, but I don't know if it should be fixed

allenss-amazon:
Can we resolve the conflicts on this so it can be reviewed?

rileydes-improving:
> Can we resolve the conflicts on this so it can be reviewed?
Conflcts resolved and rebased to the most recent main.

src/commands/ft_aggregate_exec.cc
```diff
@@ -341,6 +343,227 @@ class CountDistinct : public GroupBy::ReducerInstance {
...
+          double max_val = GetMaxVal(r);
+          size_t delta =
+              max_val > 1.0 ? static_cast<size_t>(std::floor(max_val)) - 1 : 0;
+          merged.emplace_back(buffer_[buf_idx++], 1, delta);
+        } else {
+          // Keep existing sample, advance rank
+          r += samples_[samp_idx].g;
+          merged.push_back(samples_[samp_idx++]);
+        }
+      }
```
allenss-amazon: Wouldn't this be cleaner (and faster) if you did the  loop around (buf_idx < buf-size && samp < sample_size) to handle the case with both values and then dropped out to a loop to finish the residual buffer elements and another loop to finish the residual samples?

src/commands/ft_aggregate_exec.cc
```diff
@@ -341,6 +343,227 @@ class CountDistinct : public GroupBy::ReducerInstance {
...
+    for (size_t i = 1; i < samples_.size(); ++i) {
+      const Sample& cur = samples_[i];
+      if (rank + cur.g + cur.delta >= t) {
+        break;
+      }
+      rank += cur.g;
+      prev = &cur;
+    }
+
+    return prev->value;
```
allenss-amazon: isn't prev always equal to samples_[i-1] ? Seems like prev could just be removed.

src/commands/ft_aggregate_exec.cc
```diff
@@ -341,6 +343,227 @@ class CountDistinct : public GroupBy::ReducerInstance {
...
+      if (absl::SimpleAtod(val.AsStringView(), &parsed)) {
+        buffer_.push_back(parsed);
+        n_++;
+        if (buffer_.size() >= kDefaultBufferSize) {
+          Flush();
+          Compress();
+        }
+        return true;
+      }
+    }
```
allenss-amazon: AsDouble already succeeds if the string is convertable to a number. This code can probably be deleted.

src/commands/ft_aggregate_exec.cc
```diff
@@ -341,6 +343,227 @@ class CountDistinct : public GroupBy::ReducerInstance {
...
+        has_error_ = true;
+        quantile_ = -1.0;
+        return;
+      }
+      quantile_ = *quantile_opt;
+      if (quantile_ < 0.0 || quantile_ > 1.0) {
+        has_error_ = true;
+        return;
+      }
+      quantile_initialized_ = true;
```
allenss-amazon: Right now, this code would allow a quantile value to be an expression -- which seems like a problem.

If this reducer was converted to the same custom parser as FIRST_VALUE was forced to do, then it could parse the quantile value at compile time, rejecting the error with a clean message (rather than returning nil, which is what this code does). Also, the initialization stuff wouldn't have to be inside of ProcessRecord.

testing/ft_aggregate_exec_test.cc
```diff
@@ -6,13 +6,40 @@
...
 namespace aggregate {
 
+static double CalculateGKQuantile(const std::vector<double>& sorted_values,
+                                  double quantile) {
+  size_t n = sorted_values.size();
+  if (n == 0) return std::numeric_limits<double>::quiet_NaN();
+  if (n == 1) return sorted_values[0];
+
+  double t = std::ceil(quantile * n);
+  double epsilon = 0.01;
```
allenss-amazon: Should this be using the Epsilon that's declared in the Reducer itself?

testing/ft_aggregate_exec_test.cc
```diff
@@ -285,5 +312,760 @@ TEST_F(AggregateExecTest, testHash) {
...
+    // Generate invalid quantile values (< 0 or > 1)
+    double quantile;
+    if (iteration % 2 == 0) {
+      // Generate values less than 0
+      quantile = -1.0 - dist_offset(rng) / 10.0;
+    } else {
+      // Generate values greater than 1
+      quantile = 1.1 + dist_offset(rng) / 10.0;
+    }
+
```
allenss-amazon, line 619: Is there some intermediate instrumentation/visibility that we could add to the reducer to ensure that we're exercising all of the various paths for merging/compression?

allenss-amazon (changes requested):

src/commands/ft_aggregate_exec.cc
rileydes-improving: Fixed in 1231dc0409c23a53aeac5fba9a70cc430e8de739

testing/ft_aggregate_exec_test.cc
rileydes-improving: I added some stat tracking in 3695954bc7b8a3c47df4ab4d3bfe2d25fc6c0943 :

```cpp
// Instrumentation counters for the QUANTILE reducer's GK algorithm.
// Exposed for testing to verify all internal paths are exercised.
struct QuantileStats {
  size_t flush_initial_count{0};  // First flush (empty samples)
  size_t flush_merge_count{0};    // Merge flush (into existing samples)
  size_t compress_count{0};       // Compression passes performed
  size_t insert_count{0};         // Total values inserted
  size_t samples_merged{0};       // Samples removed during compression
};
```

It allows testing to have some visibility into the internal algorithm. I added a basic test, QuantileInstrumentationPathCoverage, to verify the stats track as expected for the algorithm use. I can incorporate it more into the rest of the test suite or are there any specific tests that you think should be added?

testing/ft_aggregate_exec_test.cc
rileydes-improving: Refactored out the epsilon to a shared constant in 67d985bc18e643bc9f1fa2854cd7bcb5c5c40d16.

src/commands/ft_aggregate_exec.cc
rileydes-improving: Switched to a custom parser similar to first value in 3695954bc7b8a3c47df4ab4d3bfe2d25fc6c0943

src/commands/ft_aggregate_exec.cc
rileydes-improving: Fixed in 60e1c724f7e9f83d29567f8894cb1fb8e920c82a

src/commands/ft_aggregate_exec.cc
rileydes-improving: Yes, fixed in 1f0b7de7c6468e902af21af3d4104562abc68173.

src/commands/ft_aggregate_exec.cc
```diff
@@ -471,89 +485,46 @@ class Quantile : public GroupBy::ReducerInstance {
...
-      buffer_.push_back(*d);
-      n_++;
-      if (buffer_.size() >= kDefaultBufferSize) {
-        Flush();
-        Compress();
-      }
-      return true;
-    }
+    if (!d) return false;
+    if (std::isnan(*d)) return false;
```
allenss-amazon: std::isnan doesn't always work with -ffast-math, we have our own IsNan function that does work, please use that one.

src/commands/ft_aggregate_exec.cc
```diff
@@ -352,18 +353,27 @@ class Quantile : public GroupBy::ReducerInstance {
...
   static constexpr size_t kDefaultBufferSize = 500;
 
   mutable std::vector<double> buffer_;
   mutable std::vector<Sample> samples_;
   size_t n_{0};  // Total number of values inserted
   double quantile_{0.0};
-  bool has_error_{false};
-  bool quantile_initialized_{false};
 
+  std::shared_ptr<QuantileStats> stats_{std::make_shared<QuantileStats>()};
```
allenss-amazon: Why a shared_ptr here? Why not encapsulate the Stats?

allenss-amazon (commented):
Just a couple of minor items.

src/commands/ft_aggregate_exec.cc
rileydes-improving: I had used a shared_ptr to make the shared ownership self-documented and ensure that the lifetimes are guaranteed. Looking at it again, a member + raw pointer is cleaner, and avoids the heap allocation/atomic operations, I'll change out to that

src/commands/ft_aggregate_exec.cc
rileydes-improving: Used it, had to make it public throught the value.h header.

src/commands/ft_aggregate_exec.cc
rileydes-improving: Changed

src/commands/ft_aggregate_exec.cc
```diff
@@ -403,15 +598,95 @@ absl::StatusOr<std::unique_ptr<GroupBy::Reducer>> BasicReducerParser(
...
+                           _ << "Missing Reducer alias");
+    VMSDK_ASSIGN_OR_RETURN(auto output, parameters.MakeReference(
+                                            vmsdk::ToStringView(alias), true));
+    r->output_ =
+        std::unique_ptr<Attribute>(dynamic_cast<Attribute*>(output.release()));
+  } else {
+    std::ostringstream os;
+    os << *r << ',' << vmsdk::ToStringView(quantile_tok);
+    VMSDK_ASSIGN_OR_RETURN(auto output,
+                           parameters.MakeReference(os.str(), true));
```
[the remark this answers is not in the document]
rileydes-improving: Fixed

testing/ft_aggregate_parser_test.cc
```diff
@@ -225,6 +225,22 @@ static std::vector<TestStage> TestStages{
     {"GROUPBY 1 @n1 REDUCE COUNT 0", "GROUPBY @n1 COUNT() => COUNT()"},
     {"GROUPBY 1 @n1 REDUCE COUNT 0 AS Y", "GROUPBY @n1 COUNT() => Y"},
     {"GROUPBY 1 @n1 REDUCE MIN 1 @n2 as Z", "GROUPBY @n1 MIN(@n2) => Z"},
+    {"GROUPBY 1 @n1 REDUCE QUANTILE", nullptr},
+    {"GROUPBY 1 @n1 REDUCE QUANTILE 0", nullptr},
+    {"GROUPBY 1 @n1 REDUCE QUANTILE 1", nullptr},
+    {"GROUPBY 1 @n1 REDUCE QUANTILE 1 @n2", nullptr},
+    {"GROUPBY 1 @n1 REDUCE QUANTILE 2 @n2", nullptr},
+    {"GROUPBY 1 @n1 REDUCE QUANTILE 2 @n2 0.5",
+     "GROUPBY @n1 QUANTILE(@n2) => QUANTILE(@n2),0.5"},
```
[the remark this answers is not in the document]
rileydes-improving: test doesn't fail currently so

Aksha1812:
**Reviewers for this PR**

PR #949, opened 2026-04-01
feat: add Random Sample reducer to FT.Aggregate groupby stage

rileydes-improving:
## Summary
Adds a RANDOM_SAMPLE reducer to FT.AGGREGATE GROUPBY stage, enabling reservoir-sampled field values to be returned as RESP arrays from aggregate queries in RediSearch compatible behavior (Not testable as sampling is random)

## Changes

- `RandomSample` class in `ft_aggregate_exec.cc` implements Algorithm R reservoir sampling with nil-skipping, a 1000-element cap, and a thread_local std::mt19937; registered in GroupBy::reducerTable as RANDOM_SAMPLE with 2 arguments (@ field, sample_size).
- `ft.aggregate.json` and `ft.aggregate.md` updated to include `RANDOM_SAMPLE` as a valid reducer.
- Uses vector value type from #893 

## Testing

- Unit: ft_aggregate_exec_test.cc covers empty group, single element, sample < / == / > group size, nil exclusion, string and mixed-type inputs, multiple independent reducers, and per-group isolation
- Integration: test_non_vector.py covers basic sampling, size < / == / > group, sample size 0, multiple reducers, per-category GROUPBY, and numeric field values.
- Compatibility tests not added

## Example
```
FT.AGGREGATE products "@price:[1 1000]"
  LOAD 1 price  APPLY 1 AS all
  GROUPBY 1 @all
  REDUCE RANDOM_SAMPLE 2 @price 5 AS sample
```
Returns one group with sample as a RESP array of up to 5 randomly selected price values.


allenss-amazon:
Can we resolve the conflicts on this so it can be reviewed?

src/expr/value.h
```diff
@@ -129,7 +156,7 @@ static inline bool operator>(const Value& l, const Value& r) {
 
 static inline bool operator>=(const Value& l, const Value& r) {
   auto res = Compare(l, r);
-  return res != Ordering::kLESS;
+  return res == Ordering::kGREATER || res == Ordering::kEQUAL;
 }
```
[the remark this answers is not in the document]
rileydes-improving: This is on the array value implementation not the reducer.

src/expr/value.cc
```diff
@@ -234,9 +268,31 @@ Ordering Compare(const Value& l, const Value& r) {
...
+    // All elements equal, compare by length
+    if (lvec->size() < rvec->size()) {
+      return Ordering::kLESS;
+    } else if (lvec->size() > rvec->size()) {
+      return Ordering::kGREATER;
+    }
+    return Ordering::kEQUAL;
+  } else if (l.IsArray() || r.IsArray()) {
+    // Array vs scalar
+    return Ordering::kUNORDERED;
```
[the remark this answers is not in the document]
rileydes-improving: This is on the array value implementation not the reducer.

src/expr/value.h
```diff
@@ -71,6 +93,13 @@ class Value {
       return H::combine(std::move(h), 0);
     } else if (v.IsDouble()) {
       return H::combine(std::move(h), *v.AsDouble());
+    } else if (v.IsArray()) {
+      auto arr = v.GetArray();
+      h = H::combine(std::move(h), arr->size());
+      for (const auto& elem : *arr) {
+        h = H::combine(std::move(h), elem);
+      }
+      return h;
```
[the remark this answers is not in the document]
rileydes-improving: This is on the array value implementation (the feature branch this is based on) not the reducer.

integration/compatibility/generate.py
```diff
@@ -296,6 +296,34 @@ def test_aggregate_groupby(self, key_type, dialect):
...
+            # Negative sample size.
+            f"ft.aggregate {key_type}_idx1 * load 2 @__key @n1 groupby 1 @t1 reduce random_sample 2 @n1 -1 as s",
+            f"ft.aggregate {key_type}_idx1 * load 2 @__key @n1 groupby 1 @t1 reduce random_sample 2 @n1 -5 as s",
+            # Above cap (1000).
+            f"ft.aggregate {key_type}_idx1 * load 2 @__key @n1 groupby 1 @t1 reduce random_sample 2 @n1 1001 as s",
+            f"ft.aggregate {key_type}_idx1 * load 2 @__key @n1 groupby 1 @t1 reduce random_sample 2 @n1 10000 as s",
+        ]
+        for cmd in error_cases:
+            self.check(dialect, cmd)
+
```
allenss-amazon, line 378: Compatibility test checking for errors isn't really helpful, let's just move this to be a normal integration test.

integration/test_non_vector.py
```diff
@@ -323,6 +323,171 @@ def validate_aggregate_queries(client: Valkey):
         with pytest.raises(ResponseError, match=r"Invalid or missing expression"):
             client.execute_command(*command)
 
+def validate_random_sample_queries(client: Valkey):
```
allenss-amazon: could we pass in the dataset so that the comparisons can prove that the results come from the dataset itself?

integration/test_non_vector.py
```diff
@@ -323,6 +323,171 @@ def validate_aggregate_queries(client: Valkey):
...
+        "GROUPBY", "1", "@all",
+        "REDUCE", "RANDOM_SAMPLE", "2", "@price", "10", "AS", "sample"
+    )
+    assert result[0] == 1
+    row = dict(zip(result[1][::2], result[1][1::2]))
+    sample = row[b'sample']
+    assert isinstance(sample, list)
+    assert len(sample) == 10  # Should have exactly 10 elements
+    
+    # Sample size larger than group size
```
allenss-amazon, line 488: Probably should include a test that's exactly the same size of the dataset.

src/commands/ft_aggregate_exec.cc
```diff
@@ -241,6 +243,46 @@ class Count : public GroupBy::ReducerInstance {
...
+
+ private:
+  // Thread-local RNG shared across all RandomSample instances in a query,
+  // avoiding per-instance std::random_device overhead.
+  static std::mt19937& Rng() {
+    thread_local std::mt19937 rng(std::random_device{}());
+    return rng;
+  }
+
+  std::vector<expr::Value> samples_;
```
allenss-amazon: I think this should be shared_ptr<std::vector<Value>>, to avoid the copy when GetResult is called. The constructor can call make_shared....

allenss-amazon (approved):

integration/compatibility/generate.py
rileydes-improving: Done

integration/test_non_vector.py
rileydes-improving: Yes, passed it through and made some helpers to be able to do that.

integration/test_non_vector.py
rileydes-improving: Added a new one with a dataset comparison / verification

src/commands/ft_aggregate_exec.cc
rileydes-improving: Sounds good, implemented.

Aksha1812:
**Reviewers for this PR**

boda26:
@allenss-amazon  I saw you approved this on Jul 21. Do we want to merge this in as no conflicts and all tests passed?

src/commands/ft_aggregate_exec.cc
```diff
@@ -421,6 +463,88 @@ class ToList : public GroupBy::ReducerInstance {
...
+        std::unique_ptr<Attribute>(dynamic_cast<Attribute*>(output.release()));
+  } else {
+    // Workaround for memory allocator issue causing ostringstream to crash
+    // (see https://github.com/valkey-io/valkey-search/issues/965).
+    std::string default_name(r->name_);
+    default_name += '(';
+    default_name += field_text;
+    default_name += ',';
+    default_name += size_text;
+    default_name += ')';
```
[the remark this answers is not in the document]
rileydes-improving: Fixed

src/commands/ft_aggregate_exec.cc
```diff
@@ -245,6 +248,46 @@ class Count : public GroupBy::ReducerInstance {
...
 };
 
+class RandomSample : public GroupBy::ReducerInstance {
+ public:
+  static constexpr size_t kMaxSampleSize = 1000;
+
+  explicit RandomSample(size_t sample_size)
+      : samples_(std::make_shared<std::vector<expr::Value>>()),
+        sample_size_(sample_size) {
+    samples_->reserve(sample_size_);
```
boda26: RandomSample's constructor always initialize samples_ with sample_size_ regardless of the actual group size, so memory scales with group_count * sample_size. Recommends to delete the reserve here, as the logic in lines 265-272 already handles capping at sample_size_.

src/commands/ft_aggregate_exec.cc
```diff
@@ -421,6 +464,101 @@ class ToList : public GroupBy::ReducerInstance {
...
+// sample-size arg at parse time to validate it.
+absl::StatusOr<std::unique_ptr<GroupBy::Reducer>> RandomSampleReducerParser(
+    std::string_view name, AggregateParameters &parameters,
+    vmsdk::ArgsIterator &itr) {
+  auto r = std::make_unique<RandomSampleReducer>();
+  r->name_ = name;
+
+  uint32_t cnt{0};
+  VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, cnt));
+  if (cnt < 2) {
```
boda26: cnt < 2 guards nargs that is too small, but not nargs that are too large. If user provides a large nargs, it would eat the rest of query, like the AS clause after it. Since RandomSample has fixed arity 2, it's better to do if (cnt != 2).

src/commands/ft_aggregate_exec.cc
```diff
@@ -421,6 +464,101 @@ class ToList : public GroupBy::ReducerInstance {
...
+  // Evaluate the sample-size expression (arg 1) at parse time.
+  expr::Expression::EvalContext ctx;
+  Record record(parameters.record_info_by_index_.size());
+  auto size_opt = r->args_[1]->Evaluate(ctx, record).AsDouble();
+  if (!size_opt.has_value() || !std::isfinite(*size_opt) || *size_opt < 0 ||
+      *size_opt != std::floor(*size_opt)) {
+    return absl::InvalidArgumentError(absl::StrCat(
+        name, " sample size must be a non-negative integer constant"));
+  }
+  if (static_cast<size_t>(*size_opt) > RandomSample::kMaxSampleSize) {
```
boda26: static_cast<size_t>(*size_opt) happens before the kMaxSampleSize range check, so an out-of-range double is cast with undefined behavior. Better to do the check before static_cast
```
if (*size_opt > static_cast<double>(RandomSample::kMaxSampleSize)) {
    return absl::OutOfRangeError(absl::StrCat(
    name, " sample size must be <= ", RandomSample::kMaxSampleSize));
 }
r->sample_size_ = static_cast<size_t>(*size_opt)
```

src/commands/ft_aggregate_exec.cc
rileydes-improving: Makes sense, I'll do that.

src/commands/ft_aggregate_exec.cc
rileydes-improving: Fixed.

src/commands/ft_aggregate_exec.cc
rileydes-improving: Switched the order around

boda26 (approved):

PR #954, opened 2026-04-04
fix unit test output after #931

Aksha1812:
I observed in recent unit test failure output for the failure was supressed 
https://github.com/valkey-io/valkey-search/actions/runs/23964394078/job/69901205344#logs

we would still have full logs in artifcats
#931 goal of this was to make workflow output clean to just show only the failure. 

fixed with this change . failing a test on purpose to show how the output would look like


KarthikSubbarao:
@Aksha1812 , all builds failed here. Is this PR correct and needed?

Aksha1812:
> @Aksha1812 , all builds failed here. Is this PR correct and needed?

@KarthikSubbarao they didn't even run for some reason 

> Enabling core dumps
core.%p
/workspace/ci/build_ubuntu.sh: line 163: ./build.sh: Permission denied

Maybe I am not allowed to make changes to the build file . I guess

Earlier I made a change that supress unnecessary long workflow logs. But I discovered in local that for unit tests I am not seeing the failure log which we need , so yes this is needed . I failed a unit test intentionally so that i can compare and show the difference.

allenss-amazon:
Can we update or close this?

Aksha1812:
> Can we update or close this?

fixing today

Aksha1812:
This PR is not fixing anything . my assumption was wrong. verified. unit test failure output showing correctly as is

PR #955, opened 2026-04-06
feat: Implement index aliasing commands 

rileydes-improving:
## Summary

Adds `FT.ALIASADD`, `FT.ALIASDEL`, and `FT.ALIASUPDATE` commands with RediSearch-compatible behavior. Aliases are scoped per database, persisted in RDB, and propagated cluster-wide via the existing `MetadataManager` / `FT.INTERNAL_UPDATE` mechanism.

## Changes

- Added `FTAliasAddCmd`, `FTAliasDelCmd`, `FTAliasUpdateCmd` in `ft_alias_add.cc`, `ft_alias_del.cc`, `ft_alias_update.cc` and registered them in `module_loader.cc` with `@search @write @fast` ACL permissions
- Added `db_to_aliases_` (`flat_hash_map<uint32_t, flat_hash_map<string, string>>`) to `SchemaManager` with `AddAlias`, `RemoveAlias`, `UpdateAlias`, and `GetAliasesForIndex`. `GetIndexSchema` now falls through to alias resolution when a direct name lookup misses
- `RemoveIndexSchemaInternal` purges all aliases pointing to a dropped index and `OnSwapDB` swaps alias maps in lockstep with index maps
- Added `RDB_SECTION_ALIAS_MAP = 3`, `AliasEntry`, and `AliasMap` to `rdb_section.proto`. `SchemaManager` gains `LoadAliasMap` and `SaveAliases`, and RDB sections are sorted by type before writing so indexes are always loaded before aliases
- Extended `FT.INTERNAL_UPDATE` with an optional 5th `type_name` argument that defaults to `kSchemaManagerMetadataTypeName` for AOF backward compatibility. `SchemaManager` registers `kAliasMetadataTypeName` (`vs_alias`) with `MetadataManager` and implements `OnAliasMetadataCallback` for cluster propagation
- `ReconcileMetadata` now skips replication for types not registered locally, preventing mistaken calls on nodes that don't know the alias type
- `ft_dropindex.cc` resolves an alias to its real index name before calling `RemoveIndexSchema`, so `FT.DROPINDEX <alias>` works correctly
- `ft_info_parser.cc` calls `GetAliasesForIndex` and passes the sorted result to `RespondWithInfo`, adding an `aliases` array to `FT.INFO` output

## Testing

- Unit tests in `ft_alias_commands_test.cc` covering argument count errors, happy paths, duplicate rejection, alias-to-alias rejection, self-referential alias, and `ReplicateVerbatim` behavior in both coordinator modes
- Unit tests in `schema_manager_test.cc` (`SchemaManagerAliasTest`) covering add/remove/update round-trips, cross-database isolation, upsert semantics, and the first-alias-for-DB edge case in alias-to-alias rejection
- Unit tests in `ft_info_test.cc` (`FTInfoAliasTest`) verifying the sorted `aliases` array in `FT.INFO` and that querying via alias still reports the real index name
- Integration tests in `test_ft_alias.py` covering all three commands, DROPINDEX cleanup, FT.SEARCH/FT.AGGREGATE via alias, FLUSHDB, SWAPDB, RDB persistence, wrong argument errors, and name-collision edge cases
- Integration tests in `test_ft_alias_cluster.py` verifying that ALIASADD, ALIASDEL, ALIASUPDATE, and DROPINDEX tombstoning all propagate to every cluster primary
- Compatibility: `alias-answers.pickle.gz` added to the parametrized run in `compatibility_test.py` where alias management commands are replayed inline to reconstruct alias state before dependent search/aggregate queries

## Example

```
FT.CREATE my_idx ON HASH PREFIX 1 doc: SCHEMA category TAG
HSET doc:1 category books
FT.ALIASADD my_alias my_idx
FT.SEARCH my_alias "@category:{books}"
```

Returns 1 result (`doc:1`) via the alias without any change to the underlying index.


integration/compatibility/generate.py
```diff
@@ -78,12 +78,14 @@ def setup_data(self, data_set_name, key_type):
         self.key_type = key_type
         load_data(self.client, data_set_name, key_type)
 
-    def execute_command(self, cmd):
+    def execute_command(self, cmd, excluded=False):
```
allenss-amazon: What is the point of "excluded" ? I couldn't find where it was tested.

integration/compatibility/generate.py
```diff
@@ -468,3 +470,99 @@ def test_search_sortby(self, key_type, dialect):
...
+    def test_alias_search(self, key_type, dialect):
+        """FT.SEARCH via alias returns same results as via the real index name."""
+        self.setup_data("sortable numbers", key_type)
+        idx = f"{key_type}_idx1"
+        alias = f"{key_type}_alias_compat"
+        # Record alias management commands so replay recreates the same state.
+        self.execute_command(["FT.ALIASADD", alias, idx])
+        # Bare scan without SORTBY returns 10 of 15 keys; which 10 depends on
+        # shard placement in cluster mode, so mark excluded for cluster replay.
+        self.execute_command(["ft.search", alias, "@n1:[-inf inf]", "DIALECT", str(dialect)], excluded=True)
```
allenss-amazon: Why not just delete this case? I don't see much point past one query to prove operation for FT.SEARCH

integration/compatibility/generate.py
```diff
@@ -468,3 +470,99 @@ def test_search_sortby(self, key_type, dialect):
...
+
+    def test_alias_aggregate(self, key_type, dialect):
+        """FT.AGGREGATE via alias returns same results as via the real index name."""
+        self.setup_data("sortable numbers", key_type)
+        idx = f"{key_type}_idx1"
+        alias = f"{key_type}_alias_compat"
+        # Record alias management commands so replay recreates the same state.
+        self.execute_command(["FT.ALIASADD", alias, idx])
+        # Aggregate across all keys only works on standalone; mark excluded for cluster replay.
+        self.check(dialect, f"ft.aggregate {alias} * load 3 @__key @n1 @n2 sortby 2 @__key asc", excluded=True)
```
allenss-amazon: Same here, it seems like one aggregate command should be sufficient.

integration/compatibility/data_sets.py
```diff
@@ -591,18 +594,64 @@ def generate_doc(doc_id):
     return data
 
 ### Helper Functions ###
+def compute_alias_data():
```
allenss-amazon, line 597: This seems like a lot of logic. What is the purpose for this?

integration/compatibility/generate.py
```diff
@@ -468,3 +470,99 @@ def test_search_sortby(self, key_type, dialect):
...
+        self.execute_command(["FT.AGGREGATE", "alias_agg", "@category:{electronics}",
+            "LOAD", "1", "@category",
+            "GROUPBY", "1", "@category",
+            "REDUCE", "COUNT", "0", "AS", "count",
+        ])
+        self.execute_command([
+            "FT.AGGREGATE", "alias_agg", "@price:[0 +inf]",
+            "LOAD", "2", "@price", "@category",
+            "SORTBY", "2", "@price", "ASC",
+        ])
```
allenss-amazon: What are these testing?

integration/compatibility_test.py
```diff
@@ -311,6 +311,14 @@ def compare_results(expected, results):
         print(TEST_MARKER)
         return False
 
+    # Handle simple (non-list) results from mutation commands like ALIASADD/ALIASDEL/ALIASUPDATE.
```
allenss-amazon: Seems more straightforward to look at the command itself, not the result.

integration/compatibility_test.py
```diff
@@ -381,15 +389,49 @@ def mark_as_failed(testname):
...
 
+    # Alias management commands (ALIASADD/ALIASDEL/ALIASUPDATE) are recorded inline
+    # so that replay recreates the same alias state before the search commands that
+    # depend on it.  Execute them and verify they succeed, but don't count them as
+    # test answers in the pass/fail tally.
+    if _is_alias_management_cmd(expected['cmd']):
+        try:
+            client.execute_command(*expected['cmd'])
+            print(f"Alias setup: {expected['cmd']}")
+        except valkey.ResponseError as e:
```
allenss-amazon: This should count as an error, right?

integration/compatibility_test.py
```diff
@@ -463,6 +505,48 @@ def do_answer_cluster(cluster_client, expected, data_set, test_case):
...
+        verb = expected["cmd"][0].upper()
+        alias_name = expected["cmd"][1]
+        try:
+            # Route to a specific primary since cluster client can't determine
+            # the slot for FT.ALIAS* commands (no key argument).
+            primary0 = test_case.new_client_for_primary(0)
+            primary0.execute_command(*expected["cmd"])
+            print(f"Alias setup (cluster): {expected['cmd']}")
+        except valkey.ResponseError as e:
+            print(f"WARNING: alias management command failed during cluster replay: {e} cmd={expected['cmd']}")
```
allenss-amazon: shouldn't this be an error?

integration/compatibility_test.py
```diff
@@ -463,6 +505,48 @@ def do_answer_cluster(cluster_client, expected, data_set, test_case):
...
+                    try:
+                        node.execute_command("FT.INFO", name)
+                    except valkey.ResponseError:
+                        return False
+                return True
+            try:
+                waiters.wait_for_true(_alias_visible, timeout=10)
+            except Exception as e:
+                print(f"WARNING: alias {alias_name} propagation timed out: {e}")
+        return data_set
```
allenss-amazon: The ALIAS commands need to distribute the data AND check for consistency, like FT.CREATE and FT._DROPINDEX do, meaning that this code shouldn't be required.

integration/test_ft_alias.py
```diff
@@ -0,0 +1,772 @@
...
+class TestFTAliasAdd(ValkeySearchTestCaseBase):
+    """Tests for FT.ALIASADD."""
+
+    def test_aliasadd_happy_path(self):
+        """Adding a valid alias returns OK and the alias resolves to the index."""
+        client = self.client
+        assert client.execute_command(*CREATE_TAG_INDEX) == b"OK"
+        assert client.execute_command("FT.ALIASADD", ALIAS_NAME, INDEX_NAME) == b"OK"
+        # Alias should resolve to the underlying index via FT.INFO
+        info = client.execute_command("FT.INFO", ALIAS_NAME)
```
allenss-amazon: Let's check the returned index name.

integration/test_ft_alias_cluster.py
```diff
@@ -0,0 +1,330 @@
...
+from valkey.client import Valkey
+from valkey_search_test_case import ValkeySearchClusterTestCase
+from valkeytestframework.conftest import resource_port_tracker
+from valkeytestframework.util import waiters
+from utils import IndexingTestHelper
+
+INDEX_NAME = "alias_cluster_idx"
+INDEX_NAME_2 = "alias_cluster_idx2"
+ALIAS_NAME = "cluster_alias"
+
```
allenss-amazon, line 18: Seems like modification of the ALIAS commands to wait for cluster consistency would dramatically change these tests.

src/commands/ft_alias_add.cc
```diff
@@ -0,0 +1,39 @@
...
+  auto index_name = vmsdk::ToStringView(argv[2]);
+
+  VMSDK_RETURN_IF_ERROR(SchemaManager::Instance().AddAlias(
+      ValkeyModule_GetSelectedDb(ctx), alias, index_name));
+
+  ValkeyModule_ReplyWithSimpleString(ctx, "OK");
+
+  if (!options::GetUseCoordinator().GetValue()) {
+    ValkeyModule_ReplicateVerbatim(ctx);
+  }
```
allenss-amazon, line 68: All of the alias commands that mutate the alias map need to wait for cluster consistency, see the FT.CREATE/FT._DROPINDEX commands for examples on how to do this.

src/commands/ft_internal_update.cc
```diff
@@ -55,9 +56,17 @@ absl::Status HandleInternalUpdateFailure(
...
 absl::Status FTInternalUpdateCmd(ValkeyModuleCtx *ctx,
                                  ValkeyModuleString **argv, int argc) {
-  CHECK_EQ(argc, kFTInternalUpdateArgCount)
+  CHECK(argc >= kFTInternalUpdateMinArgCount &&
+        argc <= kFTInternalUpdateMaxArgCount)
       << "FT.INTERNAL_UPDATE called with wrong argument count: " << argc;
 
+  absl::string_view type_name = kSchemaManagerMetadataTypeName;
+  if (argc == kFTInternalUpdateMaxArgCount) {
+    type_name = vmsdk::ToStringView(argv[4]);
```
allenss-amazon: let's switch to a keyword based extension scheme. there are other potential additional fields that might get added to FT._INTERNALUPDATE in the future. Specifically, let's update this to build a map<string, string> (keyword, value) that way future keywords that aren't understood will be silently ignored.

src/rdb_section.proto
```diff
@@ -25,6 +25,19 @@ enum RDBSectionType {
   RDB_SECTION_UNSET = 0;
   RDB_SECTION_INDEX_SCHEMA = 1;
   RDB_SECTION_GLOBAL_METADATA = 2;
+  RDB_SECTION_ALIAS_MAP = 3;
```
allenss-amazon: Why do we need a separate section? Isn't this just part of the metadata?

src/rdb_serialization.cc
```diff
@@ -255,8 +257,16 @@ absl::Status PerformRDBSave(ValkeyModuleCtx *ctx, SafeRDB *rdb, int when) {
...
+  // Now do the save of the contents.
+  // Sort by section type value to guarantee a deterministic write order.
+  // In particular, RDB_SECTION_INDEX_SCHEMA (1) must be written before
+  // RDB_SECTION_ALIAS_MAP (3) so that LoadAliasMap can validate alias targets
+  // during load (sections are loaded in the order they appear in the file).
+  std::vector<std::pair<data_model::RDBSectionType, int>> ordered_sections(
+      section_counts.begin(), section_counts.end());
+  std::sort(ordered_sections.begin(), ordered_sections.end(),
+            [](const auto &a, const auto &b) { return a.first < b.first; });
+  for (auto &section_count : ordered_sections) {
```
allenss-amazon: Don't understand why this is needed. Alias table belongs as part of the global metadata, it shouldn't need a new section.

src/schema_manager.cc
```diff
@@ -128,6 +128,36 @@ SchemaManager::SchemaManager(
...
+          // Returns 1 if any DB has aliases to persist (all aliases are packed
+          // into a single RDB section), 0 otherwise.
+          .section_count = [this](ValkeyModuleCtx *ctx, int when) -> int {
+            // Only report sections for the AFTER phase, matching SaveAliases.
+            if (when == VALKEYMODULE_AUX_BEFORE_RDB) {
+              return 0;
+            }
+            absl::MutexLock lock(&db_to_index_schemas_mutex_);
+            for (const auto &kv : db_to_aliases_) {
+              if (!kv.second.empty()) return 1;
```
allenss-amazon: Same, why does this need to be changed?

src/schema_manager.cc
```diff
@@ -280,6 +463,25 @@ SchemaManager::RemoveIndexSchemaInternal(uint32_t db_num,
...
+    auto &alias_map = db_alias_it->second;
+    for (auto it = alias_map.begin(); it != alias_map.end();) {
+      if (it->second == name) {
+        // Advance before erasing: erasing invalidates the current iterator.
+        auto to_erase = it++;
+        alias_map.erase(to_erase);
+      } else {
+        ++it;
+      }
+    }
```
allenss-amazon: Why not have a reverse multi-map so that we don't have to walk these maps one by one?

src/schema_manager.cc
```diff
@@ -329,6 +556,22 @@ absl::flat_hash_set<std::string> SchemaManager::GetIndexSchemasInDB(
...
+    uint32_t db_num, absl::string_view index_name) const {
+  absl::MutexLock lock(&db_to_index_schemas_mutex_);
+  std::vector<std::string> result;
+  auto db_alias_it = db_to_aliases_.find(db_num);
+  if (db_alias_it != db_to_aliases_.end()) {
+    for (const auto &[alias, target] : db_alias_it->second) {
+      if (target == index_name) {
+        result.push_back(alias);
+      }
+    }
```
allenss-amazon, line 654: With a reverse map you don't need the loop.

src/rdb_serialization.cc
rileydes-improving: My reasoning was to guarantee that aliases would be loaded after indexes so I could validate their targets.

Architecturally, put it in the global metadata and defer the alias validation? Or just don't bother validating aliases on load?

src/rdb_section.proto
rileydes-improving: Same reasoning, it was intended for alias loading after indexes. If we fold into the global metadata, this is removed.

src/schema_manager.cc
rileydes-improving: Same cause as above, alias loading after indexes. If we fold into the global metadata, this is removed.

src/schema_manager.cc
rileydes-improving: Will do

src/rdb_serialization.cc
allenss-amazon: I'd say that no validation of the mapping table is required. If a mapping table alias points to a non-existent, I wouldn't be surprised to see this get handled just like any other reference to a non-existent index.

src/rdb_section.proto
allenss-amazon: Putting the alias table into the global metadata was the expectation.

integration/compatibility/data_sets.py
rileydes-improving: The purpose of that was just to setup and configure the alias data necessary for testing. I trimmed out some unneccessary logic and simplified a bit to make it smaller.

integration/compatibility/generate.py
rileydes-improving: They were for testing that aliases resolution through commands was the same as Redis, I trimmed them out to be more minimal.

integration/compatibility/generate.py
rileydes-improving: It had been to mark commands to be not ran in when the tests were testing clustered, if they would have nondeterministic in a clustered setting. I just removed the tests that would be nondeterministic and cut them down.

integration/compatibility/data_sets.py
```diff
@@ -10,6 +10,7 @@
 
 SETS_KEY = lambda key_type: f"{key_type} sets"
 CREATES_KEY = lambda key_type: f"{key_type} creates"
+SETUP_KEY = lambda key_type: f"{key_type} setup"
```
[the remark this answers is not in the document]
rileydes-improving: Matches existing code pattern, ignoring.

COMMANDS.md
```diff
@@ -308,3 +311,43 @@ The following query will return all books with "comedy" or "horror" genre (OR) p
 The following query will return all books that either don't have a genre field, or have a genre field not equal to "comedy", that are published between 2015 and 2024:
 
 `-@genre:[comedy] @year:[2015 2024]`
+
+## FT.ALIASADD
+
+```
+FT.ALIASADD <alias> <index-name>
+```
```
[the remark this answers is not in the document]
rileydes-improving: Matches exiting patterns

vmsdk/src/testing_infra/module.h
```diff
@@ -1333,14 +1334,16 @@ inline ValkeyModuleCallReply *TestValkeyModule_Call(ValkeyModuleCtx *ctx,
...
     const char *arg2 = va_arg(args, const char *);
     size_t arg2_len = va_arg(args, size_t);
     const char *arg3 = va_arg(args, const char *);
     size_t arg3_len = va_arg(args, size_t);
+    const char *arg4 = va_arg(args, const char *);
+    const char *arg5 = va_arg(args, const char *);
     auto ret = kMockValkeyModule->Call(ctx, cmdname, fmt, arg1, arg2, arg2_len,
-                                       arg3, arg3_len);
+                                       arg3, arg3_len, arg4, arg5);
     return ret;
```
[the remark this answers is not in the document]
rileydes-improving: Out of scope for this feature / pr, the existing branches have the same pattern.

integration/compatibility/generate.py
```diff
@@ -468,3 +468,62 @@ def test_search_sortby(self, key_type, dialect):
...
+
+    def test_aliasupdate_search(self, key_type, dialect):
+        """FT.ALIASUPDATE reassigns alias; queries via alias reflect the new target."""
+        self.setup_data("sortable numbers", key_type)
+        idx = f"{key_type}_idx1"
+        alias = f"{key_type}_alias_compat"
+        # Record alias management commands so replay recreates the same state.
+        self.execute_command(["FT.ALIASADD", alias, idx])
+        self.execute_command(["FT.ALIASUPDATE", alias, idx])
+        self.execute_command(["ft.search", alias, "@n1:[-inf inf]", "SORTBY", "n1", "ASC", "DIALECT", str(dialect)])
```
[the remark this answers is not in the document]
rileydes-improving: The simple version of test_aliasupdate_search (ALIASUPDATE to the same index) is sufficient for compatibility purposes, it confirms the command parses and executes the same way as Redis Stack. True reassignment behavior is thoroughly covered in test_ft_alias.py and test_ft_alias_cluster.py.

vmsdk/src/testing_infra/module.h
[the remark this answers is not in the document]
allenss-amazon: Meh, sure.

integration/compatibility/generate.py
```diff
@@ -468,3 +468,74 @@ def test_search_sortby(self, key_type, dialect):
...
+    def test_aliasupdate_search(self, key_type, dialect):
+        """FT.ALIASUPDATE reassigns alias; queries via alias reflect the new target."""
+        self.setup_data("sortable numbers", key_type)
+        idx = f"{key_type}_idx1"
+        alias = f"{key_type}_alias_update_compat"
+        self.execute_command(["FT.ALIASADD", alias, idx])
+        self.execute_command(["FT.ALIASUPDATE", alias, idx])
+        self.check(dialect, "ft.search", alias, "*", "LIMIT", "0", "15")
+        self.execute_command(["FT.ALIASDEL", alias])
+
```
allenss-amazon: Maybe add FT.INFO ?

integration/compatibility/generate.py
```diff
@@ -468,3 +468,74 @@ def test_search_sortby(self, key_type, dialect):
...
+        self.check(dialect, f"ft.aggregate {alias} * load 3 @__key @n1 @n2")
+        self.execute_command(["FT.ALIASDEL", alias])
+
+    def test_aliasupdate_search(self, key_type, dialect):
+        """FT.ALIASUPDATE reassigns alias; queries via alias reflect the new target."""
+        self.setup_data("sortable numbers", key_type)
+        idx = f"{key_type}_idx1"
+        alias = f"{key_type}_alias_update_compat"
+        self.execute_command(["FT.ALIASADD", alias, idx])
+        self.execute_command(["FT.ALIASUPDATE", alias, idx])
```
allenss-amazon: Do we test if we set an alias to a non-existent index? What happens to an alias if the target index is deleted?

If there's no synchronization between the alias table and the existing index table, then maybe this test case would be better if the first alias didn't exist -- that we we know that the update worked.

integration/compatibility/generate.py
```diff
@@ -468,3 +468,74 @@ def test_search_sortby(self, key_type, dialect):
...
+        self.check(dialect, "ft.search", alias, "*", "LIMIT", "0", "15")
+        self.execute_command(["FT.ALIASDEL", alias])
+
+
+
+@pytest.mark.parametrize("key_type", ["hash"])
+class TestAliasCompatibility(BaseCompatibilityTest):
+
```
allenss-amazon: This entire section doesn't test anything -- self.execute_command isn't hooked into the compatibility framework. It looks like it's redundant with the code above, maybe delete it??

Does removing this section reduce the needed changes to data_sets.py ?

integration/compatibility_test.py
```diff
@@ -381,22 +390,58 @@ def mark_as_failed(testname):
...
     global correct_answers, failed_tests, passed_tests
     if (expected['data_set_name'], expected['key_type'], expected.get('schema_type')) != data_set:
         print("Loading data set:", expected['data_set_name'], "key type:", expected['key_type'])
         client.execute_command("FLUSHALL SYNC")
         load_data(client, expected['data_set_name'], expected['key_type'], schema_type=expected.get('schema_type', 'default'))
-        waiters.wait_for_true(lambda: IndexingTestHelper.is_indexing_complete_on_node(client, f"{expected['key_type']}_idx1"))
+        # Derive the index name to wait on from the dataset's first CREATE command
+        # rather than hardcoding per-dataset names.
+        index_to_wait = _first_index_name(expected['data_set_name'], expected['key_type'])
+        waiters.wait_for_true(lambda: IndexingTestHelper.is_indexing_complete_on_node(client, index_to_wait))
```
allenss-amazon, line 443: Why do we need to change this? FT.INFO should work equally well on an aliased and a non-aliased index.

integration/compatibility_test.py
```diff
@@ -381,22 +390,58 @@ def mark_as_failed(testname):
...
+    if data_set_name == "alias":
+        data = compute_alias_data()
+        key_type = "hash"
+    elif data_set_name in TEXT_DATASETS:
+        data = compute_text_data_sets(data_set_name)
+    else:
+        data = compute_data_sets()
+    create_cmd = data[data_set_name][CREATES_KEY(key_type)][0]
+    # FT.CREATE <index_name> ... — index name is the second token.
+    return create_cmd.split()[1]
```
allenss-amazon, line 432: Pretty sure we don't need this, see below.

integration/compatibility_test.py
```diff
@@ -463,6 +508,36 @@ def do_answer_cluster(cluster_client, expected, data_set, test_case):
 
         data_set = next_data_set
 
+    # Replay alias setup commands without counting them in the pass/fail tally.
```
allenss-amazon: It's fine to count them in the total. No real reason to exclude them....

integration/compatibility_test.py
```diff
@@ -511,7 +587,7 @@ def test_answers(self, answers):
         for i in range(len(answers)):
             data_set = do_answer(client, answers[i], data_set)
 
-        expected_count = sum(1 for a in answers if not a.get('excluded'))
+        expected_count = sum(1 for a in answers if not a.get('excluded') and not _is_alias_management_cmd(a.get('cmd', [])))
```
allenss-amazon: no reason to excluded these....

integration/test_ft_alias.py
```diff
@@ -0,0 +1,785 @@
+"""Integration tests for FT.ALIASADD, FT.ALIASDEL, and FT.ALIASUPDATE."""
```
allenss-amazon, line 1: We should add FT.ALIASLIST. It's a horrible oversight on Redis's part.

integration/test_ft_alias.py
```diff
@@ -0,0 +1,785 @@
...
+        assert info_list[idx + 1] == INDEX_NAME.encode()
+
+    def test_aliasadd_duplicate_alias(self):
+        """Adding the same alias twice returns an error."""
+        client = self.client
+        assert client.execute_command(*CREATE_TAG_INDEX) == b"OK"
+        assert client.execute_command("FT.ALIASADD", ALIAS_NAME, INDEX_NAME) == b"OK"
+        with pytest.raises(ResponseError) as exc_info:
+            client.execute_command("FT.ALIASADD", ALIAS_NAME, INDEX_NAME)
+        assert "Alias already exists" in str(exc_info.value)
```
allenss-amazon: Need a compatibility test of this case.

integration/test_ft_alias.py
```diff
@@ -0,0 +1,785 @@
...
+
+    def test_aliasadd_nonexistent_index(self):
+        """Adding an alias for a non-existent index returns an error."""
+        client = self.client
+        with pytest.raises(ResponseError) as exc_info:
+            client.execute_command("FT.ALIASADD", ALIAS_NAME, "no_such_index")
+        assert "no_such_index" in str(exc_info.value)
```
allenss-amazon: same, need compatibility test.

integration/test_ft_alias.py
```diff
@@ -0,0 +1,785 @@
...
+        assert client.execute_command("FT.ALIASADD", ALIAS_NAME, INDEX_NAME) == b"OK"
+        with pytest.raises(ResponseError) as exc_info:
+            client.execute_command("FT.ALIASADD", "alias2", ALIAS_NAME)
+        assert "Unknown index name or name is an alias" in str(exc_info.value)
+
+    def test_aliasadd_multiple_aliases_same_index(self):
+        """Multiple aliases can point to the same index."""
+        client = self.client
+        assert client.execute_command(*CREATE_TAG_INDEX) == b"OK"
+        assert client.execute_command("FT.ALIASADD", "alias_a", INDEX_NAME) == b"OK"
```
allenss-amazon: same, need compatibility testing of these cases

integration/test_ft_alias.py
```diff
@@ -0,0 +1,785 @@
...
+        # Underlying index still accessible
+        assert client.execute_command("FT.INFO", INDEX_NAME) is not None
+
+    def test_aliasdel_nonexistent_alias(self):
+        """Deleting a non-existent alias returns an error."""
+        client = self.client
+        with pytest.raises(ResponseError) as exc_info:
+            client.execute_command("FT.ALIASDEL", "no_such_alias")
+        assert "Alias does not exist" in str(exc_info.value)
+
```
allenss-amazon: need compatibilty

integration/test_ft_alias.py
```diff
@@ -0,0 +1,785 @@
...
+        """After deleting an alias it can be re-added."""
+        client = self.client
+        assert client.execute_command(*CREATE_TAG_INDEX) == b"OK"
+        assert client.execute_command("FT.ALIASADD", ALIAS_NAME, INDEX_NAME) == b"OK"
+        assert client.execute_command("FT.ALIASDEL", ALIAS_NAME) == b"OK"
+        assert client.execute_command("FT.ALIASADD", ALIAS_NAME, INDEX_NAME) == b"OK"
+        assert client.execute_command("FT.INFO", ALIAS_NAME) is not None
+
+
+class TestFTAliasUpdate(ValkeySearchTestCaseBase):
```
allenss-amazon: These cases also need to be tested in CME.

integration/test_ft_alias.py
```diff
@@ -0,0 +1,778 @@
...
+        client.execute_command("HSET", "doc:3", "category", "books")
+        assert client.execute_command("FT.ALIASADD", ALIAS_NAME, INDEX_NAME) == b"OK"
+
+        result = client.execute_command(
+            "FT.AGGREGATE", ALIAS_NAME, "@category:{books}",
+            "GROUPBY", "1", "@category",
+            "REDUCE", "COUNT", "0", "AS", "count",
+        )
+        # result[0] is the number of groups
+        assert result[0] >= 1
```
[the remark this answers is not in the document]
allenss-amazon: Incorrect. writing a hash to an existing index is synchronous -- thus issuing a query from the same connection doesn't require any additional waiting and is not subject to races.

integration/test_ft_alias.py
```diff
@@ -0,0 +1,785 @@
...
+    info = client.execute_command("FT.INFO", name)
+    info_list = list(info)
+    idx = next(
+        (i for i, v in enumerate(info_list) if v == b"aliases"), None
+    )
+    assert idx is not None, "FT.INFO response missing 'aliases' field"
+    aliases = info_list[idx + 1]
+    if aliases is None:
+        return []
+    return sorted(aliases) if isinstance(aliases, list) else sorted(list(aliases))
```
allenss-amazon: There is an existing FT INFO parser object, let's update that and use it here.

integration/test_ft_alias.py
```diff
@@ -0,0 +1,785 @@
...
+        assert client.execute_command("FT.INFO", ALIAS_NAME) is not None
+
+class TestFTAliasNameBoundaries(ValkeySearchTestCaseBase):
+    """Boundary tests for alias name values."""
+
+    def test_aliasadd_empty_alias_name_accepted(self):
+        """
+        FT.ALIASADD with an empty alias name is accepted by the server, matching
+        RediSearch behaviour (verified against redis-stack-server v21020:
+        FT.ALIASADD "", FT.ALIASUPDATE "", and FT.ALIASDEL "" all return OK).
```
allenss-amazon, line 609: Can I then actually FT.SEARCH with a 0-length index name?

src/commands/ft_alias_add.cc
```diff
@@ -0,0 +1,65 @@
...
+    unsigned timeout_ms = options::GetFTInfoTimeoutMs().GetValue();
+    auto op = new AliasExistsConsistencyCheckFanoutOperation(
+        ValkeyModule_GetSelectedDb(ctx), std::string(alias), timeout_ms, ifv);
+    op->StartOperation(ctx);
+  } else {
+    if (is_loading || inside_multi_exec) {
+      VMSDK_LOG(NOTICE, nullptr) << "The server is loading AOF or inside "
+                                    "multi/exec or lua script, skip "
+                                    "fanout operation";
+    }
```
allenss-amazon: This should be a "not supported" style error message.

src/commands/ft.aliasadd.json
```diff
@@ -0,0 +1,25 @@
...
+            {
+                "key_spec_index": 0,
+                "name": "index-name",
+                "type": "key"
+            }
+        ],
+        "arity": 3,
+        "complexity": "O(1)",
+        "group": "search",
+        "module_since": "1.0.0",
```
allenss-amazon: 1.0 is the wrong value here. Put in 1.3.0

src/commands/ft.aliasdel.json
```diff
@@ -0,0 +1,20 @@
...
+        "arguments": [
+            {
+                "name": "alias",
+                "type": "string"
+            }
+        ],
+        "arity": 2,
+        "complexity": "O(1)",
+        "group": "search",
+        "module_since": "1.0.0",
```
allenss-amazon: 1.3.0

src/commands/ft.aliasupdate.json
```diff
@@ -0,0 +1,25 @@
...
+            {
+                "key_spec_index": 0,
+                "name": "index-name",
+                "type": "key"
+            }
+        ],
+        "arity": 3,
+        "complexity": "O(1)",
+        "group": "search",
+        "module_since": "1.0.0",
```
allenss-amazon: 1.3.0

src/commands/ft_alias_del.cc
```diff
@@ -0,0 +1,58 @@
...
+    unsigned timeout_ms = options::GetFTInfoTimeoutMs().GetValue();
+    auto op = new AliasRemovedConsistencyCheckFanoutOperation(
+        ValkeyModule_GetSelectedDb(ctx), std::string(alias), timeout_ms);
+    op->StartOperation(ctx);
+  } else {
+    if (is_loading || inside_multi_exec) {
+      VMSDK_LOG(NOTICE, nullptr) << "The server is loading AOF or inside "
+                                    "multi/exec or lua script, skip "
+                                    "fanout operation";
+    }
```
allenss-amazon: Should be an error, not a no-op.

src/commands/ft_alias_update.cc
```diff
@@ -0,0 +1,65 @@
...
+    ifv.set_version(schema->GetVersion());
+    unsigned timeout_ms = options::GetFTInfoTimeoutMs().GetValue();
+    auto op = new AliasExistsConsistencyCheckFanoutOperation(
+        ValkeyModule_GetSelectedDb(ctx), std::string(alias), timeout_ms, ifv);
+    op->StartOperation(ctx);
+  } else {
+    if (is_loading || inside_multi_exec) {
+      VMSDK_LOG(NOTICE, nullptr) << "The server is loading AOF or inside "
+                                    "multi/exec or lua script, skip "
+                                    "fanout operation";
```
allenss-amazon: error.

src/commands/ft_internal_update.cc
```diff
@@ -55,12 +55,33 @@ absl::Status HandleInternalUpdateFailure(
 
 absl::Status FTInternalUpdateCmd(ValkeyModuleCtx *ctx,
                                  ValkeyModuleString **argv, int argc) {
-  CHECK_EQ(argc, kFTInternalUpdateArgCount)
+  CHECK(argc >= kFTInternalUpdateMinArgCount)
```
allenss-amazon: Let's convert this from a crash to an error. It's long-standing mistake in the design of this command. It will mean that the DEATH tests for this case will need to get fixed also.

src/commands/ft_internal_update.cc
```diff
@@ -55,12 +55,33 @@ absl::Status HandleInternalUpdateFailure(
...
+
+  // Parse keyword/value pairs from argv[4..argc-1].
+  // Recognized keywords: TYPE (metadata type name).
+  // Unrecognized keywords are silently ignored for forward compatibility.
+  absl::string_view type_name = kSchemaManagerMetadataTypeName;
+  for (int i = 4; i + 1 < argc; i += 2) {
+    auto key = vmsdk::ToStringView(argv[i]);
+    if (key == "TYPE") {
+      type_name = vmsdk::ToStringView(argv[i + 1]);
+    }
```
allenss-amazon: Thanks for thinking about extensibility here. But this design only allows extensions for keyword and single value. Let's modify this so that we can extend with keyword, arg count and # args value -- which is more extensible and consistent with syntax style.

src/coordinator/coordinator.proto
```diff
@@ -172,6 +172,11 @@ message GlobalMetadata {
   GlobalMetadataVersionHeader version_header = 1;
   // Nested map. Outer map is namespace, inner map is ID to entry.
   map<string, GlobalMetadataEntryMap> type_namespace_map = 2;
+  // Serialized valkey_search.data_model.AliasMap proto for standalone
+  // (non-coordinator) mode persistence. In coordinator mode, aliases are
+  // stored in type_namespace_map via MetadataManager. In standalone mode,
+  // this field carries all aliases.
+  bytes alias_map_bytes = 3;
```
allenss-amazon, line 179: Why does this need extension here? Why isn't the alias just another type in the type_namespace_map ??

src/coordinator/coordinator.proto
```diff
@@ -234,3 +239,19 @@ message InfoIndexPartitionResponse {
...
+  uint32 db_num = 1;
+  string alias = 2;
+  string index_name = 3;
+}
+
+// Collection of alias entries for RDB persistence in standalone mode.
+// Serialized into GlobalMetadata.alias_map_bytes.
+message AliasMap {
+  repeated AliasEntry entries = 1;
+}
```
allenss-amazon, line 257: I think this object hierarchy is wrong, it should be able to extend the existing mechanism which provides versioning, collision detection, consistency resolution.

src/coordinator/metadata_manager.cc
```diff
@@ -567,9 +569,14 @@ absl::Status MetadataManager::ReconcileMetadata(const GlobalMetadata &proposed,
...
-        if (!status.ok()) {
-          return status;
+        // Only replicate via FT.INTERNAL_UPDATE if the type is registered
+        // locally — unregistered types are stored but not acted upon.
+        if (registered_types_.Get().contains(type_name)) {
+          auto status = CallFTInternalUpdateForReconciliation(
+              id, proposed_entry, type_name);
+          if (!status.ok()) {
+            return status;
+          }
```
allenss-amazon, line 579: What is the logic here? Does this bypass the resolution machinery?

src/coordinator/metadata_manager.cc
```diff
@@ -984,16 +992,27 @@ absl::Status MetadataManager::CallFTInternalUpdateForReconciliation(
...
+                           header_binary.data(), header_binary.size());
+  } else {
+    std::string type_name_str(type_name);
+    // Replicate FT.INTERNAL_UPDATE to replicas for AOF consistency
+    ValkeyModule_Replicate(detached_ctx_.get(), "FT.INTERNAL_UPDATE", "cbbcc",
+                           std::string(encoded_id).c_str(),
+                           metadata_binary.data(), metadata_binary.size(),
+                           header_binary.data(), header_binary.size(), "TYPE",
+                           type_name_str.c_str());
+  }
```
allenss-amazon: There's no requirement to support rolling upgrades in this situation.

src/index_schema.cc
```diff
@@ -1096,6 +1097,12 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
   ValkeyModule_ReplyWithSimpleString(ctx, "index_name");
   ValkeyModule_ReplyWithSimpleString(ctx, name_.data());
 
+  ValkeyModule_ReplyWithSimpleString(ctx, "aliases");
+  ValkeyModule_ReplyWithArray(ctx, aliases.size());
+  for (const auto &alias : aliases) {
+    ValkeyModule_ReplyWithSimpleString(ctx, alias.c_str());
```
allenss-amazon, line 1103: This will fail if an alias has an embedded null byte. Either we should strict them from being created OR preserve their value here.

src/schema_manager.cc
```diff
@@ -129,6 +129,33 @@ SchemaManager::SchemaManager(
...
               }});
+  // In standalone mode, register for RDB_SECTION_GLOBAL_METADATA to persist
+  // aliases. In coordinator mode, MetadataManager owns that section and aliases
+  // are already stored in its type_namespace_map — no separate save needed.
+  if (!coordinator_enabled) {
+    RegisterRDBCallback(
+        data_model::RDB_SECTION_GLOBAL_METADATA,
+        RDBSectionCallbacks{
+            .load = [this](ValkeyModuleCtx *ctx,
+                           std::unique_ptr<data_model::RDBSection> section,
```
allenss-amazon, line 141: Let's revisit this code when the mapping of alias table is redone.

src/schema_manager.cc
```diff
@@ -257,12 +296,209 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> SchemaManager::GetIndexSchema(
...
   }
   return existing_entry.value();
 }
 
+void SchemaManager::AddToReverseAliasMap(uint32_t db_num,
+                                         absl::string_view index_name,
+                                         absl::string_view alias) {
+  db_to_index_to_aliases_[db_num][std::string(index_name)].insert(
+      std::string(alias));
+}
```
allenss-amazon: This overwrites any existing entries. Is that correct? Maybe this should be "UpdateReverseAliasMap" ??
If it's guarnateed not to exist, let's CHECK for that, this isn't a critical speed path.

src/schema_manager.cc
```diff
@@ -257,12 +296,209 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> SchemaManager::GetIndexSchema(
...
+  }
+  if (db_alias_it != db_to_aliases_.end() &&
+      db_alias_it->second.contains(alias)) {
+    return absl::AlreadyExistsError("Alias already exists");
+  }
+  if (!LookupInternal(db_num, index_name).ok()) {
+    return GenerateIndexNotFoundError(db_num, index_name);
+  }
+  db_to_aliases_[db_num][std::string(alias)] = std::string(index_name);
+  AddToReverseAliasMap(db_num, index_name, alias);
```
allenss-amazon: Since the forward and reverse alias maps need to be updated in lock-step, I think it would be better to have helper routines that update both, rather than splitting the code into two places.

src/schema_manager.cc
```diff
@@ -257,12 +296,209 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> SchemaManager::GetIndexSchema(
...
+      }
+    }
+    coordinator::AliasEntry entry;
+    entry.set_index_name(std::string(index_name));
+    auto any_proto = std::make_unique<google::protobuf::Any>();
+    any_proto->PackFrom(entry);
+    return coordinator::MetadataManager::Instance()
+        .CreateEntry(kAliasMetadataTypeName,
+                     coordinator::ObjName(db_num, alias), std::move(any_proto))
+        .status();
```
allenss-amazon, line 418: This seems like it's using the type system correctly.

src/schema_manager.cc
```diff
@@ -257,12 +296,209 @@ absl::StatusOr<std::shared_ptr<IndexSchema>> SchemaManager::GetIndexSchema(
...
+  // Remove old reverse mapping if the alias already exists.
+  if (db_alias_it != db_to_aliases_.end()) {
+    auto old_it = db_alias_it->second.find(alias);
+    if (old_it != db_alias_it->second.end()) {
+      RemoveFromReverseAliasMap(db_num, old_it->second, alias);
+    }
+  }
+  // ALIASUPDATE is an upsert: creates the alias if absent, reassigns if
+  // present.
+  db_to_aliases_[db_num][std::string(alias)] = std::string(index_name);
```
allenss-amazon: again, let's combine all forward & reverse map updates into single routines.

src/schema_manager.cc
```diff
@@ -324,6 +606,21 @@ absl::flat_hash_set<std::string> SchemaManager::GetIndexSchemasInDB(
...
+  std::vector<std::string> result;
+  auto rev_db_it = db_to_index_to_aliases_.find(db_num);
+  if (rev_db_it != db_to_index_to_aliases_.end()) {
+    auto rev_idx_it = rev_db_it->second.find(index_name);
+    if (rev_idx_it != rev_db_it->second.end()) {
+      result.assign(rev_idx_it->second.begin(), rev_idx_it->second.end());
+    }
+  }
+  std::sort(result.begin(), result.end());
+  return result;
```
allenss-amazon, line 643: This silently returns an empty list if the db,index doesn't exist. Is that really the right behavior?

src/schema_manager.cc
```diff
@@ -556,6 +938,52 @@ void SchemaManager::OnSwapDB(ValkeyModuleSwapDbInfo *swap_db_info) {
   for (auto &schema : db_to_index_schemas_[swap_db_info->dbnum_second]) {
     schema.second->OnSwapDB(swap_db_info);
   }
+  // Swap alias maps in lockstep with index schemas.
```
allenss-amazon, line 956: Why a lambda that's only called once? Why not inline code?

src/schema_manager.cc
```diff
@@ -556,6 +938,52 @@ void SchemaManager::OnSwapDB(ValkeyModuleSwapDbInfo *swap_db_info) {
...
+      auto inner = std::move(it_b->second);
+      db_to_aliases_.emplace(a, std::move(inner));
+      db_to_aliases_.erase(b);
+    } else {
+      std::swap(it_a->second, it_b->second);
+    }
+  };
+  swap_aliases(swap_db_info->dbnum_first, swap_db_info->dbnum_second);
+  // Swap reverse alias maps in lockstep with forward alias maps.
+  auto swap_rev_aliases = [this](uint32_t a, uint32_t b) {
```
allenss-amazon: same.

src/schema_manager.cc
```diff
@@ -637,6 +1069,79 @@ absl::Status SchemaManager::SaveIndexes(ValkeyModuleCtx *ctx, SafeRDB *rdb,
...
+  if (when == VALKEYMODULE_AUX_BEFORE_RDB) {
+    return absl::OkStatus();
+  }
+  // All aliases across all DBs are packed into a single RDB section proto
+  // embedded in the GlobalMetadata section. This is intentional: aliases are
+  // lightweight (two strings per entry) and their total count is implicitly
+  // bounded by the max-indexes config.
+  absl::MutexLock lock(&db_to_index_schemas_mutex_);
+
+  coordinator::AliasMap alias_map_proto;
```
allenss-amazon, line 1092: I think this is better handled with a new supplemental section.

COMMANDS.md
```diff
@@ -308,3 +311,47 @@ The following query will return all books with "comedy" or "horror" genre (OR) p
...
+
+Adds an alias to an existing index. The alias can then be used in place of the index name in any command that accepts an index name (e.g., `FT.SEARCH`, `FT.AGGREGATE`).
+
+- **\<alias\>** (required): The alias name to create.
+- **\<index-name\>** (required): The name of an existing index to associate with the alias.
+
+An alias cannot point to another alias. Multiple aliases may point to the same index.
+
+> **Note:** `FT.ALIASADD` permits an alias whose name matches an existing index. Direct index lookup takes priority, so the alias is shadowed and unreachable until the index is dropped. This matches RediSearch.
```
allenss-amazon: Let's not characterize the behavior of  Redisearch.

COMMANDS.md
```diff
@@ -308,3 +311,47 @@ The following query will return all books with "comedy" or "horror" genre (OR) p
...
+```
+FT.ALIASUPDATE <alias> <index-name>
+```
+
+Adds or updates an alias. If the alias does not exist it is created (equivalent to `FT.ALIASADD`). If it already exists, it is atomically reassigned to the new index, enabling zero-downtime schema migrations.
+
+- **\<alias\>** (required): The alias name to create or update.
+- **\<index-name\>** (required): The name of an existing index to associate with the alias.
+
+> **Note:** Unlike `FT.ALIASADD`, `FT.ALIASUPDATE` rejects an alias whose name matches an existing index. This matches RediSearch.
```
allenss-amazon: Let's not characterize Redisearch.

allenss-amazon (changes requested):
1. Let's add FT.ALIASLIST. It should have always been implemented.
2. Need CME testing of all alias commands, distribution across the cluster is required.
3. I didn't see any propagation failure mode testing.
4. Why isn't the existing top-level metadata protobuf definition sufficient? It looks like maybe this is left over code from a previous edit -- because there appears to be code that properly uses the typed namespace too.
5. Should an RDB save/restore be a new supplemental section?

src/commands/ft.aliasadd.json
rileydes-improving: Fixed in fcaa151b28fc332c3fb4e0932c10dc5619983b4f

src/commands/ft.aliasdel.json
rileydes-improving: Fixed in fcaa151b28fc332c3fb4e0932c10dc5619983b4f

src/commands/ft.aliasupdate.json
rileydes-improving: Fixed in fcaa151b28fc332c3fb4e0932c10dc5619983b4f

src/commands/ft_internal_update.cc
rileydes-improving: Fixed in 3baa1ebb451e78058f09baebf48bcff17ad17930

COMMANDS.md
rileydes-improving: Fixed in 59890842d9884d04540e3666c088f64a4b410dab

COMMANDS.md
rileydes-improving: Fixed in 59890842d9884d04540e3666c088f64a4b410dab

allenss-amazon:
We need to chat about the approach that you're using for aliases. Please find some time for us to connect.

src/commands/ft_alias_del.cc
rileydes-improving: Fixed in 9857b398f437725e980ccef95ffb83a05e8a2d23. Also moved the guard before the actual modification call so that the error doesn't cause a desync.

src/commands/ft_alias_update.cc
rileydes-improving: Fixed in https://github.com/valkey-io/valkey-search/commit/9857b398f437725e980ccef95ffb83a05e8a2d23. Also moved the guard before the actual modification call so that the error doesn't cause a desync.

src/commands/ft_internal_update.cc
rileydes-improving: Implemented in f08d7d588a1f5dbcfc351aaaa608af8a41ad2cbf, should give enough extensibility for future expansions.

src/coordinator/metadata_manager.cc
rileydes-improving: Sounds good, simplified in f08d7d588a1f5dbcfc351aaaa608af8a41ad2cbf.

src/index_schema.cc
rileydes-improving: Added alias name rejecton with a null byte in 1b9a25c31a7bdbc54a6c560f9acaae4397d60126. Handles it in create or update at the schema manager level. I think being strict during attempted modifications is probably the right choice here.

src/commands/ft_alias_add.cc
rileydes-improving: Fixed in https://github.com/valkey-io/valkey-search/commit/9857b398f437725e980ccef95ffb83a05e8a2d23. Made into a proper error and made the message "MULTI/EXEC or Lua script are not supported in CME mode."

src/schema_manager.cc
rileydes-improving: Split into helper routines and used them instead in c139d763fa52531f875ab49c4bac9a2843d14168.

src/schema_manager.cc
rileydes-improving: plit into helper routines and used them instead in https://github.com/valkey-io/valkey-search/commit/c139d763fa52531f875ab49c4bac9a2843d14168.

src/schema_manager.cc
rileydes-improving: Inlined unnecesarry lambdas in 3eef501a3160c4a8785d444ae7a962f257b1faee. Kept them in scoped blocks to keep the local variables scoped without leaking into the function body.

src/schema_manager.cc
rileydes-improving: Inlined unnecesarry lambdas in 3eef501a3160c4a8785d444ae7a962f257b1faee. Kept them in scoped blocks to keep the local variables scoped without leaking into the function body.

integration/compatibility_test.py
rileydes-improving: Fix done in 46d4f46b007e0e7420fd1ca5393fbd530340ce5d. Comment updated in 8a648222b1eaaa6fe2e1755667262b884a0475b8.

integration/test_ft_alias.py
rileydes-improving: Added Alias List in a7cfabdaee964494005eb2131b7e6c5dd62ca717 and added the test suite for it in 12a78f50b6a374af6c238051c59b16a550d175af.

integration/test_ft_alias.py
rileydes-improving: Sounds good, implemented in 8cac5cad0e540314f609ed3c5658f0a4551824f2.

src/schema_manager.cc
rileydes-improving: Added in the check in c139d763fa52531f875ab49c4bac9a2843d14168

rileydes-improving:
New architecture discussed with Allen will be posted in a new branch/PR soon.

PR #959, opened 2026-04-09
Add support for FT.EXPLAINCLI 

KarthikSubbarao:
PR to add support for FT.EXPLAINCLI 

```
127.0.0.1:6379> ft.explaincli idx:products_full 'hello, how are you doing? is the weather (bad | good | great)?'
 1) "AND{"
 2) "  TEXT-TERM(\"hello\", field=*)"
 3) "  TEXT-TERM(\"how\", field=*)"
 4) "  TEXT-TERM(\"you\", field=*)"
 5) "  TEXT-TERM(\"doing\", field=*)"
 6) "  TEXT-TERM(\"weather\", field=*)"
 7) "  OR{"
 8) "    TEXT-TERM(\"bad\", field=*)"
 9) "    TEXT-TERM(\"good\", field=*)"
10) "    TEXT-TERM(\"great\", field=*)"
11) "  }"
12) "}"
127.0.0.1:6379> ft.explaincli idx:products_full 'Wireless %eraphones% with active noise canc*' 
1) "AND{"
2) "  TEXT-TERM(\"wireless\", field=*)"
3) "  TEXT-FUZZY(\"eraphones\", distance=1, field=*)"
4) "  TEXT-TERM(\"active\", field=*)"
5) "  TEXT-TERM(\"noise\", field=*)"
6) "  TEXT-PREFIX(\"canc\", field=*)"
7) "}"
```


allenss-amazon (commented):
FieldMask makes no sense to a user, we need to do a better job of handling that.

Can't we also implement the non-CLI form? 

The testing could use the non-CLI form much easer and actually look at the results of some of these commands (which I don't think the current testing actually does).

KarthikSubbarao:
@allenss-amazon , FT.EXPLAINCLI is both unit tested and integration tested. The filter parser unit test covers a variety of cases on the PrintPredicateTree fn


I also changed the fn to return the alias name and not the fieldmask

Issue #957, opened 2026-04-08
[FEATURE] FT.SEARCH Wildcard treated as invalid query string syntax.


rileydes-improving:
**Describe the bug**

`FT.SEARCH <index> "*" DIALECT 2` returns `Invalid query string syntax` in Valkey-Search for non-vector queries. Redis Stack accepts `*` as a match-all and returns every document in the index. This breaks any client code ported from Redis Stack that uses `*` to enumerate or filter documents, including combinations with `SORTBY`, `LIMIT`, `NOCONTENT`, or filter expressions.

**To Reproduce**

Add the following to `generate.py`, regenerate the answers file and run the compatibility test. 

As an alternative, a branch with the compatibility test and answers file regenerated is available at https://github.com/Bit-Quill/valkey-search/tree/bug/ft-search-bare-wildcard

```diff
--- a/integration/compatibility/generate.py
+++ b/integration/compatibility/generate.py
@@ -472,3 +472,16 @@ class TestAggregateCompatibility(BaseCompatibilityTest):
                             self.check(dialect, f"ft.search {key_type}_idx1 * SORTBY {sort_key} {direction} {return_keys} {limit} {wsk}")
 
 
+@pytest.mark.parametrize("dialect", [2])
+@pytest.mark.parametrize("key_type", ["json", "hash"])
+class TestSearchCompatibility(BaseCompatibilityTest):
+    ANSWER_FILE_NAME = "search-answers.pickle.gz"
+
+    def test_bare_wildcard(self, key_type, dialect):
+        self.setup_data("sortable numbers", key_type)
+        self.check("ft.search", f"{key_type}_idx1", "*", "DIALECT", str(dialect))
+        self.check("ft.search", f"{key_type}_idx1", "*", "NOCONTENT", "DIALECT", str(dialect))
+        self.check("ft.search", f"{key_type}_idx1", "*", "LIMIT", "0", "3", "DIALECT", str(dialect))
+        self.check("ft.search", f"{key_type}_idx1", "*", "SORTBY", "n1", "ASC", "DIALECT", str(dialect))
```

```diff
--- a/integration/compatibility_test.py
+++ b/integration/compatibility_test.py
@@ -494,7 +494,7 @@ class TestAnswersCMD(ValkeySearchTestCaseBase):
-    @pytest.mark.parametrize("answers", ["aggregate-answers.pickle.gz", "text-search-answers.pickle.gz"])
[...]

PR #960, opened 2026-04-09
Fix FT.SEARCH bare wildcard execution

bandalgomsu:
Fix non-vector FT.SEARCH so a bare "*" query executes instead of failing with invalid syntax. Treat bare wildcard queries as match-all searches and run them through the universal key fetch path.

issue : #957

### DIFF
--- a/src/query/search.cc
+++ b/src/query/search.cc
@@ -564,9 +564,17 @@ absl::StatusOr<std::vector<indexes::Neighbor>> MaybeAddIndexedContent(
 absl::StatusOr<std::vector<indexes::Neighbor>> SearchNonVectorQuery(
     const SearchParameters &parameters) {
   std::queue<std::unique_ptr<indexes::EntriesFetcherBase>> entries_fetchers;
-  size_t qualified_entries = EvaluateFilterAsPrimary(
-      parameters, parameters.filter_parse_results.root_predicate.get(),
-      entries_fetchers, false);
+  size_t qualified_entries = 0;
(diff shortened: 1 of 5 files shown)


allenss-amazon (approved):

src/query/search.cc
```diff
@@ -564,9 +564,17 @@ absl::StatusOr<std::vector<indexes::Neighbor>> MaybeAddIndexedContent(
 absl::StatusOr<std::vector<indexes::Neighbor>> SearchNonVectorQuery(
     const SearchParameters &parameters) {
   std::queue<std::unique_ptr<indexes::EntriesFetcherBase>> entries_fetchers;
-  size_t qualified_entries = EvaluateFilterAsPrimary(
-      parameters, parameters.filter_parse_results.root_predicate.get(),
-      entries_fetchers, false);
+  size_t qualified_entries = 0;
+  if (parameters.filter_parse_results.is_match_all) {
+    auto universal_fetcher = std::make_unique<indexes::UniversalSetFetcher>(
```
KarthikSubbarao, line 573: In case a user provides `*`, we should return all the returns, so `UniversalSetFetcher` makes sense.

But we should not spend time on the prefilter evaluator in the background reader thread.

You can checked the Unsolved Query fn here.

Let's add an exception to that rule to say it should not include `is_match_all`.

it will help keep the performance on reader threads better

KarthikSubbarao (approved):

PrashantAshok:
@KarthikSubbarao - when's the plan to release 1.3.0?

PR #961, opened 2026-04-09
fix test_dropindex_with_background_queries heap-use-after-free leak

Frank-Gu-81:
The `mutation_processing` pause point was never reset. Added reset to fix heap-use-after-free leak.

```
PASSED [ 61%]=================================================================
==17560==ERROR: AddressSanitizer: heap-use-after-free on address 0xffffb2319c48 at pc 0xffff95c0dc30 bp 0xffff89eeb450 sp 0xffff89eeb440
READ of size 8 at 0xffffb2319c48 thread T20
ASAN:DEADLYSIGNAL
AddressSanitizer: nested bug in the same thread, aborting.
```

### DIFF
--- a/src/server_events.cc
+++ b/src/server_events.cc
@@ -54,9 +54,8 @@ void OnServerCronCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
 
 void OnShutdownCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
                         uint64_t subevent, void *data) {
-  // Mark the module as shutting down so that RunByMain() calls from
-  // background threads execute their callbacks directly instead of
-  // posting to the (now-dead) event loop.
+  // Mark the module as shutting down so that RunByMain() stops
+  // scheduling new tasks.
   vmsdk::MarkAsShuttingDown();
   // Clear all PausePoints so any waiting worker threads wake up and exit
   // their spin loops, then join all thread pools so every in-flight task
--- a/src/server_events.cc
+++ b/src/server_events.cc
@@ -13,6 +13,7 @@
 #include "src/schema_manager.h"
 #include "src/valkey_search.h"
 #include "vmsdk/src/debug.h"
+#include "vmsdk/src/utils.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
 namespace valkey_search::server_events {
@@ -53,9 +54,20 @@ void OnServerCronCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
 
 void OnShutdownCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
                         uint64_t subevent, void *data) {
-  // Clear all PausePoints first so any waiting worker threads wake up and
-  // exit their spin loops before global destructors destroy the map.
+  // Mark the module as shutting down so that RunByMain() calls from
+  // background threads execute their callbacks directly instead of
+  // posting to the (now-dead) event loop.
+  vmsdk::MarkAsShuttingDown();
+  // Clear all PausePoints so any waiting worker threads wake up and exit
+  // their spin loops, then join all thread pools so every in-flight task
+  // completes before we tear down index schemas.  This ordering matters:
+  //   1. ClearAllPausePoints  – unblocks workers stuck in PausePoint().
+  //   2. JoinAllThreadPools   – drains task queues and waits for every
+  //      worker thread to exit.
+  //   3. OnShutdownCallback   – removes all index schemas on the main
+  //      thread.
   vmsdk::debug::ClearAllPausePoints();
+  ValkeySearch::Instance().JoinAllThreadPools();
   SchemaManager::Instance().OnShutdownCallback(ctx, eid, subevent, data);
 }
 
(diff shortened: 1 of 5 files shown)


BCathcart (approved):

BCathcart:
Do you know why #958 doesn't fix this already?

Frank-Gu-81:
#958 prevents the global pause_point_waiters map from being destroyed while threads are still spinning in PausePoint(). This issue is in the test itself: the test resets background_search_completing but never resets mutation_processing, so the hset thread stays blocked on that pausepoint holding a reference to the index. When FT.DROPINDEX frees the index and the server shuts down, that thread accesses freed memory.

BCathcart:
From a comment in #958:
>  // Clear all PausePoints first so any waiting worker threads wake up and
  // exit their spin loops before global destructors destroy the map.

I'm just pointing out that maybe we should update the shutdown logic to wait until all the worker threads exit or kill them. We're relying on the timing so that the worker threads exit the spin loop before the shutdown routine and don't access the `pause_point_waiters` map before it's destroyed. I guess this usually works in practice but it still isn't guaranteed. In this test the shutdown routine is clearly running before the worker thread exits. It has gotten out of the spin loop presumably but is still accessing other memory.

Frank-Gu-81:
Addresses the review feedback from #958 about relying on timing during shutdown.

**Problem**

1. Worker threads stuck in `PausePoint()` spin loops could still be
   accessing memory after `ClearAllPausePoints()` returned, because
   nothing waited for them to actually finish.

2. When a worker thread held the last `shared_ptr<IndexSchema>` and woke
   up during shutdown, the IndexSchema destructor called `RunByMain()`
   from the worker thread. RunByMain posted a callback to the event
   loop, but the event loop was dead. The heap-allocated callback was
   never executed or freed (32-byte LSAN leak).

**Fix**

- Add `JoinAllThreadPools()` to explicitly `pthread_join` all worker
  threads during `OnShutdownCallback`, replacing the timing dependency
  with a hard synchronization point.

- Add a `MarkAsShuttingDown()` flag checked by `RunByMain()`. During
  shutdown, `RunByMain` executes callbacks directly instead of posting
  to the dead event loop. This is safe because IO threads are already
  inactive at shutdown time. There shouldn't be any race condition under
  this scenario. 

- Remove the `PAUSEPOINT RESET` mutation_processing workaround from
  `test_dropindex_with_background_queries` since the shutdown logic now
  handles stuck workers properly.

Shutdown order:
1. `MarkAsShuttingDown` so RunByMain stops posting to the event loop
2. `ClearAllPausePoints` wake stuck workers
3. `JoinAllThreadPools` wait for all workers to fully exit
4. S`chemaManager::OnShutdownCallback` tear down schemas safely

Tested with ASan: 9/9 passed, zero leaks, zero sanitizer errors.
```
./build.sh --asan --run-integration-tests=test_fulltext_inflight_blocking.py
```

Frank-Gu-81:
The failed test in `mrmw_mutex_test.cc` looks unrelated to my change.

vmsdk/src/utils.cc
```diff
@@ -70,11 +70,25 @@ void TrackCurrentAsMainThread() {
...
   if (IsMainThread() && !force_async) {
     fn();
     return VALKEYMODULE_OK;
   }
+  // During shutdown the event loop is no longer processing one-shot
+  // callbacks.  Execute the function directly to avoid leaking the
+  // allocation.  At this point IO threads are inactive so it is safe
+  // to call ValkeyModule_FreeThreadSafeContext from a worker thread.
+  if (IsShuttingDown()) {
+    fn();
```
BCathcart: Performing an action that is meant to run from the main thread on the background thread seems like a recipe for some other strange behaviour on shutdown. How about simply adding a log and not doing anything else?

vmsdk/src/utils.cc
```diff
@@ -70,11 +70,25 @@ void TrackCurrentAsMainThread() {
...
+
 int RunByMain(absl::AnyInvocable<void()> fn, bool force_async) {
   if (IsMainThread() && !force_async) {
     fn();
     return VALKEYMODULE_OK;
   }
+  // During shutdown the event loop is no longer processing one-shot
+  // callbacks.  Execute the function directly to avoid leaking the
+  // allocation.  At this point IO threads are inactive so it is safe
+  // to call ValkeyModule_FreeThreadSafeContext from a worker thread.
```
BCathcart:
```suggestion
  // During shutdown the event loop is no longer processing one-shot
  // callbacks.
```

src/server_events.cc
```diff
@@ -53,9 +54,20 @@ void OnServerCronCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
 
 void OnShutdownCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
                         uint64_t subevent, void *data) {
-  // Clear all PausePoints first so any waiting worker threads wake up and
-  // exit their spin loops before global destructors destroy the map.
+  // Mark the module as shutting down so that RunByMain() calls from
+  // background threads execute their callbacks directly instead of
+  // posting to the (now-dead) event loop.
```
BCathcart:
```suggestion
  // Mark the module as shutting down so that RunByMain() stops
  // scheduling new tasks.
```

BCathcart (approved):

Issue #962, opened 2026-04-14
[BUG] Memory Overflow after null length string input to timefmt function.


allenss-amazon:
**Describe the bug**
The timefmt function falls into an infinite scaling loop when the format string is empty.

**To Reproduce**
```
127.0.0.1:6379> FT.CREATE testidx ON HASH PREFIX 1 doc: SCHEMA title TEXT ts NUMERIC
OK
127.0.0.1:6379> HSET doc:1 title "test" ts 1700000000
(integer) 2
127.0.0.1:6379> FT.AGGREGATE testidx "test" LOAD 1 @ts APPLY 'timefmt(@ts, "")' AS fmt
Error: Server closed the connection
(17.40s)
```
**Expected behavior**

Empty format string should yield an empty output string.

**Environment**

Additional information to reproduce the failure

- Valkey Search Module Version
- Valkey Core Version 

**Additional context**

Aksha1812:
@allenss-amazon do we need #982 to close this issue or #963  is enough . If so should we either close this issue or the stalke #982 PR ?

PR #963, opened 2026-04-14
Fix timefmt causes OOM on empty input, Issue: #962

allenss-amazon:
Fixes two problems.

1. The unit tests for expr and values somehow fell out of the "run-tests" option on build.sh
2. Fixes Issue #962 by checking for null input string and short circuiting.


src/expr/value.cc
BCathcart, line 530: Can we add a sanity upper limit so this doesn't grow indefinitely?

I asked Gemini if there are any other edge cases where the buffer is big enough and strftime would return 0, and it mentioned this one:

```
2. The "%p" Edge Case (Locale-Specific)

This is a rare but fascinating bug/feature. In some languages and locales, the "AM/PM" designation (%p) is actually an empty string.

    Example: Many locales in Europe use a 24-hour clock and don't have a concept of AM/PM.

    If a user provides the format string "%p", and the system is set to a locale where AM/PM doesn't exist, strftime will write nothing to the buffer and return 0.

    The Result: Even if your buffer is 1GB, the loop will see that 0, think "not enough space," and double the buffer again.
```

KarthikSubbarao (approved):
Approving, the size of `result` is mainly controlled by the input

PR #967, opened 2026-04-14
feat: Implement INKEYS parameter for FT.SEARCH

### LINKED ISSUE #898: [FEATURE] Support the INKEYS parameter to FT.SEARCH

**Additional information**

rileydes-improving:
Addressing issue #898 

## Summary
Adds RediSearch-compatible `INKEYS` option to `FT.SEARCH`, allowing callers to restrict results to a specified set of document keys. Also fixes a lock-ordering issue in `IndexSchema::MarkAsDestructing()` that could deadlock during client unblock callbacks.

## Changes
- `ApplyInkeysFilter()` in `ft_search.h`, post-filter that retains only neighbors in the INKEYS set, with saturating `total_count` adjustment.
- Parser (`ft_search_parser.cc`) handles `INKEYS {count} {key}...`, rejects negative/non-integer counts, deduplicates keys into `absl::flat_hash_set`.
- `RequiresCompleteResults()` returns true when INKEYS is present so LIMIT doesn't trim candidates before the post-filter runs.
- `SendReply()` restructured, skips content resolution for NOCONTENT, short-circuits on `INKEYS 0`, applies the filter before sorting.
- Cluster support: `repeated string inkeys = 20` in `coordinator.proto`, round-tripped through `search_converter.cc`. Filter runs shard-locally in `content_resolution.cc`.
- Command spec (`ft.search.json`) and docs (`ft.search.md`) updated.
- `IsMarkedDestructing()` in `src/index_schema.h` made `const` and thread-safe with `ABSL_LOCKS_EXCLUDED` annotation and a mutex guard.

## Testing
- Unit tests in `ft_search_parser_test.cc` test parser cases — single/multi/duplicate/zero/error, vector+NOCONTENT combos.
- Unit tests in `ft_search_test.cc` test filter logic — intersection, no-op on empty set, underflow saturation.
- Unit tests in `search_converter_test.cc` test proto round-trip — empty, single, multi-key sets .
- Integration test in `test_ft_search_inkeys.py` test NOCONTENT, nonexistent keys, mixed keys, LIMIT, SORTBY, RETURN, vector KNN, tag/numeric filters, error cases.
- Compatibility tests in `TestSearchCompatibility` class with regenerated `search-answers.pickle.gz`.

## Example
```
FT.SEARCH idx "@score:[-inf +inf]" INKEYS 3 doc:0 doc:1 doc:2 NOCONTENT DIALECT 2
```
Returns `[3, "doc:0", "doc:1", "doc:2"]`, count plus the matching key names.

Resolves #898


integration/compatibility/generate.py
```diff
@@ -468,3 +468,46 @@ def test_search_sortby(self, key_type, dialect):
...
+
+
+@pytest.mark.parametrize("dialect", [2])
+@pytest.mark.parametrize("key_type", ["json", "hash"])
+class TestSearchCompatibility(BaseCompatibilityTest):
+    ANSWER_FILE_NAME = "search-answers.pickle.gz"
+
+    def test_inkeys_basic(self, key_type, dialect):
+        self.setup_data("sortable numbers", key_type)
+        self.check("ft.search", f"{key_type}_idx1", "@n1:[-inf inf]", "INKEYS", "3", f"{key_type}:00", f"{key_type}:01", f"{key_type}:02", "DIALECT", str(dialect))
```
allenss-amazon: Currently, nobody is actually looking at the results from setup_data. It's returning the # of loaded keys. But can just as easily return it's array of keys/values, so that here instead of re-generating the key names, we could use a slice of the array and eliminate this extra depenency.

allenss-amazon: Some of these tests need to include vectors as the code has substantial differences in processing vectors vs non-vectors. I think if you use "*" instead of a numeric query the infrastructure will automatically execute both a vector and a non-vector query.

integration/test_ft_search_inkeys.py
```diff
@@ -0,0 +1,355 @@
+"""Integration tests for FT.SEARCH INKEYS option."""
```
allenss-amazon, line 1: Is this redundant with the compatibility tests? If so, let's trim it down.

src/query/search.h
```diff
@@ -171,6 +173,8 @@ struct SearchParameters {
   int k{0};
   std::optional<unsigned> ef;
   LimitParameter limit;
+  absl::flat_hash_set<std::string> inkeys;
```
allenss-amazon: let's use the std::optional paradigm here.

src/commands/ft_search.cc
```diff
@@ -314,18 +313,34 @@ void SearchCommand::SendReply(ValkeyModuleCtx *ctx,
...
 
-  // 2. Process neighbors for the query
-  auto status = ProcessNeighborsForQuery(ctx, search_result, *this);
-  if (!status.ok()) {
-    ++Metrics::GetStats().query_failed_requests_cnt;
-    ValkeyModule_ReplyWithError(ctx, status.message().data());
+  // 1.5. INKEYS 0 means no keys can match — return empty result
+  if (has_inkeys && inkeys.empty()) {
+    ValkeyModule_ReplyWithArray(ctx, 1);
+    ValkeyModule_ReplyWithLongLong(ctx, 0);
```
allenss-amazon: Optimizing for the INKEYS 0 case seems like a increase in complexity with no real gain -- it seems like a highly  unlikely scenario.

src/commands/ft_search.cc
```diff
@@ -260,11 +264,6 @@ bool HandleEarlyReplyScenarios(ValkeyModuleCtx *ctx,
     return true;  // Early reply sent, stop processing
   }
 
-  if (command.no_content) {
```
allenss-amazon, line 264: Optimizing for no-content is important, why is this removed?

src/commands/ft_search.h
```diff
@@ -8,8 +8,35 @@
...
+#include "absl/container/flat_hash_set.h"
+#include "src/indexes/vector_base.h"
+
 namespace valkey_search {
 class ValkeySearch;
 
+// Remove neighbors whose key is not in the inkeys set.
+// Adjusts total_count down by the number removed (floors at 0).
+// No-op when inkeys is empty.
+inline void ApplyInkeysFilter(std::vector<indexes::Neighbor> &neighbors,
```
allenss-amazon: I'd rather see this pass in the SearchResult struct than the parts of it.

src/commands/ft_search_parser.cc
```diff
@@ -201,6 +201,25 @@ std::unique_ptr<vmsdk::ParamParser<SearchCommand>> ConstructReturnParser() {
       });
 }
 
+std::unique_ptr<vmsdk::ParamParser<SearchCommand>> ConstructInkeysParser() {
+  return std::make_unique<vmsdk::ParamParser<SearchCommand>>(
+      [](SearchCommand &parameters, vmsdk::ArgsIterator &itr) -> absl::Status {
+        int count{0};
```
allenss-amazon: I believe you could make this be a uint32_t and avoid the negative value check.

src/commands/ft_search_parser.cc
```diff
@@ -201,6 +201,25 @@ std::unique_ptr<vmsdk::ParamParser<SearchCommand>> ConstructReturnParser() {
...
+        VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, count));
+        if (count < 0) {
+          return absl::InvalidArgumentError(
+              "INKEYS count must not be negative");
+        }
+        parameters.has_inkeys = true;
+        for (int i = 0; i < count; ++i) {
+          VMSDK_ASSIGN_OR_RETURN(auto key, itr.Get());
```
allenss-amazon: Replace itr.Get() with itr.PopNext() and you can delete the itr.Next call.

src/query/search.cc
```diff
@@ -785,6 +797,12 @@ SerializationRange SearchResult::GetSerializationRange(
 }
 
 absl::Status Search(SearchParameters &parameters, SearchMode search_mode) {
+  // INKEYS 0 (has_inkeys=true, empty set) means no keys can match.
+  // Short-circuit before acquiring the lock to avoid unnecessary contention.
+  if (parameters.has_inkeys && parameters.inkeys.empty()) {
+    parameters.search_result = SearchResult(0, {}, parameters);
+    return absl::OkStatus();
+  }
```
allenss-amazon: Again, why optimize INKEYS 0?

src/index_schema.cc
```diff
@@ -1909,29 +1909,38 @@ bool IndexSchema::TrackMutatedRecord(ValkeyModuleCtx *ctx, const Key &key,
...
+    // Collect waiting queries so we can notify them outside the lock.
+    for (auto &[key, mutation] : tracked_mutated_records_) {
+      for (auto &params : mutation.waiting_queries) {
+        if (params) {
+          params->search_result.status =
+              GenerateIndexNotFoundError(db_num_, name_);
+          to_notify.push_back(std::move(params));
+        }
       }
     }
```
allenss-amazon: What problem is this change solving? It seems totally unrelated to INKEYS.

src/commands/ft_search_parser.cc
rileydes-improving: I had it as a uint32_t while making it but changed it to be able to have the explicit error handling. I will switch it back.

integration/compatibility/generate.py
rileydes-improving: Switched `load_data` and both `setup_data` overrides to return the load list, and updated the tests to slice keys from it instead of formatting them by hand.

Added a small `_run_both` helper that swaps `*` for either a match-all numeric filter or a KNN clause, and routed the INKEYS tests through it so both paths get coverage.

integration/test_ft_search_inkeys.py
rileydes-improving: I'll trim the redundant ones and make sure the compatibility suite covers every happy path.

src/query/search.h
rileydes-improving: Collapsed `has_inkeys` + `inkeys` into `std::optional<flat_hash_set<string>>`. On the wire, empty-repeated vs unset collapse to the same bytes, should be fine in practice since the only case that loses round-trip fidelity (INKEYS 0 on a shard) is already the case your later comments suggest not optimizing for. The coordinator's post-filter still returns the correct (empty) result.

src/commands/ft_search.cc
rileydes-improving: Fair. Dropped both short-circuits, the slow path handles an engaged empty set correctly. One knock-on: `ApplyInkeysFilter` used to treat an empty set as a no-op (consistent with the old "empty means unset" meaning), so I removed that early return too.

src/commands/ft_search.cc
rileydes-improving: Main had NOCONTENT early-returning from `HandleEarlyReplyScenarios`, during the refactor I replaced that with an `if (!no_content)` gate around the content fetch, which left the rest of `SendReply` running linearly.

I put the fast path back, just in `SendReply` instead — INKEYS needs to run before the reply (otherwise `SendReplyNoContent` emits excluded keys). Apply INKEYS, `SendReplyNoContent`, return. 

It matches main's behavior just with avoiding the issue of emitting excluded keys, happy to restructure if you'd rather it live elsewhere.

src/commands/ft_search.h
rileydes-improving: Done, `ApplyInkeysFilter` takes `SearchResult&` now.

src/commands/ft_search_parser.cc
rileydes-improving: Done, `ParseParamValue` into a `uint32_t` handles it, so the hand-rolled negative check is gone. The error message from the param parser is likely good enough

src/commands/ft_search_parser.cc
rileydes-improving: Done.

src/query/search.cc
rileydes-improving: Addressed in reply above.

src/index_schema.cc
rileydes-improving: Yeah, sorry,  I had noticed a potential lock-scope issue while debugging an INKEYS test and fixed it in the same branch temporarily, forgot to remove it. Reverted back to `main`

allenss-amazon (commented):
Claude says this:

```
I asked: Can you make the repeated field optional? 

  Not directly — proto3 forbids optional repeated (it's a compile error). optional adds an explicit presence bit, but repeated fields already lack a "set vs. unset" distinction on the wire: zero elements and "never set" serialize identically. So
  optional repeated string inkeys won't compile, and even conceptually it can't express what you want.

  The question behind the question is real, though. The C++ side is tri-state (std::optional<flat_hash_set>):

  ┌────────────────────┬───────────────────────────────┐
  │       State        │            Meaning            │
  ├────────────────────┼───────────────────────────────┤
  │ disengaged         │ no INKEYS → return everything │
  ├────────────────────┼───────────────────────────────┤
  │ engaged, empty     │ INKEYS 0 → return nothing     │
  ├────────────────────┼───────────────────────────────┤
  │ engaged, non-empty │ filter to the set             │
  └────────────────────┴───────────────────────────────┘

  The wire is bi-state. The middle row collapses into the top row across the coordinator→shard boundary. The PR's own round-trip test comments on this ("the empty/unset case is indistinguishable on the wire"). The standard fix for
  presence-on-a-collection is a wrapper message:

  message InkeysFilter { repeated string keys = 1; }
  optional InkeysFilter inkeys = 20;   // has_inkeys() now distinguishes all three states

  Why it actually matters — a verified bug, not just tidiness

  I traced INKEYS 0 through cluster mode, and the wire ambiguity produces a wrong total_count, not just wasted work:

  1. Coordinator parses INKEYS 0 → inkeys engaged+empty; RequiresCompleteResults() true locally only.
  2. ParametersToGRPCSearchRequest iterates the empty set → adds nothing. Wire = "unset."
  3. GRPCSearchRequestToParameters gates on if (!request.inkeys().empty()) → shard's inkeys stays disengaged.
  4. Shard therefore: skips the in-scan filter, skips content_resolution.cc's ApplyInkeysFilter, and — because its RequiresCompleteResults() is false — applies LIMIT. It returns a trimmed neighbor list but total_count = borrowed.size() = the full
  pre-limit match count (e.g. 10000).
  5. Coordinator sums shard total_counts → ~10000, holds only the ~LIMIT neighbors.
  6. Coordinator's SendReply re-runs ApplyInkeysFilter with the empty set: it removes the handful of held neighbors and does total_count -= removed, where removed is only what it physically holds (~20), not the grand total.

  Result: INKEYS 0 in cluster mode returns 0 rows but total_count ≈ 9980. The round-trip test's claim that "the coordinator still returns zero rows" is true for rows but false for the count. Standalone is fine (it short-circuits INKEYS 0 directly);
  the bug is cluster-only and is a direct symptom of the missing presence bit. A wrapper message fixes it at the source (shard engages empty set → filters → decrements correctly → RequiresCompleteResults true so no premature LIMIT).

  Note the non-empty cluster case is correct: shards filter locally and decrement their own totals; the coordinator's re-filter finds all merged neighbors in-set (removed=0), so no double-decrement. Only the engaged-empty state is broken.

  Interesting test cases the suite misses

  Ordered by how likely they are to catch a real bug:

  1. INKEYS 0 end-to-end in cluster mode — assert total_count == 0, not just zero rows. This is the bug above. The current round-trip test deliberately skips the empty set; nothing exercises it through a real fan-out.
  2. total_count correctness in cluster mode under a partial INKEYS — keys spread across shards, result set exceeding LIMIT. Verify the reported count equals the true number of in-set matches across shards, not a per-shard or pre-filter artifact.
  3. total_count on the vector/KNN path — for non-vector, total_count is already correct before the post-filter (in-scan pre-filter + borrowed.size()), so the post-filter is a no-op there. For KNN the post-filter is the only corrector. There's a
  unit test for underflow saturation, but an end-to-end "KNN + INKEYS returns N rows and total_count == N" assertion would lock the path that actually depends on it.
  
```

rileydes-improving:
I added the missing tests, made RequireCompleteResults account for inkeys, and made the change to the proto wrapper to make inkeys properly optional.

src/commands/ft_search_parser.cc
```diff
@@ -201,6 +201,20 @@ std::unique_ptr<vmsdk::ParamParser<SearchCommand>> ConstructReturnParser() {
       });
 }
 
+std::unique_ptr<vmsdk::ParamParser<SearchCommand>> ConstructInkeysParser() {
+  return std::make_unique<vmsdk::ParamParser<SearchCommand>>(
+      [](SearchCommand &parameters, vmsdk::ArgsIterator &itr) -> absl::Status {
+        uint32_t count{0};
+        VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, count));
```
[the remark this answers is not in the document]
allenss-amazon: don't care.

src/query/search.cc
```diff
@@ -603,6 +609,12 @@ absl::StatusOr<std::vector<indexes::Neighbor>> SearchNonVectorQuery(
       while (!iterator->Done()) {
         const auto &key = **iterator;
         BACKGROUND_PAUSEPOINT("search_entries_fetcher");
+        // Skip keys not in the INKEYS set (when specified).
+        if (parameters.inkeys.has_value() &&
+            !parameters.inkeys->contains(key->Str())) {
+          iterator->Next();
+          continue;
+        }
```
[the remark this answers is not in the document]
rileydes-improving: By doing INKEYs first, it gives consistent ordering and better performance (allowing early skips and not adding seen_keys).

Aksha1812:
**Reviewers for this PR**

src/query/search.cc
Frank-Gu-81, line 1205: This looks correct for pre-filter. But pure/inline-filtered KNN (`PerformVectorSearch`) builds its `InlineVectorFilter` only from the query predicate and never consults `inkeys`, so INKEYS is applied only as a post-filter after top-K selection. That means in-set docs outside the global top-K are silently dropped:
  
`KNN 10 @v $q INKEYS 3 a b c` returns nothing if a/b/c are the 40th-nearest, even though they match.

integration/test_ft_search_inkeys_cluster.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+
+        # total_count should be <= 3 (only keys in INKEYS that match)
+        assert result[0] <= 3, (
+            f"Expected total_count<=3, got {result[0]}"
+        )
+        # All returned keys must be in the target set
+        returned_keys = {result[i].decode() for i in range(1, len(result), 2)}
+        assert returned_keys.issubset(set(target_keys))
+        assert result[0] == len(returned_keys)
```
Frank-Gu-81: Echoing the comment I had earlier in `DoSearchNonVector`: This only asserts `total_count <= 3`, which passes even when KNN+INKEYS under-returns (e.g. returns 0 while a/b/c all exist and match). To pin the intended semantics, can we construct the case so the in-set keys are not among the global top-K and assert the query still returns them?

src/index_schema.h
```diff
@@ -277,7 +277,10 @@ class IndexSchema : public KeyspaceEventSubscription,
     return time_sliced_mutex_;
   }
   void MarkAsDestructing();
-  bool IsMarkedDestructing() { return is_destructing_; };
+  bool IsMarkedDestructing() const ABSL_LOCKS_EXCLUDED(mutated_records_mutex_) {
+    absl::MutexLock lock(&mutated_records_mutex_);
+    return is_destructing_;
+  }
```
Frank-Gu-81: is this needed? `is_destructing_` is already an `atomic<bool>`.

Frank-Gu-81:
Note on failed check: 

The ASAN issue is not related to this change. Fix is in this PR: https://github.com/valkey-io/valkey-search/pull/1291

integration/compatibility/generate_search.py
Frank-Gu-81, line 1: is `generate_search.py` intended as the general FT.SEARCH compatibility suite that future search PRs extend, or just INKEYS? If the latter, i think we should fold these cases into `generate.py`.

rileydes-improving: That was the intention as I also used the same infrastructure naming in my INFIELDs PR (Another search subcommand). Some search subcommands do currently have minimal compatibility tests and as we add more, I thought it easier to manage rather than making generate.py enormous.

It is up for discussion though, if we would like to just keep generate.py as a very large single source of truth, I can change.

Long term, Allen has mentioned that we may try to adjust the answer files to allow for proper diffs/automatic conflict resolution and not requiring regeneration after every feature PR is merged. I saw this as a natural extension, having a different file for different FT.* commands.

Frank-Gu-81: Makes sense. Thanks for the context. Keeping a per-command generator is reasonable, especially given the direction toward diffable answer files.

src/index_schema.h
rileydes-improving: Removed. Initially I had put it on to ensure teardown ordering but both the reader and writer should only run on the main thread, so removing it should prevent unnecessary contention.

integration/test_ft_search_inkeys_cluster.py
rileydes-improving: The test now targets the three farthest docs (vdoc:27/28/29) so they're outside the global top-K, and asserts exact results (result[0] == 3, exact key set) instead of <= 3. 

Added some extra tests into the inkeys integration test to cover a couple other cases.

src/query/search.cc
rileydes-improving: Fixed. INKEYS now feeds candidate selection instead of post-filtering a global top-K. Pure KNN goes through CalcBestMatchingInkeys (exact search over just the in-set keys), predicate+KNN takes the exact prefilter path when INKEYS is present. InlineVectorFilter also drops out-of-set keys.

I added some compatibility tests to verify the behaviour against RediSearch and ensure no regressions.

Frank-Gu-81 (approved):

PR #973, opened 2026-04-16
fix aggregate recursion depth

boda26:
1. Moving query-string-depth config to valkey_search_options, enable reuse through multiple files
2. Use query-string-depth to limit aggregate query, prevent stack overflow
3. Added unit tests for aggregate query depth

### DIFF
--- a/src/expr/expr.cc
+++ b/src/expr/expr.cc
@@ -281,6 +281,8 @@ bool IsIdentifierChar(int c) {
 }
 
 struct Compiler {
+  int depth_ = 0;
+  static constexpr int kMaxDepth = 1000;
   utils::Scanner s_;
   Compiler(absl::string_view sv) : s_(sv) {}
 
@@ -349,6 +351,9 @@ struct Compiler {
   }
 
   absl::StatusOr<ExprPtr> Primary(CompileContext& ctx) {
+    if (++depth_ > kMaxDepth) {
+      return absl::InvalidArgumentError("Expression too complex");
+    }
     s_.SkipWhiteSpace();
     DBG << "Primary: '" << s_.GetUnscanned() << "'\n";
     switch (s_.PeekByte()) {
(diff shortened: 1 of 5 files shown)


src/expr/expr.cc
```diff
@@ -281,6 +281,8 @@ bool IsIdentifierChar(int c) {
 }
 
 struct Compiler {
+  int depth_ = 0;
+  static constexpr int kMaxDepth = 1000;
```
allenss-amazon: Let's make the limit configurable.

src/expr/expr.cc
```diff
@@ -349,6 +351,9 @@ struct Compiler {
   }
 
   absl::StatusOr<ExprPtr> Primary(CompileContext& ctx) {
+    if (++depth_ > kMaxDepth) {
```
allenss-amazon: Shouldn't you be decrementing depth when you return?

src/expr/expr.cc
KarthikSubbarao: Is there a general recursion depth config which we want to use here?

I know we have one that is used in different query parsing functions

boda26: Yes, there is a configureable, query-string-depth, which is in filter_parser.cc. It is default to 1000. We can reuse it here if necessary

src/expr/expr.cc
allenss-amazon: As written, this is a limit on the number of "primaries" that you can have in the expression. Not the same as a nesting limit.

src/expr/expr.cc
boda26: query-string-depth is written in filter_parser.cc. Including that would create a dependency to expr.cc which is not good. I would move the query-string-depth config to src/valkey_search_options.cc, which is a natural home for all configs and make it reusable across multiple files

allenss-amazon (approved):

PR #974, opened 2026-04-17
Update search_number_of_indexes while restoring

Aksha1812:
During RDB restore `search_number_of_indexes` INFO metric reported 0 because indexes are not added to the live schema map until after the restore completes. Changing this behaviour so that during restore we use `rdb_restore_total_indexes`  - `rdb_restore_completed_indexes` i.e indexes pending rdb load to update it. Will allow us to identify clusters with search indexes early and ensures the metric never drops to 0 while indexes are mid-load.

Added integration test to confirm if above works. 

Added a Dev metric as well to give us more visibility on search RDB restore overhead duration

- `rdb_last_restore_aux_load_duration_ms` — Duration of the last search index schema deserialization from RDB

### DIFF
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -392,10 +392,9 @@ uint64_t SchemaManager::GetNumberOfIndexSchemas() const {
   // Use rdb_restore_total_indexes while rdb_restore_in_progress
   if (Metrics::GetStats().rdb_restore_in_progress.load(
           std::memory_order_relaxed)) {
-    num_schemas = std::max(
-        static_cast<uint64_t>(num_schemas),
-        Metrics::GetStats().rdb_restore_total_indexes.load(
-            std::memory_order_relaxed));
+    num_schemas = std::max(static_cast<uint64_t>(num_schemas),
+                           Metrics::GetStats().rdb_restore_total_indexes.load(
+                               std::memory_order_relaxed));
   }
   return num_schemas;
 }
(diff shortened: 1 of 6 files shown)


src/schema_manager.cc
```diff
@@ -388,6 +389,13 @@ uint64_t SchemaManager::GetNumberOfIndexSchemas() const {
   for (const auto &[db_num, schema_map] : db_to_index_schemas_) {
     num_schemas += schema_map.size();
   }
+  // Use rdb_restore_total_indexes while rdb_restore_in_progress
+  if (Metrics::GetStats().rdb_restore_in_progress.load(
+          std::memory_order_relaxed)) {
+    num_schemas = std::max(static_cast<uint64_t>(num_schemas),
+                           Metrics::GetStats().rdb_restore_total_indexes.load(
+                               std::memory_order_relaxed));
+  }
```
allenss-amazon: This seems like an odd place to put this functionality. At a minimum, the comment is inadequate.

src/valkey_search.cc
```diff
@@ -357,6 +358,37 @@ static vmsdk::info_field::Integer rdb_restore_backpressure_wait_cycles(
...
+          // In progress - compute elapsed
+          if (SchemaManager::Instance().IsIndexingInProgress()) {
+            return absl::ToUnixMillis(absl::Now()) - start_ms;
+          }
+          // Finished
+          auto duration = absl::ToUnixMillis(absl::Now()) - start_ms;
+          Metrics::GetStats().rdb_last_restore_backfill_duration_ms = duration;
+          return duration;
+        }));
+
```
allenss-amazon: This metric is only correct if it's polled at a high frequency, right? What is the use-case for that?

allenss-amazon (approved):

Issue #975, opened 2026-04-20
[TEST] Flaky test_fulltext_inflight_blocking


Aksha1812:
**Resolve Flakiness**

- Recent failure : https://github.com/valkey-io/valkey-search/actions/runs/24662207052/job/72110821066?pr=974
- Mostly flaky in ASAN workflows
- https://github.com/valkey-io/valkey-search/pull/950 this didn't fix flakiness.
- Needs deep dive

Aksha1812:
Can we add this to https://github.com/orgs/valkey-io/projects/58/views/1?filterQuery=is%3Aopen . so this can get picked up soon. failing very frequently recently.

BCathcart:
I'll add it to the 1.3 board that was just created: https://github.com/orgs/valkey-io/projects/69

Aksha1812:
That sounds better thanks

Aksha1812:
related PR : https://github.com/valkey-io/valkey-search/pull/978

Aksha1812:
FYI . Just testing slash command . /label test

Frank-Gu-81:
#1234 is fixing this issue.

PR #978, opened 2026-04-21
Fix intermittent failure in ASAN tests w.r.t. Pausepoint and shutdown

allenss-amazon:
Failure was seen here: https://github.com/valkey-io/valkey-search/actions/runs/24662207052/job/72110821066?pr=974

Problem is that if the server executes a shutdown, it looks like the module doesn't shutdown it's worker threads, meaning that a worker thread could be in an arbitrary state. Here it was polling a PAUSEPOINT.

Fix is to have the shutdown callback wait until all worker threads are suspended.


src/server_events.cc
```diff
@@ -68,6 +68,7 @@ void OnShutdownCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
   vmsdk::debug::ClearAllPausePoints();
   ValkeySearch::Instance().JoinAllThreadPools();
```
Aksha1812, line 69: Does JoinWorkers() already terminate threads 

```
void ThreadPool::JoinWorkers() {
  {
    absl::MutexLock lock(&queue_mutex_);
    if (!stop_mode_.has_value()) {
      stop_mode_ = StopMode::kGraceful;
      condition_.SignalAll();
    }
    suspend_workers_ = false;
  }

  threads_.ClearWithCallback(
      [](auto thread) { pthread_join(thread->thread_id, nullptr); });
  started_ = false;

  JoinTerminatedWorkers();
}
```

Not sure how suspending worker threads after would help in that case .  Let me know if i am wrong , not too familiar with the thread pooling infrstructure

Aksha1812:
Relevant Issue : https://github.com/valkey-io/valkey-search/issues/975

src/server_events.cc
allenss-amazon: Hmm, for this fix, actually terminating the threads doesn't matter. What matters is to avoid them polling on PAUSEPOINT stuff.

Aksha1812: I was more confused about the placement of suspending worker threads here at the end using ValkeySearch::Instance().OnShutdownCallback(); . Will it be a no-op after ValkeySearch::Instance().JoinAllThreadPools(); exits  all worker threads .

allenss-amazon: it should be.

src/valkey_search.cc
```diff
@@ -1021,25 +1021,33 @@ void ValkeySearch::Info(ValkeyModuleInfoCtx *ctx, bool for_crash_report) const {
...
   // Sanity: fork can occur (by example: calling to "popen") before the thread
   // pool is initialized
+  SuspendWorkers("At prepare fork callback");
+}
+
+void ValkeySearch::SuspendWorkers(std::string_view reason) {
   if (writer_thread_pool_ == nullptr || reader_thread_pool_ == nullptr ||
       utility_thread_pool_ == nullptr) {
     return;
   }
```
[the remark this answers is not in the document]
allenss-amazon: meaningless these three pools are always in lock-step.

allenss-amazon:
appears unneeded.

PR #979, opened 2026-04-21
Fix tag query parsing to support escaped closing brace

### LINKED ISSUE #454: [BUG] Tag Query Parsing Logic doesn't allow pipe or right brace characters in tags
**Describe the bug**
Using the SEPARATOR option of a TAG field definition allows the ingestion of tags that contain up to 255 of the 256 possible byte patterns. However, regardless of this, it's not possible to construct a query for a tag that contains a pipe (|) or a right brace (}) character. This effectively means that the SEPARATOR option isn't really useful.

**To Reproduce**
a query string like this:
```
@tag-field:{ tag-containing-}-and-} }
```
Will cause a syntax error because the tag query is terminated at the first occurrence of the brace.

**Expected behavior**
Parsing of tag field should support escaping of pipe and right hand brace characters.

### LINKED ISSUE #518: [BUG] Can't use escape to quote special characters in TAG query syntax
**Describe the bug**

In order to query for a tag that contains a pipe character or other special characters (quotes, brackets, etc.) they need to be escaped. This isn't handled properly.

**To Reproduce**

Fix the query parsing to support escaped characters.

Frank-Gu-81:
Fixes #454
Fixes #518

**Problem**

`ParseTagString()` used `find('}')` to locate the end of a tag expression, which stopped at the first `}` without considering
backslash escaping. This made it impossible to query tags containing literal `}` characters.

For example, `@tags:{ a\}b }` would be parsed as `@tags:{ a\}` with `b }` treated as leftover text, producing a syntax error.

**Fix**

Replace the naive `find('}')` with a character-by-character scan that skips escaped characters (`\}`, `\\`, etc.) before matching the closing brace. This follows the same escape convention already used by `ParseSearchTags()` for pipe characters and matches the RedisSearch tag query escaping spec.

**Tests**

Added 3 integration tests:
- Escaped closing brace (`\}`) with multiple braces in one tag (reproducer from #454)
- Escaped pipe (`\|`) — end-to-end coverage
- Multiple special characters combined (`\|` and `\}` together)


integration/test_filter_expressions.py
```diff
@@ -930,3 +930,89 @@ def test_10k_tag_filters(self):
...
+        assert client.execute_command("HSET", "doc:2", "tags", "other") == 1
+
+        # Query with escaped braces — matches the issue's example
+        result = client.execute_command(
+            "FT.SEARCH", "esc_brace_idx",
+            r"@tags:{ tag-containing-\}-and-\} }",
+            "NOCONTENT"
+        )
+        assert result[0] == 1
+        assert result[1] == b"doc:1"
```
VoletiRam: Is the result order guaranteed?
To avoid potential flakiness, i would suggest change this to
```
assert result[0] == 1
keys = set(result[i].decode('utf-8') for i in range(1, len(result)))
assert keys == {"doc:1"}
```

integration/test_filter_expressions.py
VoletiRam, line 1: We have unit test coverage for filter parser in testing/filter_test.cc. We already have `tag_escaped_pipe_*` cases there that follow the same pattern. We should add a couple parallel cases for `\}` at minimum a match case, a no-match case, and the missing-close-bracket error case. 

__`test_tag_escaped_closing_brace`__ keep this one, it directly reproduces the issue.

__`test_tag_escaped_pipe`__ and __`test_tag_multiple_special_characters`__ the `\|` behavior is already covered end-to-end by the existing unit tests. These two don't exercise any new code path that isn't already tested. Suggest dropping them to stay lean.

VoletiRam (approved):

integration/test_filter_expressions.py
```diff
@@ -930,3 +930,35 @@ def test_10k_tag_filters(self):
         assert result[0] == 3
         keys = set(result[i].decode('utf-8') for i in range(1, len(result)))
         assert keys == {"item:1", "item:2", "item:3"}
+
+    def test_tag_escaped_closing_brace(self):
+        """
+        Test that a tag containing a literal '}' can be queried using
+        backslash escaping: @field:{value\\}more}
+
```
allenss-amazon, line 938: Let's add some tests to the compatibility suite to ensure this matches Redis behavior.

integration/compatibility/data_sets.py
```diff
@@ -496,6 +496,28 @@ def make_field_definition(key_type, name, typ, i):
...
+            create_cmd = ("FT.CREATE json_idx1 ON JSON PREFIX 1 json: "
+                          "SCHEMA $.tags AS tags TAG SEPARATOR ,")
+        data["tag special chars"][CREATES_KEY(key_type)] = [create_cmd]
+        data["tag special chars"][SETS_KEY(key_type)] = [
+            (f"{key_type}:1", {"tags": "a}b"}),
+            (f"{key_type}:2", {"tags": "a|b"}),
+            (f"{key_type}:3", {"tags": "normal"}),
+            (f"{key_type}:4", {"tags": "tag-containing-}-and-}"}),
+            (f"{key_type}:5", {"tags": "a\\b"}),
+        ]
```
allenss-amazon: How about we also add some cases with multiple tags, plus a dataset that has a non-default separator character.

I'd recommend generating the multiple tags cases simply by generating all possible combinations of the strings that you're using above.

integration/compatibility/data_sets.py
```diff
@@ -634,6 +663,7 @@ def load_data(client, data_set, key_type, data_source=None, schema_type="default
 
 def load_data_cluster(cluster_client, test_case, data_set, key_type):
     data = compute_data_sets()
+    data.update(compute_tag_special_chars_data())
```
allenss-amazon: ? Why isn't this in compute_data_sets ?

integration/compatibility/generate.py
```diff
@@ -468,3 +473,21 @@ def test_search_sortby(self, key_type, dialect):
...
+                              r"@tags:{ a\}b }", "DIALECT", str(dialect)])
+        # Escaped pipe: should match doc with literal tag "a|b"
+        self.execute_command(["FT.SEARCH", f"{key_type}_idx1",
+                              r"@tags:{ a\|b }", "DIALECT", str(dialect)])
+        # Multiple escaped braces: reproducer from issue #454
+        self.execute_command(["FT.SEARCH", f"{key_type}_idx1",
+                              r"@tags:{ tag-containing-\}-and-\} }",
+                              "DIALECT", str(dialect)])
+        # Normal tag (no escaping needed): baseline
+        self.execute_command(["FT.SEARCH", f"{key_type}_idx1",
```
allenss-amazon, line 494: Same comment here, let's have some multi-tag queries.

integration/compatibility/data_sets.py
Frank-Gu-81: Will move move this into `compute_data_sets`

Frank-Gu-81:
Resolved all merge conflicts

allenss-amazon:
Obsoleted by #1148

PR #983, opened 2026-04-22
feat: Implement INFIELDs parameter for FT.SEARCH

### LINKED ISSUE #899: [FEATURE] Support the INFIELDS parameter to FT.SEARCH

**Additional information**

rileydes-improving:
Addressing issue #899 

## Summary

Adds `INFIELDS <count> <field> [...]` to `FT.SEARCH` with Redis Stack-compatible behavior.
Restricts full-text term matching to a caller-specified set of indexed TEXT fields via the
`FieldMaskPredicate` bitmask.

Semantics (Verified against Redis Stack):
- Non-existent and non-TEXT fields are silently ignored.
- `INFIELDS 0` is a no-op — searches all text fields.
- Explicit `@field:` prefixes are intersected with INFIELDS. Field not in INFIELDS → term matches nothing.
- Inert for pure KNN vector queries.

## Changes
- `search.h` — `SearchParameters::infields` is `std::optional<flat_hash_set<string>>`.
- `ft_search_parser.cc`: `ConstructInfieldsParser()`: `uint32_t` count (rejects negatives), `itr.PopNext()` loop, registered under `kInfieldsParam`.
- `filter_parser.cc/h`: `TextParsingOptions::infields` is a raw pointer populated when the optional is engaged and non-empty. `SetupTextFieldConfiguration()` extended in both branches: bare terms restricted to INFIELDS fields, explicit `@field:` intersected with INFIELDS (mask=0 if not present).
- `ft_search.cc`: Fixed `SORTBY + NOCONTENT` ordering: `ProcessNeighborsForQuery` and `ApplySorting` now run before the NOCONTENT reply path. NOCONTENT without SORTBY still skips content fetch.
- `coordinator.proto`: `repeated string infields = 20` added to `SearchIndexPartitionRequest`.
- `search_converter.cc/h`: `InfieldsFromGRPC()` added as a named function for testability. `ParametersToGRPCSearchRequest` serializes when `has_value()`,`GRPCSearchRequestToParameters` deserializes into the optional only when non-empty. Removed duplicate `set_db_num` call.
- `search.cc`: `ParsePreFilter()` populates `TextParsingOptions::infields` from the optional.
- `ft.search.json` / `docs/commands/ft.search.md` — spec and docs updated.
- `testing/CMakeLists.txt`: `search_converter_test.cc` added.

## Tests

- Unit Tests: parser (`ft_search_parser_test.cc`), field mask (`filter_test.cc`), wire round-trip (`search_converter_test.cc`).
- Compatibility Tests (`search-answers.pickle.gz`, `TestInfieldsCompatibility`): basic scoping, invalid fields, LIMIT, zero-count no-op, prefix, multi-field, SORTBY, non-text predicates, duplicates, mixed filters, VERBATIM, SLOP+INORDER, SORTBY+RETURN, error cases (bad count, negative, mismatch), INFIELDS+KNN inert.
- `TestAggregateCompatibility` additions: `LIMIT 0 0` (ShouldReturnNoResults path), offset-beyond-results, INFIELDS+KNN inert.
- Compatibility infra: `unpack_search_result` uses command-token NOCONTENT detection instead of shape inference. `load_data`/`setup_data` return the load list.
- Integration Tests (`test_ft_search_infields.py`): trimmed to INFIELDS+KNN (not exercisable via text-schema compat suite).

## Example

```
FT.CREATE idx ON HASH SCHEMA title TEXT body TEXT category TAG
HSET doc:1 title "python programming" body "learn to code" category "tech"
HSET doc:2 title "advanced coding" body "python is great" category "tech"

FT.SEARCH idx "python" INFIELDS 1 title NOCONTENT DIALECT 2

FT.SEARCH idx "@title:python" INFIELDS 1 body NOCONTENT DIALECT 2
```
Returns 2 results (`doc:1`, `doc:2` — both have "python" in `title`), but `FT.SEARCH idx "programming" INFIELDS 1 body NOCONTENT DIALECT 2` returns empty since the term only appears in `title`.

Resolves #899


src/commands/ft_search_parser.cc
```diff
@@ -137,6 +137,23 @@ std::unique_ptr<vmsdk::ParamParser<SearchCommand>> ConstructParamsParser() {
         return absl::OkStatus();
       });
 }
+std::unique_ptr<vmsdk::ParamParser<SearchCommand>> ConstructInfieldsParser() {
+  return std::make_unique<vmsdk::ParamParser<SearchCommand>>(
+      [](SearchCommand &parameters, vmsdk::ArgsIterator &itr) -> absl::Status {
+        uint32_t count{0};
+        VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, count));
```
allenss-amazon, line 144: Should reject if more than max supported fields kMaxTextFieldsCount

src/coordinator/search_converter.cc
```diff
@@ -43,6 +43,17 @@ void SortByToGRPC(const std::optional<query::SortByParameter>& sortby,
                        : coordinator::SORT_ORDER_DESCENDING);
 }
 
+std::optional<absl::flat_hash_set<std::string>> InfieldsFromGRPC(
+    const SearchIndexPartitionRequest& request) {
+  if (request.infields().empty()) {
+    return std::nullopt;
+  }
+  absl::flat_hash_set<std::string> infields;
```
allenss-amazon, line 56: If > kMaxTextFieldsCount, put out a log message and ignore.

src/commands/filter_parser.cc
```diff
@@ -765,24 +783,61 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
...
+      auto text_identifiers = index_schema_.GetAllTextIdentifiers(with_suffix);
+      // Set field mask to include all text fields in the index schema.
+      field_mask = index_schema_.GetAllTextFieldMask(with_suffix);
+      if (text_identifiers.size() == 0 || field_mask == 0ULL) {
+        if (with_suffix) {
+          return absl::InvalidArgumentError("No fields support suffix search");
+        }
+        return absl::InvalidArgumentError("Index does not have any text field");
+      }
+      filter_identifiers_.reserve(filter_identifiers_.size() +
```
allenss-amazon: The conversion from field names to field mask of the INFIELDS stuff should be done once and stored in TextParsingOptions. If there's no INFIELDS, then that mask defaults to all 1's. Here you should be able to just AND the default mask from TextParsingOptions into the local setup. It's inefficient to reparse the INFIELDS stuff for every call to this function.

rileydes-improving: Fixed

src/coordinator/search_converter.cc
rileydes-improving: Fixed

src/commands/ft_search_parser.cc
rileydes-improving: Fixed

src/commands/filter_parser.h
```diff
@@ -28,6 +28,15 @@ struct TextParsingOptions {
   bool verbatim = false;
   bool inorder = false;
   std::optional<uint32_t> slop = std::nullopt;
+  const absl::flat_hash_set<std::string>* infields = nullptr;
```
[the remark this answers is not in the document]
rileydes-improving: Leaving as-is. TextParsingOptions is a short-lived stack struct constructed and consumed within the same function scope, search_params always outlives it. Copying the flat_hash_set would add an allocation on the query parsing hot path for no current benefit. The comment on lines 37–38 already documents the lifetime contract.

integration/compatibility/generate.py
```diff
@@ -464,35 +457,177 @@ def test_aggregate_dyadic_ops(self, key_type, dialect):
...
+    def test_search_limit_zero_number(self, key_type, dialect):
+        """LIMIT 0 0 — ShouldReturnNoResults path: reply is [total_count] only."""
+        self.setup_data("sortable numbers", key_type)
+        self.check(dialect, f"ft.search {key_type}_idx1 * LIMIT 0 0")
+        self.check(dialect, f"ft.search {key_type}_idx1 @n1:[-inf inf] LIMIT 0 0")
+
+    def test_search_limit_offset_beyond_results(self, key_type, dialect):
+        """LIMIT first_index > total results — reply is [total_count] only."""
+        self.setup_data("sortable numbers", key_type)
+        self.check(dialect, f"ft.search {key_type}_idx1 * LIMIT 10000 10")
```
allenss-amazon, line 541: These don't belong to this PR, correct? I think they're stale.

integration/compatibility/generate.py
```diff
@@ -464,35 +457,177 @@ def test_aggregate_dyadic_ops(self, key_type, dialect):
...
+        self.check("ft.search", f"{key_type}_idx1", "apple",
+                   "INFIELDS", "-1",
+                   "DIALECT", str(dialect))
+
+    def test_infields_error_count_mismatch(self, key_type, dialect):
+        """INFIELDS count > args — both engines raise."""
+        self.setup_data("pure text small", key_type)
+        self.check("ft.search", f"{key_type}_idx1", "apple",
+                   "INFIELDS", "5", "f1", "f2",
+                   "DIALECT", str(dialect))
```
allenss-amazon: compatibility isn't a good place for these kinds of checks, they should have a separate integraiton test.

src/commands/filter_parser.cc
```diff
@@ -34,8 +34,55 @@
...
+    return;
+  }
+  FieldMaskPredicate mask = 0ULL;
+  FieldMaskPredicate suffix_mask = 0ULL;
+  for (const auto& field : *infields) {
+    auto index = index_schema.GetIndex(field);
+    if (!index.ok() ||
+        index.value()->GetIndexerType() != indexes::IndexerType::kText) {
+      continue;
+    }
```
allenss-amazon: Seems like this should be an error.

src/commands/filter_parser.cc
```diff
@@ -34,8 +34,55 @@
...
+    auto* text_index = static_cast<const indexes::Text*>(index.value().get());
+    auto identifier = index_schema.GetIdentifier(field);
+    if (!identifier.ok()) {
+      continue;
+    }
```
allenss-amazon, line 79: Don't understand why this is duplicated with GetIndex. Seems like we should only be accepting the alias.

src/commands/filter_parser.cc
```diff
@@ -34,8 +34,55 @@
...
+    mask |= 1ULL << text_index->GetTextFieldNumber();
+    infields_identifiers.insert(identifier.value());
+    if (text_index->WithSuffixTrie()) {
+      suffix_mask |= 1ULL << text_index->GetTextFieldNumber();
+      infields_suffix_identifiers.insert(identifier.value());
+    }
+  }
+  infields_field_mask = mask;
+  infields_suffix_field_mask = suffix_mask;
```
allenss-amazon, line 88: Why do we need a suffix mask different than a normal mask?

src/commands/filter_parser.cc
```diff
@@ -792,24 +816,51 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
...
-      if (with_suffix) {
-        return absl::InvalidArgumentError("No fields support suffix search");
+    // INFIELDS: use precomputed field mask and identifiers.
+    if (options_.infields && !options_.infields->empty()) {
+      const auto& precomputed_identifiers =
+          with_suffix ? options_.infields_suffix_identifiers
+                      : options_.infields_identifiers;
+      FieldMaskPredicate precomputed_mask =
+          with_suffix ? options_.infields_suffix_field_mask
+                      : options_.infields_field_mask;
```
allenss-amazon: This strongly suggests the uselessness of the suffix and not suffix forms.

src/commands/filter_parser.cc
```diff
@@ -792,24 +816,51 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
...
       return absl::InvalidArgumentError("Field does not support suffix search");
     }
+    // INFIELDS restricts the allowed field set for every term, including
+    // explicit @field: terms. If the explicit field is not in INFIELDS, the
+    // intersection is empty and the term matches nothing (Redis Stack parity).
+    if (options_.infields && !options_.infields->empty() &&
+        !options_.infields->contains(*field_name)) {
+      field_mask = 0ULL;
+      return absl::OkStatus();
+    }
```
allenss-amazon, line 861: Why bother detecting this case if we're not going to treat it as an error? 

I'm fine diverging from Redisearch in a non-sensical case like this.

src/commands/filter_parser.cc
```diff
@@ -792,24 +816,51 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
     if (with_suffix && !text_index->WithSuffixTrie()) {
       return absl::InvalidArgumentError("Field does not support suffix search");
```
allenss-amazon, line 793: Not sure why we aren't just ANDing a global "precomputed-infields-mask" here, which is -1 when the option isn't specified. Why change all of this code?

src/commands/ft.search.json
```diff
@@ -7,7 +7,7 @@
     ],
     "history": [
       [
-        "1.2.0", 
+        "1.2.0",
```
allenss-amazon, line 10: 2.0.0

src/commands/ft_search.cc
```diff
@@ -309,22 +309,37 @@ void SearchCommand::SendReply(ValkeyModuleCtx *ctx,
...
+  if (query::ShouldReturnNoResults(*this)) {
+    ValkeyModule_ReplyWithArray(ctx, 1);
+    ValkeyModule_ReplyWithLongLong(ctx, search_result.total_count);
+    return;
+  }
+
+  // 2. NOCONTENT without SORTBY: skip content fetching entirely.
+  if (no_content && !sortby.has_value()) {
+    SendReplyNoContent(ctx, search_result, *this);
     return;
```
allenss-amazon, line 314: Why change any of this?

src/commands/ft_search.cc
```diff
@@ -309,22 +309,37 @@ void SearchCommand::SendReply(ValkeyModuleCtx *ctx,
...
   ApplySorting(search_result.neighbors, *this);
 
-  // 3. Serialize neighbors based on query type
+  // 5. NOCONTENT with SORTBY: reply without field content.
+  if (no_content) {
+    SendReplyNoContent(ctx, search_result, *this);
+    return;
+  }
+
+  // 6. Serialize neighbors based on query type.
```
allenss-amazon: If this is a bug outside of INFIELDS, then let's make a separate PR. Otherwise why would this change?

rileydes-improving: Sounds good, it was a bug with NOCONTENT+SORTBY not properly sorting.  Reverted all SendReply/response_generator changes from this PR. I'll double check if it has been already reported and if not make a bug report.

src/commands/ft_search.cc
rileydes-improving: Reverted. The SendReply refactor was fixing a SORTBY+NOCONTENT ordering bug (results unsorted when both flags are present).

src/commands/ft.search.json
rileydes-improving: I added another element to the history array that to record the changes in 2.0 being to add infields support (Inkeys will also be added into there). Let me know if you want that changed to just overwrite the previous 1.2 entry.

src/commands/filter_parser.cc
rileydes-improving: Refactored so default masks are `~0ULL`. `SetupTextFieldConfiguration` now always computes the full text field mask, then ANDs with the precomputed INFIELDS mask. When INFIELDS is absent the AND is a no-op. Eliminates all the conditional branching.

src/commands/filter_parser.cc
rileydes-improving: Done. `SetupTextFieldConfiguration` now returns `InvalidArgumentError("Field '<name>' is not in INFIELDS list")` when an explicit `@field:` term references a field not in the INFIELDS set. Consistent with the Comment #3 change (both now error on invalid field usage).

I'll make sure to document the divergence into the valkey docs repo once ready to merge.

src/commands/filter_parser.cc
rileydes-improving: Not all TEXT fields have suffix trie support. When a query uses suffix matching (e.g., `*foo`), only fields with `WithSuffixTrie() == true` can participate. The suffix mask is the subset of INFIELDS-listed TEXT fields that actually support suffix search.

Example: `INFIELDS 2 title body` where `title` has suffix trie and `body` doesn't. Normal mask covers both, suffix mask covers only `title`. Without the distinction, suffix queries would match against fields that lack the trie.

The distinction only matters when the INFIELDS list contains a mix of suffix-capable and non-suffix-capable TEXT fields. If all INFIELDS fields have WITHSUFFIXTRIE (or none do), the two masks are equivalent. The separate mask prevents the query engine from trying to search a suffix tree that doesn't exist on a given field.

I can remove it if the case isn't a concern.

src/commands/filter_parser.cc
rileydes-improving: The ternary at the consumption site is just selecting which precomputed mask to apply based on `with_suffix`. The two forms exist because suffix queries operate on a restricted field subset (see previous comment). After the AND-mask refactor this is simplified a bit.

src/commands/filter_parser.cc
rileydes-improving: So the index object is needed for GetTextFieldNumber() and WithSuffixTrie() and the identifier is needed for the alias. Though they are hitting the same map, I don't see anyway with the existing API to get all the required information in one call.

A GetAttribute() function that returns the Attribute object itself would allow that, but adding that should be another PR.

src/commands/filter_parser.cc
rileydes-improving: Done. `PrecomputeInfieldsMasks` now returns `absl::Status` and raises `InvalidArgumentError` if any field in INFIELDS doesn't exist or is not a TEXT field. This diverges from Redis Stack (which silently ignores non-text fields, I had done this for compatibility) but catches user mistakes early.

Also removed the `test_infields_invalid_field_ignored` compatibility test since we're explicitly diverging here.

I'll make sure the valkey docs have this divergence recorded once this PR is ready to merge

integration/compatibility/generate.py
rileydes-improving: Done. Moved `test_infields_error_bad_count`, `test_infields_error_negative_count`, and `test_infields_error_count_mismatch` to `integration/test_ft_search_infields.py` as standalone behavioral tests.

integration/compatibility/generate.py
rileydes-improving: Removed. These were leftover from testing the `ShouldReturnNoResults` path and aren't INFIELDS-specific.

Aksha1812:
**Reviewers for this PR**

src/commands/filter_parser.cc
```diff
@@ -34,6 +34,55 @@
...
+  for (const auto& field : *infields) {
+    // GetIndex validates the field exists and is a TEXT index (type check).
+    auto index = index_schema.GetIndex(field);
+    if (!index.ok()) {
+      return absl::InvalidArgumentError(absl::StrCat(
+          "INFIELDS field '", field, "' does not exist in the index"));
+    }
+    if (index.value()->GetIndexerType() != indexes::IndexerType::kText) {
+      return absl::InvalidArgumentError(
+          absl::StrCat("INFIELDS field '", field, "' is not a TEXT field"));
```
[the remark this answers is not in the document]
rileydes-improving: This erroring behavior was explicitly requested by @allenss-amazon in review in a previous comment.

src/commands/filter_parser.cc
```diff
@@ -783,24 +832,52 @@ absl::Status FilterParser::SetupTextFieldConfiguration(
...
       return absl::InvalidArgumentError("Field does not support suffix search");
     }
+    // INFIELDS restricts the allowed field set for every term, including
+    // explicit @field: terms. If the explicit field is not in INFIELDS, the
+    // query is invalid — the user asked to search a field they excluded.
+    if (options_.infields && !options_.infields->empty() &&
+        !options_.infields->contains(*field_name)) {
+      return absl::InvalidArgumentError(
+          absl::StrCat("Field '", *field_name, "' is not in INFIELDS list"));
+    }
```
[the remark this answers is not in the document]
rileydes-improving: This erroring behavior was explicitly requested by @allenss-amazon in review, though the discussion is still open so I will not change the documentation until confirmed.

src/commands/ft.search.json
allenss-amazon: This should be 1.3.0

allenss-amazon:
Claude review:

The PR description doesn't match the code

- Unknown or non-TEXT fields: the description and docs/commands/ft.search.md say they are "silently ignored". The code returns an error for both, and the unit and integration tests check for that error.
- @field not in INFIELDS: the description says the term "matches nothing" (Redis Stack behaviour). The code returns an error.
- SORTBY + NOCONTENT fix in ft_search.cc: the description lists it, but ft_search.cc is not in the diff.
- Proto field number: the description says 20, but 20 is already scorer. The code uses 22 and reserves 21 for INKEYS.
- Compatibility tests: the description lists "invalid fields" and error cases, but generate_search.py has neither.

*These all look like simple doc bugs that need to be fixed*

# Additional Issues/tests needed

1. Suffix, fuzzy, exact-phrase and @field inside the list are never run end to end. Some end-to-end testing is needed.
2. Cluster mode isn't tested at all.
3. Scoring isn't tested. Do we know for certain that INFIELDS doesn't affect the scoring?
4. Newer syntax isn't tested with INFIELDS: field-scoped groups @title:(a|b)

PR #984, opened 2026-04-23
Implement FILTER option on FT.CREATE

allenss-amazon:


src/index_schema.cc
```diff
@@ -2144,4 +2182,46 @@ absl::StatusOr<vmsdk::ValkeyVersion> IndexSchema::GetMinVersion(
...
+
+bool IndexSchema::EvaluateFilter(
+    const MutatedAttributes &mutated_attributes) const {
+  FilterRecord record(&mutated_attributes);
+  expr::Expression::EvalContext ctx;
+  auto result = compiled_filter_->Evaluate(ctx, record);
+  // A Nil ("unknown") result means the filter referenced a missing field;
+  // matching Redisearch, such a document is kept. Only a definite false
+  // excludes it.
+  return result.IsNil() || result.IsTrue();
```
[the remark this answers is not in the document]
allenss-amazon: Compatible with Redisearch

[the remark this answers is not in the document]
allenss-amazon: compatible with Redisearch

src/index_schema.cc
```diff
@@ -647,6 +656,24 @@ void IndexSchema::SyncProcessMutation(ValkeyModuleCtx *ctx,
     // front
     text_index_schema_->DeleteKeyData(key);
   }
+  // Apply filter if present: check if the key has actual data (not all
```
yairgott: Ideally this ingestion filtering should be done way earlier in the ingestion pipeline flow. We can safely early return if there is no tracking on any of the record attributes which are indexed and the key is not actively tracked by the mutation tracker.

src/index_schema.cc
```diff
@@ -2144,4 +2182,46 @@ absl::StatusOr<vmsdk::ValkeyVersion> IndexSchema::GetMinVersion(
   }
 }
 
+absl::StatusOr<std::unique_ptr<expr::Expression::AttributeReference>>
+IndexSchema::MakeReference(absl::string_view name, bool create) {
```
yairgott, line 2199: It's not clear from the method name what type reference it makes.

src/filter_expr.h
```diff
@@ -0,0 +1,55 @@
...
+#include "src/expr/expr.h"
+#include "src/expr/value.h"
+#include "src/index_schema.pb.h"
+#include "src/indexes/index_base.h"
+
+namespace valkey_search {
+
+// A compiled reference to an attribute field for filter evaluation.
+// Holds the alias used to look up values in MutatedAttributes.
+class FilterAttributeReference : public expr::Expression::AttributeReference {
```
yairgott, line 22: I wonder if the create filter any different than the aggregate one and to what extent can the aggregate code be reused here.

src/filter_expr.h
```diff
@@ -0,0 +1,55 @@
...
+  indexes::IndexerType type_;
+  data_model::AttributeDataType data_type_;
+};
+
+// Adapter that makes MutatedAttributes accessible through the expression
+// Record interface. Stores a type-erased pointer to MutatedAttributes
+// to avoid circular header dependencies with index_schema.h.
+class FilterRecord : public expr::Expression::Record {
+ public:
+  explicit FilterRecord(const void* mutated_attributes)
```
yairgott: why void*?

src/index_schema.cc
```diff
@@ -2144,4 +2182,46 @@ absl::StatusOr<vmsdk::ValkeyVersion> IndexSchema::GetMinVersion(
...
+  }
+  return std::make_unique<FilterAttributeReference>(
+      std::string(name), indexes::IndexerType::kNone, data_type);
+}
+
+absl::StatusOr<expr::Value> IndexSchema::GetParam(absl::string_view s) const {
+  return absl::NotFoundError(absl::StrCat("Parameter `", s, "` not found"));
+}
+
+bool IndexSchema::EvaluateFilter(
```
yairgott: Can't the existing filter evaluator be reused? If not, the method name should indicate which filter it's evaluating.

src/commands/ft_create_parser.cc
```diff
@@ -646,7 +646,12 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
 
   VMSDK_ASSIGN_OR_RETURN(res, vmsdk::IsParamKeyMatch(kFilterParam, false, itr));
   if (res) {
-    return absl::InvalidArgumentError(NotSupportedParamErrorMsg(kFilterParam));
+    absl::string_view filter_expr;
+    VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, filter_expr));
+    if (filter_expr.empty()) {
+      return absl::InvalidArgumentError("FILTER expression cannot be empty");
+    }
+    index_schema_proto.set_filter(std::string(filter_expr));
```
yairgott, line 654: I might had missed that but I couldn't find a place where you set the filter proto field.

src/expr/expr.cc
```diff
@@ -279,6 +288,53 @@ struct Dyadic : Expression {
...
 
+// Short-circuit && / || for FILTER context, implementing Redisearch's
+// three-valued (SQL NULL-like) logic. A Nil ("unknown", from a missing-field
+// comparison -- see FilterFuncEq) on the left propagates as Nil and keeps the
+// document (IndexSchema::EvaluateFilter keeps on Nil). Otherwise the standard
+// short-circuit applies, and the right operand's value (which may itself be
+// Nil) becomes the result when it is reached. This is order-sensitive, exactly
+// as Redisearch is: `false && Nil` -> false (excludes), but `Nil && false`
+// -> Nil (keeps). The APPLY path keeps the eager FuncLand/FuncLor behavior.
+struct FilterLogical : Expression {
```
yairgott, line 299: 1. Why can't we reuse the aggregate filter existing machinery?
2. The type name should indicate that it's aimed to be used by the create filter.

allenss-amazon: It IS reusing the entire aggregate expression compiler and execution engine.  The only thing that's different is that the core expression compiler/execution engine are parameterized to enable both schema-less and schema-full records to operate on. In the FILTER context it's schema-less. In the aggregate context there's a schema ahead of time.

Unfortunately, the compatible handling of Nil in aggregation isn't the same as during FILTER, so there is some difference in the execution of some of the operators.

src/index_schema.cc
allenss-amazon: The core expression compiler/execution engine are shared with the filter and apply logic in aggregation. However, aggregation uses a schema-based record format that's incompatible with the record format used in the ingestion engine (which is schema-less), so there's some glue logic that's used to map the compiler/execution engine into the two different environments.

integration/test_apply_vs_filter.py
```diff
@@ -0,0 +1,177 @@
...
+
+        results = _run_apply(client, "@status!='inactive'")
+        print(f"\n[{key_type}] APPLY @status!='inactive' results:")
+        for k in sorted(results.keys()):
+            print(f"  {k} => {results[k]}")
+
+        # doc:08 has no status — should evaluate to 1 (true) if missing != value
+        assert b"doc:08" in results, "doc:08 missing from APPLY results entirely"
+        val = results[b"doc:08"]
+        print(f"\n  doc:08 (no status field) => {val}  (expected: 1 = true)")
```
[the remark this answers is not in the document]
allenss-amazon: compatibility tests have been added for this case

src/index_schema.cc
allenss-amazon: It's assumed that any FILTER expression will use data coming from the current key. So the filter test can't happen until we've fetched at least some of the contents of the key. It's certainly true that we could split this into two data fetch phases. And the schema-based orientation of Aggregation would make that better. But that's an optimization that can be done later.

src/index_schema.cc
allenss-amazon: This the implementation of a virtual function, the name is meaningful in the context of the base class...

src/filter_expr.h
allenss-amazon: The aggregation engine's native record format is different than the existing ingestion engine's native record format. I didn't want to redo the entire ingestion engine's basic record format just for FILTER. You might not end up saving as much code as you think.

src/filter_expr.h
allenss-amazon: The comment preceding explains that. This was AI generated. We can certainly restructure the headers to avoid the circular dependency. Once that's done, then the void can be properly typed.

src/filter_expr.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#ifndef VALKEYSEARCH_SRC_FILTER_EXPR_H_
```
yairgott, line 8: We should be using `#pragma once`

src/filter_expr.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
```
yairgott, line 2: 2026

src/filter_expr.cc
```diff
@@ -0,0 +1,93 @@
...
+    filter_record.RecordNumericConversionFailure();
+  }
+  return expr::Value(data_view);
+}
+
+void FilterAttributeReference::Dump(std::ostream& os) const {
+  os << "@" << alias_;
+}
+
+bool IndexSchema::EvaluateFilter(
```
yairgott: This should be under index_schema.cc.

src/filter_expr.cc
```diff
@@ -0,0 +1,93 @@
...
+//
+// FilterRecord is confined to this translation unit:
+// IndexSchema::EvaluateFilter (defined below) is its only constructor and
+// FilterAttributeReference its only consumer, so it can hold a fully-typed
+// pointer to MutatedAttributes without forcing that type into filter_expr.h
+// (which would couple it to index_schema.h).
+class FilterRecord : public expr::Expression::Record {
+ public:
+  FilterRecord(const IndexSchema::MutatedAttributes* mutated_attributes,
+               std::atomic<uint64_t>* numeric_conversion_failures)
```
yairgott: why can't we use `IndexSchema::GetStats()` while removing the const from the returned stats?

Aksha1812:
**Reviewers for this PR**

docs/commands/ft.create.md
```diff
@@ -36,6 +37,10 @@ FT.CREATE <index-name>
 
 - `PREFIX <prefix-count> <prefix>` (optional): If this clause is specified, then only keys that begin with the same bytes as one or more of the specified prefixes will be included into this index. If this clause is omitted, all keys of the correct type will be included. A zero-length prefix would also match all keys of the correct type.
 
+- `FILTER <expression>` (optional): A boolean expression evaluated for each candidate key during ingestion; only keys for which it evaluates to true (or to an unknown/`NULL` result, following three-valued logic) are included in the index. A comparison involving a missing field yields `NULL`, so, for example, a negation such as `@status != 'active'` still admits a key that has no `status` field. The number of keys excluded by the filter is reported by `FT.INFO` as `filter_rejected_keys`.
```
mohanrajendran, line 40: Should we also add a link to the expressions syntax from here? https://valkey.io/topics/search-expressions/

src/filter_expr.h
allenss-amazon: We should be consistent across the codebase on this one. Most of the code uses guards like this. I'm fine with switching, but let's switch everywhere. or remove the #pragma once.

docs/commands/ft.create.md
allenss-amazon: fixed.

Frank-Gu-81:

src/expr/value.cc
```diff
@@ -384,6 +384,66 @@ Value FuncGt(const Value& l, const Value& r) { return Value(l > r); }
...
+static bool EitherNil(const Value& l, const Value& r) {
+  return l.IsNil() || r.IsNil();
+}
+
+Value FilterFuncEq(const Value& l, const Value& r) {
+  if (EitherNil(l, r)) {
+    return Value(Value::Nil("filter ==: missing field"));
+  }
+  return Value(l == r);
+}
```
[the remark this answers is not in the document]
allenss-amazon: This is tested and matches Redisearch.

mohanrajendran (commented):
Could you add integration coverage for:

1. A FILTER that references an alias different from its source field, for example $.status AS state with FILTER @state == 'active'.
2. An RDB save/restore round trip for a filtered index, verifying that the filter is restored and still controls subsequent indexing.

mohanrajendran (approved):

src/commands/ft_create_parser.cc
```diff
@@ -270,6 +282,26 @@ absl::Status ParsePrefixes(vmsdk::ArgsIterator &itr,
...
+                         vmsdk::IsParamKeyMatch(kFilterParam, false, itr));
+  if (!res) {
+    return absl::OkStatus();
+  }
+  absl::string_view filter_expr;
+  VMSDK_RETURN_IF_ERROR(vmsdk::ParseParamValue(itr, filter_expr));
+  if (filter_expr.empty()) {
+    return absl::InvalidArgumentError("FILTER expression cannot be empty");
+  }
+  index_schema_proto.set_filter(std::string(filter_expr));
```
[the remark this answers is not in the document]
allenss-amazon: this is fine.

PR #988, opened 2026-04-24
Run github workflows in 1.0

BCathcart:
I updated the existing github workflows in 1.0 with the latest versions in main and enabled them to run on 1.0 PRs. Right now the integration test workflow fails on the stability tests. @Nivesh-01 will advise on what we want to do with them here. If they're fundamentally broken in 1.0, let's just skip them. Otherwise we can merge a fix for them before we commit this PR. We should follow this up by doing the same for 1.1.


BCathcart:
Looks like there are some existing memory leaks. Previously the ASAN workflows didn't fail if there was a leak, so it's not surprising there are some hiding in 1.0.

Aksha1812:
**Reviewers for this PR**

BCathcart:
This was fixed elsewhere I believe

PR #990, opened 2026-04-24
[DOCUMENTATION] Updating QUICK_START.md

Aksha1812:
Stale Quick start guide . Updating.


QUICK_START.md
allenss-amazon: 9.0.1, not 7.2.6

QUICK_START.md
```diff
@@ -12,48 +12,237 @@ Follow these steps to set up, build, and run the Valkey server with vector searc
...
+For optimal performance, valkey-search matches worker threads to the number of CPU cores on the host. You can override this explicitly:
 
-## Step 3: Create a Vector Index
+```bash
+valkey-server "--loadmodule /path/to/libsearch.so --reader-threads 64 --writer-threads 64"
+```
 
-To enable vector search functionality, you need to create an index for storing vector data.
-Start a Valkey CLI session:
+To enable JSON support, load the JSON module as well:
```
allenss-amazon: Let's put a reference to the JSON module, perhaps directly to the build instructions for it.

QUICK_START.md
```diff
@@ -12,48 +12,237 @@ Follow these steps to set up, build, and run the Valkey server with vector searc
...
+- `vector` is the vector field name for storing the vectors.
+- `VECTOR HNSW` uses the Hierarchical Navigable Small World algorithm for approximate nearest neighbor search. The other option is `VECTOR FLAT` for exact brute-force search.
 - `DIM 3` sets the vector dimensionality to 3.
-- `DISTANCE_METRIC COSINE` sets the distance metric to cosine similarity.
+- `DISTANCE_METRIC COSINE` sets the distance metric to cosine similarity. Other options are `L2` and `IP`.
 
-## Step 4: Insert Some Vectors
 
-To insert vectors, use the `HSET` command:
+### Insert Some Vectors
```
allenss-amazon, line 47: Let's move the comment about vector encoding up to here. the current placement makes it difficult to read in a single pass.

QUICK_START.md
```diff
@@ -12,48 +12,237 @@ Follow these steps to set up, build, and run the Valkey server with vector searc
...
 ```
 
-Replace the `\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80?` and `\x00\xaa\x00\x00\x00\x00\x00\x00\x00\x00\x80?` with actual vectors.
+Replace the `\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80?` and `\x00\xaa\x00\x00\x00\x00\x00\x00\x00\x00\x80?` with actual vectors. Vectors must be encoded as 32-bit IEEE 754 floats in little-endian byte order.
 
-## Step 5: Issue Vector Queries
+### Issue a Vector Query
 
-Now that you've created an index and inserted vectors, you can perform a vector search. Use the `FT.SEARCH` command to find similar vectors:
+Perform a KNN search returning the top 5 nearest vectors:
```
allenss-amazon: KNN Hasn't been defined yet.

QUICK_START.md
```diff
@@ -12,48 +12,237 @@ Follow these steps to set up, build, and run the Valkey server with vector searc
...
 FT.SEARCH myIndex "*=>[KNN 5 @vector $query_vector]" PARAMS 2 query_vector "\xcd\xccL?\x00\x00\x00\x00\x00\x00\x00\x00"
 ```
 
-This command performs a K-nearest neighbors search and returns the top 5 closest vectors to the provided query vector.
+The `*` before `=>` means no pre-filtering — all vectors in the index are searched. Replace `*` with a filter expression (e.g. `@category:{electronics}`) to run a hybrid query that narrows candidates before KNN.
+
+---
+
+## Working with Tag and Numeric Fields
+
```
allenss-amazon, line 77: A sentence or two about use cases for Tag and Numeric would be really helpful.

QUICK_START.md
```diff
@@ -1,59 +1,250 @@
...
+### Create a Vector Index
 
 ```bash
 FT.CREATE myIndex SCHEMA vector VECTOR HNSW 6 TYPE FLOAT32 DIM 3 DISTANCE_METRIC COSINE
 ```
 
-- `vector` is the vector field for storing the vectors.
-- `VECTOR HNSW` specifies the use of the HNSW (Hierarchical Navigable Small World) algorithm for vector search.
+- `vector` is the vector field name for storing the vectors.
+- `VECTOR HNSW` uses the Hierarchical Navigable Small World algorithm for approximate nearest neighbor search. The other option is `VECTOR FLAT` for exact brute-force search.
```
allenss-amazon: let's change the name "vector" to something that's not ambiguous: vector-field-name ?? name?

QUICK_START.md
```diff
@@ -1,59 +1,250 @@
...
+```
+
+Parentheses make a bound exclusive. For example, to find products with a price strictly less than 500:
+
+```bash
+FT.SEARCH products "@price:[-inf (500]"
+```
+
+### Combine Tag and Numeric Filters
+
```
allenss-amazon, line 127: Maybe add a comment that multiple filters separate by spaces are and'ed together.

allenss-amazon: Maybe add an "Or" example too?

QUICK_START.md
```diff
@@ -1,59 +1,250 @@
...
+### Combining Text with Other Filters
+
+Text matchers can be combined with tag and numeric filters in the same query. Separate predicates with a space for `AND`, or `|` for `OR`:
+
+```bash
+FT.CREATE docs ON HASH PREFIX 1 doc: SCHEMA content TEXT category TAG year NUMERIC
+HSET doc:1 content "great introduction to databases" category "tech" year 2024
+HSET doc:2 content "great recipes for summer" category "cooking" year 2023
+
+FT.SEARCH docs "@category:{tech} database*"
```
allenss-amazon, line 205: Maybe add an "or" example too?

QUICK_START.md
```diff
@@ -1,59 +1,264 @@
 # Quick Start
 
-Follow these steps to set up, build, and run the Valkey server with vector search capabilities. This guide will walk you through creating a vector index, inserting vectors, and issuing queries.
+Follow these steps to set up, build, and run the Valkey server with the valkey-search module. This guide walks you through creating indexes, inserting data, and issuing queries across all supported field types — vector, tag, numeric, and full-text.
 
 ## Step 1: Install Valkey and valkey-search
 
-1. Follow the [instructions to build Valkey from source](https://github.com/valkey-io/valkey?tab=readme-ov-file#building-valkey-using-makefile). Make sure to use Valkey version 7.2.6 or later as the previous versions have Valkey module API bugs.
```
allenss-amazon, line 7: I think we should also mention the bundle

QUICK_START.md
```diff
@@ -1,59 +1,264 @@
...
 
 Once valkey-search is built, run the Valkey server with the valkey-search module loaded:
 
 ```bash
-./valkey-server "--loadmodule libsearch.so  --reader-threads 64 --writer-threads 64"
+valkey-server --loadmodule /path/to/libsearch.so
 ```
 
-You should see the Valkey server start, and it will be ready to accept commands.
+For optimal performance, valkey-search matches worker threads to the number of CPU cores on the host. You can override this explicitly:
```
allenss-amazon: This is the quick start, we shouldn't need options -- especially for overriding features that probably do exactly what is desired.

QUICK_START.md
```diff
@@ -1,59 +1,264 @@
...
 
-## Step 3: Create a Vector Index
+```bash
+valkey-server "--loadmodule /path/to/libsearch.so --reader-threads 64 --writer-threads 64"
```
allenss-amazon: remove the thread options.

QUICK_START.md
```diff
@@ -1,59 +1,264 @@
...
+```
+
+To enable JSON support, load the [valkey-json](https://github.com/valkey-io/valkey-json) module as well:
```
allenss-amazon: To enable JSON support, you'll need to load the JSON module. See [here] for instructions on how to build it also....

QUICK_START.md
```diff
@@ -1,59 +1,264 @@
...
-## Step 4: Insert Some Vectors
 
-To insert vectors, use the `HSET` command:
+### Insert Some Vectors
+
+Vectors must be encoded as 32-bit IEEE 754 floats in little-endian byte order. Each vector must have exactly `DIM` elements.
+
+```bash
+HSET my_hash_key_1 embedding "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80?"
+HSET my_hash_key_2 embedding "\x00\xaa\x00\x00\x00\x00\x00\x00\x00\x00\x80?"
```
allenss-amazon, line 55: I think it would be clarifying to have a comment about what the actual numeric value is of these obscure hex blobs.

QUICK_START.md
```diff
@@ -1,59 +1,264 @@
...
+
+Perform a K-Nearest Neighbors (KNN) search returning the top 5 nearest vectors:
 
 ```bash
-HSET my_hash_key_1 vector "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80?"
-HSET my_hash_key_2 vector "\x00\xaa\x00\x00\x00\x00\x00\x00\x00\x00\x80?"
+FT.SEARCH myIndex "*=>[KNN 5 @embedding $query_vector]" PARAMS 2 query_vector "\xcd\xccL?\x00\x00\x00\x00\x00\x00\x00\x00"
 ```
 
-Replace the `\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80?` and `\x00\xaa\x00\x00\x00\x00\x00\x00\x00\x00\x80?` with actual vectors.
```
allenss-amazon, line 49: I don't think dropping this entirely is correct. Rather, I'd update the early text to say something like "Performat a K-Nearest...... which will return the 5 vectors in the database which are nearest to a supplied vector" or something like that.

QUICK_START.md
```diff
@@ -1,59 +1,264 @@
...
+The `*` before `=>` means no pre-filtering — all vectors in the index are searched. Replace `*` with a filter expression (e.g. `@category:{electronics}`) to run a hybrid query that narrows candidates before KNN.
+
+---
+
```
allenss-amazon, line 75: Let's include a hybrid example. OR add a reference to the documentation that has a section on hybrid queries. https://valkey.io/topics/search-query/

allenss-amazon (commented):
I'm wondering if the "quick start" is too long, because there are too many examples? Do they duplicate what's in the docs? Would be make more sense to reduce the size here and move some of these examples to docs and just reference them here?

Aksha1812:
Addressed all comments. I don't think the quick start quide is getting too long , it just we have different indexes we support and will we helpful to give examples of them all for a person who is just starting with working on this repository. Also i have added the https://valkey.io/topics/search-query/ reference at the end of the file. I don't think just referencing that is too helpful. I see other places withing the organization where we reiterate the docs . e.g. https://github.com/valkey-io/valkey-bloom/blob/c6c1558fe045cfd291a032da075fd087a995b9a3/QUICK_START.md?plain=1 . Let me know if you still think we need to make this compact and refer the documentation. let me know if need to address anythign else as well

allenss-amazon (approved):

PR #994, opened 2026-04-29
Fix Tag Search's EntriesFetcherIterator logic to not rely on RootIterator (full tree walk) on standard cases

KarthikSubbarao:
The Tag performance was looking pretty bad in comparison with Text and Numeric.... For Context, with NOCONTENT, we were getting ~50-70K RPS with text term / text prefix searches which had ~1000 matches. Tag was showing 44 RPS for queries with ~100 matches.

In particular, we noticed it was not scaling with the index size.

From looking at the code, we are always creating a root iterator which makes tag searches degrade as the index grows larger. This root iterator is only really needed in negate.

We used the amazon-science data set: https://github.com/amazon-science/esci-data
The Tag Searches were fair and were just having ~100 matches (as the COUNT) per FT.SEARCH command in both the index size of 100K and 1.2M cases.

---
### TAG Scaling: 100K vs 1.2M Documents
Standalone TAG query: `@brand:{brand}` (verified 5-100 matches)  
Tests whether TAG overhead scales with index size or result count.
#### 100K Documents (idx:products_100k)
| Instance | C=1 P50 | QPS C=1 | QPS C=300 | QPS C=1000 |
|----------|---------|---------|-----------|------------|
| **r7g.8xlarge** | 1.855 | 520 | 10,490 | 10,309 |
| **r7g.2xlarge** | 2.031 | 480 | 2,545 | 2,555 |
| **r7g.large** | 1.783 | 520 | 561 | 544 |
#### 1.2M Documents
| Instance | C=1 P50 | QPS C=1 | QPS C=300 | QPS C=1000 |
|----------|---------|---------|-----------|------------|
| **r7g.8xlarge** | 19.631 | 51 | 1,376 | 1,374 |
| **r7g.2xlarge** | 20.431 | 49 | 309 | 316 |
| **r7g.large** | 22.287 | 39 | 44 | 44 |
#### Scaling Impact
| Instance | 100K P50 | 1.2M P50 | Latency increase | 100K QPS (C=300) | 1.2M QPS (C=300) | QPS drop |
|----------|----------|----------|-----------------|-----------------|-----------------|----------|
| **r7g.8xlarge** | 1.855ms | 19.631ms | **10.6×** | 10,490 | 1,376 | **7.6×** |
| **r7g.2xlarge** | 2.031ms | 20.431ms | **10.1×** | 2,545 | 309 | **8.2×** |
| **r7g.large** | 1.783ms | 22.287ms | **12.5×** | 561 | 44 | **12.8×** |

------

From local testing, just with my PR's change of moving the root iterator off the constructor and only into negate path, we see this in perf with using 1000 clients:
```
Summary:
  throughput summary: 76923.08 requests per second
  latency summary (msec):
          avg       min       p50       p95       p99       max
       11.221     5.368    11.055    15.295    18.143    20.703
```

Without my change:
```
Summary:
  throughput summary: 1284.19 requests per second
  latency summary (msec):
          avg       min       p50       p95       p99       max
      739.236    28.848   772.095   798.207   801.791   803.327
```

### DIFF
--- a/src/indexes/tag.cc
+++ b/src/indexes/tag.cc
@@ -263,15 +263,15 @@ Tag::EntriesFetcherIterator::EntriesFetcherIterator(
       untracked_keys_(untracked_keys),
       negate_(negate) {}
 
-void Tag::EntriesFetcherIterator::EnsureTreeIter() {
-  if (!tree_iter_.has_value()) {
-    tree_iter_.emplace(tree_.RootIterator());
+void Tag::EntriesFetcherIterator::EnsureNegateRootIter() {
+  if (!negate_root_iter_.has_value()) {
+    negate_root_iter_.emplace(tree_.RootIterator());
   }
 }
 
 bool Tag::EntriesFetcherIterator::Done() const {
   if (negate_) {
-    return tree_iter_.has_value() && tree_iter_->Done() &&
+    return negate_root_iter_.has_value() && negate_root_iter_->Done() &&
            (untracked_keys_.empty() ||
             (untracked_keys_iter_.has_value() &&
              untracked_keys_iter_.value() == untracked_keys_.end()));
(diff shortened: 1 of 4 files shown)


src/indexes/tag.cc
```diff
@@ -258,14 +258,20 @@ Tag::EntriesFetcherIterator::EntriesFetcherIterator(
     const PatriciaTreeIndex& tree,
     absl::flat_hash_set<PatriciaNodeIndex*>& entries,
     const InternedStringSet& untracked_keys, bool negate)
-    : tree_iter_(tree.RootIterator()),
```
murphyjacob4, line 261: It's a bit surprising to me that construction of an iterator is so expensive.

Should we optimize the iterator construction in the PatriciaTree class, such that this would be O(1)? Basically, instead of deferring creation of the iterator, make the iterator defer traversal until the first element is retrieved?

This change also works, but future users of PatriciaTree may fall into the same trap

KarthikSubbarao: Both work, but for Tag Searches, we don't need to be creating a root iterator for a standard search. It is only needed for negate paths.

I'd say even if the optimization was made in PatriciaTree, we still only need to create the root iterator in the negate path, so the current change holds.

However, I don't mind looking into updating PatriciaTree. I just updated it in the same PR now. 

So, we can keep both changes

murphyjacob4 (approved):

PR #997, opened 2026-04-29
Fix GROUPBY to implicitly load group key fields without explicit LOAD

### LINKED ISSUE #919: [BUG] FT.AGGREGATE GROUPBY Without LOAD Produces Missing Group Key Fields
**Describe the bug**
`FT.AGGREGATE ... GROUPBY 1 @field REDUCE COUNT 0 AS count` returns only the
reducer output (`count`) and omits the group key field (`field`) when no `LOAD`
clause is present. Redis Stack implicitly loads GROUPBY key fields; Valkey Search
did not.

**To Reproduce**
Add the following to `generate.py`, regenerate the answers file and run compatibility test. A branch with the compatibility test and answers file regenerated is available at https://github.com/Bit-Quill/valkey-search/tree/bugfix/groupby-no-load.

```
--- a/integration/compatibility/generate.py
+++ b/integration/compatibility/generate.py
@@ -263,6 +263,35 @@ class TestAggregateCompatibility(BaseCompatibilityTest):
             f"ft.aggregate {key_type}_idx1 * load 6 @__key @n1 @n2 @t1 @t2 @t3 groupby 1 @t3 reduce max 1 @n1 as nmax"
         )
         self.check(dialect, f'ft.aggregate {key_type}_idx1 * load 6 @__key @n1 @n2 @t1 @t2 @t3 groupby 1 @t1 reduce max 1 @n2 as nmax')
+    def test_aggregate_groupby_no_load(self, key_type, dialect):
+        """GROUPBY without LOAD: group key fields must appear in output.
+
+        Redis Stack implicitly loads GROUPBY key fields even when no LOAD
+        clause is present.  This test ensures Valkey Search matches that
+        behaviour.
+        """
+        # Tag field as group key, COUNT reducer — the canonical failing case
+        self.check(dialect,
+            f"ft.aggregate {key_type}_idx1 * groupby 1 @t3 reduce count 0 as count"
+        )
+        # Numeric field as group key, COUNT reducer
+        self.check(dialect,
+            f"ft.aggregate {key_type}_idx1 * groupby 1 @n1 reduce count 0 as count"
+        )
+        # Tag field as group key, SUM reducer on a numeric field
+        self.check(dialect,
+            f"ft.aggregate {key_type}_idx1 * groupby 1 @t3 reduce sum 1 @n1 as total"
+        )
+        # Multiple group keys, no LOAD
+        self.check(dialect,
+            f"ft.aggregate {key_type}_idx1 * groupby 2 @t3 @n1 reduce count 0 as count"
+        )
+        # Regression guard: explicit LOAD must still work correctly
+        self.check(dialect,
+            f"ft.aggregate {key_type}_idx1 * load 1 @t3 groupby 1 @t3 reduce count 0 as count"
+        )
+
     def test_aggregate_limit(self, key_type, dialect):
         self.check(dialect, f"ft.aggregate {key_type}_idx1  * load 3 @__key @n1 @n2")
```

**Expected behavior**
If drop-in compatibility with RediSearch is required, implicitly loading the GROUPBY keys even without a load.

**Environment**
Valkey Search Module Version: 1.2.0-rc2
Valkey Core Version: 9.0.1

**Proposed Fix**
```
--- a/src/commands/ft_aggregate.cc
+++ b/src/commands/ft_aggregate.cc
@@ -54,7 +54,30 @@ absl::Status ManipulateReturnsClause(AggregateParameters &params) {
     CHECK(params.return_attributes.empty());
     return absl::OkStatus();
   } else {
-    for (const auto &load : params.loads_) {
+    std::vector<std::string> loads_to_process = params.loads_;
+
+    // When no LOAD clause is given, implicitly load any fields referenced as
+    // GROUPBY keys so they are fetched from the data store and appear in the
+    // output rows (matching observed Redis behaviour).
+    if (loads_to_process.empty()) {
+      for (const auto &stage : params.stages_) {
+        const auto *groupby = dynamic_cast<const GroupBy *>(stage.get());
+        if (groupby == nullptr) {
+          continue;
+        }
+        for (const auto &group_attr : groupby->groups_) {
+          const auto &name = group_attr->name_;
+          if (name == "__key" ||
+              name == vmsdk::ToStringView(params.score_as.get())) {
+            continue;
+          }
+          loads_to_process.push_back(name);
+        }
+      }
+    }
+
+    for (const auto &load : loads_to_process) {
```

cnuthalapati:
## Summary
* Fix FT.AGGREGATE GROUPBY to implicitly load group key fields into the output when no explicit LOAD clause covers them
* Only implicitly loads fields that exist in the index schema, so chained GROUPBYs using derived fields (reducer outputs) don't break
* Without this fix, GROUPBY results omit the group key fields unless the user redundantly specifies them in LOAD

## Test plan
* Run existing compatibility tests with regenerated answers from the issue's test additions
* Verify `FT.AGGREGATE idx * GROUPBY 1 @field REDUCE COUNT 0 AS count` returns both `field` and `count`
* Verify partial LOAD case: `FT.AGGREGATE idx * LOAD 1 @price GROUPBY 1 @category REDUCE COUNT 0 AS count` returns all three fields
* Verify chained GROUPBY: `FT.AGGREGATE idx * GROUPBY 1 @category REDUCE COUNT 0 AS count GROUPBY 1 @count REDUCE COUNT 0 AS num` doesn't error
* Verify explicit LOAD of GROUPBY key still works (no duplication)

Fixes #919


allenss-amazon:
I wonder if this is a more generic problem. Will fields used in an expression (like in an APPLY) also require being in the LOAD?

cnuthalapati:
Yes, APPLY expressions referencing fields not in LOAD will evaluate against empty values. However, the two cases warrant different treatment.

GROUPBY keys are required to interpret the data. They define what each output row represents. Without the grouping key in output, results are uninterpretable: you get aggregates with no labels. Implicitly loading them is the correct behavior because the user cannot make sense of the output otherwise.

APPLY inputs are only computational. The user asked for the computed result (the AS field), not the source fields. Whether source fields also appear in output is an explicit choice via LOAD.

This PR fixes the GROUPBY case. The APPLY behavior (requiring explicit LOAD for expression inputs) is by design. We should not implicitly load computation fields.

KarthikSubbarao:
Can we add a test in `test_non_vector.py` with the commands we want to run here? @cnuthalapati

Aksha1812:
@allenss-amazon anything remaining here. ?

Aksha1812:
**Reviewers for this PR**

- **First Pass Reviewer: @Aksha1812** — Please do your best to do a detailed review on the PR and get a response on your feedback. Once the first pass is done, notify the maintainer assigned to this PR to follow up on the final review and getting the PR merged. You can reach out to the people owning the relevant code paths for more help on the review.

Aksha1812:
> Can we add a test in `test_non_vector.py` with the commands we want to run here? @cnuthalapati

@cnuthalapati Any update on @KarthikSubbarao's comment above ?

src/commands/ft_aggregate.cc
```diff
@@ -54,7 +56,26 @@ absl::Status ManipulateReturnsClause(AggregateParameters &params) {
     CHECK(params.return_attributes.empty());
     return absl::OkStatus();
   } else {
-    for (const auto &load : params.loads_) {
+    std::vector<std::string> loads_to_process = params.loads_;
+
+    for (const auto &stage : params.stages_) {
+      const auto *groupby = dynamic_cast<const GroupBy *>(stage.get());
```
Aksha1812, line 62: dynamic_cast per stage works but a virtual Stage::CollectImplicitLoads() would avoid RTTI and let SORTBY/future stages opt in without another special case in ManipulateReturnsClause.

Aksha1812 (commented):
Does SORTBY have similar issues https://github.com/valkey-io/valkey-search/blob/2432e1e4f08cf8ddc60c707df37515be941c0bd3/src/commands/ft_aggregate_parser.cc#L162-L166 


But SORTBY resolves @field through the same MakeReference mechanism (ft_aggregate_parser.cc:162, via Expression::Compile), and nothing loads those fields either. So:
```
FT.AGGREGATE idx * SORTBY 2 @price ASC
```
leaves price nil on every record; the comparator sees kEQUAL/kUNORDERED for every pair, continues, returns false, and stable_sort leaves the order untouched — silently unsorted output. Same wrong-answer class as the GROUPBY bug, same fix location, ~3 extra lines.

Aksha1812:
Obsoleted by #1341 can we get this PR closed @allenss-amazon @Frank-Gu-81

allenss-amazon:
FIxed in #1341

PR #999, opened 2026-04-30
Fix crash when loading values of type text from JSON in FT.AGGREGATE

### LINKED ISSUE #998: [BUG] Crash in FT.AGGREGATE when LOAD of a text field with JSON
**Describe the bug**

If a declared TEXT field for a JSON key is used on a LOAD in FT.AGGREGATE, it crashes.

**To Reproduce**

  FT.CREATE  j ON JSON PREFIX 1 j: SCHEMA $.t AS t TEXT
  JSON.SET   j:1 $ '{"t":"x"}'
  FT.AGGREGATE j '@t:x' LOAD 1 @t DIALECT 2     # crashes the server

**Expected behavior**

no crash.

**Environment**

- Valkey Core Version 

**Additional context**

allenss-amazon:
(#998)


src/commands/ft_aggregate.cc
```diff
@@ -68,6 +68,12 @@ absl::Status ManipulateReturnsClause(AggregateParameters &params) {
       content = true;
       VMSDK_ASSIGN_OR_RETURN(auto indexer, params.index_schema->GetIndex(load));
       auto indexer_type = indexer->GetIndexerType();
+      if (indexer_type == indexes::IndexerType::kVector ||
+          indexer_type == indexes::IndexerType::kHNSW ||
+          indexer_type == indexes::IndexerType::kFlat) {
+        return absl::InvalidArgumentError(absl::StrCat(
+            "Loading of vector fields is not supported (field `", load, "`)"));
+      }
```
KarthikSubbarao: This looks like a different crash case than what the PR calls out.. Maybe we can update the PR title/desc to people to know it covers both

But as a nit, we can use `IsVectorIndex()` to cover all 3 cases

src/commands/ft_aggregate.cc
```diff
@@ -68,6 +68,10 @@ absl::Status ManipulateReturnsClause(AggregateParameters &params) {
       content = true;
       VMSDK_ASSIGN_OR_RETURN(auto indexer, params.index_schema->GetIndex(load));
       auto indexer_type = indexer->GetIndexerType();
+      if (indexer->IsVectorIndex()) {
+        return absl::InvalidArgumentError(absl::StrCat(
+            "Loading of vector fields is not supported (field `", load, "`)"));
+      }
```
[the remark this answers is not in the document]
allenss-amazon: That get's handled at load time, dynamically.

src/commands/ft_aggregate.cc
```diff
@@ -143,6 +147,7 @@ bool ReplyWithValue(ValkeyModuleCtx *ctx,
     } else {
       switch (indexer_type) {
         case indexes::IndexerType::kTag:
+        case indexes::IndexerType::kText:
         case indexes::IndexerType::kNone: {
           value_view = value.AsStringView();
           break;
```
Aksha1812, line 153: for the deafult case which crashed due to text being not there `(CHECK .....`  do we want exit more gracefully here such that it doesn't crash?

integration/test_aggregate_load_text_vector.py
```diff
@@ -0,0 +1,99 @@
...
+                assert client.execute_command(
+                    "JSON.SET", f"doc:{i}", "$", json.dumps(doc),
+                ) == b"OK"
+
+    @pytest.mark.parametrize("key_type", ["HASH", "JSON"])
+    def test_load_text_field_does_not_crash(self, key_type):
+        client: Valkey = self.server.get_new_client()
+        self._create_index(client, key_type)
+
+        result = client.execute_command(
```
Aksha1812, line 73: can we extend test to check  `LOAD *` case  and more aggregation stages `GROUPBY..` etc... . also just for my knowledge what happens when we load both text and vector fields together ?

Aksha1812 (commented):
just two small comments . otherwise LGTM ! will approve post your response .

Aksha1812:
Also i guess this bug fix is a behaviour change. Are we sure this should go in patch release . Does the documention say we support text for aggregate load ?

src/commands/ft_aggregate.cc
allenss-amazon: No, for this kind of logic error, the sooner you find out the sooner you fix it. IMO, when you detect a logic error or contract violation, it's best to crash as this provides the most context and allows it to get fixed quickly. I find that silently ignoring errors at one place in the code usually causes failures elsewhere that are hard to diagnose.

allenss-amazon:
Fixing crashes definitely belongs in patch releases.

integration/test_aggregate_load_text_vector.py
allenss-amazon: There's quite a bit of testing of those stages in the compatibility test suite.

PR #1001, opened 2026-04-30
Implement FLOAT16 and BFLOAT16

allenss-amazon:
Adds `FLOAT16` and `BFLOAT16` vector element types for both HASH and JSON documents, across FLAT and HNSW, for all three distance metrics.

All float16/bfloat16 distance computations perform their intermediate computation in fp32 or wider. Unit tests are included for all distance metrics to verify correct intermediate computation sizes. Correct execution was verified against X86-EPYC, Graviton-2, -3 and -4.

## Measured results

Cohere wiki-embeddings, 10k x 768, HNSW M=16 EF_CONSTRUCTION=200 EF_RUNTIME=100, recall@10 against an exact per-metric ground truth over 100 queries. Identical protocol on all four CPUs: server pinned to 32 physical cores with 32 reader/writer threads, load generator on a disjoint core set. Latency is 500 single-connection queries drawn at random, round-robin interleaved across all 9 resident indexes over 5 passes; throughput is the peak of a 16-128 connection sweep.

Recall@10 is identical on all four CPUs for all nine (element type, metric) combinations -- 0.995 to 1.000.

CPU coverage: Graviton 2 (Neoverse N1, `asimdhp` only), Graviton 3 (Neoverse V1, `+asimdfhm +bf16 +sve`), Graviton 4 (Neoverse V2, `+sve2`), x86-64 (AMD EPYC 7R13).

### Averages across the 9 (element type, metric) configurations

| CPU | mean p50 ms | mean max qps | mean ingest s |
|---|---:|---:|---:|
| Graviton 2 | 0.747 | 13,373 | 13.4 |
| Graviton 3 | 0.448 | 18,194 | 8.5 |
| Graviton 4 | 0.329 | 28,927 | 7.1 |
| x86 EPYC | 0.843 | 19,745 | 9.9 |

At equal core count, Graviton 3 is 1.36x Graviton 2 on throughput and 1.67x on latency; Graviton 4 is 1.59x Graviton 3 on throughput and 2.16x Graviton 2 overall. The x86 host was shared and carrying other load, so its row is indicative only; the three Gravitons were idle and dedicated.

The performance measurements were taken using a harness on the same machine, thus these numbers should just be an indicator of relative performance.

### 16-bit vs FLOAT32, same CPU and metric

| CPU | metric | FP16 qps | FP16 p50 | BF16 qps | BF16 p50 |
|---|---|---:|---:|---:|---:|
| Graviton 2 | L2 | +5.2% | 1.04x | +6.2% | 0.91x |
| Graviton 2 | IP | +4.1% | 1.05x | +5.5% | 0.85x |
| Graviton 2 | COSINE | +6.0% | 1.01x | +6.2% | 0.83x |
| Graviton 3 | L2 | +6.5% | 1.10x | +6.1% | 1.03x |
| Graviton 3 | IP | +7.1% | 0.87x | +8.4% | 0.87x |
| Graviton 3 | COSINE | +6.6% | 0.86x | +6.4% | 0.86x |
| Graviton 4 | L2 | +1.3% | 0.96x | +2.6% | 0.87x |
| Graviton 4 | IP | +4.1% | 0.73x | +4.7% | 0.73x |
| Graviton 4 | COSINE | +2.0% | 0.73x | +2.9% | 0.73x |
| x86 EPYC | L2 | +3.9% | 1.10x | +4.0% | 1.08x |
| x86 EPYC | IP | +0.8% | 0.93x | -1.4% | 0.89x |
| x86 EPYC | COSINE | +4.9% | 0.87x | +5.0% | 1.00x |

qps columns are the change in peak throughput; p50 columns are the ratio of median single-query latency to FLOAT32 (lower is better). The 16-bit types are net-positive on every Graviton. BF16 edges out FP16 nearly everywhere, most visibly on Graviton 2, where BF16 uses the new shift-based NEON kernels while FP16 falls back to plain NEON widening. The exception is L2 latency on Graviton 3 and x86: L2 has no FHM kernel (`l2sq` is routed to the widening NEON path), so it pays the conversion without a matching instruction win and only the halved memory traffic remains.

### Memory consumption (x86, 10k x 768 vectors, HNSW M=16)

Measured on a fresh server per configuration: vectors loaded as hashes with no index defined, memory snapshotted, then `FT.CREATE` and backfill, and snapshotted again, so index cost is isolated from keyspace cost. Results were identical across all three distance metrics and identical between FLOAT16 and BFLOAT16.

| component | FLOAT32 | FLOAT16 | BFLOAT16 | 16-bit vs FP32 |
|---|---:|---:|---:|---:|
| HNSW index | 43.4 MiB | 28.1 MiB | 28.1 MiB | **1.545x smaller (35.3% less)** |
| Hash keyspace | 36.4 MiB | 19.3 MiB | 19.3 MiB | **1.887x smaller (47.0% less)** |
| Total (keyspace + index) | 45.8 MiB | 30.5 MiB | 30.5 MiB | **1.502x smaller (33.4% less)** |

The index reduction is 1.545x rather than the raw 2x of the vector payload because the HNSW graph links are a fixed cost independent of element width: at M=16 the level-0 link block is 132 bytes per element against 3072 bytes of FLOAT32 payload or 1536 bytes of 16-bit payload.


src/indexes/vector_base.cc
```diff
@@ -110,38 +130,55 @@ query::EvaluationResult PrefilterEvaluator::EvaluateText(
...
+                                               (bfloat16 *)record.data(),
+                                               ret.size() / sizeof(bfloat16));
+      if (magnitude) *magnitude = result;
+      return ret;
+    }
+    default:
+      CHECK(false) << "unsupported vector data type";
   }
-  CHECK(false) << "unsupported type size";
 }
```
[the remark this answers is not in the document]
allenss-amazon: these are known aligned.

src/indexes/vector_base.cc
```diff
@@ -110,38 +130,55 @@ query::EvaluationResult PrefilterEvaluator::EvaluateText(
...
 
-template <typename T>
+template <typename StorageT>
 void VectorBase::Init(int dimensions,
                       valkey_search::data_model::DistanceMetric distance_metric,
-                      std::unique_ptr<hnswlib::SpaceInterface<T>> &space) {
-  space = CreateSpace<T>(dimensions, distance_metric);
+                      std::unique_ptr<hnswlib::SpaceInterface<float>> &space) {
```
yairgott: This should use the template type, e.g.: `hnswlib::SpaceInterface<StorageT>`

src/indexes/vector_base.cc
```diff
@@ -110,38 +130,55 @@ query::EvaluationResult PrefilterEvaluator::EvaluateText(
 }
 
 template <typename T>
-T CopyAndNormalizeEmbedding(T *dst, T *src, size_t size) {
-  T magnitude = 0.0f;
+float CopyAndNormalizeEmbedding(T *dst, T *src, size_t size) {
```
yairgott, line 88: This should remain to use the template type.

yairgott (commented):
I think it would be cleaner if we were to introduce a vector interface, .e.g. VectorInterface, type and change VectorBase to become templatized.

src/indexes/vector_flat.cc
```diff
@@ -244,16 +239,16 @@ absl::StatusOr<std::vector<Neighbor>> VectorFlat<T>::Search(
       return absl::InternalError(e.what());
     }
   };
-  if (normalize_) {
-    auto norm_record = NormalizeEmbedding(query, GetVectorDataType());
+  if (this->normalize_) {
```
yairgott, line 242: We shouldn't be using `this->` to access members or methods.

yairgott (commented):
Looking at vector_base.cc/h, it still uses in many places references to float. My expectation is that after this change, vector_base.h would remain just an interface while the current vector_base implementation would move to vector_type with the appropriate changes.

yairgott:
> Looking at vector_base.cc/h, it still uses in many places references to float. My expectation is that after this change, vector_base.h would remain just an interface while the current vector_base implementation would move to vector_type with the appropriate changes.

In addition, with this change, I think that we should be able to eliminate the method GetDataTypeSize.

Aksha1812:
**Reviewers for this PR**

src/indexes/vector_base.cc
```diff
@@ -579,12 +538,35 @@ absl::Status VectorBase::ForEachUnTrackedKey(
...
-template void VectorBase::Init<float>(
-    int dimensions, data_model::DistanceMetric distance_metric,
-    std::unique_ptr<hnswlib::SpaceInterface<float>> &space);
-
 template absl::StatusOr<std::vector<Neighbor>> VectorBase::CreateReply<float>(
     std::priority_queue<std::pair<float, hnswlib::labeltype>> &knn_res);
+
+absl::Status CheckSimsimdBf16Capability() {
+#if defined(__SSE2__) || defined(__AVX512F__) || \
+    defined(__ARM_BF16_FORMAT_ALTERNATIVE__)
```
Frank-Gu-81, line 600: Not sure if this is intended but seems like there's an inconsistency where lib.c selects the BF16 path with __APPLE__ carved out first:
  
```
#if defined(__APPLE__)
#define SIMSIMD_NATIVE_BF16 0        // serial path is BIT-CORRECT here
#elif defined(__SSE2__) || defined(__AVX512F__) || defined(__ARM_BF16_FORMAT_ALTERNATIVE__)
#define SIMSIMD_NATIVE_BF16 1
```
But CheckSimsimdBf16Capability() re-derives the predicate without the __APPLE__ carve-out:
```
#if defined(__SSE2__) || defined(__AVX512F__) || defined(__ARM_BF16_FORMAT_ALTERNATIVE__)
// require simsimd_uses_haswell()||genoa()||sapphire()||neon_bf16()||sve_bf16()
#endif
```

Frank-Gu-81 (commented):
Should we also update `ft.create.md` documenting FLOAT16/BFLOAT16?

src/indexes/vector_base.cc
```diff
@@ -582,8 +635,43 @@ template void VectorBase::Init<float>(
...
+  const bool has_simd_safe_bf16_path =
+      simsimd_uses_haswell() || simsimd_uses_genoa() ||
+      simsimd_uses_sapphire() || simsimd_uses_neon_bf16() ||
+      simsimd_uses_sve_bf16();
+  if (!has_simd_safe_bf16_path) {
+    return absl::FailedPreconditionError(
+        "BFLOAT16 indexes require a SIMD-targeted BF16 path "
+        "(Haswell/Genoa/Sapphire on x86, NEON-BF16/SVE-BF16 on ARM). "
+        "simsimd's serial BF16 fallback is unsafe under the current build "
+        "(SIMSIMD_NATIVE_BF16=1 in third_party/simsimd/c/lib.c). Either "
```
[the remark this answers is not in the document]
allenss-amazon: AppleOS isn't a supported platform. As long as it compiles that's fine.

src/query/search.cc
```diff
@@ -586,20 +565,38 @@ absl::StatusOr<std::vector<indexes::Neighbor>> MaybeAddIndexedContent(
...
-              attribute_value = vmsdk::MakeUniqueValkeyString(
-                  StringFormatVector(vector.value()));
-            } else {
-              attribute_value =
-                  vmsdk::UniqueValkeyString(ValkeyModule_CreateString(
-                      nullptr, vector->data(), vector->size()));
-            }
+            attribute_value =
+                vmsdk::UniqueValkeyString(ValkeyModule_CreateString(
+                    nullptr, vector->data(), vector->size()));
```
Frank-Gu-81, line 599: The current return value is constructed from `StringFormatVector` with the new one is the exact original JSON text fetched from the document. Would this cause a behavioral change that could be treated as a breaking change? 

i.e, supposed a FLOAT32 JSON index witha document like : `{"embedding": [1.0, 2.5, 3.0]}` and the client runs search on this:

Current result: "[1,2.5,3]" vs new result: "[1.0, 2.5, 3.0]"

This seems fine for most cases if client does a `json.loads(), but seems like it would break if client is doing a string compare or something similar.

src/indexes/bfloat16.h
```diff
@@ -0,0 +1,54 @@
...
+
+  bfloat16() = default;
+
+  explicit bfloat16(float v) {
+    uint32_t u;
+    std::memcpy(&u, &v, sizeof(u));
+    uint32_t rounding_bias = 0x7FFFu + ((u >> 16) & 1u);
+    u += rounding_bias;
+    bits = static_cast<uint16_t>(u >> 16);
+  }
```
Frank-Gu-81, line 39: For the comment above:
> // NaN/Inf caveat: the rounding bias can promote a finite-with-saturated-
// exponent value into a NaN; the engine never feeds those through here
// (vectors are validated upstream), so we don't special-case them.

I asked claude to do a scan in this repo and looks like `NormalizeStringRecord` only checks that `SimpleAtof`, which accepts nan and inf, parses. I don't have too much context on this and the guard might have existed in the upstream and I just missed it. In that case, please ignore this comment. Just wanted to flag here in case we actually don't have the guard in upstream.

src/version.h
```diff
@@ -13,7 +13,7 @@
 //
 // Set the module version to the current release
 //
-constexpr auto kModuleVersion = vmsdk::ValkeyVersion(1, 2, 1);
+constexpr auto kModuleVersion = vmsdk::ValkeyVersion(1, 3, 0);
```
Frank-Gu-81, line 16: Should we do the version bump within this PR? I was thinking separating them might be cleaner for tracking purposes. @VoletiRam has a separate PR for this: https://github.com/valkey-io/valkey-search/pull/1302/changes

src/indexes/bfloat16.h
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
```
Frank-Gu-81: Nit: since this is a newly created file, should the copyright year be 2026? Same as the other new files created by this PR.

src/indexes/vector_flat.cc
Frank-Gu-81, line 211: This is not part of this PR, but i thought address of a reference can never be NULL, so looks like this is redundant? Anyways, not a big issue.

src/indexes/vector_hnsw.cc
Frank-Gu-81, line 40: Nit: duplicate import statements. also not introduced in this PR, but I wanted to flag it so that we can clean up later.

src/indexes/vector_type.cc
```diff
@@ -0,0 +1,170 @@
...
+void VectorType<T>::Init(data_model::DistanceMetric distance_metric) {
+  space_ = CreateSpace<T>(dimensions_, distance_metric);
+  distance_metric_ = distance_metric;
+  if (distance_metric == data_model::DistanceMetric::DISTANCE_METRIC_COSINE) {
+    normalize_ = true;
+  }
+}
+
+template <typename T>
+float VectorType<T>::ComputeReciprocalMagnitude(
```
yairgott, line 111: Should ComputeReciprocalMagnitude be a free function rather than a method of VectorType<T>? Since it likely doesn't need to mutate or access the private state of the class, extracting it would improve encapsulation and make it easier to reuse or test independently.

src/index_schema.cc
```diff
@@ -2209,13 +2225,22 @@ absl::StatusOr<vmsdk::ValkeyVersion> IndexSchema::GetMinVersion(
         "calculation");
   }
   bool has_text_index = false;
+  bool has_low_precision_vector = false;
   for (const auto &attr : unpacked->attributes()) {
     if (attr.index().has_text_index()) {
       has_text_index = true;
-      break;
+    }
+    if (attr.index().has_vector_index()) {
```
yairgott, line 2249: I wonder if it would be cleaner to encapsulate this check inside VectorBase. We could add a pure virtual method, HasLowPrecisionVectors, and have the VectorType implement it.

src/indexes/vector_base.cc
```diff
@@ -589,13 +592,35 @@ VectorBase::~VectorBase() {
...
 }
 
-template void VectorBase::Init<float>(
-    int dimensions, data_model::DistanceMetric distance_metric,
-    std::unique_ptr<hnswlib::SpaceInterface<float>> &space);
-
 template absl::StatusOr<std::vector<Neighbor>> VectorBase::CreateReply<float>(
     std::priority_queue<std::pair<float, hnswlib::labeltype>> &knn_res);
 
+absl::Status CheckSimsimdBf16Capability() {
```
yairgott, line 598: This may cause  `BFLOAT16` index creation and RDB loading to fail on common server CPUs (e.g., AWS Graviton 2 / Neoverse N1, which lack `FEAT_BF16`, as well as older x86 machines/VMs).

Instead of gating hardware at runtime, we can fix the scalar decompression in `third_party/simsimd/include/simsimd/types.h:310-316`:

```
#ifndef SIMSIMD_UNCOMPRESS_BF16
#if SIMSIMD_NATIVE_BF16
#define SIMSIMD_UNCOMPRESS_BF16(x) (simsimd_uncompress_bf16(*(const unsigned short*)&(x)))
#else
#define SIMSIMD_UNCOMPRESS_BF16(x) (simsimd_uncompress_bf16(x))
#endif
#endif

File: src/indexes/vector_type.cc [hunk 14]
```diff
@@ -0,0 +1,170 @@
...
+}
+
+template <typename T>
+vmsdk::UniqueValkeyString VectorType<T>::NormalizeStringRecord(
+    vmsdk::UniqueValkeyString record) const {
+  auto record_str = vmsdk::ToStringView(record.get());
+  if (absl::ConsumePrefix(&record_str, "[")) {
+    absl::ConsumeSuffix(&record_str, "]");
+  }
+  std::vector<std::string> float_strings =
```
yairgott: There are two sources of unnecessary heap allocations here:
1. `std::vector<std::string> float_strings = absl::StrSplit(...)` allocates both a vector and temporary `std::string` buffers for every dimension.
2. `binary_string += std::string((char *)&value, sizeof(T));` constructs and destroys a temporary `std::string` on *every* element. For a 1536-dim vector, this performs 1536 allocations.
We can iterate over `absl::string_view` directly and use `append`, while also validating that the parsed dimension count matches `dimensions_` early

src/indexes/vector_type.cc
```diff
@@ -0,0 +1,170 @@
...
+                                 record.size() / sizeof(T));
+}
+
+template <typename T>
+void VectorType<T>::EmitDataTypeInfo(ValkeyModuleCtx *ctx) const {
+  ValkeyModule_ReplyWithSimpleString(ctx, "data_type");
+  // LookupKeyByValue returns a string_view, whose data() is not guaranteed
+  // NUL-terminated; materialize it before handing it to the C API.
+  ValkeyModule_ReplyWithSimpleString(
+      ctx, std::string(LookupKeyByValue(*kVectorDataTypeByStr,
```
yairgott, line 130: Nit: Avoid the associated allocation of `std::string` by using `ValkeyModule_ReplyWithStringBuffer` with the returned absl::string_view from the call to the LookupKeyByValue

third_party/hnswlib/space_ip_bfloat16.h
```diff
@@ -0,0 +1,82 @@
...
+#endif
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
+#pragma GCC diagnostic ignored "-Wunused-variable"
+namespace hnswlib {
+
+static float
+InnerProductBF16(const void *pVect1, const void *pVect2, const void *qty_ptr) {
+    bfloat16 *pVect1f = (bfloat16 *) pVect1;
```
yairgott: Here and some other places below, use C++ style cast, e.g. `const bfloat16 *pVect1f = static_cast<const bfloat16 *> pVect1;`

src/indexes/vector_hnsw.cc
```diff
@@ -165,7 +167,7 @@ absl::Status VectorHNSW<T>::AddRecordImpl(
       absl::ReaderMutexLock lock(&resize_mutex_);
 
       algo_->addPoint(QueryVector(std::move(vector_record), GetVectorDataSize(),
-                                  normalize_),
+                                  normalize_, this->GetVectorDataType()),
```
yairgott: why do you use `this->`?

src/vector_registry.cc
```diff
@@ -131,9 +132,8 @@ std::shared_ptr<indexes::VectorRecord> VectorRegistry::Track(
       }
 
       if (!vector_record) {
-        float reciprocal_magnitude = indexes::CalcReciprocalMagnitude(
-            reinterpret_cast<const float *>(vector_str.data()),
-            vector_str.size() / sizeof(float));
+        float reciprocal_magnitude =
```
yairgott, line 135: If multiple schemas are configured over the same key with different vector data types , `VectorRegistry` would share or overwrite the record without distinguishing element type, leading to incorrect `reciprocal_magnitude_` calculations.

While this scenario is unlikely, I suggest we defend against it by including the vector data type in the vector registry key. WDYT?

src/query/search.cc
allenss-amazon: This path is removed. Now when returning a JSON vector it always uses the original JSON value. This properly handles the case where the JSON value isn't even an array, it's just a string.

src/version.h
allenss-amazon: I need 1.3, since I generate 1.3 RDB files (FP16) I can't load them into a 1.2.1 module for testing. We'll just have to collide on the PRs.

src/indexes/vector_flat.cc
allenss-amazon: removed

src/indexes/vector_hnsw.cc
allenss-amazon: removed

src/indexes/vector_base.cc
allenss-amazon: It's probably broken on Apple. It's not a supported platform and I don't have a way to test it.

src/indexes/bfloat16.h
allenss-amazon: Bfloat16 supports inf and nan. The code above properly implements the conversion.

src/vector_registry.cc
allenss-amazon: No, it has no usefulness and that will kill the performance. I'm going to make it an error.

src/indexes/vector_type.cc
allenss-amazon: This change causes a wire-format difference, so not going to do it.

src/indexes/vector_type.cc
allenss-amazon: Fixed

src/indexes/vector_hnsw.cc
allenss-amazon: fixed

allenss-amazon:
## Memory consumption: FLOAT32 vs 16-bit element types

Measured on x86-64 (AMD EPYC 7R13), 10,000 Cohere wiki-embedding vectors of 768 dimensions, HNSW `M=16 EF_CONSTRUCTION=200 EF_RUNTIME=100`.

Method: a fresh server per configuration, so no allocator state leaks between runs. Vectors are loaded as hashes with **no index defined**, memory is snapshotted, then `FT.CREATE` runs and backfill builds the index, and memory is snapshotted again. That isolates index cost from keyspace cost. Keyspace is the `used_memory` delta; index is the delta in the module's own `search_used_memory_bytes`.

Results were identical across all three distance metrics (L2, IP, COSINE) and identical between FLOAT16 and BFLOAT16, as expected for equal element width.

| component | FLOAT32 | FLOAT16 | BFLOAT16 | 16-bit vs FP32 |
|---|---:|---:|---:|---:|
| HNSW index | 43.4 MiB | 28.1 MiB | 28.1 MiB | **1.545x smaller (35.3% less)** |
| Hash keyspace | 36.4 MiB | 19.3 MiB | 19.3 MiB | **1.887x smaller (47.0% less)** |
| Total (keyspace + index) | 45.8 MiB | 30.5 MiB | 30.5 MiB | **1.502x smaller (33.4% less)** |

Exact byte counts: index 45,505,376 -> 29,446,256 (ratio 1.5454); total 48,039,077 -> 31,980,107 (ratio 1.5022).

The index shrinks by 1.545x rather than the raw 2x of the vector payload because the HNSW graph links are a fixed cost independent of element width. At `M=16`, hnswlib's level-0 block is `maxM0 * sizeof(tableint) + sizeof(linklistsizeint)` = 132 bytes per element, plus an 8-byte label, against 3072 bytes of FLOAT32 payload or 1536 bytes of 16-bit payload -- so per element the stored record goes from 3212 to 1676 bytes. The keyspace, which carries no such per-element overhead, does realize the full ~1.9x.

src/index_schema.cc
allenss-amazon: This code operates on the protobuf, not an object. So no virtual dispatch available.

src/indexes/vector_type.cc
allenss-amazon: it'll just mean a virtual dispatch elsewhere.

src/indexes/bfloat16.h
allenss-amazon: fixed

src/indexes/vector_base.cc
allenss-amazon: Current code fully supports and is optimized for Grav 2, 3 and 4. variants. If the code fails the CPU processor check then the RDB load is aborted.

third_party/hnswlib/space_ip_bfloat16.h
allenss-amazon: Fixed the ones added here.

src/schema_manager.cc
```diff
@@ -188,6 +190,111 @@ absl::Status SchemaManager::ImportIndexSchema(
...
+      data_model::AttributeDataType::ATTRIBUTE_DATA_TYPE_HASH) {
+    return absl::OkStatus();
+  }
+
+  absl::flat_hash_map<absl::string_view, data_model::VectorDataType>
+      new_vector_fields;
+  for (const auto &attribute : new_proto.attributes()) {
+    if (attribute.index().has_vector_index()) {
+      new_vector_fields[attribute.identifier()] =
+          attribute.index().vector_index().vector_data_type();
```
Frank-Gu-81: I think this only guards the cross-schema case. Within a single `FT.CREATE`, the same identifier can now appear twice under different aliases (due to recent changes in #1195), and this map collapses vec→FLOAT16 / vec→BFLOAT16 to one entry before the check runs and we only compare against `existing_schemas`, never the new proto against itself. Since `RegistryKey` doesn't include the element type, both attributes share one record and the second gets wrong distances. Should we detect a repeated identifier with conflicting types inside new_proto here (or reject it in `IndexSchema::AddIndex`), and add a same-schema case to `test_vector_registry.py`?

```
FT.CREATE idx ON HASH SCHEMA  vec AS v1 VECTOR ... TYPE FLOAT16 ...  vec AS v2 VECTOR ... TYPE BFLOAT16 ...
```

allenss-amazon: good catch.

Frank-Gu-81 (approved):

PR #1008, opened 2026-05-01
Add SCORE_FIELD support and SCORE range validation to FT.CREATE

Frank-Gu-81:
## Summary

Adds support for the `SCORE_FIELD` parameter in `FT.CREATE`, changes `SCORE` validation from accepting only `1.0` to accepting any value in `[0.0, 1.0]`, and reads per-document score field values during ingestion.

## Changes

- **index_schema.proto**: Added `optional string score_field = 14` to persist the score field name across RDB save/load
- **ft_create_parser.cc**: Parse `SCORE_FIELD <field_name>` instead of rejecting it; validate `SCORE` range [0.0, 1.0]
- **index_schema.h/cc**: Added `score_field_`, `score_`, and `document_score` member variables with getters, constructor initialization, `ToProto()` serialization, and `RespondWithInfo()` output. Added `GetDocumentScore()` API for the query path.
- **index_schema.cc (ingestion)**: Read the score field value from the hash during `ProcessKeyspaceNotification`, thread it through `ProcessMutation` → `TrackMutatedRecord` → `ConsumeTrackedMutatedAttribute`, and store it in `IndexKeyInfo` for query-time access.
- **ft.create.json**: Added `SCORE_FIELD` to the command spec
- **testing/common.h**: Extended `MockIndexSchema::Create()` with `score` and `score_field` parameters
- **ft_create_parser_test.cc**: Added `score_field_supported` success test, updated score error messages, added negative score test, added score_field assertion
- **ft_info_test.cc**: Updated expected FT.INFO output to include `score_field_name` and dynamic `default_score`
- **index_schema_test.cc**: Added `IndexSchemaScoreFieldTest` with tests for score field ingestion (happy path + fallback to default)

## What this enables

The system now fully tracks per-document custom scores: it knows which hash field to look at (`SCORE_FIELD`), reads the value during ingestion, and stores it per-document in `IndexKeyInfo`. The scoring algorithm (TF-IDF/BM25) can retrieve document scores via `GetDocumentScore()` at query time.

## Testing

- Full suites: `commands_test` (375 passed), `indexes_test` (169 passed)
- Manual testing: verified FT.CREATE with SCORE_FIELD, FT.INFO output, SCORE range validation, and SCORE+SCORE_FIELD together against a live valkey-server


src/index_schema.h
```diff
@@ -395,6 +398,8 @@ class IndexSchema : public KeyspaceEventSubscription,
   bool with_offsets_{true};
   std::vector<std::string> stop_words_;
   uint32_t min_stem_size_{4};
+  float score_{1.0};
```
KarthikSubbarao: Is the default 1.0 in Re Search? Is the default value fixed or is the default configurable?

src/index_schema.cc
```diff
@@ -1094,10 +1098,12 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
...
   }
-  // hard-code default score of 1 as it's the only value we currently
-  // supported.
   ValkeyModule_ReplyWithSimpleString(ctx, "default_score");
-  ValkeyModule_ReplyWithCString(ctx, "1");
+  ValkeyModule_ReplyWithCString(ctx, absl::StrCat(score_).c_str());
+
+  ValkeyModule_ReplyWithSimpleString(ctx, "score_field_name");
+  ValkeyModule_ReplyWithSimpleString(
+      ctx, score_field_.empty() ? "" : score_field_.c_str());
```
KarthikSubbarao: 1. Does this format match Re Search?
2. Does Re Search always include `score_field_name` even when not provided in index schema creation? or does it hide it?

src/commands/ft_create_parser.cc
```diff
@@ -287,16 +287,21 @@ absl::Status ParseScore(vmsdk::ArgsIterator &itr,
   float score{1.0};
   VMSDK_ASSIGN_OR_RETURN(auto res,
                          vmsdk::ParseParam(kScoreParam, false, itr, score));
-  if (res && score != 1.0) {
-    return absl::InvalidArgumentError(
-        absl::StrCat("`", kScoreParam, "` parameter with a value `", score,
-                     "` is not supported. The only supported value is `1.0`"));
+  if (res) {
+    if (score < 0.0 || score > 1.0) {
```
KarthikSubbarao, line 292: Is this range controlled by a config in Re Search? Or is it always fixed to between 0 and 1?

src/index_schema.proto
```diff
@@ -36,6 +36,7 @@ message IndexSchema {
   repeated string stop_words = 11;
   uint32 min_stem_size = 12;
   bool skip_initial_scan = 13;
+  optional string score_field = 14;
```
KarthikSubbarao, line 43: Is there any de-serialization needed?

src/index_schema.h
allenss-amazon: I believe score is multiplicative, so a default of 1.0 is correct and no reason to make it configurable.

src/index_schema.cc
allenss-amazon: We don't provide compatibility for FT.INFO.

My preference would be to always include it.

src/index_schema.h
Frank-Gu-81: >SCORE {default_score}
is default score for documents in the index. Default score is 1.0.

Default is 1.0 according to Re Search's doc. I do not think this is configuration.

src/index_schema.cc
Frank-Gu-81: The Redis `FT.INFO` docs and example output show `index_definition` only containing `key_type`,`prefixes`, and `default_score`, no `score_field_name`. The example index wasn't created with `SCORE_FIELD`, and the docs don't mention `score_field_name` at all. I will do some testing locally. 

https://redis.io/docs/latest/commands/ft.info/

src/commands/ft_create_parser.cc
Frank-Gu-81: It's always fixed to [0.0, 1.0] in Redis. Does not look like it is configurable. From the Redis `FT.CREATE` docs (https://redis.io/docs/latest/commands/ft.create/): 

>SCORE {default_score} — "is default score for documents in the index. Default score is 1.0."
SCORE_FIELD {score_attribute} — "Ranking must be between 0.0 and 1.0."

Looks like it's hardcoded validation at FT.CREATE parse time.

src/index_schema.proto
Frank-Gu-81: My understanding is that protobuf handles serialization/deserialization automatically via `ToProto()`. `optional` is added for backward-compatibility.

src/index_schema.cc
Frank-Gu-81: Tested it out locally. Redis search include `score_field` if the index is created with it. Otherwise it doesn't. I will change the naming in the next revision.

```
localhost:6379> FT.CREATE idx2 ON HASH PREFIX 1 item: SCORE_FIELD my_score SCHEMA title TEXT
OK

localhost:6379> FT.INFO idx2
 1) index_name
 2) idx2
 3) index_options
 4) (empty array)
 5) index_definition
 6) 1) key_type
    2) HASH
    3) prefixes
    4) 1) item:
    5) default_score
    6) "1"
    7) score_field
    8) my_score
```


```
localhost:6379> FT.CREATE idx3 ON HASH PREFIX 1 doc: SCHEMA title TEXT
OK

localhost:6379> ft.info idx3
 1) index_name
 2) idx3
 3) index_options
 4) (empty array)
 5) index_definition
 6) 1) key_type
    2) HASH
    3) prefixes
    4) 1) doc:
    5) default_score
    6) "1"
```

Frank-Gu-81:
In terms of the score values accepted by Redis Search. It appears that the effective range for `SCORE_FIELD` is [0, +inf). All negative scores are clamped to 0. 

```
localhost:6379> HSET st:1 title "negative score" my_score -100.5
(integer) 2
localhost:6379> HSET st:2 title "zero score" my_score 0
(integer) 2
localhost:6379> HSET st:3 title "huge score" my_score 999999
(integer) 2
localhost:6379> HSET st:4 title "tiny score" my_score 0.001
(integer) 2
localhost:6379> HSET st:5 title "negative inf" my_score -inf
(integer) 2
localhost:6379> HSET st:6 title "positive inf" my_score inf
(integer) 2
localhost:6379> FT.SEARCH score_test "*" WITHSCORES
 1) (integer) 6
 2) "st:6"
 3) "inf"
 4) 1) "title"
    2) "positive inf"
 5) "st:3"
 6) "999999"
 7) 1) "title"
    2) "huge score"
 8) "st:4"
 9) "0.0010000000474974513"
10) 1) "title"
    2) "tiny score"
11) "st:1"
12) "0"
13) 1) "title"
    2) "negative score"
14) "st:2"
15) "0"
16) 1) "title"
    2) "zero score"
17) "st:5"
18) "0"
19) 1) "title"
    2) "negative inf"
```

src/index_schema.cc
```diff
@@ -586,6 +590,26 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
...
+                         score_field_.c_str(), &score_record, NULL);
+    if (score_record) {
+      vmsdk::UniqueValkeyString score_str(score_record);
+      auto parsed = vmsdk::To<float>(vmsdk::ToStringView(score_record));
+      // Use the parsed value if succeeded; Otherwise, we fall back to the
+      // default score silently.
+      // Negative values are clamped to 0. The effective range of scores
+      // is [0, inf)
+      if (parsed.ok()) {
+        document_score = std::max(0.0f, parsed.value());
```
allenss-amazon: This is the ingestion path for every key, Let's avoid copying the data into a ValkeyString just for the purposes of validating that it's a parseable number. We can refactor the vmsdk conversions if we're worried about divergence.

testing/common.h
```diff
@@ -96,45 +96,45 @@ class MockIndex : public indexes::IndexBase {
...
-      (absl::AnyInvocable<absl::Status(const InternedStringPtr& key)> fn),
+      (absl::AnyInvocable<absl::Status(const InternedStringPtr &key)> fn),
       (const, override));
   MOCK_METHOD(
       (absl::Status), ForEachUnTrackedKey,
-      (absl::AnyInvocable<absl::Status(const InternedStringPtr& key)> fn),
+      (absl::AnyInvocable<absl::Status(const InternedStringPtr &key)> fn),
       (const, override));
   MOCK_METHOD(uint32_t, GetMutationWeight, (), (const, override));
 };
```
allenss-amazon, line 114: Looks like maybe you have a local override for text formatting. Let's fix that.

allenss-amazon (commented):
Too many formatting-only changes to review in depth.

Also, does the command/ft.create.md need updating?

testing/common.h
Frank-Gu-81: These were from `.clang-format` rules in this repo. Not sure why it wasn't caught previously. 
```
PointerAlignment: Right
```

Frank-Gu-81:
I've updated `command/ft.create.md`. The score range related sections are subject to change at this moment depending on what scoring algorithm we are going with. 

For TFIDF, the score range is [0.0, inf)
for BM25STD, the score range is (-inf, inf)

src/index_schema.cc
```diff
@@ -298,6 +300,10 @@ IndexSchema::IndexSchema(ValkeyModuleCtx *ctx,
       min_stem_size_(index_schema_proto.min_stem_size() > 0
                          ? index_schema_proto.min_stem_size()
                          : 4),
+      score_(index_schema_proto.score() > 0 ? index_schema_proto.score() : 1.0),
```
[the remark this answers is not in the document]
Frank-Gu-81: The suggested fix (`>= 0 && <= 1.0`) doesn't solve the problem though. proto3 defaults unset float fields to 0.0, so we still can't distinguish "user didn't specify SCORE" (should default to 1.0) from "user explicitly set SCORE 0.0" (should store 0.0). Both cases would produce 0.0 from the proto.

I tested Redis's behavior: it accepts `SCORE 0.0` but then omits `default_score` entirely from FT.INFO output, so it's unclear what value it actually uses internally.

But, since score field is multiplicative,`SCORE 0.0` = "zero relevance" is not a practical use case.

docs/commands/ft.create.md
```diff
@@ -47,7 +48,9 @@ FT.CREATE <index-name>
...
 
-- `SCORE` (optional): The current implementation only allows the value to be 1.0. This parameter is accepted to make valkey-search more interoperable with RediSearch. (default: 1.0)
+- `SCORE` (optional): Sets the default document score used for text search ranking. The value must be between 0.0 and 1.0. When `SCORE_FIELD` is configured, this value is used as the fallback if a document's score field is missing or cannot be parsed. (default: 1.0)
+
+- `SCORE_FIELD <field_name>` (optional): Specifies the name of a hash field whose numeric value is used as the per-document score. When configured, the value of this field is read during ingestion and used as the document's relevance score for text search ranking. If the field is missing or cannot be parsed as a valid number, the index-level `SCORE` default is used. Negative values are clamped to 0. The effective range of scores is [0, infinity).
```
boda26: Just a note that the SCORE_FIELD range is different for TFIDF and BM25STD. This depends on the default, but we can change it anytime later.

src/index_schema.cc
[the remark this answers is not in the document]
boda26: I think it should be index_schema_proto.score() >= 0 here. Setting SCORE to 0.0 is a valid case. I tested in redis and the final score is always 0 if SCORE is set to 0.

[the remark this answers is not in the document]
Frank-Gu-81: Made the proto field `optional float score = 5;` and use `has_score()` to distinguish the two cases.

docs/commands/ft.create.md
Frank-Gu-81: Yes, I had a note [here](https://github.com/valkey-io/valkey-search/pull/1008#issuecomment-4401350422). Will make sure to use the current one once we made the decision.

src/index_schema.cc
```diff
@@ -1106,10 +1135,12 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
...
   }
-  // hard-code default score of 1 as it's the only value we currently
-  // supported.
   ValkeyModule_ReplyWithSimpleString(ctx, "default_score");
-  ValkeyModule_ReplyWithCString(ctx, "1");
+  ValkeyModule_ReplyWithCString(ctx, absl::StrCat(score_).c_str());
+
+  ValkeyModule_ReplyWithSimpleString(ctx, "score_field");
+  ValkeyModule_ReplyWithSimpleString(
+      ctx, score_field_.empty() ? "" : score_field_.c_str());
```
zackcam: Nit: I think this might be redundant, as c_str will return "" if it was empty and we called c_str on it

src/index_schema.cc
```diff
@@ -586,6 +593,27 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
       added = true;
     }
   }
+
+  // Read the per-document score from SCORE_FIELD if configured
+  float document_score = score_;
+  if (key_obj && HasScoreField()) {
+    ValkeyModuleString *score_record = nullptr;
+    ValkeyModule_HashGet(key_obj.get(), VALKEYMODULE_HASH_CFIELDS,
```
zackcam: Both AI reviewers left a note for the same thing here, from my understanding of the code this isn't a dangerous issue but could help reduce work. But do we want to add the JSON path here now instead of being more selective and doing a basic fallback?

src/index_schema.cc
Frank-Gu-81: Will update this part to be:
```
ValkeyModule_ReplyWithSimpleString(ctx, score_field_.c_str());
```

src/index_schema.cc
Frank-Gu-81: That's a good point. Will fix this.

zackcam (approved):

src/index_schema.h
```diff
@@ -88,6 +88,7 @@ class IndexSchema : public KeyspaceEventSubscription,
  public:
   struct IndexKeyInfo {
     MutationSequenceNumber mutation_sequence_number_{0};
+    float document_score{1.0};
```
KarthikSubbarao: Is this the right type to use?

> float document_score{1.0};

Could you clarify if all documents must have a default score of 1.0 in the scoring formulas?

Frank-Gu-81: Default score is 1.0 according to Mile's doc. 
image

I do not see any redis doc mentioning the type used. Would you want me to switch to double instead for better precision?

src/index_schema.cc
```diff
@@ -298,6 +300,11 @@ IndexSchema::IndexSchema(ValkeyModuleCtx *ctx,
       min_stem_size_(index_schema_proto.min_stem_size() > 0
                          ? index_schema_proto.min_stem_size()
                          : 4),
+      score_(index_schema_proto.has_score() > 0 ? index_schema_proto.score()
+                                                : 1.0),
```
KarthikSubbarao: Can we have 1.0 stored in a default?

src/index_schema.cc
```diff
@@ -298,6 +300,11 @@ IndexSchema::IndexSchema(ValkeyModuleCtx *ctx,
...
+      score_field_(index_schema_proto.has_score_field()
+                       ? index_schema_proto.score_field()
+                       : ""),
```
KarthikSubbarao: Empty? Does this set it to Null or are we storing an empty string?

Using Null or an optional would be clearer when not used

src/index_schema.cc
Frank-Gu-81: Yes, I can add a constant for the default value.

```
static constexpr double kDefaultDocumentScore = 1.0;
```

src/index_schema.cc
Frank-Gu-81: It currently sets to an empty string. I will change to `std::optional<std::string>`. This makes the "not configured" state explicit rather than relying on empty string as a sentinel.

src/index_schema.h
Frank-Gu-81: Tested it out with redis docker:
```
FT.SEARCH testidx "hello" WITHSCORES SCORER BM25STD
doc:7 → "-1.8232160519019482e-4"
```

This exceeds float's precision. So I will change it to double.

src/index_schema.cc
```diff
@@ -586,6 +593,24 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
...
+  if (key_obj && HasScoreField()) {
+    auto score_record = attribute_data_type_->GetRecord(ctx, key_obj.get(),
+                                                        key_cstr, score_field_);
+    if (score_record.ok()) {
+      // Parse the value as a float. If it fails (non-numeric), fall back to
+      // the default score silently. Raw value is stored without clamping.
+      // The scoring algorithm decides how to handle values at query time.
+      float value;
+      auto str_view = vmsdk::ToStringView(score_record.value().get());
+      if (absl::SimpleAtof(str_view, &value)) {
```
KarthikSubbarao: Should we accept `inf`, `-inf` and `nan`? @boda26

Frank-Gu-81: From my testing:
- `inf`/`-inf`: Redis accepts and stores them as-is. We do the same (SimpleAtod parses them successfully).
- `nan`/`NaN`/`NAN`: Redis falls back to the default score (1.0). We match. SimpleAtod returns false for "nan" strings, so we fall back to score_ (the configured default). No explicit NaN check needed.

@boda26 please confirm if this matches what you saw as well.

boda26 (approved):

src/index_schema.h
```diff
@@ -172,6 +175,21 @@ class IndexSchema : public KeyspaceEventSubscription,
 
   inline const std::string &GetName() const { return name_; }
   inline std::uint32_t GetDBNum() const { return db_num_; }
+  inline double GetScore() const { return score_; }
+  inline const std::string &GetScoreField() const {
+    DCHECK(score_field_.has_value())
+        << "GetScoreField() called without HasScoreField() guard";
```
KarthikSubbarao: DCHECK is only in debug mode. CHECK is in release mode which is what we use on default builds

src/index_schema.cc
```diff
@@ -586,6 +593,24 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
...
+  if (key_obj && HasScoreField()) {
+    auto score_record = attribute_data_type_->GetRecord(
+        ctx, key_obj.get(), key_cstr, score_field_.value());
+    if (score_record.ok()) {
+      // Parse the value as a double. If it fails (non-numeric), fall back to
+      // the default score silently. Raw value is stored without clamping.
+      // The scoring algorithm decides how to handle values at query time.
+      double value;
+      auto str_view = vmsdk::ToStringView(score_record.value().get());
+      if (absl::SimpleAtod(str_view, &value)) {
```
KarthikSubbarao: SimpleAtod from a quick search accepts nan. let's check if we want this or not

src/commands/ft_create_parser.cc
```diff
@@ -287,18 +287,23 @@ absl::Status ParseScore(vmsdk::ArgsIterator &itr,
...
+      return absl::InvalidArgumentError(
+          absl::StrCat("`", kScoreParam, "` parameter with a value `", score,
+                       "` is not supported. The value must be between "
+                       "0.0 and 1.0"));
+    }
+    index_schema_proto.set_score(score);
```
KarthikSubbarao, line 298: What's the type? Let's ensure it matches the needed precision

src/index_schema.cc
Frank-Gu-81: My understanding was wrong. `SimpleAtod` actually returns `true` for `nan` and stores it as `nan`.

src/index_schema.cc
Frank-Gu-81: added `value == value` to handle this case. See [IEEE 754 standard for floating-point arithmetic](https://en.wikipedia.org/wiki/NaN)

src/commands/ft_create_parser.cc
Frank-Gu-81: Changed to double as it matches redis's behavior. 

```
localhost:6380> FT.CREATE scoreprec3 ON HASH PREFIX 1 b: SCORE 0.1234567891234567 SCHEMA title TEXT
OK
localhost:6380> FT.INFO scoreprec3
 1) index_name
 2) scoreprec3
 3) index_options
 4) (empty array)
 5) index_definition
 6) 1) key_type
    2) HASH
    3) prefixes
    4) 1) b:
    5) default_score
    6) "0.1234567891234567"
```

Frank-Gu-81:
Discussed offline and we will be storing per-index and per-document score in `float`.

src/commands/ft_create_parser.cc
```diff
@@ -284,21 +284,26 @@ absl::Status ParseLanguage(vmsdk::ArgsIterator &itr,
 }
 absl::Status ParseScore(vmsdk::ArgsIterator &itr,
                         data_model::IndexSchema &index_schema_proto) {
-  float score{1.0};
+  float score{1.0f};
```
KarthikSubbarao: Nit: Use a constant for 1.0 instead of hard coding

src/index_schema.cc
```diff
@@ -586,6 +593,24 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
...
+  if (key_obj && HasScoreField()) {
+    auto score_record = attribute_data_type_->GetRecord(
+        ctx, key_obj.get(), key_cstr, score_field_.value());
+    if (score_record.ok()) {
+      // Parse the value as a float. If it fails (non-numeric), fall back to
+      // the default score silently. Raw value is stored without clamping.
+      // The scoring algorithm decides how to handle values at query time.
+      float value;
+      auto str_view = vmsdk::ToStringView(score_record.value().get());
+      if (absl::SimpleAtof(str_view, &value) && value == value) {  // NaN check
```
KarthikSubbarao, line 608: Nit: Here is an alternative:
```
      float value = 0.0f;
      const auto str_view = vmsdk::ToStringView(score_record.value().get());
      // Fall back to default score silently if parsing fails or returns NaN
      if (absl::SimpleAtof(str_view, &value) && !std::isnan(value)) {
```

src/index_schema.cc
```diff
@@ -1106,10 +1132,13 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
   for (const auto &prefix : subscribed_key_prefixes_) {
     ValkeyModule_ReplyWithSimpleString(ctx, prefix.c_str());
   }
-  // hard-code default score of 1 as it's the only value we currently
-  // supported.
   ValkeyModule_ReplyWithSimpleString(ctx, "default_score");
-  ValkeyModule_ReplyWithCString(ctx, "1");
+  ValkeyModule_ReplyWithCString(
+      ctx, absl::StrFormat("%.17g", static_cast<double>(score_)).c_str());
```
KarthikSubbarao: Question: Do we want `ValkeyModule_ReplyWithCString` or `ValkeyModule_ReplyWithDouble`?

src/index_schema.cc
Frank-Gu-81: `std::isnan` would hit this warning, which was why I switched to using `value == value` check:
```
Use of NaN is undefined behavior due to the currently enabled floating-point optionsclang(-Wnan-infinity-disabled)
```

src/index_schema.cc
Frank-Gu-81: Tested it out for both `ValkeyModule_ReplyWithCString` and `ValkeyModule_ReplyWithDouble`. They produce exactly the same result.

```
+----------------------------+--------------------------------+--------------------------+
| Input                      | ReplyWithCString (%.17g)       | ReplyWithDouble          |
+----------------------------+--------------------------------+--------------------------+
| SCORE 0.123456789          | "0.12345679104328156"          | "0.12345679104328156"    |
| SCORE 0.5                  | "0.5"                          | "0.5"                    |
| (no SCORE, default 1.0)    | "1"                            | "1"                      |
+----------------------------+--------------------------------+--------------------------+
```

src/index_schema.cc
allenss-amazon: We tend to compile this code with -ffast-floating which doesn't handle NaN correctly in all cases. There is a function in value.cc which will correctly detect a Nan regardless of how the code is compiled.

KarthikSubbarao (approved):

src/index_schema.cc
Frank-Gu-81: The function is currently static though :(

```
// Built-in isnan doesn't work when compiling with fast-math, which is what we
// want to do.
static bool IsNan(const double& d) {
  uint64_t v = *(uint64_t*)&d;
  return ((v & kExponentMask) == kExponentMask) && ((v & kMantissaMask) != 0);
}
```

PR #1009, opened 2026-05-03
Fix compatibility test suite

allenss-amazon:
The compatibility test suite has always relied on manual procedures to generate the answer files (aka pickle files). This has proven inadequate as the current test suite will actually fail if you regenerate the answer files. meaning that test cases were committed to the test suite that weren't actually being tested.

This PR does three things.

1. Centralizes the definition of generate modules. This simplifies the addition of new compatibility test suites.
2. Creates a mechanism to ensure that any change to a compatibility test suite will force a regeneration of the answer files.
3. Temporarily comment out failing test cases:

Separate PRs will need to be created to fix the newly commented out test cases.


integration/compatibility/generate.py
```diff
@@ -164,11 +169,10 @@ def test_search_reverse(self, key_type, dialect):
         self.setup_data("reverse vector numbers", key_type)
         self.checkall(dialect, f"ft.search {key_type}_idx1 *")
         self.checkall(dialect, f"ft.search {key_type}_idx1 * limit 0 5")
-
     def test_search(self, key_type, dialect):
         self.setup_data("sortable numbers", key_type)
         self.checkall(dialect, f"ft.search {key_type}_idx1 *")
-        self.checkall(dialect, f"ft.search {key_type}_idx1 * limit 0 5")
+    '''
```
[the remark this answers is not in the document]
allenss-amazon: This is intentional.

integration/compatibility/generate_text.py
```diff
@@ -422,6 +422,7 @@ def test_text_search_group_depth2_inorder_slop(self, key_type, dialect, schema_t
     def test_text_search_group_depth3_inorder_slop(self, key_type, dialect, schema_type):
         self._run_test(gen_depth3, "pure text", key_type, dialect, schema_type, inorder=True, slop=True, check_parsing=True)
 
+    '''
```
KarthikSubbarao: Can we use `pytest.mark.skip` to skip since this gets reported and is less likely to get forgotten ?

Same comment for other skipped tests

integration/compatibility/generate.py
```diff
@@ -466,5 +470,5 @@ def test_search_sortby(self, key_type, dialect):
                 for return_keys in ["", "RETURN 2 @n1 @t1"]:
                     for wsk in ["", "WITHSORTKEYS"]:
                         for limit in ["LIMIT 0 5", "LIMIT 2 3", ""]:
-                            self.check(dialect, f"ft.search {key_type}_idx1 * SORTBY {sort_key} {direction} {return_keys} {limit} {wsk}")
-
+                            self.check(dialect, f"ft.search {key_type}_idx1 * sortby {sort_key} {direction} {return_keys} {limit} {wsk}")
```
KarthikSubbarao: Is the `SORTBY` -> `sortby` change needed?

integration/compatibility/regenerate.sh
```diff
@@ -0,0 +1,64 @@
...
+#
+# Requires Docker: the generators spin up redis/redis-stack-server on port 6380
+# to capture reference answers.
+#
+# Usage:
+#   ./integration/compatibility/regenerate.sh [extra pytest args...]
+#
+# After it finishes, git add and commit the updated *.pickle.gz files.
+
+ROOT_DIR=$(readlink -f "$(dirname "$0")/../..")
```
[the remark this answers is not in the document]
KarthikSubbarao: Should we follow this suggestion?

```
-ROOT_DIR=$(readlink -f "$(dirname "$0")/../..")
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+ROOT_DIR="$(cd "${SCRIPT_DIR}/../.." && pwd)"
```

integration/compatibility_test.py
```diff
@@ -5,7 +5,11 @@
 from itertools import chain, combinations
 import pickle
 import compatibility
-from compatibility.data_sets import * 
+from compatibility import GENERATORS, compute_sources_hash
+from compatibility.data_sets import *
```
KarthikSubbarao, line 9: Can be simplified to:
```
from compatibility import GENERATORS, compute_sources_hash
from compatibility.data_sets import *
```

integration/compatibility/__init__.py
```diff
@@ -1,3 +1,32 @@
...
+
+
+def compute_sources_hash():
+    """SHA256 of every .py file in this directory.
+
+    Stored inside the generated pickle answer files so compatibility_test.py
+    can detect when a pickle is stale relative to the generators and helpers.
+    """
+    h = hashlib.sha256()
+    for fname in sorted(os.listdir(_COMPAT_DIR)):
```
KarthikSubbarao, line 24: Is it sufficient to check the compat directory? What if a helper / utility test function changes?

integration/compatibility/generate.py
KarthikSubbarao, line 48: One concern I have is this is not pinned to a specific version. Through patches, it might change behavior


The bigger question I have is which version we want to perform compat testing against? 

Redis Stack has version from 6.2 (2022) to 7.2 (2024). After this, we have Redis Open Source which has 7.4 (2024) to 8.6 (current 2026)

integration/compatibility/generate_text.py
allenss-amazon: ack.

integration/compatibility/__init__.py
allenss-amazon: I think this is enough to catch the vast vast majority of failure modes.

integration/compatibility/generate.py
allenss-amazon: Agreed. I will open an issue for the community to decide what version to standardize on. Once that's done we can have a followup CR to nail it here.

allenss-amazon: Done.

integration/compatibility/generate_text.py
```diff
@@ -399,15 +399,15 @@ def test_text_search_group_depth2(self, key_type, dialect, schema_type):
     def test_text_search_group_depth3(self, key_type, dialect, schema_type):
         """Test grouped queries with depth 3."""
         self._run_test(gen_depth3, "pure text", key_type, dialect, schema_type)
-    
+    '''
```
KarthikSubbarao: Can we use `@pytest.mark.skip(` here?

KarthikSubbarao (approved):

integration/compatibility/generate_text.py
allenss-amazon: done.

Issue #993, opened 2026-04-28
[BUG] Integration Tests Broken on MacOS


rileydes-improving:
**Describe the bug**
A recent commit broke the integration tests on MacOS, specifically in
https://github.com/valkey-io/valkey-search/commit/63574f75e6279c2a77b6c2f6729843aad8c74a88#diff-18d4fa406fcd85c52dc88c0c3ccb13f7cc7705d95a31f34392327d9b451a59c4R164

On macOS, script takes the form script [-flags] [file [command ...]], the command goes at the end as positional args, not after -c.

Therefore this line breaks on MacOs
```
script -q -e -c "${PYTHON_PATH} -m pytest ${FILTER_ARGS} ${CAPTURE_ARG} --cache-clear -v ${ROOT_DIR}/integration/" ${PYTEST_OUTPUT_LOG}
```

**To Reproduce**
On a Mac/Darwin system

`./build.sh --run-integration-tests`

**Expected behavior**
The integration tests to run.

**Possible Fix/Temporary Fix for Local Development**
Replace the line mentioned above.
```
if [[ "$(uname)" == "Darwin" ]]; then
    ${PYTHON_PATH} -m pytest ${FILTER_ARGS} ${CAPTURE_ARG} --color=yes --cache-clear -v ${ROOT_DIR}/integration/ 2>&1 | tee ${PYTEST_OUTPUT_LOG}
    RUN_SUCCESS=${PIPESTATUS[0]}
  else
    script -q -e -c "${PYTHON_PATH} -m pytest ${FILTER_ARGS} ${CAPTURE_ARG} --cache-clear -v ${ROOT_DIR}/integration/" ${PYTEST_OUTPUT_LOG}
    RUN_SUCCESS=$?
  fi
```

PR #1014, opened 2026-05-04
fix: Handle macOS-incompatible script command in integration test runner

smkher:
On macOS, the script command uses different syntax than Linux. Linux uses 'script -q -e -c "command" outfile' while macOS expects the command as positional args with no -c or -e flags.

Add a uname check in run_pytest() to use tee-based output capture on Darwin.

For #993


smkher:
ASAN and UT failures seem unrelated

allenss-amazon (approved):

PR #1015, opened 2026-05-05
Replace deque with flat hash map for tracked_vectors

### LINKED ISSUE #500: [BUG] alllow-replace-deleted does not work as expected
## Summary

This issue describes two separate problems in the HNSW vector index implementation:

1. **Issue 1: `allow-replace-deleted` is not working as intended** - The vector tracking data structure (`std::deque`) cannot support the slot reuse mechanism required by this feature.

2. **Issue 2: `reclaimable_memory` is incorrect** - The current metric counts each HNSW node deletion independently, without accounting for string interning deduplication. This leads to over-counting when multiple keys share identical vector data.

These are distinct issues with different root causes and solutions, though both relate to vector lifecycle management.

[A more accurate formulation of the states of a vector - [Formal Model: Interned String Memory States and Transitions](#here)]

# Issue 1: 

## Background

The `allow-replace-deleted` feature allows HNSW to reuse slots from deleted vectors for new insertions, reducing memory fragmentation. Currently, this feature is explicitly disabled:

```cpp
// src/indexes/vector_hnsw.cc lines 97-98
// Notes:
// 1. Not allowing replace delete is aligned with RediSearch
// 2. Consider making allow_replace_deleted_ configurable
index->algo_->allow_replace_deleted_ = false;
```

### Current Implementation

```cpp
// src/indexes/vector_hnsw.h lines 113-114
std::deque<InternedStringPtr> tracked_vectors_
    ABSL_GUARDED_BY(tracked_vectors_mutex_);
```

```cpp
// src/indexes/vector_hnsw.cc lines 108-109
template <typename T>
void VectorHNSW<T>::TrackVector(uint64_t internal_id,
                                const InternedStringPtr &vector) {
  absl::MutexLock lock(&tracked_vectors_mutex_);
  tracked_vectors_.push_back(vector);  // Just append, loses internal_id!
}
```

### Why This Breaks `allow-replace-deleted`

The `std::deque` is append-only with no association between `internal_id` and the stored vector. When HNSW reuses a slot:

1. HNSW calls `TrackVector(internal_id, new_vector)` for the new vector
2. The new vector is appended to the deque
3. **The old vector at that `internal_id` is never removed**
4. The deque grows unboundedly, holding `shared_ptr`s to all vectors ever tracked

```
Example:
1. HSET doc:1 vec [0.1, 0.2, 0.3, 0.4]  → slot 5
   tracked_vectors_ = [vector_A]

2. DEL doc:1  → slot 5 marked deleted
   tracked_vectors_ = [vector_A]  // Still there

3. HSET doc:2 vec [0.5, 0.6, 0.7, 0.8]  → reuses slot 5 (if allow-replace-deleted=true)
   TrackVector(5, vector_B) called
   tracked_vectors_ = [vector_A, vector_B]  // BOTH in deque!

Result: vector_A is never freed - memory leak
After 1000 cycles: deque has 1000 entries, all held in memory
```

## [Suggested]Solution: Replace Deque with Map

The fix is straightforward: use a map that associates `internal_id` with the vector.

```cpp
// Proposed change
absl::flat_hash_map<uint64_t, InternedStringPtr> tracked_vectors_;

void TrackVector(uint64_t internal_id, const InternedStringPtr &vector) {
  absl::MutexLock lock(&tracked_vectors_mutex_);
  // Insert or replace - if slot was previously used, old shared_ptr is released
  tracked_vectors_[internal_id] = vector;
}
```

### Why This Works

**Key insight:** We do NOT need to remove vectors from the map when keys are deleted. We only need to replace them when slots are reused.

```
1. HSET doc:1 vec [0.1, 0.2, 0.3, 0.4]  → slot 5
   tracked_vectors_ = {5: vector_A}

2. DEL doc:1  → slot 5 marked deleted
   tracked_vectors_ = {5: vector_A}  // Keep it! HNSW still has raw pointer

3. HSET doc:2 vec [0.5, 0.6, 0.7, 0.8]  → reuses slot 5
   tracked_vectors_[5] = vector_B  // Replace!
   - Old shared_ptr to vector_A is released
   - If no other references, vector_A is freed
   - This happens exactly when HNSW overwrites the slot - safe!

Result: Memory properly managed, no leaks
```

### String Deduplication is Handled Automatically

The map naturally handles deduplication because each slot independently holds its own `shared_ptr`. A shared vector is only freed when ALL slots referencing it have been reused.

### `UnTrackVector()` Can Remain Empty

The current empty `UnTrackVector()` is actually fine with the map approach:

```cpp
// src/indexes/vector_hnsw.cc line 131
template <typename T>
void VectorHNSW<T>::UnTrackVector(uint64_t internal_id) {}  // Empty is OK!
```

When a key is deleted:
- HNSW marks the slot as deleted
- The vector stays in `tracked_vectors_` (keeping it alive)
- HNSW's raw pointer to the vector data remains valid
- When the slot is eventually reused, `TrackVector()` replaces the entry

### No Changes Needed to HNSW or VectorTracker Interface

The fix is entirely within `VectorHNSW`:
- Change `std::deque` to `flat_hash_map`
- `TrackVector()` does insert-or-replace
- No new callbacks needed
- No changes to `third_party/hnswlib/`

## Summary of Required Changes for Issue 1

| Component | Current | Required Change |
|-----------|---------|-----------------|
| `tracked_vectors_` | `std::deque<InternedStringPtr>` | `flat_hash_map<uint64_t, InternedStringPtr>` |
| `TrackVector()` | `push_back(vector)` | `tracked_vectors_[internal_id] = vector` |
| `UnTrackVector()` | Empty | No change needed |
| `allow_replace_deleted_` | `false` | Can be set to `true` after fix |

---

# Issue 2: Incorrect Metering of Marked-Deleted Vectors

## Description

The `reclaimable_memory` metric tracks memory that could be freed if all indexes were rebuilt. 

The current metering adds `vector_size_` for each HNSW node marked deleted. But with string interning, multiple nodes in the same (less likely - but possible) or different indexes can share the same vector data. The reclaimable memory should only consider vectors with `MarkedDeleted` occurrences that don't have `Active` occurrences. 

In addition, the `StringInternStore` uses a single global pool for all interned strings. This means if a `KEY` name (or `TAG` or any other interned string) happens to equal the serialized vector data, they will share the **same** `InternedString` object. This creates a problem for metering: when the object is destroyed, how do we know if it was ever used as a vector (and thus should update `reclaimable_memory`)

## Solution: TBD

---

## Additional per-index metric
In the context of MarkedDeleted phenomena in HNSW, there are 2 useful, actionable metrics. 
1. **Global**: How much memory is "locked" in marked deleted and can be reused by new writes or by rebuilding the index. 
2. **Per HNSW Index**: How many MarkedDeleted nodes are there in the index, this will impact query performance and user could rebuild the index or tune the ef_runtime

`reclaimable_memory` - once fixed, is the implementation of issue 1
we should consider adding the per-index info.

-----
## Formal Model: Interned String Memory States and Transitions

This section provides a rigorous mathematical formulation of interned string memory lifecycle to precisely define the problems and solutions.

### Important: String Interning is Shared Across All Types

The `StringInternStore` uses a **single global pool** for all interned strings. This means the same underlying string data `S` can be referenced by:
- HNSW vector indexes (as vector embeddings)
- FLAT vector indexes (as vector embeddings)
- Key names (the Redis/Valkey key itself)
- Tag field values
- Text field values

When any of these happen to have identical string content, they share the **same** `InternedString` object in memory.

### Definitions

**String Reference Tuple:**
```
R_ij := (Ctx_i, K_j, S, slot_ij)
```
Where:
- `Ctx_i` = Context i - can be:
  - An HNSW vector index (has slot, has MarkedDeleted state)
  - A FLAT vector index (no slot, no MarkedDeleted state)
  - A KEY reference (synthetic - the key name itself)
  - A TAG field reference (synthetic)
  - A TEXT field reference (synthetic)
- `K_j` = Key j (a Redis/Valkey key, e.g., "doc:123")
- `S` = The actual string data (the interned string content)
- `slot_ij` = The HNSW node position (only meaningful for HNSW contexts, otherwise ∅)

**String Reference Set:**
```
R_set(S) := { R_ij | Ctx_i references string S via key K_j }
```

This is the set of all references to the same underlying string data `S`.

**Context Type:**
```
Type(Ctx_i) ∈ { HNSW_VECTOR, FLAT_VECTOR, KEY, TAG, TEXT }
```

**Simplification for Non-HNSW Contexts:**

For KEY, TAG, TEXT, and FLAT_VECTOR contexts:
- There is no "slot" concept
- There is no "MarkedDeleted" state - deletion immediately removes the reference
- We can model these as having a synthetic unique context per reference

This simplifies the model: only HNSW_VECTOR contexts have the Active/MarkedDeleted distinction.

### State Function

Each element in R_set has a state (only meaningful for HNSW contexts):
```
State: R_ij → { Active, MarkedDeleted, N/A }
```

**State by Context Type:**
| Context Type | Possible States | Notes |
|--------------|-----------------|-------|
| HNSW_VECTOR | Active, MarkedDeleted | Full lifecycle |
| FLAT_VECTOR | N/A (always Active until removed) | Deletion removes immediately |
| KEY | N/A (always Active until removed) | Deletion removes immediately |
| TAG | N/A (always Active until removed) | Deletion removes immediately |
| TEXT | N/A (always Active until removed) | Deletion removes immediately |

**Initial State:**
```
When R_ij is created:
  if Type(Ctx_i) = HNSW_VECTOR: State(R_ij) = Active
  else: State(R_ij) = N/A (implicitly Active)
```

### State Transitions

| Transition | Trigger | Action |
|------------|---------|--------|
| **T1: Key Deletion** | `DEL K_j` | For HNSW_VECTOR: `State(R_ij) := MarkedDeleted` for all i<br>For others: `R_set(S) := R_set(S) - {R_ij}` |
| **T2: Index Deletion** | `FT.DROPINDEX Ctx_i` | `R_set(S) := R_set(S) - {R_ij}` for all j |
| **T3: Slot Reuse** | HNSW reuses slot | `R_set(S) := R_set(S) - {R_ij}` (requires HNSW_VECTOR and MarkedDeleted) |
| **T4: String Addition** | `HSET K_j field S` | `R_set(S) := R_set(S) ∪ {R_ij}`, `State(R_ij) := Active` (if HNSW) |

### Memory Reclamation Rules

```
Memory(S) can be reclaimed ⟺ R_set(S) = ∅
```

```
HNSW slot can be reused ⟺ State(R_ij) = MarkedDeleted ∧ allow-replace-deleted = yes
```

### Reclaimable Memory Formulas

**Total String Memory (for a given type, e.g., vectors):**
```
TotalMemory = Σ_S [ len(S) × 𝟙(R_set(S) ≠ ∅) ]
```

**Reclaimable Memory (correct definition for vectors):**

A string is "reclaimable" only when ALL HNSW references to it are MarkedDeleted AND there are no non-HNSW references keeping it active:

```
ReclaimableMemory = Σ_S [ len(S) × 𝟙(R_set(S) ≠ ∅) × 𝟙(∀ R_ij ∈ R_set(S) : 
                          Type(Ctx_i) = HNSW_VECTOR ∧ State(R_ij) = MarkedDeleted) ]
```

**Important:** If a string S is used as both a vector AND a key/tag/text, it can only be "reclaimable" if:
1. All HNSW references are MarkedDeleted
2. All non-HNSW references have been removed (key deleted)

In practice, this cross-type sharing is rare but must be handled correctly.

### State Transition Diagram

```
                    ┌────────────────────────────────────────────────────────────────┐
                    │            R_set(S) - All references to string S               │
                    │                                                                │
                    │  ┌──────────────────────────────────────────────────────────┐  │
                    │  │  HNSW_VECTOR References (have slot, have state)          │  │
                    │  │                                                          │  │
    HSET K_j ───────┼──┼──► R_ij [Active] ◄────────────────────────────────────┐  │  │
    (HNSW index)    │  │         │                                             │  │  │
                    │  │         │ DEL K_j                                     │  │  │
                    │  │         ▼                                             │  │  │
                    │  │    R_ij [MarkedDeleted]                               │  │  │
                    │  │         │                                             │  │  │
                    │  │         │ Slot Reuse OR FT.DROPINDEX                  │  │  │
                    │  │         ▼                                             │  │  │
                    │  │    R_ij removed from R_set ───────────────────────────┘  │  │
                    │  └──────────────────────────────────────────────────────────┘  │
                    │                                                                │
                    │  ┌──────────────────────────────────────────────────────────┐  │
                    │  │  Non-HNSW References (KEY, TAG, TEXT, FLAT_VECTOR)       │  │
                    │  │  No MarkedDeleted state - deletion removes immediately.  │  │
                    │  │                                                          │  │
                    │  │  HSET K_j ──► R_ij [Active]                              │  │
                    │  │                    │                                     │  │
                    │  │                    │ DEL K_j                             │  │
                    │  │                    ▼                                     │  │
                    │  │              R_ij removed from R_set                     │  │
                    │  └──────────────────────────────────────────────────────────┘  │
                    │                                                                │
                    │             When R_set(S) = ∅ ──► Memory(S) freed              │
                    └────────────────────────────────────────────────────────────────┘

Note: The R_set of a vector represent it's actual state, it can have arbitrary number of MarkDeleted 
          occurrences and Active occurrences at the same time.
```

### How This Model Exposes the Bugs

**Issue 1 (Deque Bug):**
- Transition T3 (Slot Reuse) requires removing R_ij from R_set
- The `std::deque` cannot perform this removal (no ID association)
- Result: R_set never shrinks for HNSW vectors, Memory(S) never freed

**Issue 2 (Metering Bug):**
- Current code: `reclaimable_memory += vector_size_` on each `markDelete()`
- This counts per-HNSW-node, not per-string
- With deduplication, multiple R_ij share the same S
- Result: Over-counting when `|R_set(S)| > 1`

**Cross-Type Consideration:**
- If string S is used as both a vector AND a key/tag, the metering must account for this
- A string is only truly "reclaimable" when ALL references (of all types) allow it
- Current code doesn't distinguish string types, which could cause metering errors

---

Aksha1812:
## Fix unbounded memory growth in VectorHNSW tracked_vectors when allow replace deleted enabled

Replace `std::deque<InternedStringPtr>` with `absl::flat_hash_map<uint64_t, InternedStringPtr>` for `VectorHNSW::tracked_vectors_`, mirroring the approach already used by `VectorFlat`. The deque blindly appended on every `TrackVector` call and ignored the `internal_id`, so when slots were reused (via `allow_replace_deleted` or vector modifications) the old `InternedStringPtr` was never released causing memory to grow without bound. With a flat_hash_map keyed by `internal_id`, `TrackVector` naturally replaces the old entry on slot reuse, freeing the stale shared_ptr. `UnTrackVector` remains a no-op since HNSW deleted nodes still need their vector data alive for graph routing.

Based on design https://github.com/valkey-io/valkey-search/issues/500 part 1.  
Added and deleted integration test which proves above theory.

Pending attach proof of that integration test run + micro benchmark + perf comparison when using these two data structures.

Resolves #500


allenss-amazon (approved):

Aksha1812:
undrafted PR , cleaned comment, fixed DCO . Can you merge @allenss-amazon

src/indexes/vector_hnsw.cc
```diff
@@ -109,28 +108,25 @@ template <typename T>
...
-    char *data_ptrv = algo_->getDataByInternalId(*id);
-    size_t dim = *((size_t *)algo_->dist_func_param_);
-    absl::string_view record(data_ptrv, dim * sizeof(T));
-    return vector->Str() == record;
+  absl::MutexLock lock(&tracked_vectors_mutex_);
+  auto it = tracked_vectors_.find(internal_id);
+  if (it == tracked_vectors_.end()) {
+    return false;
   }
+  return it->second->Str() == vector->Str();
```
allenss-amazon: Shouldn't need to compare ->Str(), comparing the InternedStringPtr directly is sufficient -- they are unique by definition.

src/indexes/vector_hnsw.cc
```diff
@@ -109,28 +108,25 @@ template <typename T>
...
 }
-// UnTrackVector does not delete the vector in VectorHNSW, as vectors are never
-// physically removed from the graph—only marked as deleted.
+
+// UnTrackVector does not erase the entry in VectorHNSW because vectors are
+// never physically removed from the HNSW graph—only marked as deleted. When
+// a slot is later reused via allow_replace_deleted, TrackVector naturally
+// replaces the old entry via operator[] on the flat_hash_map, releasing the
+// old shared_ptr.
```
allenss-amazon: This implies that the internal id is somehow related to the slot number. That isn't true is it? Aren't internal_ids just an incrementing variable?

src/indexes/vector_hnsw.cc
Aksha1812: Fixed this

src/indexes/vector_hnsw.cc
Aksha1812: updated the comment

Aksha1812:
> Vector data must be tracked even after deletion because HNSW only marks nodes as deleted without reclaiming space. Physical node removal is not currently supported, as it requires complex re-linking to ensure the graph remains intact and navigable for routing. The flat algorithm doesn't suffer from this issue and therefor node removal is supported.

Yep, exactly. The no-op is intentional to keep vector data alive for graph routing. This PR just fixes the modification path — same key modified N times no longer leaks N entries.   @yairgott Could you review /approve the PR as well . More details here https://github.com/valkey-io/valkey-search/issues/500. this PR is based on part 1 of this design in this issue. Thank you

src/indexes/vector_hnsw.cc
```diff
@@ -109,28 +108,24 @@ template <typename T>
 void VectorHNSW<T>::TrackVector(uint64_t internal_id,
                                 const InternedStringPtr &vector) {
   absl::MutexLock lock(&tracked_vectors_mutex_);
-  tracked_vectors_.push_back(vector);
+  tracked_vectors_[internal_id] = vector;
```
yairgott, line 111: On an update of a vector value, doing so would lead to the release of the old vector which would lead to having a dangling pointer in HNSW. Keep in mind that `UpdateMetadata`is called prior to updating the HNSW algo. See https://github.com/valkey-io/valkey-search/blob/147b0b2bf99c110e7fd97351d5c9b0d6cd3dda17/src/indexes/vector_base.cc#L233
Maybe it can be addressed by just changing the order between the calls to `UpdateMetadata` and `ModifyRecordImpl`?

Aksha1812: Good catch. moved TrackVector to after ModifyRecordImpl so the old InternedStringPtr stays alive until HNSW has replaced its data pointer. Lmk if that is accurate.

yairgott: Makes sense. I wonder if the call to `UnTrackKey` from `ModifyRecord` is still required.

yairgott:
I've been working on the [task](https://github.com/valkey-io/valkey-search/pull/675) to enable memory sharing for vectors which touches similar place as this fix. My PR is pretty big and It would be helpful for me if this is landed after my PR. Can this work?

Aksha1812:
> I've been working on the https://github.com/valkey-io/valkey-search/pull/675 to enable memory sharing for vectors which touches similar place as this fix. My PR is pretty big and It would be helpful for me if this is landed after my PR. Can this work?

Sure once your task is finished will update this PR. will keep this PR on hold till then . I also have a draft PR open dependent on this PR https://github.com/valkey-io/valkey-search/pull/1034 which fixes some reclaimable memory overcounting as mentioned in part 2 here https://github.com/valkey-io/valkey-search/issues/500 . Once your change is done . please help review both these PRs . thanks. 

In the meantime could you help review https://github.com/valkey-io/valkey-search/pull/1079 which was discussed here https://github.com/valkey-io/valkey-search/issues/1070 . Thanks

src/indexes/vector_hnsw.cc
allenss-amazon: I wouldn't worry about it. I'm working on a PR that collapses all of the tracked/untracked keys stuff into a common mechanism in IndexBase.

Aksha1812: Ack. Resolving conversation , lmk if above still an issue.

Aksha1812:
Sorry for raising review. Converting to draft . As per discussion above . PR to be merged after memory sharing for vectors #675

Aksha1812:
@allenss-amazon can you give this a final check and merge as we discussed

Aksha1812:
@allenss-amazon reminder to merge .

Aksha1812:
@allenss-amazon reminder to merge .

allenss-amazon (approved):

PR #1026, opened 2026-05-10
Implement space savings for Tag and Numeric

allenss-amazon:
The class BagOfInternedStringPtrs is created as a replacement for flat_hash_set\<InternedStringPtr\>. The new class is optimized for small numbers of keys. Numeric and Tag indexes were updated to use the new class.

Implementation has four representations:

```
 The bag now has four heap layouts discriminated by the bottom 2 bits of storage_:

  ┌─────┬───────────────────┬──────────┬────────────────────────────────────────────────────────────────────────┐
  │ Tag │       Mode        │ Capacity │                           Heap (per bucket)                            │
  ├─────┼───────────────────┼──────────┼────────────────────────────────────────────────────────────────────────┤
  │  00 │ Single (or Empty) │        1 │                                                                0 bytes │
  ├─────┼───────────────────┼──────────┼────────────────────────────────────────────────────────────────────────┤
  │  01 │ Array4            │        4 │ std::array<InternedStringPtr, 4> (32 B + ~16 B malloc overhead ≈ 48 B) │
  ├─────┼───────────────────┼──────────┼────────────────────────────────────────────────────────────────────────┤
  │  10 │ Array8            │        8 │                 std::array<InternedStringPtr, 8> (64 B + ~16 B ≈ 80 B) │
  ├─────┼───────────────────┼──────────┼────────────────────────────────────────────────────────────────────────┤
  │  11 │ Set               │        ∞ │     absl::flat_hash_set<InternedStringPtr> (~176 B for ≤ ~12 elements) │
  └─────┴───────────────────┴──────────┴────────────────────────────────────────────────────────────────────────┘

Space savings:


  Numeric Index

  ┌────────────┬──────────────────┬─────────────┬─────────┬─────────┬───────────────┐
  │ keys/value │ Baseline (B/key) │ Bag (B/key) │ Δ B/key │   Δ %   │ Bag mode used │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          1 │           174.63 │      146.83 │  −27.81 │ −15.92% │ Single        │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          2 │           146.78 │      124.80 │  −21.99 │ −14.98% │ Array4        │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          3 │           107.05 │       92.32 │  −14.73 │ −13.76% │ Array4        │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          4 │            95.13 │       76.07 │  −19.06 │ −20.04% │ Array4        │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          5 │            81.50 │       72.71 │   −8.79 │ −10.79% │ Array8        │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          6 │            72.53 │       65.16 │   −7.37 │ −10.16% │ Array8        │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          7 │            66.12 │       59.76 │   −6.36 │  −9.62% │ Array8        │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          8 │            71.20 │       55.75 │  −15.45 │ −21.71% │ Array8        │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │          9 │            66.35 │       68.59 │   +2.23 │  +3.36% │ Set           │
  ├────────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │         10 │            62.44 │       64.48 │   +2.04 │  +3.27% │ Set           │
  └────────────┴──────────────────┴─────────────┴─────────┴─────────┴───────────────┘

  Tag Index

  ┌──────────┬──────────────────┬─────────────┬─────────┬─────────┬───────────────┐
  │ keys/tag │ Baseline (B/key) │ Bag (B/key) │ Δ B/key │   Δ %   │ Bag mode used │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        1 │           223.12 │      191.13 │  −31.99 │ −14.34% │ Single        │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        2 │           208.55 │      184.59 │  −23.96 │ −11.49% │ Array4        │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        3 │           165.60 │      149.64 │  −15.96 │  −9.64% │ Array4        │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        4 │           152.19 │      132.19 │  −20.00 │ −13.14% │ Array4        │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        5 │           137.71 │      128.07 │   −9.63 │  −7.00% │ Array8        │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        6 │           128.02 │      119.98 │   −8.05 │  −6.29% │ Array8        │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        7 │           121.18 │      114.27 │   −6.91 │  −5.70% │ Array8        │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        8 │           125.91 │      109.95 │  −15.96 │ −12.67% │ Array8        │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │        9 │           120.78 │      122.55 │   +1.77 │  +1.46% │ Set           │
  ├──────────┼──────────────────┼─────────────┼─────────┼─────────┼───────────────┤
  │       10 │           116.68 │      118.27 │   +1.60 │  +1.36% │ Set           │
  └──────────┴──────────────────┴─────────────┴─────────┴─────────┴───────────────┘
```


src/utils/string_interning.h
```diff
@@ -171,13 +175,657 @@ inline std::ostream &operator<<(std::ostream &os,
...
+  InternedStringSet *GetSet() const {
+    return reinterpret_cast<InternedStringSet *>(storage_ & ~kTagMask);
+  }
+  void SetArray4(Array4 *p) {
+    storage_ = reinterpret_cast<uintptr_t>(p) | kArray4Tag;
+  }
+  void SetArray8(Array8 *p) {
+    storage_ = reinterpret_cast<uintptr_t>(p) | kArray8Tag;
+  }
+  void SetSet(InternedStringSet *p) {
```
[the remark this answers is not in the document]
allenss-amazon: Iterators aren't valid after erase.

src/utils/string_interning.h
```diff
@@ -171,13 +175,658 @@ inline std::ostream &operator<<(std::ostream &os,
...
+  BagOfInternedStringPtrs(BagOfInternedStringPtrs &&other) noexcept
+      : storage_(other.storage_) {
+    other.storage_ = 0;
+  }
+  BagOfInternedStringPtrs &operator=(BagOfInternedStringPtrs &&other) noexcept {
+    if (this != &other) {
+      clear();
+      storage_ = other.storage_;
+      other.storage_ = 0;
+    }
```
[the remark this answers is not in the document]
allenss-amazon: The raw move is safe, because the source is nuked to 0 without affecting the refcount.

src/utils/string_interning.h
```diff
@@ -171,13 +175,658 @@ inline std::ostream &operator<<(std::ostream &os,
...
+  // Demote Set -> Array8 when set size drops to 8.
+  void DemoteSetIfPossible() {
+    auto *set = GetSet();
+    if (set->size() > kArray8Cap) {
+      return;
+    }
+    auto *arr8 = new Array8();
+    std::size_t i = 0;
+    for (const auto &elem : *set) {
+      (*arr8)[i++] = elem;  // copy bumps ref count
```
Aksha1812: How often do we expect mode operations ? here we see 16 atomic operations which we wouldn't do earlier. Can we add a section for performance impact analysis in the PR description to make sure it is not too much/ for what scenarios performance will slightly degrade?

Aksha1812: AI suggested below

The fix: Use const_cast (safe here because we're about to destroy the set) or extract:
```
for (auto &elem : *set) {
    (*arr8)[i++] = std::move(const_cast<InternedStringPtr&>(elem));
}
```
Or more correctly using node_handle extraction (but flat_hash_set doesn't have extract-by-iterator efficiently).

src/utils/string_interning.h
```diff
@@ -171,13 +175,658 @@ inline std::ostream &operator<<(std::ostream &os,
...
+    }
+    delete set;  // dec ref count for each element
+    SetArray8(arr8);
+  }
+};
+static_assert(sizeof(BagOfInternedStringPtrs) == 8,
```
Aksha1812, line 900: Can sizeof(InternedStringPtr) be gretaer than 8 in any case ? If not worth adding a static assert for it as well to be extra sure

Aksha1812:
Some mode transition diagrams i had AI make for my own understanding while reviewing this PR. thought it would be helpful to share https://filepaste.app/p/r13hy7so

src/utils/string_interning.h
```diff
@@ -171,13 +175,658 @@ inline std::ostream &operator<<(std::ostream &os,
...
+    std::size_t i = idx;
+    while (i + 1 < N && arr[i + 1]) {
+      arr[i] = std::move(arr[i + 1]);
+      ++i;
+    }
+    arr[i] = nullptr;
+  }
+
+  template <typename Key>
+  std::pair<const_iterator, bool> InsertImpl(Key &&key, bool is_rvalue) {
```
Aksha1812: Was reading about this don't know how this translates to larger functions e.g InsertIntoArray4 . you can let the compiler figure this out at compile time using "perfect forwarding" instead of adding a parameter :
```
  template <typename Key>
  std::pair<const_iterator, bool> InsertImpl(Key &&key) {
      // std::forward<Key>(key) automatically does the right thing:
      // - If Key deduced as InternedStringPtr& → forwards as lvalue (copy)
      // - If Key deduced as InternedStringPtr  → forwards as rvalue (move)
      (*arr)[i] = std::forward<Key>(key);
  }
```

testing/utils/string_interning_test.cc
```diff
@@ -178,6 +184,669 @@ TEST_F(StringInterningMultithreadTest, ConcurrentInterning) {
...
+TEST_F(BagOfInternedStringPtrsTest, InsertDuplicateInSingleMode) {
+  auto k = StringInternStore::Intern("dup");
+  BagOfInternedStringPtrs bag;
+  bag.insert(k);
+  EXPECT_EQ(k.RefCount(), 2);
+  auto [it, inserted] = bag.insert(k);
+  EXPECT_FALSE(inserted);
+  EXPECT_EQ(bag.size(), 1u);
+  EXPECT_EQ(k.RefCount(), 2);
+  EXPECT_NE(it, bag.end());
```
Aksha1812, line 299: I expect this check is to ensure only the refcount increments and we don't transition modes. Probably good to adda sizeof sanity check here to ensure BagOfInternedStringPtrs takes only 8 bytes

Aksha1812: And maybe add size checks in other tests below as well for multi mode , wherevr missing. to ensure this functionality doesn't break. especially demote / promote changes

testing/utils/string_interning_test.cc
```diff
@@ -178,6 +184,669 @@ TEST_F(StringInterningMultithreadTest, ConcurrentInterning) {
...
+      auto lhs_size = lhs.size();
+      auto rhs_size = rhs.size();
+      lhs.swap(rhs);
+      EXPECT_EQ(lhs.size(), rhs_size);
+      EXPECT_EQ(rhs.size(), lhs_size);
+    }
+  }
+}
+
+TEST_F(BagOfInternedStringPtrsTest, RvalueInsertAdoptsRefCount) {
```
Aksha1812, line 573: Can we add test for Rvalue Insert into all Array/Set Modes

```
TEST_F(BagOfInternedStringPtrsTest, RvalueInsertIntoArrayModes) {
    auto keys = MakeKeys(6, "rva_");
    BagOfInternedStringPtrs bag;
    bag.insert(keys[0]);  // Single
    
    InternedStringPtr local = keys[1];
    EXPECT_EQ(keys[1].RefCount(), 2);
    bag.insert(std::move(local));  // Array4 via rvalue
    EXPECT_EQ(keys[1].RefCount(), 2);  // moved, not copied
    
    // Fill to Array8
    for (int i = 2; i < 5; ++i) {
        InternedStringPtr l = keys[i];
        bag.insert(std::move(l));
    }
    
    InternedStringPtr l5 = keys[5];
    bag.insert(std::move(l5));  // Into Array8 via rvalue
    EXPECT_EQ(keys[5].RefCount(), 2);
}
```

testing/utils/string_interning_test.cc
```diff
@@ -178,6 +184,669 @@ TEST_F(StringInterningMultithreadTest, ConcurrentInterning) {
...
+  EXPECT_EQ(std::distance(bag.begin(), bag.end()), 1);
+  EXPECT_EQ((*bag.begin())->Str(), "demoteA");
+  // Removing the lone element returns to empty.
+  EXPECT_EQ(bag.erase(a), 1u);
+  EXPECT_TRUE(bag.empty());
+  EXPECT_EQ(a.RefCount(), 1);
+}
+
+TEST_F(BagOfInternedStringPtrsTest, EraseByIterator) {
+  // Single mode.
```
Aksha1812, line 390: can we add a test to erase non existent keys in array modes , if not added already ?

Aksha1812:
Is it possible to add a reproducible simple benchmark test like https://github.com/valkey-io/valkey-search/blob/main/integration/test_fulltext_space_performance.py ? which we can run on command , just so in future if we are making any changes around this architecture , we can quickly run that script ourselves to see if we are within bounds. this doesn't need to run with every PR but just for our awareness. Is that also possible for some of the performance changes you made @KarthikSubbarao  ?

Aksha1812:
What would happen during batch inserts? or during RDB restore / backfilling. Will all posting lists holding >8 elements go through 
```
Empty → Single → Array4 → Array8 → Set
```
 
Is it possible to reserve enough memory from the beginning and directly create a set and avoid what i guess would be 3 intermediate allocations + 2 data migrations + 1 final allocation.

src/utils/string_interning.h
```diff
@@ -625,6 +625,51 @@ class BagOfInternedStringPtrs {
...
+        return TestMode::kArray4;
+      case kArray8Tag:
+        return TestMode::kArray8;
+      case kSetTag:
+        return TestMode::kSet;
+    }
+    return TestMode::kEmpty;
+  }
+
+  // Hint: reserve space for an expected element count and pre-pick the
```
Aksha1812, line 648: Nit : We can remove these long AI comments now and replace with more concise ones wherever required.

Aksha1812 (approved):
Approving. comments addressed. LGTM !

PR #1027, opened 2026-05-10
Fix crash in unittests introduced in previous PR

allenss-amazon:
Not sure how the previous PR actually made it through the regression testing.


Aksha1812:
which PR is this dependent on. Is it #963 ?

allenss-amazon:
Not sure, but we're not running into it now.

PR #1028, opened 2026-05-10
Revise integration workflows to consolidate unit&integration tests in…

allenss-amazon:
…to the same run inorder to avoid double checkout/build times


Aksha1812:
I tried something like this earlier but doesn't this make the workflow runtime longer? assuming you are running them sequentially . also it is already a bit difficult to go through logs sometimes for one workflow , would be harder in this case. To save on build time . would something like this would help https://github.com/valkey-io/valkey-search/pull/945 .?

allenss-amazon:
I merged #945 in preference to this approach

PR #1029, opened 2026-05-11
Fix RunByMain shutdown leak by draining pending one-shots

allenss-amazon:
## Summary
- `RunByMain()` in `vmsdk/src/utils.cc` allocates an `absl::AnyInvocable` and hands it to `ValkeyModule_EventLoopAddOneShot`. If a worker passes the `IsShuttingDown()` check before the main thread sets the flag, the one-shot is enqueued but the event loop never runs it after `JoinAllThreadPools()`. The allocation, plus the lambda's captured `SearchParameters` and the index graph it transitively owns, leak at process exit.
- This is the failure surfaced as a 32-byte `Direct leak` rooted at `vmsdk::RunByMain` / `src/query/search.cc:829` with a long tail of `Indirect leak`s for HNSW data and `IndexSchema` hashmaps. Most visible in #1026's ASAN runs after `test_cancel.py::TestCancelCMD::test_timeoutCMD` (2 of 3 attempts on the same job).
- Track every allocated callback in a mutex-protected set; `RunAnyInvocable()` removes itself before delete. Add `vmsdk::DrainPendingMainCallbacks()`, called from `OnShutdownCallback` after `JoinAllThreadPools()` (no workers running ⇒ no new inserts can race in) and before the module tears down.

## Test plan
- [x] `./build.sh --run-tests` — all unit tests pass, including new `UtilsTest.DrainPendingMainCallbacksFreesUninvokedCallback`.
- [x] New test uses a `std::shared_ptr` `use_count` tracker to prove the captured state is freed when the one-shot is never invoked, and that the drain is idempotent.
- [ ] Run the integration ASAN job that hit this leak to confirm it no longer triggers under `test_timeoutCMD`.

### DIFF
--- a/vmsdk/src/utils.cc
+++ b/vmsdk/src/utils.cc
@@ -11,11 +11,13 @@
 #include <string>
 #include <utility>
 
+#include "absl/container/flat_hash_set.h"
 #include "absl/functional/any_invocable.h"
 #include "absl/log/check.h"
 #include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/string_view.h"
+#include "absl/synchronization/mutex.h"
 #include "vmsdk/src/log.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
@@ -24,8 +26,20 @@ namespace {
 static bool set_main_thread = false;
 thread_local static bool is_main_thread = false;
 
+// Tracks every RunByMain() callback that has been allocated and handed to
+// ValkeyModule_EventLoopAddOneShot() but not yet invoked. RunAnyInvocable()
+// removes its own entry before deleting. DrainPendingMainCallbacks() deletes
+// whatever is still here at shutdown.
+absl::Mutex outstanding_callbacks_mu;
+absl::flat_hash_set<absl::AnyInvocable<void()> *> outstanding_callbacks
+    ABSL_GUARDED_BY(outstanding_callbacks_mu);
+
 void RunAnyInvocable(void *invocable) {
   absl::AnyInvocable<void()> *fn = (absl::AnyInvocable<void()> *)invocable;
+  {
+    absl::MutexLock lock(&outstanding_callbacks_mu);
+    outstanding_callbacks.erase(fn);
+  }
   (*fn)();
   delete fn;
 }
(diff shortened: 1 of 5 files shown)


vmsdk/src/utils.cc
```diff
@@ -88,9 +102,26 @@ int RunByMain(absl::AnyInvocable<void()> fn, bool force_async) {
...
+    absl::MutexLock lock(&outstanding_callbacks_mu);
+    outstanding_callbacks.insert(call_by_main);
+  }
   return ValkeyModule_EventLoopAddOneShot(RunAnyInvocable, call_by_main);
 }
 
+void DrainPendingMainCallbacks() {
+  absl::MutexLock lock(&outstanding_callbacks_mu);
+  for (absl::AnyInvocable<void()> *fn : outstanding_callbacks) {
+    delete fn;
```
KarthikSubbarao, line 120: Question: Is this delete/free needed because on shutdown, the core valkey's module interface does not drop this? If so, is there some odd chance that the core tries to access this fn pointer and it does not exist?

allenss-amazon: this list isn't actually shared with the core, it's owned by us. When the list is non-zero we invoke the core's one shot callback mechanism. What causes the ASAN failure is that if the core shuts down before the oneshot happens then this list never gets emptied and that causes leaks.

boda26 (approved):

vmsdk/src/utils.cc
KarthikSubbarao: Yes, the list is maintained by the Module, but the fns in the outstanding_callbacks are the private data which the core has a reference to (it is the private data provided in the one shot callback)

KarthikSubbarao: Read the code and looks good:
```
The shutdown event callback is invoked synchronously inside finishShutdown() via moduleFireServerEvent(). After all subscribed
  modules' shutdown callbacks return, finishShutdown() continues its synchronous cleanup and returns C_OK, then the caller does
  exit(0). The event loop never iterates again.
  
  So from the perspective of a module that subscribed to VALKEYMODULE_EVENT_SHUTDOWN:
  
  - When your shutdown callback is invoked, you are guaranteed that no future one-shot callbacks will execute after your callback
  returns.
  - Any one-shot events already queued but not yet processed will never fire.
  - Any one-shot events you queue from within your shutdown callback will also never fire.
```

KarthikSubbarao (approved):

PR #1030, opened 2026-05-11
Reimplement Numeric field for 3x space reduction

allenss-amazon:
Replace existing btree + segment tree with ordered statistic B+ tree.

```
    Space Savings vs Redisearch

      ┌─────────────────────────────────────────────┬────────────┬─────────────┬──────────────────────┐
      │                  Scenario                   │ RediSearch │ valkey main │ valkey (B+-tree+Bag) │
      ├─────────────────────────────────────────────┼────────────┼─────────────┼──────────────────────┤
      │ kpv=1 — 10 000 distinct values/field        │      287.9 │     2 080.1 │                864.4 │
      ├─────────────────────────────────────────────┼────────────┼─────────────┼──────────────────────┤
      │ kpv=2 — 5 000 distinct values/field         │      289.4 │     1 641.8 │                869.8 │
      ├─────────────────────────────────────────────┼────────────┼─────────────┼──────────────────────┤
      │ kpv=3 — 3 333 distinct values/field         │      293.3 │     1 283.1 │                773.4 │
      ├─────────────────────────────────────────────┼────────────┼─────────────┼──────────────────────┤
      │ kpv=10 — 1 000 distinct values/field        │      293.9 │       874.7 │                785.8 │
      ├─────────────────────────────────────────────┼────────────┼─────────────┼──────────────────────┤
      │ kpv=10 000 — 1 value/field (all keys share) │      262.9 │       726.7 │                727.2 │
      └─────────────────────────────────────────────┴────────────┴─────────────┴──────────────────────┘

      Ratio vs RediSearch (lower = more compact)

      ┌────────────┬──────────────────┬───────────────────────────┬────────────┐
      │  Scenario  │ valkey main / RS │ valkey (B+-tree+Bag) / RS │ New ÷ Main │
      ├────────────┼──────────────────┼───────────────────────────┼────────────┤
      │ kpv=1      │            7.23× │                     3.00× │      0.42× │
      ├────────────┼──────────────────┼───────────────────────────┼────────────┤
      │ kpv=2      │            5.67× │                     3.00× │      0.53× │
      ├────────────┼──────────────────┼───────────────────────────┼────────────┤
      │ kpv=3      │            4.38× │                     2.64× │      0.60× │
      ├────────────┼──────────────────┼───────────────────────────┼────────────┤
      │ kpv=10     │            2.98× │                     2.67× │      0.90× │
      ├────────────┼──────────────────┼───────────────────────────┼────────────┤
      │ kpv=10 000 │            2.76× │                     2.77× │     ≈1.00× │
      └────────────┴──────────────────┴───────────────────────────┴────────────┘

    Dynamic Performance:

     Net dynamic-performance improvement using a microbenchmark.

      ┌──────────────────────────┬───────────────────────────────────────┐
      │        Operation         │          New tree vs legacy           │
      ├──────────────────────────┼───────────────────────────────────────┤
      │ Insert                   │ 5–8× faster                           │
      ├──────────────────────────┼───────────────────────────────────────┤
      │ Erase                    │ 4–6× faster                           │
      ├──────────────────────────┼───────────────────────────────────────┤
      │ Range count, any range   │ 1.5–2× faster                         │
      ├──────────────────────────┼───────────────────────────────────────┤
      │ Range scan, small ranges │ within 5–15% (essentially equivalent) │
      ├──────────────────────────┼───────────────────────────────────────┤
      │ Range scan, large/full   │ 1.2–1.7× faster                       │
      └──────────────────────────┴───────────────────────────────────────┘

```


BCathcart:
At a high level your approach makes sense to me. From what I understand so far, we're taking elements of the two previous data structures and combining them into one data structure that should be more efficient in pretty much all the dimensions I can think of and it looks like the performance numbers back that up. Space is obvious: saving a tree worth of pointers, only storing the actual numbers once instead of twice, etc. Range scans make sense to not have to traverse internal nodes after finding the start and end nodes. Looks good.

Let's do a rebase to get rid of a lot of the code that was already merged and then I'll give it a proper review.

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,800 @@
...
+
+  // Postings with value in [start, end] respecting inclusive flags.
+  uint64_t Count(double start, double end, bool start_inclusive,
+                 bool end_inclusive) const;
+
+ private:
+  static constexpr int kLeafCap = 32;
+  static constexpr int kInternalCap = 32;
+  static constexpr int kLeafMin = kLeafCap / 2;          // 16
+  static constexpr int kInternalMin = kInternalCap / 2;  // 16
```
BCathcart, line 69: Not doubting these but curious the reasons for choosing them

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,800 @@
...
+
+  struct NodeBase {
+    Kind kind;
+    uint16_t n;
+    NodeBase* parent;
+    bool IsLeaf() const { return kind == Kind::kLeaf; }
+  };
+
+  struct Leaf : NodeBase {
+    Leaf* prev;
```
BCathcart: Grepping the PR I'm not seeing any case where we actually use this pointer to traverse backwards

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,800 @@
...
+  static constexpr int kInternalCap = 32;
+  static constexpr int kLeafMin = kLeafCap / 2;          // 16
+  static constexpr int kInternalMin = kInternalCap / 2;  // 16
+
+  enum class Kind : uint8_t { kLeaf, kInternal };
+
+  struct NodeBase {
+    Kind kind;
+    uint16_t n;
+    NodeBase* parent;
```
BCathcart: Is this parent pointer ever used?

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,800 @@
...
+
+  // For value-only RankLT (count of values strictly less than v).
+  static int InternalChildForValueLT(const Internal* in, double v) {
+    int i = 0;
+    while (i < in->n && in->separators[i] < v) {
+      ++i;
+    }
+    return i;
+  }
+  static int InternalChildForValueLE(const Internal* in, double v) {
```
BCathcart: This function is the same as InternalChildForValue

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,800 @@
...
+    if (pos < l->n && l->entries[pos].value == v) {
+      auto [_, ins] = l->entries[pos].keys.insert(k);
+      return NoOp(ins);
+    }
+    // Need a new LeafEntry at pos.
+    if (l->n < kLeafCap) {
+      for (int i = l->n; i > pos; --i) {
+        l->entries[i] = std::move(l->entries[i - 1]);
+      }
+      l->entries[pos] = LeafEntry();
```
BCathcart, line 179: This line isn't needed

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,800 @@
...
+      }
+      for (int i = in->n + 1; i > ci + 1; --i) {
+        in->children[i] = in->children[i - 1];
+        in->subtree_count[i] = in->subtree_count[i - 1];
+      }
+      in->separators[ci] = sub.split_sep;
+      in->children[ci + 1] = sub.split_right;
+      // Recompute counts for the two children touched by the split.
+      in->subtree_count[ci] = SubtreePostings(in->children[ci]);
+      in->subtree_count[ci + 1] = SubtreePostings(sub.split_right);
```
BCathcart: Claude recommendation to avoid walking tree subtrees:

**Step 1: Add `right_postings` to `InsertOut`**

```cpp
struct InsertOut {
  bool new_posting;
  NodeBase* split_right;
  double split_sep;
  uint64_t right_postings;  // postings in split_right; 0 if no split
};
```

**Step 2: Fill it in `InsertLeaf` during split**

During the split, sum the bag sizes as you move entries into `right`:

```cpp
uint64_t rp = 0;
for (int i = mid; i < kLeafCap; ++i) {
  rp += l->entries[i].keys.size();   // count before moving
  right->entries[i - mid] = std::move(l->entries[i]);
}
// ... (rest of split unchanged)
return InsertOut{true, right, right->entries[0].value, rp};
```

**Step 3: Use it in `InsertInternal` instead of `SubtreePostings`**

At the point where `sub.split_right` has just been attached:

```cpp
// At the top of the function, subtree_count[ci] was already incremented by 1
// for the new posting. Split moved `sub.right_postings` of those into right.
in->subtree_count[ci + 1] = sub.right_postings;
in->subtree_count[ci]    -= sub.right_postings;
```

That's it — no tree walk, O(kLeafCap) during the split loop rather than O(all leaves in subtree).

The same `right_postings` value is also available for the internal-node split path below it, replacing the two `SubtreePostings` calls in the `cnts[]` staging array.

BCathcart (approved):
LGTM. Just a few nits

src/utils/numeric_btree.h
allenss-amazon: Claude's doing, I think they're related to cache line size.

src/utils/numeric_btree.h
allenss-amazon: done, removed.

src/utils/numeric_btree.h
BCathcart: Not important. Resolving

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,773 @@
...
+    auto* new_root = new Internal();
+    new_root->separators[0] = r.split_sep;
+    new_root->children[0] = root_;
+    new_root->children[1] = r.split_right;
+    new_root->subtree_count[0] = SubtreePostings(root_);
+    new_root->subtree_count[1] = r.right_postings;
+    new_root->n = 1;
+    root_ = new_root;
+  }
+  ++total_postings_;
```
BCathcart:
```suggestion
  ++total_postings_;
  if (r.split_right) {
    auto *new_root = new Internal();
    new_root->separators[0] = r.split_sep;
    new_root->children[0] = root_;
    new_root->children[1] = r.split_right;
    new_root->subtree_count[1] = r.right_postings;
    new_root->subtree_count[0] =
        static_cast<uint64_t>(total_postings_) - r.right_postings;
    new_root->n = 1;
    root_ = new_root;
  }
```

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,773 @@
...
+      return s;
+    }
+    auto* in = static_cast<Internal*>(node);
+    uint64_t s = 0;
+    for (int i = 0; i <= in->n; ++i) {
+      s += in->subtree_count[i];
+    }
+    return s;
+  }
+
```
BCathcart:
```suggestion
```

src/indexes/tag.cc
```diff
@@ -293,165 +234,155 @@ uint32_t Tag::GetMutationWeight() const {
...
-    for (const auto& k : untracked_keys_) {
-      extras.push_back(k);
-    }
-    out_size = negate_total + extras.size();
-    return std::make_unique<EntriesFetcher>(std::move(negate_slots),
-                                            std::move(extras), out_size);
+    size = tracked_tags_by_keys_.size() > size
+               ? tracked_tags_by_keys_.size() - size
+               : tracked_tags_by_keys_.size();
+    size += untracked_keys_.size();
```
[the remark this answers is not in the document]
allenss-amazon: understood. the pre-filtering logic should catch there.

src/utils/numeric_btree.h
```diff
@@ -0,0 +1,754 @@
...
+    cached_max_ = value;
+    return true;
+  }
+  InsertOut r = InsertRec(root_, value, key);
+  if (!r.new_posting) {
+    return false;
+  }
+  ++total_postings_;
+  if (r.split_right) {
+    auto *new_root = new Internal();
```
BCathcart:
```suggestion
    auto* new_root = new Internal();
```

PR #1031, opened 2026-05-11
Reimplement TAG field using RAX tree to save space.

allenss-amazon:
(This is build on top of PR-1026)

5 scenarios were tested:

``` ┌─────┬─────────────────────────────────┬──────────┬─────────────┬─────────────────────┬────────────────────┐
  │  #  │            Workload             │ Tags/doc │ Unique tags │     Cardinality     │      Stresses      │
  ├─────┼─────────────────────────────────┼──────────┼─────────────┼─────────────────────┼────────────────────┤
  │ S1  │ 10k docs, mixed singleton-heavy │        3 │     ~20 050 │ mostly 1, some ~200 │ Realistic baseline │
  ├─────┼─────────────────────────────────┼──────────┼─────────────┼─────────────────────┼────────────────────┤
  │ S2  │ 10k docs, all shared            │        3 │          50 │           ~600 each │ Big posting lists  │
  ├─────┼─────────────────────────────────┼──────────┼─────────────┼─────────────────────┼────────────────────┤
  │ S3  │ 10k docs, all unique            │        5 │      50 000 │            always 1 │ Per-tag overhead   │
  ├─────┼─────────────────────────────────┼──────────┼─────────────┼─────────────────────┼────────────────────┤
  │ S4  │ 50k docs, mixed                 │        4 │    ~100 160 │ mostly 1, some ~625 │ Scale + mixed      │
  ├─────┼─────────────────────────────────┼──────────┼─────────────┼─────────────────────┼────────────────────┤
  │ S5  │ 10k docs, small shared groups   │        3 │       6 000 │            2–8 each │ Bag Array4/Array8  │
  └─────┴─────────────────────────────────┴──────────┴─────────────┴─────────────────────┴────────────────────┘

Space savings are:

  ┌──────────────────────────────┬────────────┬─────────────┬────────────┐
  │           Scenario           │ Legacy Tag │ Current Tag │ RediSearch │
  ├──────────────────────────────┼────────────┼─────────────┼────────────┤
  │ S1 10k mixed singleton-heavy │   15.98 MB │     1.69 MB │    3.85 MB │
  ├──────────────────────────────┼────────────┼─────────────┼────────────┤
  │ S2 10k all shared            │    3.99 MB │     0.74 MB │    1.25 MB │
  ├──────────────────────────────┼────────────┼─────────────┼────────────┤
  │ S3 10k all unique singletons │   22.33 MB │     2.73 MB │    7.28 MB │
  ├──────────────────────────────┼────────────┼─────────────┼────────────┤
  │ S4 50k mixed                 │   79.95 MB │     9.04 MB │   19.36 MB │
  ├──────────────────────────────┼────────────┼─────────────┼────────────┤
  │ S5 10k small shared (2–8)    │    5.25 MB │     0.85 MB │    1.94 MB │
  └──────────────────────────────┴────────────┴─────────────┴────────────┘

  Ratios:

  ┌──────────┬──────────────────┬──────────────────────┐
  │ Scenario │ Current / Legacy │ Current / RediSearch │
  ├──────────┼──────────────────┼──────────────────────┤
  │ S1       │    0.106× (−89%) │        0.439× (−56%) │
  ├──────────┼──────────────────┼──────────────────────┤
  │ S2       │    0.186× (−81%) │        0.595× (−40%) │
  ├──────────┼──────────────────┼──────────────────────┤
  │ S3       │    0.122× (−88%) │        0.375× (−63%) │
  ├──────────┼──────────────────┼──────────────────────┤
  │ S4       │    0.113× (−89%) │        0.467× (−53%) │
  ├──────────┼──────────────────┼──────────────────────┤
  │ S5       │    0.162× (−84%) │        0.440× (−56%) │
  └──────────┴──────────────────┴──────────────────────┘

  Average reduction: 86% vs Legacy Tag, 53% vs RediSearch.
```
Performance benchmark — Tag (Patricia) vs TagRax (rax)

  Per-op wall time, all five scenarios:

  S1: 10k keys, mostly singletons                   Tag         TagRax      Rax/Tag
    AddRecord                                  1726.2 ns/op   1444.2 ns/op   0.84x
    RemoveRecord                               1100.5 ns/op   1235.3 ns/op   1.12x
    ExactSearch                                 404.5 ns/op    401.9 ns/op   0.99x
    PrefixSearch+iter                          4985.9 ns/op   7290.7 ns/op   1.46x
    PrefixCountEstim                             85.2 ns/op     65.3 ns/op   0.77x

  S2: 10k keys, all shared categorical
    AddRecord                                   674.6 ns/op    725.4 ns/op   1.08x
    RemoveRecord                                608.3 ns/op    627.8 ns/op   1.03x
    ExactSearch                                 301.5 ns/op    353.8 ns/op   1.17x
    PrefixSearch+iter                         55236.9 ns/op  77953.0 ns/op   1.41x
    PrefixSearch ns/yield                         9.4 ns         13.2 ns
    PrefixCountEstim                             23.9 ns/op     32.3 ns/op   1.35x

  S3: 10k keys, all unique singleton tags
    AddRecord                                  2379.8 ns/op   2282.8 ns/op   0.96x
    RemoveRecord                               2293.3 ns/op   1959.0 ns/op   0.85x
    ExactSearch                                 478.3 ns/op    412.1 ns/op   0.86x
    PrefixSearch+iter                          1000.1 ns/op    646.3 ns/op   0.65x
    PrefixCountEstim                            111.6 ns/op     65.9 ns/op   0.59x

  S4: 50k keys, mixed
    AddRecord                                  2203.8 ns/op   1992.5 ns/op   0.90x
    RemoveRecord                               1992.1 ns/op   1880.0 ns/op   0.94x
    ExactSearch                                 438.2 ns/op    449.5 ns/op   1.03x
    PrefixSearch+iter                         40829.0 ns/op  89333.0 ns/op   2.19x
    PrefixSearch ns/yield                         9.4 ns         20.6 ns
    PrefixCountEstim                            134.2 ns/op     72.6 ns/op   0.54x

  S5: 10k keys, small shared groups (2-8 keys/tag)
    AddRecord                                  1244.3 ns/op   1214.2 ns/op   0.98x
    RemoveRecord                               1168.9 ns/op   1059.5 ns/op   0.91x
    ExactSearch                                 422.4 ns/op    404.4 ns/op   0.96x
    PrefixSearch+iter                          2510.7 ns/op   3057.2 ns/op   1.22x
    PrefixCountEstim                             90.3 ns/op     55.1 ns/op   0.61x


src/indexes/tag.cc
```diff
@@ -233,156 +277,161 @@ uint32_t Tag::GetMutationWeight() const {
...
-}
 
-std::unique_ptr<EntriesFetcherIteratorBase> Tag::EntriesFetcher::Begin() {
-  auto itr = std::make_unique<EntriesFetcherIterator>(tree_, entries_,
-                                                      untracked_keys_, negate_);
-  itr->Next();
-  return itr;
+  return std::make_unique<EntriesFetcher>(std::move(matched_slots),
+                                          std::move(extras), out_size);
 }
```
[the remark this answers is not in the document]
allenss-amazon: Overestimation for positive queries is known and tolerable. The negative case is actually fixed later, i.e., it's ok for false positives to escape at this point as the filtering will get caught later.

src/indexes/text/rax/rax.c
```diff
@@ -1338,8 +1338,6 @@ int raxMutate(rax *rax, unsigned char *s, size_t len, raxMutateCallback callback
     }
     return success;
 }
-/* END SEARCH */
```
BCathcart, line 1341: I don't think we should remove this. I was using it to indicate the start and end of search-specific code added/modified in the rax code.

src/indexes/tag.cc
```diff
@@ -1,17 +1,18 @@
 /*
- * Copyright (c) 2025, valkey-search contributors
+ * Copyright (c) 2026, valkey-search contributors
```
BCathcart: Just for my understanding, if we make new changes and forget to update the year, is the new code not protected or is it more a formality to keep the year up to date?

src/indexes/tag.cc
```diff
@@ -1,17 +1,18 @@
...
  * All rights reserved.
  * SPDX-License-Identifier: BSD 3-Clause
- *
  */
 
 #include "src/indexes/tag.h"
 
+#include <array>
```
BCathcart: Is this used?

src/indexes/tag.h
```diff
@@ -21,17 +22,32 @@
...
 #include "src/utils/string_interning.h"
 #include "vmsdk/src/valkey_module_api/valkey_module.h"
 
 namespace valkey_search::indexes {
 
+// Tag index backed by an in-tree vs_rax radix tree.
+//
+// Storage model:
+//   - tracked_tags_by_keys_: doc-key → interned raw tag string.
+//   - untracked_keys_: keys present in the dataset but without any tag.
```
BCathcart: nit: I'd be consistent with using "doc-keys" so it's obvious they are the same as in the other two structures

src/indexes/tag.h
```diff
@@ -70,74 +86,69 @@ class Tag : public IndexBase {
...
-                   std::equal_to<InternedStringPtr>, KeySet>;
-  using PatriciaNodeIndex =
-      PatriciaNode<InternedStringPtr, absl::Hash<InternedStringPtr>,
-                   std::equal_to<InternedStringPtr>, KeySet>;
 
+  // Cheap estimate of the number of (tag, key) posting entries whose tag
+  // begins with `prefix`. O(prefix length) — does not iterate the matching
+  // keys. The result counts postings, not unique keys or unique tags.
+  size_t EstimatePrefixCount(absl::string_view prefix) const
+      ABSL_LOCKS_EXCLUDED(index_mutex_);
```
BCathcart: Where is this function used?

src/indexes/tag.cc
```diff
@@ -233,156 +277,161 @@ uint32_t Tag::GetMutationWeight() const {
...
-      if (node != nullptr) {
-        auto res = entries.insert(node);
-        if (res.second && node->value.has_value()) {
-          size += node->value.value().size();
-        }
+      raxIterator it;
+      raxStart(&it, tree_);
+      raxSeekSubTree(&it, qbytes, norm.size());
+      while (raxNext(&it)) {
+        consume_slot(it.data);
```
BCathcart: Did you consider using the Rax C++ class in `rax_wrapper.h/.cc`? I'm sure it would need improvement first, but it seems we're essentially re-implementing the WordIterator here and having to lifecycle manage the rax object. The rax wrapper definitely got a little messy to support closure callbacks for text, but we could add simpler APIs for mutating

src/indexes/tag.cc
```diff
@@ -69,7 +132,6 @@ std::string Tag::UnescapeTag(absl::string_view tag) {
   result.reserve(tag.size());
   for (size_t i = 0; i < tag.size(); ++i) {
     if (tag[i] == '\\' && i + 1 < tag.size()) {
-      // Escape sequence: consume next character literally
```
BCathcart, line 72: We're removing a lot of existing comments. Was that intentional?

src/indexes/tag.cc
```diff
@@ -233,156 +277,161 @@ uint32_t Tag::GetMutationWeight() const {
...
     const query::TagPredicate& predicate, bool negate) const {
-  absl::flat_hash_set<PatriciaNodeIndex*> entries;
-  size_t size = 0;
+  // Collect matched rax slots (each slot's 8 bytes encode a bag) without
+  // iterating their postings; the iterator yields lazily during Begin().
+  absl::flat_hash_set<void*> seen;
+  std::vector<void*> matched_slots;
+  size_t total = 0;
+
+  auto consume_slot = [&](void* slot) {
```
BCathcart: Can we name this `collect_slot`? Consume initially sounded like it was being removed from the tree to me

src/indexes/tag.cc
```diff
@@ -233,156 +277,161 @@ uint32_t Tag::GetMutationWeight() const {
...
+    if (slot == nullptr) return;
+    if (!seen.insert(slot).second) return;
+    matched_slots.push_back(slot);
+    auto bag = BagOfInternedStringPtrs::Adopt(SlotToStorage(slot));
+    total += bag.size();
+    (void)bag.Release();
```
BCathcart, line 395: Another pattern would be to have a static `BagOfInternedStringPtrs::PeekSize(uintptr_t)` function so that we don't transfer ownership at all. It's very insignificant but popped into my head as being a bit more intuitive to peek the object where it is than taking it then giving it back.

BCathcart: Nevermind we also have this pattern for entries fetcher. We could make a BagView class I suppose but that's probably over-engineering.

src/indexes/tag.cc
allenss-amazon: Depends on the amount of change. If the change is minor / formatting, bug fix, etc. Then no. But if the change is larger, then it becomes a "derived" work and then yes.

allenss-amazon: Fixed.

src/indexes/tag.h
allenss-amazon: fixed

src/indexes/tag.h
allenss-amazon: removed

src/indexes/tag.cc
allenss-amazon: Conceptually, I like the idea. here's what claude had to say about it:

```
 The Rax wrapper in
  src/indexes/text/rax_wrapper.h already exposes everything Tag needs. The real work isn't adding capabilities — it's removing the
  wrapper's text-specific coupling so Tag can use it without dragging in posting.h / Postings / StemParents.

  Capability check — what Tag uses from raw rax vs what the wrapper provides

  ┌───────────────────────────────────────────┬─────────────────────────────────────────────┬─────────────────────────────────────────┐
  │            Tag's raw-rax site             │                 Wrapper API                 │                 Status                  │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ raxNew() + raxFreeWithCallback(tree_,     │ Rax(free_callback) ctor + ~Rax()            │ ✓ already there                         │
  │ free_cb)                                  │                                             │                                         │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ raxMutate(tree_, bytes, len, trampoline,  │ Rax::MutateTarget(word,                     │                                         │
  │ ctx, ADD) (insert path)                   │ FunctionRef<void*(void*)>,                  │ ✓ already there                         │
  │                                           │ item_count_op::ADD)                         │                                         │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ raxMutate(... SUBTRACT) (erase path)      │ same MutateTarget with SUBTRACT             │ ✓ already there                         │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ raxFind(tree_, bytes, len, &p) for exact  │ Rax::FindTarget(word) returns void*         │ ✓ already there                         │
  │ lookup                                    │                                             │                                         │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ raxIterator + raxStart + raxSeekSubTree + │ Rax::WordIterator(prefix) —                 │ ✓ already there                         │
  │  raxNext + raxStop for prefix walk        │ Next/Done/GetWord/GetTarget                 │                                         │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────────────────┤
  │ Full-tree walk for the negate branch      │ Rax::WordIterator("") walks lexically from  │ ✓ already there                         │
  │ (raxSeekSubTree(&empty, 0))               │ root                                        │                                         │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────┼─────────────────────────────────────────┤
  │                                           │                                             │ ✓ already there (no longer needed by    │
  │ raxGetSubtreeItemCount(...)               │ Rax::GetSubtreeItemCount(prefix)            │ Tag; was used by the deleted            │
  │                                           │                                             │ EstimatePrefixCount)                    │
  └───────────────────────────────────────────┴─────────────────────────────────────────────┴─────────────────────────────────────────┘

  No new wrapper APIs are required. What I told the reviewer in the previous reply was incorrect — sorry. The right reply, and what
  justifies actually doing the migration, is the coupling cleanup below.

  Actual gaps (none about missing functionality)

  ┌───────────────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────┐
  │                                  Gap                                  │                       Impact on Tag                       │
  ├───────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ Wrapper lives in src/indexes/text/rax_wrapper.{h,cc} and the type is  │ Tag (valkey_search::indexes::Tag) would have to import    │
  │ valkey_search::indexes::text::Rax                                     │ text::Rax, which reads weirdly: Tag is not a text         │
  │                                                                       │ feature.                                                  │
  ├───────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ Wrapper transitively includes posting.h for InvasivePtr<Postings> and │ Pulls text-only types into the Tag translation unit.      │
  │  defines StemParents, exposes FindPostingsTarget /                    │ Compile-time noise, but no real harm.                     │
  │ WordIterator::GetPostingsTarget() / GetStemParentsTarget()            │                                                           │
  ├───────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────┤
  │ MutateTarget operates on void* slots, callers do their own            │ Fine for Tag (BagOfInternedStringPtrs::Adopt/Release      │
  │ adopt/release                                                         │ already exists), but verbose.                             │
  └───────────────────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────┘

  Phase plan

  Phase 1 — Neutralize the wrapper (no new capabilities; pure refactor)

  Goal: The wrapper becomes safe for Tag to use without dragging in text-specific types.

  1. Relocate the wrapper.
    - Move src/indexes/text/rax_wrapper.{h,cc} → src/indexes/rax/rax_cpp.{h,cc} (or src/indexes/rax_wrapper.{h,cc} — name TBD).
    - Move the type from valkey_search::indexes::text::Rax → valkey_search::indexes::Rax.
    - Update all text/ include sites (rax_wrapper.h → new path) and the using aliases inside text/.
    - CMake: the wrapper currently sits inside the text target. Carve it into its own static lib rax_cpp (or fold into rax_lib). text and
  tag both link rax_cpp.
  2. Split text-specific accessors out of the wrapper.
    - Delete from Rax: FindPostingsTarget, WordIterator::GetPostingsTarget, WordIterator::GetStemParentsTarget,
  PathIterator::GetPostingsTarget, the Defrag shortcut.
    - Delete the StemParents typedef from the wrapper header.
    - Delete the using StemParents = std::vector<std::string> from rax_wrapper.h.
    - Drop #include "posting.h" from the wrapper header.
    - In text/, add thin call-site helpers that do the type-specific cast where today's accessor lives. Example:
  // In text-side helper
  inline InvasivePtr<Postings> FindPostingsTarget(const Rax& tree,
                                                 absl::string_view word) {
    return InvasivePtr<Postings>::CopyRaw(
        static_cast<InvasivePtrRaw<Postings>>(tree.FindTarget(word)));
  }
    - Net: the wrapper's public surface is type-agnostic (void* only). text/ regains its convenience accessors as free functions in its
  own namespace.
  3. Wrapper-internal cleanup (touch lightly, defer rewrites):
    - The TODO at rax_wrapper.h:21-23 about prefix-count is satisfied (GetSubtreeItemCount exists). Remove the TODO.
    - The "(TODO) This function is explicitly multi-thread safe" claim on MutateTarget (rax_wrapper.h:83) is misleading — it isn't today
  and Tag won't rely on it being so. Soften to "currently single-mutator; future work TBD."

  Risk for Phase 1: Low. It's a move + a header split. text/ keeps working via thin shims. Build CI will catch any include miss. ~1 hour,
  plus diff for review.

  Phase 2 — Migrate Tag

  Goal: Replace every raw-rax call in tag.{h,cc} with the relocated wrapper.

  Touch points in tag.{h,cc}:

  - tag.h:
    - rax* tree_; → Rax tree_;
    - Drop #include "src/indexes/text/rax/rax.h"; add #include "src/indexes/rax_wrapper.h".
    - Drop the trailing raxFreeWithCallback from ~Tag() (wrapper dtor handles it).
  - tag.cc:
    - Remove TagFreeCallback, SlotToStorage/StorageToSlot helpers (the former passed to Rax ctor; the latter inlined into the mutate
  lambdas).
    - Remove the MutateCtx struct and TagMutateTrampoline extern-C function.
    - IndexTagForKey / DeindexTagForKey become:
  void Tag::IndexTagForKey(absl::string_view tag,
                           const InternedStringPtr& key) {
    tree_.MutateTarget(Normalize(tag),
      [&](void* current) -> void* {
        auto bag = BagOfInternedStringPtrs::Adopt(
            reinterpret_cast<uintptr_t>(current));
        bag.insert(key);
        return reinterpret_cast<void*>(bag.Release());
      }, ADD);
  }
  // DeindexTagForKey: same shape with bag.erase + SUBTRACT
    - Tag::Search exact branch:
  void* p = tree_.FindTarget(Normalize(q));   // was raxFind(...)
  collect_slot(p);
    - Tag::Search prefix branch:
  for (auto it = tree_.GetWordIterator(Normalize(q)); !it.Done(); it.Next()) {
    collect_slot(it.GetTarget());
  }
    - Tag::Search negate branch (full-tree walk):
  for (auto it = tree_.GetWordIterator(""); !it.Done(); it.Next()) {
    void* slot = it.GetTarget();
    if (slot && !seen.contains(slot)) { ... }
  }
    - Constructor: tree_(TagFreeCallback) → wrapper takes the bag-freeing callback once. The callback body stays the same (Adopt + drop):
  extern "C" void TagFreeCallback(void* p) {
    (void)BagOfInternedStringPtrs::Adopt(reinterpret_cast<uintptr_t>(p));
  }
  Tag::Tag(...) : ..., tree_(&TagFreeCallback) {}

  Expected net diff: tag.cc shrinks by ~40 lines (no extern-C trampoline, no manual raxStart/raxStop, no MutateCtx), with no behavior
  change.

  Risk for Phase 2: Low-medium. The MutateTarget callback signature (void*(void*)) maps 1:1 to what the trampoline does, so semantics are
  preserved. The only thing to watch is that the wrapper's MutateTarget does CHECK(res) on failure where today Tag silently assumed
  success — that's a tightening, not a regression.

  Phase 3 (deferred) — Optional Rax<T> template

  Out of scope for the immediate migration but worth flagging:

  - A class template Rax<T> where T knows how to ctor/dtor itself from the slot's bits would eliminate the Adopt/Release boilerplate at
  every call site. For Tag this would mean a RaxSlot<BagOfInternedStringPtrs> RAII wrapper that auto-releases. For text it'd be
  RaxSlot<InvasivePtr<Postings>>.
  - Real value, but bigger surface area — affects text/ too, and InvasivePtr is itself a wrapper around the same pattern. Track as
  separate follow-up.

```

src/indexes/tag.cc
allenss-amazon: restored

src/indexes/tag.cc
allenss-amazon: changed

src/indexes/tag.cc
BCathcart: From a quick glance I think AI is going in the right direction here. I don't mind owning this as a follow-up to not crowd this current PR with the refactoring. I'm familiar enough with the wrapper code it shouldn't take me long.

BCathcart (approved):
I figure we can get this in and I can own the rax wrapper refactor as a follow-up if that sounds good to you. Let's just add the `/* END SEARCH */` back and then we're good to go.

allenss-amazon:
fixed  the missin comment.

PR #1039, opened 2026-05-14
FT.SEARCH and FT.AGGREGATE SCORER argument update

zackcam:
Adds basic scoring support to search, enabling both FT.SEARCH and FT.AGGREGATE to accept a SCORER parameter and FT.SEARCH to support WITHSCORES.

**Commit 1**: Adding scorer and withscores argument to ft.search command
Parse WITHSCORES and SCORER  parameter in FT.SEARCH. Add score field to the Neighbor struct (initialized from distance for vector queries, computed by scorer for text queries). Emit score in the response: as a top-level element for non-vector queries (WITHSCORES), and as a named attribute for vector queries (via the AS alias). Add parser tests covering WITHSCORES, SCORER, and their combinations

**Commit 2**: Adding scorer field to ft.aggregate
Add SCORER parameter parsing to FT.AGGREGATE. Move scorer field from SearchCommand to the shared SearchParameters base class so both commands inherit it. Use n.score instead of n.distance for vector query results in aggregate


src/commands/ft_search.cc
```diff
@@ -66,7 +66,16 @@ void SendReplyNoContent(ValkeyModuleCtx *ctx,
 void ReplyScore(ValkeyModuleCtx *ctx, ValkeyModuleString &score_as,
                 const indexes::Neighbor &neighbor) {
   ValkeyModule_ReplyWithString(ctx, &score_as);
-  auto score_value = absl::StrFormat("%.12g", neighbor.distance);
+  auto score_value = absl::StrFormat("%.12g", neighbor.score);
```
Frank-Gu-81, line 69: nit: maybe adding a comment here to indicate that non-vector queries will always emit score of 0.0 at the moment until we have a scoring function properly implemented.

src/commands/ft_aggregate_parser.cc
```diff
@@ -248,6 +249,8 @@ vmsdk::KeyValueParser<AggregateParameters> CreateAggregateParser() {
                         GENERATE_FLAG_PARSER(AggregateParameters, inorder));
   parser.AddParamParser(kVerbatim,
                         GENERATE_FLAG_PARSER(AggregateParameters, verbatim));
+  parser.AddParamParser(kScorerParam,
+                        GENERATE_VALUE_PARSER(AggregateParameters, scorer));
```
Frank-Gu-81: Should we add a validation here to check scorer value against a list of supported scoring functions (i.e. `BM25STD` and `TFIDF` for now) instead of accepting any arbitrary value?

src/commands/ft_search_parser.cc
```diff
@@ -235,6 +237,8 @@ vmsdk::KeyValueParser<SearchCommand> CreateSearchParser() {
                         GENERATE_FLAG_PARSER(SearchCommand, verbatim));
   parser.AddParamParser(query::kSlop,
                         GENERATE_VALUE_PARSER(SearchCommand, slop));
+  parser.AddParamParser(query::kScorer,
+                        GENERATE_VALUE_PARSER(SearchCommand, scorer));
```
Frank-Gu-81: Same here. I think it'd be nice to add a check for supported scorers instead of allowing any arbitrary value.

src/commands/ft_aggregate_parser.cc
zackcam: I think we should leave this open more and just add the failing later if we don't have a matching scorer. The end goal should be allowing user provided scorers which can be named anything. So I was thinking we should limit after we have defined what algorithms we will support as well as how we will deal with user defined scorers.

src/commands/ft_search_parser.cc
zackcam: Same response as above. Will add a comment in the code so that it is documented we will want to restrict later. Either by seeing if we have a list of supported scorers or associated scorers

Frank-Gu-81 (approved):

src/coordinator/server.cc
```diff
@@ -99,7 +99,7 @@ void SerializeNeighbors(SearchIndexPartitionResponse* response,
   for (const auto& neighbor : neighbors) {
     auto* neighbor_proto = response->add_neighbors();
     neighbor_proto->set_key(std::move(*neighbor.external_id));
-    neighbor_proto->set_score(neighbor.distance);
+    neighbor_proto->set_score(neighbor.score);
```
KarthikSubbarao, line 102: From tests, we might need to store both score and distance for text->vector queries

src/indexes/vector_base.h
```diff
@@ -51,34 +51,41 @@ std::vector<char> NormalizeEmbedding(absl::string_view record, size_t type_size,
 struct Neighbor {
   InternedStringPtr external_id;
   float distance;
+  float score{0.0f};
```
KarthikSubbarao: Is this (float) precision sufficient? @boda26

KarthikSubbarao: And what's the default score? 0? 

Which ever it is, maybe a constant will be good to use

src/query/search.h
```diff
@@ -178,6 +180,11 @@ struct SearchParameters {
   bool inorder{false};
   std::optional<uint32_t> slop;
   bool verbatim{false};
+  // TODO: Scorer is currently a placeholder. The selected scoring function is
+  // not yet invoked; non-vector queries will report a score of 0.0.
+  // TODO: Validate the scorer value during parsing and reject invalid/unknown
+  // scorer names with an appropriate error.
+  std::string scorer{"BM25STD"};
```
KarthikSubbarao: Should we have a enum for the scorer which is limited to certain types? The benefit here are:
1. Limited types. We should be rejecting any unknown type other than BM25STD and TDIDF
2. It can implement a method that generates the str form. e.g. "BM25STD"
3. The enum can default to BM25STD
4. This can be the scorer which Miles ends up extending for adding the scoring core logic

testing/ft_search_parser_test.cc
```diff
@@ -79,6 +79,9 @@ struct FTSearchParserTestCase {
   query::SortOrder sortby_order{query::SortOrder::kAscending};
   bool sortby_enabled{false};
   bool with_sort_keys{false};
+  // WITHSCORES and SCORER test fields
+  bool with_scores{false};
+  std::string scorer = "BM25STD";
```
KarthikSubbarao: I see "BM25STD" in a few different places. Let's add the const string above the Scorer Enum suggested [here](https://github.com/valkey-io/valkey-search/pull/1039/changes#r3270771226)

src/indexes/vector_base.h
KarthikSubbarao: Also, let us not use a defaulted value in the structure.

It can hide the usages of `Neighbor` where we need to assign a score.

It would be good to manually assign the score in this PR , so that we identify the places to assign a score which will need an update later on

zackcam:
@coderabbitai review

zackcam:
Closing as changes are in main scoring pr now

PR #1043, opened 2026-05-16
Support Borrowed InternedString Ptr to avoid ref counting 

KarthikSubbarao:
This PR aims to eliminate ref counting overhead in read path via a new Borrowed impl of InternedStringPtr

## Problem

After fixing the PausePoint mutex bottleneck (#1037), `DecrementRefCount` becomes the dominant CPU consumer at **44%** of all cycles under multi-threaded read load. When users execute searches which match 10K results, the bottle neck becomes `SearchNonVectorQuery` building a vector of up to 10K `Neighbor` objects, each copying an `InternedStringPtr` (atomic increment). Then `TrimResults` destroys ~9,990 of them (atomic decrement each). A single query performs ~20,000 atomic operations just for result collection — all unnecessary since the reader lock guarantees no mutations occur during the search.

## Fix

Support Borrowed InternedString Ptr to avoid ref counting.  Borrowed pointers skip ref counting on copy and destroy. During `SearchNonVectorQuery`, results are collected as borrowed pointers. After `TrimResults` keeps only the final `offset + limit` results, those survivors are materialized (ref count incremented) before the reader lock is released.

This reduces atomic operations per query from ~20,000 to ~10-15 (only the final returned results).

## Data

Benchmark: 500 clients, 20s duration, core-pinned (server 0-31, client 32-63).

### TEXT — 10K matches (req/s)

| Threads | PausePoint Fix Only | + Borrowed Ptr | Improvement |
|:-------:|:-------------------:|:--------------:|:-----------:|
| 1 | 1,479 | 1,747 | +18% |
| 2 | 2,829 | 3,413 | +21% |
| 4 | 5,125 | 6,074 | +19% |
| 8 | 7,185 | 10,636 | +48% |
| 16 | 11,483 | 19,732 | **+72%** |

### NUMERIC — 10K matches (req/s)

| Threads | PausePoint Fix Only | + Borrowed Ptr | Improvement |
|:-------:|:-------------------:|:--------------:|:-----------:|
| 1 | 3,854 | 6,791 | +76% |
| 2 | 6,807 | 13,450 | +98% |
| 4 | 12,943 | 26,844 | +107% |
| 8 | 17,259 | 54,119 | **+214%** |
| 16 | 17,286 | 72,660 | **+320%** |

### TAG — 10K matches (req/s)

| Threads | PausePoint Fix Only | + Borrowed Ptr | Improvement |
|:-------:|:-------------------:|:--------------:|:-----------:|
| 1 | 2,161 | 2,566 | +19% |
| 2 | 3,055 | 5,147 | +68% |
| 4 | 5,108 | 10,291 | +101% |
| 8 | 8,871 | 20,325 | **+129%** |
| 16 | 12,093 | 40,725 | **+237%** |

### 100 matches — small regression at mid-thread counts

| Threads | TEXT before | TEXT after | NUMERIC before | NUMERIC after | TAG before | TAG after |
|:-------:|:----------:|:---------:|:--------------:|:------------:|:----------:|:---------:|
| 1 | 89,743 | 86,533 | 96,101 | 92,920 | 93,081 | 93,344 |
| 4 | 67,175 | 68,786 | 87,985 | 87,092 | 91,718 | 88,204 |
| 8 | 69,315 | 66,724 | 82,989 | 82,393 | 85,091 | 83,354 |
| 16 | 69,669 | 66,046 | 84,104 | 81,126 | 83,567 | 82,292 |

Light workloads show ±3-5% variation across runs — which I believe is measurement noise as the `IsBorrowed()` check is a  single predicted branch (always false on the owning path), so no meaningful overhead is expected.

**Note:** NUMERIC and TAG see larger improvements than TEXT  because for them, the Neighbor collection was the *only* source of ref counting contention, so eliminating it gives better scaling. For Text,  the `TermIterator` currently causes additional ref counting due to how it utilizes the inlined heap iterator that currently accepts `InternedStringPtr` copies.  I validated that updating it to store raw `const Key*` pointers in the heap eliminates another ~20K atomic ops per query and doubles throughput again to ~36K on TEXT 16T. This is a separate change to `term.h`/`term.cc` which I will make next


src/utils/string_interning.h
```diff
@@ -160,12 +163,69 @@ class InternedStringPtr {
...
+//
+// Safety contract: The caller MUST guarantee that the underlying
+// InternedString outlives this pointer. Typically this means the reader lock
+// that protects the index must be held for the lifetime of all
+// BorrowedInternedStringPtr instances.
+//
+// Call Materialize() to convert to an owning InternedStringPtr (increments
+// ref count) before releasing the lock.
+//
+class BorrowedInternedStringPtr {
```
Aksha1812, line 188: Tests missing. Can we add some simple functional tests here https://github.com/valkey-io/valkey-search/blob/main/testing/utils/string_interning_test.cc

KarthikSubbarao: This should have already been covered using existing integration and unit tests for the background reader thread search path. But we did not have borrowed string ptr class tests. I will add them to the file you shared. Thanks for checking

src/query/search.cc
```diff
@@ -719,24 +709,39 @@ SearchResult::SearchResult(size_t total_count,
     : total_count(total_count),
       is_limited_with_buffer(false),
       is_offsetted(false) {
-  // Clear neighbors if no results should be returned
+  this->neighbors = std::move(neighbors);
   if (ShouldReturnNoResults(parameters)) {
```
Aksha1812, line 713: Maybe we can move neighbours after this check rather than before. I think this is trivial though as it just moving a pointer to vecotr of neighbours, i guess it is not that expensive. you can probably ignore this comment

src/utils/string_interning.h
```diff
@@ -160,12 +163,69 @@ class InternedStringPtr {
...
+  // Trivially copyable, trivially destructible — no ref counting.
+  BorrowedInternedStringPtr(const BorrowedInternedStringPtr &) = default;
+  BorrowedInternedStringPtr &operator=(const BorrowedInternedStringPtr &) =
+      default;
+  ~BorrowedInternedStringPtr() = default;
+
+  absl::string_view Str() const { return ptr_->Str(); }
+  const InternedString *operator->() const { return ptr_; }
+  const InternedString &operator*() const { return *ptr_; }
+  operator bool() const { return ptr_ != nullptr; }
```
Aksha1812: AI nit : `operator bool()` should be `explicit` — prevents accidental implicit
conversions in arithmetic/comparison contexts. Same as shared_ptr/optional do.

Aksha1812 (approved):
LGTM ! Very straightforward changes. Approving. I asked AI where else this can be used in valkey-search code. you can see if below is helpful. I think it mostly says half of them are already covered by #1044 which i have not reviewed yet

```
Other Places in valkey-search Where BorrowedInternedStringPtr Could Be Used
High Value (same pattern, high call frequency)
1. VectorBase::GetKeyDuringSearch — returns InternedStringPtr by value
File: src/indexes/vector_base.cc:211-217

Currently returns absl::StatusOr<InternedStringPtr> which COPIES from the internal key_by_internal_id_ map (atomic increment). Called in the HNSW search hot loop — once per candidate vector checked.

For a KNN query with K=10 on a million-doc index, HNSW might check 200-1000 candidates. Each check copies the InternedStringPtr to check if it passes the filter. Most candidates are discarded.

Could return a borrowed reference instead, since the reader lock is held for the entire search. The caller only needs the key to check against a filter set and (for the K survivors) to add to results. Filter-check can use borrowed; only survivors need materialization.

Estimated savings: 200-1000 atomic ops per vector query → ~10 (only K survivors).

2. TermIterator + OrProximityIterator (already addressed by PR 1044)
PR 1044 fixes TermIterator (raw pointers in heap) and OrProximityIterator (replaces multiset with InlinedPriorityQueue). These were the other high-value sites.

Medium Value
3. ProximityIterator::FindCommonKey local variables (already addressed by PR 1044)
PR 1044 changes Key min_key; Key max_key; to const Key* pointers. Already fixed.

Low Value / Not Applicable
4. PredicateEvaluator::target_key_ in response_generator.cc
Takes InternedStringPtr by value. But this runs on the main thread AFTER the reader lock is released (during content fetch). The copy is NEEDED because the lock isn't held anymore. Cannot borrow.

5. Fanout/cluster path
Creates new InternedStringPtrs from deserialized protobuf data (remote shards). There's no existing owning pointer to borrow FROM. Not applicable.

6. PopulateIndexMutationSequenceNumbers
Accesses n.external_id by reference in a find() call. No copy. Already optimal.
```

src/query/search.cc
```diff
@@ -748,65 +753,82 @@ void SearchResult::TrimResults(std::vector<indexes::Neighbor> &neighbors,
...
-      SearchResult(total_count, std::move(result), parameters);
+  ++Metrics::GetStats().time_slice_queries;
+  // Handle OOM for search requests, defends against request
+  // coming from the coordinator
+  if (search_mode == SearchMode::kRemote) {
+    auto ctx = vmsdk::MakeUniqueValkeyThreadSafeContext(nullptr);
+    auto ctx_flags = ValkeyModule_GetContextFlags(ctx.get());
+    if (ctx_flags & VALKEYMODULE_CTX_FLAGS_OOM) {
+      return absl::ResourceExhaustedError(kOOMMsg);
+    }
```
allenss-amazon, line 816: I don't think this is correct behavior.

src/query/search.cc
```diff
@@ -719,24 +709,39 @@ SearchResult::SearchResult(size_t total_count,
...
+  }
+  // Materialize only the survivors into owning Neighbor vector.
+  neighbors.reserve(borrowed.size());
+  for (auto &b : borrowed) {
+    neighbors.emplace_back(b.key.Materialize(), b.distance);
+  }
+}
+
+template <typename T>
+void SearchResult::TrimResults(std::vector<T> &vec,
```
allenss-amazon, line 741: This is inefficient when offset > 0. Long term, I suspect we want to combine this with the un-borrowing code so as to avoid needless erases from the start of a vector.

src/utils/string_interning.h
```diff
@@ -160,12 +163,69 @@ class InternedStringPtr {
...
+    }
+    return InternedStringPtr(const_cast<InternedString *>(ptr_));
+  }
+
+  auto operator<=>(const BorrowedInternedStringPtr &other) const {
+    return ptr_ <=> other.ptr_;
+  }
+  bool operator==(const BorrowedInternedStringPtr &other) const {
+    return ptr_ == other.ptr_;
+  }
```
allenss-amazon: Doesn't this duplicate the spaceship version?

allenss-amazon (changes requested):

src/query/search.cc
KarthikSubbarao: This is retaining the existing behavior we had today

src/utils/string_interning.h
KarthikSubbarao: Yes, it is redundant .  I will remove the `operator==` as it is already synthesized from the `operator<=>`

src/query/search.cc
KarthikSubbarao: For this PR, I would prefer keeping the changes specific to the borrowed interned string changes. (This PR retains the existing TrimResult logic)

The SearchResults and the trimming logic has optimizations applied to it for standalone vs CME.
The trimming from the front needs to happen in the coordinator node during the merge in case of CME. For standalone, we can likely optimize it in the TrimResults fn (to avoid memmove), but I would prefer we keep this logical change separate to be done in a follow up PR for the large offset case

PR #1045, opened 2026-05-17
Optimize content fetch: eliminate double key open, use HashGet for specific field access

KarthikSubbarao:
## Problem

When FT.SEARCH returns content, the main thread fetches hash fields for each result. In this PR, I describe two areas which are slowing down the path:

1. Redundant double key open - For each result, `GetContent` opens the key (hashtable lookup + zmalloc), checks expiry, then passes the handle to `FetchAllRecords` which ignores it and opens the same key again. At 31K queries/sec with LIMIT 10, this is 310K unnecessary opens/sec.

2. Full scan when only specific fields are needed - When the user specifies `RETURN field1 field2`, the response generator path is currently scanning ALL fields via `ValkeyModule_ScanKey` (allocating a string object per field+value) and filters in the callback based on specifier. For a 50-field hash with `RETURN 1`, this creates 100 objects when only 2 are needed.

## Fix

1. `FetchAllRecords` now uses the caller's already-open key handle directly (eliminates the second open). Caller flags was changed from `NOEFFECTS` to `NOEXPIRE` and this allows us to continue having key stats/LRU updates while keeping the crash-prevention safety (See X).

2. When RETURN is specified and `identifiers.size() <= hash_field_count / 2`, use `ValkeyModule_HashGet` per field instead of scanning. Otherwise fall through to scan.

The threshold exists because for listpack-encoded hashes, each HashGet call does a linear scan (`lpFind`) of the entire listpack. N calls = O(N×M) comparisons vs one scan pass = O(M). At N ≈ M, HashGet is slower. Using `hash_len / 2` (`ValkeyModule_ValueLength` is O(1)) as the cutoff eliminates all regressions in the bench mark test, but feel free to provide other suggestions. 

| 50-field hash, LIMIT 10 | RETURN 1 | RETURN 10 | RETURN 50 (all) |
|:---|:---:|:---:|:---:|
| HashGet | 28,452 | 10,393 | 2,136 |
| Scan | 10,485 | 7,521 | 3,407 |
| Difference | **+171%** | **+38%** | **-37%** |

## Data

Benchmark: NUMERIC query matching 10K docs, 16 reader threads, 500 clients, core-pinned (server 0-15, client 16-31). Each hash field ~50 bytes.

| Fields/doc | LIMIT | RETURN | Baseline | Optimized | Improvement |
|:---:|:---:|:---:|:---:|:---:|:---:|
| 50 | 100 | 1 | 1,277 | 5,152 | **+303%** |
| 50 | 100 | 5 | 1,022 | 2,205 | **+116%** |
| 50 | 100 | 10 | 841 | 1,229 | **+46%** |
| 50 | 10 | 1 | 10,485 | 28,136 | **+168%** |
| 50 | 10 | 5 | 8,767 | 16,030 | **+83%** |
| 25 | 100 | 1 | 2,020 | 5,184 | **+157%** |
| 25 | 10 | 1 | 15,330 | 28,393 | **+85%** |
| 10 | 100 | 1 | 3,232 | 5,060 | **+57%** |
| 10 | 10 | 1 | 21,615 | 28,434 | **+32%** |

No regression when RETURN == all fields, because at this threshold we fallback to the existing scan approach:

| Fields/doc | LIMIT | RETURN (all) | Baseline | Optimized |
|:---:|:---:|:---:|:---:|:---:|
| 50 | 100 | 50 | 310 | 313 |
| 25 | 100 | 25 | 642 | 682 |
| 10 | 100 | 10 | 1,375 | 1,465 |


src/query/response_generator.cc
```diff
@@ -219,8 +219,11 @@ absl::StatusOr<RecordsMap> GetContentNoReturnJson(
     identifiers.insert(sortby_identifier);
   }
   auto key_str = vmsdk::MakeUniqueValkeyString(key);
+  // NOEXPIRE prevents lazy expiry deletion which could cause
+  // server.also_propagate.numops == 0 crash. The key handle is reused
+  // by FetchAllRecords to avoid a redundant second open.
   auto key_obj = vmsdk::MakeUniqueValkeyOpenKey(
-      ctx, key_str.get(), VALKEYMODULE_OPEN_KEY_NOEFFECTS | VALKEYMODULE_READ);
+      ctx, key_str.get(), VALKEYMODULE_OPEN_KEY_NOEXPIRE | VALKEYMODULE_READ);
```
Aksha1812, line 226: Didn't NOEFFECTS also prevent lazy expiry. Can we add a comment here that now FT.SEARCH content fetches would update LRU/ LFU time,   appear in MONITOR, and may trigger keyspace notifications and other possible implications. AI told this is the difference between NOEFFECTS and NOEXPIRE

src/attribute_data_type.cc
```diff
@@ -78,37 +78,61 @@ bool HashHasRecord(ValkeyModuleKey *key, absl::string_view identifier) {
...
+// Fetch only the requested fields by name using HashGet, avoiding a full scan.
+// This is faster than scanning when the hash has many more fields than
+// requested. Fields that don't exist in the hash are silently skipped.
+RecordsMap HashAttributeDataType::FetchSpecificFields(
+    ValkeyModuleKey *open_key,
+    const absl::flat_hash_set<absl::string_view> &identifiers) const {
+  RecordsMap content;
+  for (const auto &id : identifiers) {
+    ValkeyModuleString *value = nullptr;
+    ValkeyModule_HashGet(open_key, VALKEYMODULE_HASH_CFIELDS, id.data(), &value,
```
Aksha1812, line 123: Do we need a sanity check before this to verify if id's are null-terminated since HASH_CFIELDS expects null-terminated C strings.

src/attribute_data_type.cc
```diff
@@ -78,37 +78,61 @@ bool HashHasRecord(ValkeyModuleKey *key, absl::string_view identifier) {
 
 absl::StatusOr<RecordsMap> HashAttributeDataType::FetchAllRecords(
     ValkeyModuleCtx *ctx, const std::optional<std::string> &vector_identifier,
```
Aksha1812, line 80: Nit : `ctx` and `key` no longer used in the function , maybe we can remove ?

src/attribute_data_type.cc
```diff
@@ -78,37 +78,61 @@ bool HashHasRecord(ValkeyModuleKey *key, absl::string_view identifier) {
...
   if (vector_identifier.has_value() &&
-      !HashHasRecord(key_obj.get(), vector_identifier.value())) {
+      !HashHasRecord(open_key, vector_identifier.value())) {
     return absl::NotFoundError(absl::StrCat("No such record with identifier: `",
                                             vector_identifier.value_or(""),
                                             "`"));
   }
+  if (!identifiers.empty()) {
+    size_t hash_len = ValkeyModule_ValueLength(open_key);
+    if (identifiers.size() <= hash_len / 2) {
```
Aksha1812, line 96: I think having only one field is pretty common . in this case when hash_len == 1, hash_len/2 == 0. Can we do a simple comparison taking only one field hash and compare performance between both options, so we handle the edge case as well.

Aksha1812:
Also it would be nice to see performance numbers against the benchmarks we setup https://github.com/valkey-io/valkey-search/blob/main/.github/benchmark_configs/fts-benchmarks-arm.json . To verify if we do not accidently cause performance degradation for a particular scenario. If we don't want to do that maybe we could just look at the benchmark workflows after merging to confirm above.

src/query/response_generator.cc
KarthikSubbarao: We are using VALKEYMODULE_OPEN_KEY_NOEXPIRE already in the code base.

Previously, we used `NOEFFECTS` only on the first call to the VM_Openkey, the second call to VM_Openkey used `VALKEYMODULE_OPEN_KEY_NOEXPIRE `.


Now, we just got rid of the first call. 

So, it is the same with regard to the matched keys (LRU/LFU time,keystats) - nothing new

KarthikSubbarao:
>  Also it would be nice to see performance numbers against the benchmarks we setup https://github.com/valkey-io/valkey-search/blob/main/.github/benchmark_configs/fts-benchmarks-arm.json . To verify if we do not accidently cause performance degradation for a particular scenario. If we don't want to do that maybe we could just look at the benchmark workflows after merging to confirm above.


The scenarios which matter are varying the number of fields (from single to several) and using CONTENT. This is already tested and you can read it from the PR description

src/attribute_data_type.cc
KarthikSubbarao: This is in the PR description already (`RETURN == all fields`). The part to remember is that performance for FetchAllrecords was bad when we were just scanning to find the few fields we need to return. When we need to fetch all the fields (e.g. doc with only 1 field), using  FetchAllrecords works fine. Benchmarks confirm that

src/attribute_data_type.cc
KarthikSubbarao: Good question. Need to check the whether it is ever possible for hash field names to contain such values. If it is not possible, then we should be good

KarthikSubbarao: Checked...

the identifiers here come from `std::string` objects in the schema field definitions, which are populated from FT.CREATE parsing. id.data() is safe with HASH_CFIELDS and this is used today already with the GetRecord api

src/attribute_data_type.cc
KarthikSubbarao: Yes, ctx and key are  unused now. However, this is a virtual method override, so the signature must match the base class declaration. Can't remove them without changing the interface in attribute_data_type.h and all other implementations (like the JSON variant).

Aksha1812: Ah. Thanks for checking

Aksha1812 (approved):
All comments addressed . Approving.

src/query/response_generator.cc
BCathcart: I'm re-considering this choice. By using NOEXPIRE we are ignoring these:

```
VALKEYMODULE_OPEN_KEY_NOTOUCH - Avoid touching the LRU/LFU of the key when opened.
VALKEYMODULE_OPEN_KEY_NONOTIFY - Don't trigger keyspace event on key misses.
VALKEYMODULE_OPEN_KEY_NOSTATS - Don't update keyspace hits/misses counters.
```

Touching the LRU/LFU seems fine, but the other two I'm not so sure

1. Triggering a keyspace event feels a bit weird. The customer didn't explicitly try to access a key that didn't exist. There would have been a race between a asynchronous search and keyspace delete and we now drop the key from the results. Is the keyspace notification providing useful information that the customer can act on? I don't feel like it is.
2. Updating the keyspace hits/misses counter also feels like we could be muddying the meaning of these counters if today they always represent explicit customer access attempts.

BCathcart: I'll merge and create a follow-up PR

BCathcart (approved):

BCathcart:
Merging but I'll follow-up on https://github.com/valkey-io/valkey-search/pull/1045#discussion_r3358987920

src/query/response_generator.cc
KarthikSubbarao: > Triggering a keyspace event feels a bit weird. The customer didn't explicitly try to access a key that didn't exist. There would have been a race between a asynchronous search and keyspace delete and we now drop the key from the results. Is the keyspace notification providing useful information that the customer can act on? I don't feel like it is.
Updating the keyspace hits/misses counter also feels like we could be muddying the meaning of these counters if today they always represent explicit customer access attempts.


```
Extra flags that can be pass to the API under the mode argument:
VALKEYMODULE_OPEN_KEY_NOTOUCH - Avoid touching the LRU/LFU of the key when opened.
VALKEYMODULE_OPEN_KEY_NONOTIFY - Don't trigger keyspace event on key misses.
VALKEYMODULE_OPEN_KEY_NOSTATS - Don't update keyspace hits/misses counters.
VALKEYMODULE_OPEN_KEY_NOEXPIRE - Avoid deleting lazy expired keys.
VALKEYMODULE_OPEN_KEY_NOEFFECTS - Avoid any effects from fetching the key.
```

We were always using the other flags on the keys returned in the past.

This is the normal pattern used by other standard commands like GET/HGETALL, etc - where they update keyspace counters, and also have keyspace events for misses.

Removing the two would be a breaking change and we should not do this in a patch.

But also, IMO, it makes sense to continue following the convention of what commands in Valkey do unless there is a reason why it does not make sense to do so in Search Module commands. But before a PR, we should have a discussion about this

BCathcart: Ok I thought it was something we changed in 1.2. I agree with the hesitation then.

In my opinion, returning keys from a search command is inherently different than a user explicitly operating on a given key, and combining the two obscures the meaning of these metrics. If a customer requests a key and it doesn't exist, it's the classic example of a cache miss and they can use the keyspace notifications to take some action on it. If a search command finds a key, and then that key is deleted by the time it gets to the main thread, what is that really telling the customer? It just tells them about the timing of a key deletion in relation to when their search ran in the background. A search request is saying "find me all the keys with these characteristics" which is much different than "fetch me this particular key" IMO. But anyways, I'm not sure how important this is to customers and there were talks of creating the responses in the background with shared memory at which point this issue goes away anyways.

allenss-amazon: IMO the question of whether a key in a search result ends up affecting the LRU/LFU is likely application specific, i.e., some apps might want it to update the LRU/LFU, other's might want not to, while others perhaps don't care. 

It might get more complicated when you look at the different cases of return values, i.e., NOCONTENT, etc. We aren't currently taking advantage of the ability to generate RETURN values from background threads -- but I don't want to block that ability as it seems like a useful optimization (when possible to honor it).

If there's no clear requirement to update the LRU/LFU, then I'd like to avoid it. But probably more important than that is consistency. IMO, either we ALWAYS update the bits or we never update the bits -- but it should be consistent regardless of the options selected.

So, I come down on the never modify the bits case.

allenss-amazon: In any case, we can always add some future option flag to explicitly control whether we do or don't update the flags....

Issue #1042, opened 2026-05-16
[BUG] coordinator_bytes_in field incorrect.


allenss-amazon:
**Describe the bug**

In [src/coordinator/client.cc](https://github.com/valkey-io/valkey-search/pull/1040#discussion_r3251395988):
>            Metrics::GetStats().coordinator_bytes_in.fetch_add(
-              args->response.ByteSizeLong(), std::memory_order_relaxed);
+              args->response->ByteSizeLong(), std::memory_order_relaxed);

 coordinator_bytes_in measures post-drain response size
ByteSizeLong() is called after callback(s, *args->response) returns. The callback routes into SearchPartitionResultsTracker::HandleResponse, which loops on response.mutable_neighbors()->ReleaseLast() until all neighbor entries are removed from the repeated field. By the time ByteSizeLong() runs, the neighbors field is empty, so the metric captures only the residual fields (e.g. total_count) and omits the bulk of the payload — the neighbors themselves. This under-counting existed before this PR too, but now that the arena optimization makes large result sets the intended hot-path, the discrepancy will be proportionally larger. Consider capturing ByteSizeLong() before invoking the callback, or storing the pre-drain byte count alongside the response.

**To Reproduce**
Steps to reproduce the behavior. Ideally, a script to reproduce the problem.

**Expected behavior**

**Environment**

Additional information to reproduce the failure

- Valkey Search Module Version
- Valkey Core Version 

**Additional context**

Frank-Gu-81:
Fix is merged in #1046 . Closing this issue.

PR #1046, opened 2026-05-18
Fix coordinator bytes-in counting before response mutation

bandalgomsu:
Capture SearchIndexPartition response size before invoking callbacks so coordinator_bytes_in reflects the full inbound payload.

issue : #1042


testing/coordinator/client_test.cc
```diff
@@ -8,35 +8,219 @@
...
+  stub->async_stub()->search_response = response;
+  ClientImpl client(nullptr, "test_address", std::move(stub));
+
+  bool callback_called = false;
+  client.SearchIndexPartition(
+      std::move(request),
+      [&](grpc::Status status, SearchIndexPartitionResponse &resp) {
+        EXPECT_TRUE(status.ok());
+        callback_called = true;
+        resp.clear_neighbors();
```
Aksha1812, line 213: This unit test mimics the drain instead of using SearchPartitionResultsTracker::HandleResponse . it is fine for a unit test . bbut having seeing a integration test would be good to verify the info field works correctly

Aksha1812 (approved):
Changes look good. Approving. integration test would be a nice to have . @allenss-amazon to do final review & merge

Frank-Gu-81:
Fixes #1042 .

PR #1056, opened 2026-05-22
Make text tokenization, fuzzy search, and query parsing multi-byte aware

VoletiRam:
Tokenization, fuzzy search DP, and query token loops now operate on Unicode code points instead of raw bytes:

- Lexer PunctuationSet keyed by code point (was bitset<256> indexed by byte). Custom multi-byte PUNCTUATION strings work; previously stray bytes were marked as punctuation and shredded Arabic/CJK words.
- min_stem_size compared against code-point count, not byte length.
- Fuzzy search DP matrix sized in code points; é→è costs 1 edit. Edges containing partial UTF-8 sequences are buffered across the next edge so multi-byte code points split by radix-tree compression decode correctly.
- FilterParser quoted/unquoted token loops decode code points.

Adds src/utils/utf8_iterator.h as the single source of truth for UTF-8 decoding (Next, ExpectedLen, IsAscii). Scanner delegates to it, dropping the transitive valkey_module.h dependency.

Tests added:
- testing/text_test.cc: fuzzy across multi-byte edge split, fuzzy code-point distance.
- testing/lexer_test.cc: multi-byte custom punctuation.
- integration/test_fulltext.py: %Hola% updated to match code-point fuzzy semantics.


KarthikSubbarao:
Can we perform a microbenchmark on this PR focusing on the parsing (query or ingestion) to see how much of an impact this has?

Before making major updates, at this point in the project, it will be useful to understand what impact it might have

src/utils/utf8_iterator.h
```diff
@@ -0,0 +1,132 @@
...
+// On invalid bytes returns the raw byte value with byte_len=1.
+// Scanner::NextUtf8()/PeekUtf8() delegate to this class.
+class Utf8Iterator {
+ public:
+  // Result of decoding one code point.
+  struct Result {
+    uint32_t codepoint;  // Unicode scalar value (e.g., 0x00E9 for 'é')
+    uint8_t byte_len;    // Bytes consumed: 1 for ASCII or invalid, 2-4 for
+                         // valid multi-byte sequences.
+  };
```
allenss-amazon: Rather than returning these two items, and requiring the caller to save them, I'd rather see an iterator that behaves more like what existing iterators do. Specifically, the paradigm for iterators is that they are references to a position within their container and you can use the iterator just like a pointer (by overriding operator * and ->). 

Here, the "result" is really associated with the current position. It should be "inside" of the iterator and you would use the iterator to reference it. You see this pattern with stl iterators (see the map iterator, which is similar to the current situation).

I think the iterator should contain a "Result" inside of itself. Then the operator * and -> should generate const references to that result. So from a notation:
```
UTF8Iterator x;
...
if (x->codepoint) ....

len += x->byte_len;

x.Next();
```
Are more natural and make the code easier to read.

src/utils/utf8_iterator.h
```diff
@@ -0,0 +1,132 @@
...
+  // (U+D800..U+DFFF), and code points > U+10FFFF. Invalid sequences advance
+  // one byte and return the raw lead byte with byte_len=1.
+  //
+  // UTF-8 byte patterns:
+  //   0xxxxxxx              (0x00..0x7F)  → 1 byte,  U+0000..U+007F  (ASCII)
+  //   110xxxxx 10xxxxxx     (0xC2..0xDF)  → 2 bytes, U+0080..U+07FF
+  //   1110xxxx 10xxxxxx×2   (0xE0..0xEF)  → 3 bytes, U+0800..U+FFFF
+  //   \ surrogates 11110xxx 10xxxxxx×3   (0xF0..0xF4)  → 4 bytes,
+  //   U+10000..U+10FFFF
+  Result Next() {
```
allenss-amazon: In the spirit of the comment above, this is the equivalent of "operator++". However, I'm fine with using Next. But since I've proposed that the "Result" be contained inside of the iterator itself, then I would use the return value to indicate "Done". Thus the coding becomes much clearer:
```
UTF8Iterator x(...);

do {
    code.push_back(x->codepoint);
    pos += x->byte_len;

} while (x.Next());

// Or another form:

while (x.Next()) {
   code.push_back(x->codepoint);
   pos += x->byte_len;
}
```

Much cleaner.

src/utils/utf8_iterator.h
```diff
@@ -0,0 +1,132 @@
...
+
+  // Returns true when there are no more bytes to decode.
+  bool Done() const { return pos_ >= text_.size(); }
+
+  // Current byte offset into the underlying string_view.
+  size_t pos() const { return pos_; }
+
+  // Decode and advance past one code point. Must only be called when !Done().
+  //
+  // Rejects RFC 3629 violations: overlong encodings, surrogate code points
```
allenss-amazon: What is a "Rejection", specifically, what is the state of the iterator after a Rejection? Can I detect it? What are the guarantees after a Rejection?

allenss-amazon (commented):
The UTF8 iterator needs a more exhaustive test. There are lots of bit masks and comparison that we're not sure are correct at the edge cases. Are we certain that with invalid combinations that it behaves well (doesn't address off the end of the string_view) ?

src/utils/utf8_iterator.h
VoletiRam: Ack, will address it

src/utils/utf8_iterator.h
VoletiRam: If the iterator encountered a byte sequence that didn't match any valid UTF-8 pattern, it would silently return the raw lead byte as the code point with `byte_len = 1` and advance 1 byte. This was called a "rejection". The multi-byte interpretation was rejected and the byte was treated as opaque data. 

Note that all the callers validate text via `Lexer::IsValidUtf8()` before constructing this iterator at tokenize(). The above change is defensive code to prevent potential corruption if an invalid utf ever slips through despite the upstream check.


I plan to handle this with `DCHECK(false)` plus keeping the `++pos_` advance to prevent an infinite loop in release builds

src/commands/filter_parser.cc
```diff
@@ -580,12 +580,12 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseQuotedTextToken(
...
-      if (cp == '\\') continue;  // Don't break on backslash
-      if (lexer.IsPunctuation(cp)) break;
-      processed_content.append(expression_.data() + pos_, byte_len);
-      pos_ += byte_len;
+      cp_it.Next();
+      if (cp_it.codepoint() == '"') break;
+      if (cp_it.codepoint() == '\\') continue;  // Don't break on backslash
+      if (lexer.IsPunctuation(cp_it.codepoint())) break;
+      processed_content.append(expression_.data() + pos_, cp_it.byte_len());
+      pos_ += cp_it.byte_len();
```
allenss-amazon: Why isn't the Utf8Iterator integrated into the Parser? the parser has it's own stream of characters with it's own position, next and done, etc.. It seems to me that logically the FilterParser itself should be enhanced to provide the bookkeeping for utf8 characters. It's possibly more changes, but they'll be easier to maintain over the long run.

VoletiRam:
Closing the current PR as we moved our changes to the fork branch : https://github.com/VoletiRam/valkey-search/pull/6

PR #1058, opened 2026-05-22
Run CI workflows on all release branches

Aksha1812:
Add 1.0, 1.1, and 1.2 release branches to CI triggers so that pushes and PRs targeting them get the same CI coverage as main. Previously only main was covered, which allowed build breaks on release branches to go undetected.


.github/workflows/integration_tests-asan.yml
```diff
@@ -4,11 +4,15 @@ on:
   pull_request: # This triggers the workflow for pull requests
     branches:
       - main    # Run tests for pull requests targeting the "main" branch
-      - fulltext
+      - '1.0'
```
KarthikSubbarao: Nit: The number of yml files will keep increasing. If we start a template where all yml files read the versions from a single file (that contains versions listed in new lines, command separated, etc), that will reduce changes when we want to add a new release version and avoid missing some files accidentally

.github/workflows/macos.yml
```diff
@@ -1,20 +1,30 @@
...
+      should_run: ${{ steps.check.outputs.run }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: check
+        run: |
+          if grep -qx "${{ github.ref_name }}" .github/release-branches.txt; then
+            echo "run=true" >> "$GITHUB_OUTPUT"
+          fi
+
   cmake-tests-macos:
```
KarthikSubbarao: Instead of all of this, can we do the following?


Add a step after the checkout step.
```
    - name: Verify Branch
      run: grep -qx "${{ github.ref_name }}" .github/release-branches.txt || exit 0
```

allenss-amazon (approved):

Aksha1812:
@allenss-amazon @BCathcart can we add this PR to these three boards with Status "To be backported". I dont have the permissions to . valkey-ci-agent should raise a backport PR 

- 1.0 - https://github.com/orgs/valkey-io/projects/36/views/1
- 1.1 - https://github.com/orgs/valkey-io/projects/34/views/1
- 1.2 - https://github.com/orgs/valkey-io/projects/58/views/1

PR #1060, opened 2026-05-23
Fix race in UnTrackedKeys ingestion access vs negative queries

allenss-amazon:
Race discovered by claude. Extremely difficult to repro.

The problem is mutation of UnTrackedKeys by the main thread. This violates the assumption that Index data structures are immutable during a read-phase of the TimeSlicedMutex. Here, if a query is running with a negated EntriesFetcher, the entries fetcher will retain an iterator to the UnTrackedKeys (absl::flat_hash_set<InternedStringPtr>) in the reader thread. If a key is ingested in the foreground that satisfies this early optimization code, then it will insert into the UnTrackedKeys which violates the assumption of immutability that applies to the iteration in the negated EntriesFetcher case.

The actual implementation of flat_hash_set will usually allow this violation to proceed and most of the time it will be benign as the mutated key will either appear or not -- and that will get addressed by the validation logic later. However, IFF the mutation by the main thread causes a rehash (to expand the table), THEN the iterator becomes wildly invalid and will likely segfault.

This probably needs to be backported to 1.2 and 1.0.

There are two fixes. Either have the mainthread acquire a reader timesliced mutex OR drop the optimizations. I've elected to drop the optimization -- it doesn't seem very important.  

I was working on completely reimplementing tracked/untracked logic to save space when this issue was diagnosed by claude.


src/index_schema.cc
```diff
@@ -586,13 +586,6 @@ void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
     vmsdk::UniqueValkeyString record = VectorExternalizer::Instance().GetRecord(
         ctx, attribute_data_type_.get(), key_obj.get(), key_cstr,
         attribute.GetIdentifier(), is_module_owned);
-    // Early return on record not found just if the record not tracked.
-    // Otherwise, it will be processed as a delete
-    if (!record && !attribute.GetIndex()->IsTracked(interned_key) &&
-        !InTrackedMutationRecords(interned_key, attribute.GetIdentifier())) {
-      attribute.GetIndex()->UnTrack(interned_key);
-      continue;
-    }
```
BCathcart, line 595: I doubled checked the tracked/untracked bookkeeping and I don't see any change caused by removing this

BCathcart (approved):

PR #1061, opened 2026-05-24
Fix shutdown hang with many worker threads

allenss-amazon:
Hang happened because of ThreadPool::JoinWorkers was holding a mutex while calling pthread_join as part of the shutdown logic. The deadlock would happen when the worker thread noticed the shutdown request and moved itself from threads_ array to pending_join_threads_ array. 

The fix is two fold. One, remove the code in JoinWorkers that waits with the lock held. Instead, poll the threads_ array until it's empty. this requires that all paths that shutdown a workerthread result in moving that thread from threads_ array into the pending_join_threads_ array. the graceful shutdown path already did that, but the stop_mode_ path didn't and now is updated to do the same upon shudown, i.e., the join behavior is now the same regardless of how the workerthread stops.


allenss-amazon:
supercede by a redesign

PR #1066, opened 2026-05-26
Add BM25STD Scorer and Scoring Session

boda26:
## Summary

  Adds the math + per-query collection layer for BM25STD scoring.
  Algorithm-agnostic `Scorer` / `ScoringStats` interfaces so TFIDF
  (and later tag / numeric scorers) can plug in without changes.
  Driving the session from the predicate tree is a follow-up.

  ## Changes

  - `src/indexes/scoring/scoring_stats.h` — `ScoringStats` base +
    `Bm25StdStats` / `TfidfStats` subclasses
  - `src/indexes/scoring/scorer.{h,cc}` — `Scorer` interface,
    default `CombineGroup`
  - `src/indexes/scoring/bm25std_scorer.{h,cc}` — Okapi BM25
    (k1=1.2, b=0.75). Uses a bit-pattern `IsInf` because
    `-ffast-math` makes `std::isinf` unreliable.
  - `src/indexes/scoring/scoring_session.{h,cc}` — per-query
    collector. `RecordLeaf` / `EnterGroup` / `ExitGroup` / `Rank`.
  - Wired as `scoring` static lib in `src/indexes/CMakeLists.txt`.

  ## Tests

  New `scoring_test` binary (26 tests):

  - `testing/scoring/scoring_test_data.h` — shared 8-doc, 4-term
    corpus designed to isolate TF saturation, length norm, IDF
    differentiation, tie-break, partial-match admission.
  - `testing/scoring/bm25std_scorer_test.cc` — scorer math,
    edge cases, death tests for invariant violations.
  - `testing/scoring/scoring_session_test.cc` — all query shapes
    the corpus unlocks (single leaf, AND, OR, weighted/nested
    groups, mixed admission, ties).


src/indexes/scoring/scoring_stats.h
```diff
@@ -0,0 +1,89 @@
...
+//   positions  - positional offsets of `term` within the document, sorted
+//                ascending; empty when SLOP is not used
+struct TfidfStats : ScoringStats {
+  std::unique_ptr<ScoringStats> Clone() const override {
+    return std::make_unique<TfidfStats>(*this);
+  }
+
+  uint32_t norm = 0;
+  absl::Span<const uint32_t> positions;
+};
```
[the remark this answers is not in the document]
boda26: positions is for TFIDF only, which is out of scope of this PR. Will fix this in the future TFIDF implementation.

src/indexes/scoring/bm25std_scorer.h
```diff
@@ -0,0 +1,49 @@
...
+// k1 = 1.2, b = 0.75 — aligned with Redis 8.6 / OpenSearch / Wikipedia
+// standard.
+//
+// Expects Bm25StdStats as the ScoringStats subtype. Mismatched subtypes
+// trigger a DCHECK in debug builds.
+class Bm25StdScorer : public Scorer {
+ public:
+  static constexpr std::string_view kName = "BM25STD";
+  static constexpr float kK1 = 1.2f;
+  static constexpr float kB = 0.75f;
```
Frank-Gu-81, line 29: I believe redis's returned score is in double. Since score is calculated with multiple multiplications, maybe we should store the constants in double as well.

src/indexes/scoring/scorer.h
```diff
@@ -0,0 +1,82 @@
...
+// scorers DCHECK this contract.
+class Scorer {
+ public:
+  virtual ~Scorer() = default;
+
+  // Identity string surfaced by FT.SEARCH SCORER and EXPLAINSCORE.
+  virtual std::string_view Name() const = 0;
+
+  // Concrete-class identity for dispatch.
+  virtual ScorerType Type() const = 0;
```
Frank-Gu-81: nit: is there a specific reason we need to have both name and type as identifiers?

boda26: Yes, the name can be used in EXPLAINSCORES, and type is used for internal identification.

src/indexes/scoring/bm25std_scorer.h
boda26: Right now we are storing score as float, I think internally in the calculation process we should keep consistent on the data type. We should either (1) use all variables and constants as double (2) use all variables and constants as float, and cast to double when printing

testing/CMakeLists.txt
```diff
@@ -178,5 +178,14 @@ target_link_libraries(text_index_test PRIVATE testing_common_base)
...
+# 8. Scoring Test Suite
+add_executable(scoring_test
+    ${CMAKE_CURRENT_LIST_DIR}/scoring/bm25std_scorer_test.cc
+    ${CMAKE_CURRENT_LIST_DIR}/scoring/scoring_session_test.cc)
+target_include_directories(scoring_test PUBLIC ${CMAKE_CURRENT_LIST_DIR})
+target_link_libraries(scoring_test PRIVATE testing_common_base)
+target_link_libraries(scoring_test PRIVATE scoring)
+finalize_test_flags(scoring_test)
+
 # 8. Expr Test Suite
```
zackcam: Should this now be 9?

boda26: Yes, fixing in next commit

PR #1071, opened 2026-05-28
chore(multi-language support): add kRelease13 and kRelease14

neerajr0:
## Motivation
Multi-language FTS needs to be gated to version 1.4.

## Summary of Changes
- Added new kRelease13 and kRelease14 in version.h. Multi-language FTS will be introduced in version 1.4, also added version 1.3 for continuity from 1.2.
- Added logic to return version 1.4 in `IndexSchema::GetMinVersion()` if the schema uses a text index and specifies a non-English language

## Testing
Added unit tests in `index_schema_test.cc` to assert that correct versions are returned for different index/language/has_db_num combinations.


testing/index_schema_test.cc
```diff
@@ -2633,4 +2635,114 @@ TEST_F(IndexSchemaScoreFieldTest, FallsBackToDefaultScoreWhenFieldMissing) {
...
+    GetMinVersionTests, GetMinVersionTest,
+    ValuesIn<GetMinVersionTestCase>({
+        {
+            .test_name = "EnglishTextIndex",
+            .index_name = "english_text_idx",
+            .language = data_model::LANGUAGE_ENGLISH,
+            .attribute_type = IndexAttributeType::kText,
+            .db_num = 0,
+            .expected_version = kRelease12,
+        },
```
VoletiRam, line 2694: I like the regression tests for other cases which ensures we don't break on non-multi language code paths.

VoletiRam (approved):

src/version.h
```diff
@@ -98,6 +98,16 @@ constexpr vmsdk::ValkeyVersion kRelease11(1, 1, 0);
 //
 constexpr vmsdk::ValkeyVersion kRelease12(1, 2, 0);
 
+//
+// Release 1.3
+//
+constexpr vmsdk::ValkeyVersion kRelease13(1, 3, 0);
+
+//
```
VoletiRam: I would've left adding the 1.3 constant in this PR as release owner when bumping the version would usually do it. There is no issue adding the constant though as we are no bumping version.

VoletiRam (commented):
We need to guard the changes of multi language to only enable in version 1.4.

src/version.h
neerajr0: Makes sense, I figured I'd add it here for continuity of the version constants so we don't have 1.4 without 1.3.

src/version.h
```diff
@@ -98,6 +98,16 @@ constexpr vmsdk::ValkeyVersion kRelease11(1, 1, 0);
 //
 constexpr vmsdk::ValkeyVersion kRelease12(1, 2, 0);
 
+//
+// Release 1.3
+//
+constexpr vmsdk::ValkeyVersion kRelease13(1, 3, 0);
```
BCathcart: If 1.3 doesn't contain metadata changes, I don't believe we'll need this. In that case we can either use 1.3.0 metadata version for valkey search version 1.4 or simply skip 1.3.0 and go straight to 1.4.0 if we want to keep it in lockstep with the module version

Based on the comment above, sounds like it was intended that they may diverge:
>Generally, they correspond to releases of ValkeySearch, but minor
// releases may not increment the version if no metadata changes were made.

neerajr0:
Closing - preferring merge into fork before direct merge into upstream for development.

PR #1074, opened 2026-05-28
Add ILanguageProcessor interface, SnowballProcessor class and multi-language enum for multi-language FTS support. No existing code paths modified.

linbran5123:
# Summary

Introduces the foundational abstractions for multi-language FTS support **without** modifying **any** existing code paths. This is the interface layer that **subsequent PRs** will wire into the pipeline and build against:

- **ILanguageProcessor interface:** pluggable abstraction that decouples language-specific behavior (stemming, tokenization) from the Lexer. Enables adding new language families without modifying the core pipeline.
- **SnowballProcessor:** concrete implementation for 12 Snowball-stemmed languages, extracted from Lexer's existing stemmer logic.
- **Language enum + parser map:** adds all 17 planned supported languages to the proto definition and FT.CREATE parser.
- **Factory method:** `ILanguageProcessor::Create(language)` returns the appropriate processor for a given language.

# Test plan
- Existing unit tests pass (no behavioral changes)
- Existing integration tests pass (no code paths modified)

### DIFF
--- a/src/commands/filter_parser.cc
+++ b/src/commands/filter_parser.cc
@@ -252,13 +252,29 @@ absl::StatusOr<double> FilterParser::ParseNumber() {
   }
   std::string number_str;
   double value;
-  int multiplier = Match('-', false) ? -1 : 1;
-  while (!IsEnd() && (std::isdigit(Peek()) || Peek() == '.')) {
+  if (!IsEnd() && Peek() == '-') {
     number_str += expression_[pos_++];
   }
-  if (absl::AsciiStrToLower(number_str) != "nan" &&
-      absl::SimpleAtod(number_str, &value)) {
-    return value * multiplier;
+  bool exponent_seen = false;
+  bool exponent_sign_allowed = false;
+  while (!IsEnd()) {
+    const auto next = Peek();
+    if (std::isdigit(static_cast<unsigned char>(next)) || next == '.') {
+      number_str += expression_[pos_++];
+      exponent_sign_allowed = false;
+    } else if ((next == 'e' || next == 'E') && !exponent_seen) {
+      number_str += expression_[pos_++];
+      exponent_seen = true;
+      exponent_sign_allowed = true;
+    } else if ((next == '+' || next == '-') && exponent_sign_allowed) {
+      number_str += expression_[pos_++];
+      exponent_sign_allowed = false;
+    } else {
+      break;
+    }
+  }
+  if (!number_str.empty() && absl::SimpleAtod(number_str, &value)) {
+    return value;
   }
   return absl::InvalidArgumentError(
       absl::StrCat("Invalid number: ", number_str));
--- a/testing/filter_test.cc
+++ b/testing/filter_test.cc
@@ -202,6 +202,53 @@ INSTANTIATE_TEST_SUITE_P(
                                        "}\n",
         },
         {
+            .test_name = "numeric_happy_path_3",
+            .filter = "@num_field_2.0:[-1.5 2.5] @num_field_1.5:[-1.0 2.0]",
+            .create_success = true,
+            .evaluate_success = true,
+            .expected_tree_structure = "AND{\n"
+                                       "  NUMERIC(num_field_2.0)\n"
+                                       "  NUMERIC(num_field_1.5)\n"
+                                       "}\n",
+        },
+        {
+            .test_name = "numeric_happy_path_scientific_notation",
+            .filter = "@num_field_2.0:[1.0 1e2] @num_field_1.5:[1e0 1.5]",
+            .create_success = true,
+            .evaluate_success = true,
+            .expected_tree_structure = "AND{\n"
+                                       "  NUMERIC(num_field_2.0)\n"
+                                       "  NUMERIC(num_field_1.5)\n"
+                                       "}\n",
+        },
+        {
+            .test_name = "numeric_happy_path_scientific_notation_negative",
+            .filter = "@num_field_2.0:[1e-2 1e+2] "
+                      "@num_field_1.5:[-1e2 1.5] "
+                      "@num_field_1.5:[-1e-2 1.5]",
+            .create_success = true,
+            .evaluate_success = true,
+            .expected_tree_structure = "AND{\n"
+                                       "  NUMERIC(num_field_2.0)\n"
+                                       "  NUMERIC(num_field_1.5)\n"
+                                       "  NUMERIC(num_field_1.5)\n"
+                                       "}\n",
+        },
+        {
+            .test_name = "numeric_invalid_embedded_minus",
+            .filter = "@num_field_1.5:[1-2 2.0]",
+            .create_success = false,
+            .create_expected_error_message =
+                "Expected space or `,` between start and end values of a "
+                "numeric field. Position: 17",
+        },
+        {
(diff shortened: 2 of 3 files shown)


src/indexes/text/snowball_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+#include "absl/strings/ascii.h"
+#include "libstemmer.h"
+#include "src/indexes/text/unicode_normalizer.h"
+
+namespace valkey_search::indexes::text {
+
+namespace {
+
+constexpr absl::string_view kDefaultPunctuation =
+    ".,<>{}[]\"':;!@#$%^&*()-+=~/ ";
```
neerajr0: This looks different from `kDefaultPunctuation` currently in `ft_create_parser.h`: https://github.com/valkey-io/valkey-search/blob/5ee719c5cee6b8d2b647e2d7b51e57ee084f8c65/src/commands/ft_create_parser.h#L27.

src/indexes/text/language_processor.cc
```diff
@@ -0,0 +1,45 @@
...
+    data_model::Language language) {
+  switch (language) {
+    case data_model::LANGUAGE_CHINESE:
+    case data_model::LANGUAGE_JAPANESE:
+    case data_model::LANGUAGE_KOREAN:
+    case data_model::LANGUAGE_POLISH:
+    case data_model::LANGUAGE_VIETNAMESE:
+      // TODO: ICUSegmenter class to be implemented
+      // return std::make_shared<ICUSegmenter>(language);
+      return nullptr;
```
neerajr0: Would it make more sense to add CJKPV later when we add the ICUSegmenter class? That'll remove the need for the `nullptr` placeholder so we can focus on ILanguageProcessor and SnowballProcessor in this PR.

src/indexes/text/snowball_processor.h
```diff
@@ -0,0 +1,52 @@
...
+#include "src/index_schema.pb.h"
+#include "src/indexes/text/language_processor.h"
+
+struct sb_stemmer;
+
+namespace valkey_search::indexes::text {
+
+class SnowballProcessor : public ILanguageProcessor {
+ public:
+  explicit SnowballProcessor(data_model::Language language);
```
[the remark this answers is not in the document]
neerajr0: I think this is a reasonable suggestion, something like

```
class SnowballProcessor : public ILanguageProcessor {
  friend class ILanguageProcessor;  // allows Create() to call the constructor
 private:
  explicit SnowballProcessor(data_model::Language language);
 public:
  // ... interface methods
};
```

src/indexes/text/language_processor.cc
VoletiRam: +1

src/indexes/text/snowball_processor.cc
VoletiRam: Lets use existing and integrate to rather than adding in separate place.

src/indexes/text/snowball_processor.cc
```diff
@@ -0,0 +1,147 @@
...
+    stemmers_[language_] = std::move(stemmer);
+    return raw_ptr;
+  }
+  return it->second.get();
+}
+
+std::string_view SnowballProcessor::DoStemming(absl::string_view word,
+                                               sb_stemmer* stemmer,
+                                               uint32_t min_stem_size) const {
+  if (word.empty() || word.length() < min_stem_size) {
```
VoletiRam: This check is no longer works for multibyte characters which we see in multi-language non-aascii characters. I am fixing it as part of this PR: https://github.com/valkey-io/valkey-search/pull/1056.

src/commands/ft_create_parser.cc
VoletiRam, line 1: Can we add tests to ensure the parsing works for multi language?

src/commands/ft_create_parser.cc
```diff
@@ -214,7 +214,23 @@ static auto default_timeout_ms =
...
+                    {"GERMAN", data_model::LANGUAGE_GERMAN},
+                    {"SPANISH", data_model::LANGUAGE_SPANISH},
+                    {"ITALIAN", data_model::LANGUAGE_ITALIAN},
+                    {"PORTUGUESE", data_model::LANGUAGE_PORTUGUESE},
+                    {"RUSSIAN", data_model::LANGUAGE_RUSSIAN},
+                    {"SWEDISH", data_model::LANGUAGE_SWEDISH},
+                    {"TURKISH", data_model::LANGUAGE_TURKISH},
+                    {"DUTCH", data_model::LANGUAGE_DUTCH},
+                    {"INDONESIAN", data_model::LANGUAGE_INDONESIAN},
+                    {"ARABIC", data_model::LANGUAGE_ARABIC},
```
VoletiRam: The `kLanguageByStr` additions expose non-English `FT.CREATE` in any binary with this PR, including 1.3. We need a guard so 1.3 behaves identically to 1.2 for multi-language creation.


My suggestion is to add a dev flag `multi-language-enabled` (default off, tied to `kModuleVersion >= 1.4`) in `valkey_search_options`, then build `kLanguageByStr` conditionally, only populate the non-English entries when the flag is enabled. That way the `ParseParam` lookup for `"FRENCH"` fails through the same code path as 1.2, producing the same error, with no new error message or different failure mode in 1.3. When we're ready to ship in 1.4, flip the flag default or remove the condition.

src/indexes/text/snowball_processor.h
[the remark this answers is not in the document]
linbran5123: Fixing

src/indexes/text/language_processor.cc
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+
+#include "src/indexes/text/language_processor.h"
+
+#include "src/indexes/text/snowball_processor.h"
+
+namespace valkey_search::indexes::text {
+
+std::shared_ptr<ILanguageProcessor> ILanguageProcessor::Create(
```
VoletiRam: Lets remove predecessor I* and keep just LanguageProcessor as we don't follow this pattern in our code base.

src/indexes/text/snowball_processor.cc
linbran5123: Fixing

src/indexes/text/language_processor.cc
linbran5123: Fixing

src/commands/ft_create_parser.cc
linbran5123: Fixing; will grab version declaration from Neeraj's PR as well

src/indexes/text/language_processor.cc
linbran5123: Ack, fixing

src/commands/ft_create_parser.cc
linbran5123: 1. The parser recognizes the new languages in `kLanguageByStr`, but the version guard I'm putting in immediately rejects them since `kMultiLanguageEnabled` is false. 
2. The SnowballProcessor isn't wired into the pipeline yet (Task 8), and the non-English Snowball stemmers aren't compiled yet (Task 6 — `add_language.sh` hasn't been run). There's no path from user input to the new code, and no binary for non-English stemmer tests to run on.

I can add both unit and integration tests for Task 8. I can put my name down for that.

PR #1083, opened 2026-05-30
FT.Hybrid Initial Implementation

allenss-amazon:
FT.HYBRID is implemented. 

FT.HYBRID is implemented as a special case of the more generic multi-arm search case.

A new gRPC call is created to support multi-arm searches. Specifically, This new gRPC call consists of an array of searches. Each element of the array is an existing search request. The new call parses the array and creates an separate work item for each of the arms. Each arm performs it's index query in parallel and the results are transmitted. Once the results are combined, they are fed into the aggregation framework.

This machinery ensures that individual query operations are re-using the entire query chain. Also the existing aggregation logic is used, unchanged. Future enhancements of new query operators or new aggregation stages will automatically be available in FT.HYBRID.

Only a very limited amount of testing has been done. As until scoring is completed, the compatibility framework is unavailable.


yairgott:
Based on the PR description, the name `FT.Hybrid` doesn't seem to align well with the introduced capability. An alternative which makes better sense to me could be `FT.AggregateMulti`. On this regards, can you please elaborate why can't  the `FT.Aggregate` supported syntax can't be enhanced to support this new capability?

allenss-amazon:
FT.HYBRID is an existing Redisearch command. This PR is a superset of what's implemented for that command. 

I fully support extending the capability along the lines you describe and I'm looking at this as the foundation of next-generation re-ranking capabilities. So I think deferring the extended capability makes the most sense in order to reduce the API surface going forward.

src/commands/ft.hybrid.json
```diff
@@ -0,0 +1,228 @@
+{
+  "FT.HYBRID": {
+    "acl_categories": [
+      "READ",
+      "SLOW",
+      "SEARCH"
+    ],
+    "history": [
+      [
+        "8.4.4",
```
yairgott, line 10: Does this refer Redis or Valkey version?

src/commands/commands.cc
```diff
@@ -16,6 +16,7 @@
 #include "src/commands/ft_search.h"
 #include "src/metrics.h"
 #include "src/query/fanout.h"
+#include "src/query/multi_search.h"  // for ExecuteCommand<MultiSearchParameters>
```
yairgott, line 19: this comment should be removed.

src/commands/commands.cc
```diff
@@ -173,83 +139,121 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
+absl::Status QueryCommand::DispatchLocalAsync(ValkeyModuleCtx *ctx,
+                                              std::unique_ptr<QueryCommand> cmd,
+                                              vmsdk::ThreadPool *pool) {
+  cmd->blocked_client = vmsdk::BlockedClient(ctx, async::Reply, async::Timeout,
+                                             async::Free, cmd->timeout_ms);
+  cmd->blocked_client->MeasureTimeStart();
+  return query::SearchAsync(std::move(cmd), pool, query::SearchMode::kLocal);
+}
+
+absl::Status QueryCommand::DispatchFanoutAsync(
```
yairgott, line 269: I think this function should reside under the query dir, probably inside the fanout.  Commands should mainly maintain command handling logic/infra and minimize any additional not in scope changes.

src/commands/commands.cc
```diff
@@ -173,83 +139,121 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
+  }
+  cmd->SendReply(ctx, cmd->search_result);
+  ValkeySearch::Instance().ScheduleSearchResultCleanup(
+      [neighbors = std::move(cmd->search_result.neighbors)]() mutable {
+        // destructor runs when lambda completes
+      });
+  return absl::OkStatus();
+}
+
+absl::Status QueryCommand::DispatchLocalAsync(ValkeyModuleCtx *ctx,
```
yairgott, line 260: Same as the previous comment but this should probably reside in fanout.h/cc

src/commands/commands.cc
```diff
@@ -173,83 +139,121 @@ absl::Status QueryCommand::Execute(ValkeyModuleCtx *ctx,
...
+absl::Status QueryCommand::ParseAfterIndex(QueryCommand &cmd,
+                                           vmsdk::ArgsIterator &itr) {
+  VMSDK_RETURN_IF_ERROR(
+      vmsdk::ParseParamValue(itr, cmd.parse_vars.query_string));
+  VMSDK_RETURN_IF_ERROR(cmd.ParseCommand(itr));
+  cmd.parse_vars.ClearAtEndOfParse();
+  return absl::OkStatus();
+}
+
+absl::Status QueryCommand::ExecuteSyncLocal(ValkeyModuleCtx *ctx,
```
yairgott, line 237: I think this function should reside under the query dir, probably inside the search.h/cc.  The commands file should mainly maintain command handling logic/infra and minimize any additional out of scope changes.

src/commands/ft_aggregate_exec.cc
```diff
@@ -413,5 +408,267 @@ absl::flat_hash_map<std::string, GroupBy::ReducerInfo> GroupBy::reducerTable{
...
+
+  query::ProcessNeighborsForReply(
+      ctx, parameters.index_schema->GetAttributeDataType(), neighbors,
+      parameters, vector_identifier);
+
+  return std::make_pair(key_index, scores_index);
+}
+
+// Process a single field value and convert it to the appropriate type
+absl::StatusOr<expr::Value> ProcessFieldValue(
```
yairgott, line 554: A cleaner approach is if the index_base exposes a virtual function by the name of ToExprValue.

src/commands/ft_aggregate_exec.cc
```diff
@@ -413,5 +408,267 @@ absl::flat_hash_map<std::string, GroupBy::ReducerInfo> GroupBy::reducerTable{
...
+    } else {
+      std::string s = absl::StrCat("[", value_view, "]");
+      ValkeyModule_ReplyWithStringBuffer(ctx, s.data(), s.size());
+    }
+  }
+  return true;
+}
+
+// Process the query setup for vector vs non-vector queries and set up indices
+absl::StatusOr<std::pair<size_t, size_t>> ProcessNeighborsForProcessing(
```
yairgott, line 525: Function name not ideal.

src/commands/ft_aggregate_exec.cc
yairgott, line 1: I find the name ft_aggregate_exec very confusing and not indicative of what it actually contains. It sounds to me that its content should reside under the query dir. The part which is specific to the reply, I suggest to maintain it under  query/aggregate_response_generator.h/cc while the rest under query/aggregate.h/cc.

src/commands/ft_aggregate_exec.cc
```diff
@@ -413,5 +408,267 @@ absl::flat_hash_map<std::string, GroupBy::ReducerInfo> GroupBy::reducerTable{
...
+          return expr::Value(std::move(*v));
+        } else {
+          return absl::InvalidArgumentError("Failed to unquote JSON value");
+        }
+      }
+  }
+}
+
+// Create records from neighbors and populate their fields
+absl::Status CreateRecordsFromNeighbors(
```
yairgott, line 575: I assume this function is not changed, just moved, right?

src/commands/ft_aggregate_exec.cc
yairgott, line 525: I assume this function is not changed, just moved, right?

src/commands/ft_aggregate_exec.cc
```diff
@@ -413,5 +408,267 @@ absl::flat_hash_map<std::string, GroupBy::ReducerInfo> GroupBy::reducerTable{
...
+    }
+
+    records.push_back(std::move(rec));
+  }
+
+  return absl::OkStatus();
+}
+
+// Execute all aggregation stages on the record set
+absl::Status ExecuteAggregationStages(AggregateParameters& parameters,
```
yairgott, line 643: I assume this was also not changed just moved, right?

src/commands/ft_aggregate_exec.h
```diff
@@ -8,15 +8,31 @@
...
 
+// Runs the full aggregate reply pipeline on the supplied neighbors:
+//   1. Process query setup (key/score record indices)
+//   2. Convert neighbors -> Record objects (LOAD pass)
+//   3. Execute aggregation stages (APPLY/FILTER/GROUPBY/SORTBY/LIMIT)
+//   4. Emit the response
+//
+// Used by aggregate::AggregateParameters::SendReply and by FT.HYBRID's
+// MultiSearchParameters::SendReply (which feeds the post-fusion neighbor list).
+absl::Status RunAggregatePipeline(ValkeyModuleCtx* ctx,
```
yairgott, line 32: This should reside under the query dir, perhaps under aggregate.h/cc.

src/commands/ft_hybrid_parser.cc
```diff
@@ -0,0 +1,605 @@
...
+                       ": `", next_or.value(), "`"));
+    }
+  }
+
+  // Now that PARAMS (if any) are populated on env.agg->parse_vars.params,
+  // share the params map with each arm so $name resolves identically. Also
+  // run the per-arm pre/post-parse so vector arms (or vector-in-SEARCH per
+  // the Valkey super-set) get their k/ef/query-blob populated.
+  for (auto &arm : env.arms) {
+    arm->parse_vars.params = env.agg->parse_vars.params;
```
yairgott, line 546: I believe you wanted to maintain a reference or a pointer to the `env.agg->parse_vars.params` as otherwise when individual arms increment parameter usage counters, during parameter substitution, they only update their local copy which would later lead to a failure on a subsequent call to VerifyQueryString by ft_search_parser.

src/query/fusion.cc
```diff
@@ -0,0 +1,243 @@
...
+
+    for (size_t rank = 0; rank < cap; ++rank) {
+      const auto& n = (*arm.neighbors)[rank];
+      const std::string key = std::string(n.external_id->Str());
+      auto [it, inserted] = accum.try_emplace(key);
+      if (inserted) {
+        it->second.representative = &n;
+        it->second.per_arm_distance.assign(arms.size(), std::nullopt);
+      }
+      double normalized =
```
yairgott, line 200: This code suggests that smaller distances indicate better matches  which holds for vector metrics. However, for text/secondary search arms where higher scores represent better matches. The formula unconditionally subtracts the normalized distance from 1.0, the best-matching document with the highest text score (10.0) receives a normalized score of 0.0, completely inverting its ranking.

 It seems that we should differentiate whether an arm uses distance-based or score-based ranking before applying normalization.

src/commands/ft_hybrid.cc
```diff
@@ -0,0 +1,363 @@
...
+  for (auto &n : neighbors) {
+    auto it = saved.find(n.external_id->Str());
+    if (it == saved.end()) {
+      continue;
+    }
+    if (!n.attribute_contents.has_value()) {
+      n.attribute_contents.emplace();
+    }
+    for (auto &[k, v] : it->second) {
+      n.attribute_contents->emplace(k, std::move(v));
```
yairgott, line 148: using emplace will silently fail if a user's `YIELD_SCORE_AS` alias collides with an existing database attribute key on the hash. In that scenario, `emplace` does not overwrite the existing key, causing the score alias to be lost and replaced by the raw database field string. Consider using `insert_or_assign` or explicitly overwriting existing keys.

src/commands/ft_hybrid_parser.cc
```diff
@@ -0,0 +1,605 @@
...
+//
+// Note: NOCONTENT is intentionally NOT in this list — we want SEARCH-scope
+// scan to see it and emit a precise rejection ("NOCONTENT is not supported
+// by FT.HYBRID") rather than letting it slip through as a top-level token
+// that subsequent handlers misinterpret.
+bool IsTopLevelKeyword(absl::string_view tok) {
+  return absl::EqualsIgnoreCase(tok, kSearchKw) ||
+         absl::EqualsIgnoreCase(tok, kVsimKw) ||
+         absl::EqualsIgnoreCase(tok, kCombineKw) ||
+         absl::EqualsIgnoreCase(tok, kPolicyKw) ||
```
yairgott, line 71: Including `kPolicyKw` in `IsTopLevelKeyword` causes `ParseSearchClause` and `ParseVsimClause` to prematurely terminate subclause token scanning if `POLICY` is placed inside a `SEARCH` or `VSIM` block. Any subclause tokens following `POLICY` (such as `YIELD_SCORE_AS`) will then be misparsed as invalid outer tokens.
Since `POLICY` is handled by the outer loop at L505, it should be removed from `IsTopLevelKeyword` or handled within subclause scanners.

src/coordinator/server.cc
```diff
@@ -198,48 +209,141 @@ void Service::EnqueueSearchRequest(
...
+    SearchOneArm(
+        context, request->sub_requests(i), sub_resp->mutable_response(),
+        [sub_resp, completion](grpc::Status s) {
+          sub_resp->set_grpc_code(static_cast<uint32_t>(s.error_code()));
+          if (!s.ok()) {
+            sub_resp->set_error_message(s.error_message());
+          }
+          if (completion->remaining.fetch_sub(1, std::memory_order_acq_rel) ==
+              1) {
+            completion->reactor->Finish(grpc::Status::OK);
```
yairgott, line 344: `reactor->Finish(grpc::Status::OK)` is called regardless of whether individual sub-request arms failed due to internal errors.

Encoding errors exclusively inside `sub_resp->grpc_code()` while returning transport-level `grpc::Status::OK` prevents standard gRPC interceptors and node failure counters.

src/query/fusion.cc
yairgott, line 1: The name fusion is to generic IMO. How about rank_fusion or multi_search_fusion?

src/query/fusion.h
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) 2026, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#ifndef VALKEYSEARCH_SRC_QUERY_FUSION_H_
```
yairgott, line 8: Here and else where, at least for new header files, we should use `#pragma once` instead.

src/query/fusion.h
```diff
@@ -0,0 +1,80 @@
...
+// `1 / (rrf_constant + rank+1)` to a hash-map accumulator keyed by external_id.
+// Returns the merged neighbors sorted by descending fused score (stored in
+// Neighbor::distance — overloaded to mean "score, higher=better" for fused
+// results, the OPPOSITE of the vector-distance convention).
+//
+// For each arm whose `score_alias` is set, the surviving fused neighbor's
+// `attribute_contents` map gains a key/value pair: alias -> raw arm distance
+// (formatted as "%.12g"). Docs absent from a given arm have no entry for that
+// arm's alias.
+std::vector<indexes::Neighbor> FuseRRF(std::vector<ArmInput> arms);
```
yairgott, line 53: As this is already inside the namespace fusion, I suggest to rename this to simply RRF. Similarly, I suggest to maintain the following function names: Linear and Function.

src/query/fusion.cc
```diff
@@ -0,0 +1,243 @@
...
+  const indexes::Neighbor* representative = nullptr;  // for attribute_contents
+  double score = 0.0;
+  // Per-arm raw distances. nullopt means "doc absent from this arm". (Using
+  // optional rather than a NaN sentinel because the build uses -ffast-math,
+  // under which NaN comparisons are undefined.)
+  std::vector<std::optional<double>> per_arm_distance;
+};
+
+// Format a double the way the rest of the code formats reply doubles.
+std::string FormatScore(double v) {
```
yairgott, line 42: This can be implemented with: `return absl::StrFormat("%.12g", v);`

src/query/fusion.cc
```diff
@@ -0,0 +1,243 @@
...
+    n.external_id = entry.representative->external_id;
+    n.distance = static_cast<float>(entry.score);
+    // Carry over the representative's attribute_contents as the base. We
+    // shallow-copy via move-fresh; if we want to merge from multiple arms in
+    // the future, this is the place.
+    if (entry.representative->attribute_contents.has_value()) {
+      // Construct a fresh RecordsMap by emplacing copies of the keys/values;
+      // RecordsMapValue holds a UniqueValkeyString, so we re-create those.
+      n.attribute_contents.emplace();
+      for (const auto& [identifier, value] :
```
yairgott, line 88: It would be more efficient to use vmsdk::RetainUniqueValkeyStringinstead of maintaining separate ValkeyModuleString objects. This code block should look something like:

````
for (const auto& [identifier, value] : *entry.representative->attribute_contents) {
  auto id_str = vmsdk::RetainUniqueValkeyString(value.GetIdentifier());
  auto val_str = vmsdk::RetainUniqueValkeyString(value.value.get());
  auto id_view = vmsdk::ToStringView(id_str.get());
  n.attribute_contents->emplace(
      id_view, RecordsMapValue(std::move(id_str), std::move(val_str)));
}
````

File: src/query/fusion.cc [hunk 20]
```diff
@@ -0,0 +1,243 @@
...
+      continue;
+    }
+    const size_t cap =
+        arm.window == 0
+            ? arm.neighbors->size()
+            : std::min(static_cast<size_t>(arm.window), arm.neighbors->size());
+    for (size_t rank = 0; rank < cap; ++rank) {
+      const auto& n = (*arm.neighbors)[rank];
+      const std::string key = std::string(n.external_id->Str());
+      auto [it, inserted] = accum.try_emplace(key);
```
yairgott, line 141: key is a std::string which redundant, instead the code should look like:
`auto [it, inserted] = accum.try_emplace(n.external_id->Str());`

src/query/fanout.cc
```diff
@@ -414,6 +443,199 @@ absl::Status PerformSearchFanoutAsync(
...
+
+void PerformRemoteMultiSearchRequestAsync(
+    std::unique_ptr<coordinator::MultiSearchIndexPartitionRequest> request,
+    const std::string &address,
+    coordinator::ClientPool *coordinator_client_pool,
+    std::vector<std::shared_ptr<SearchPartitionResultsTracker>>
+        per_arm_trackers,
+    vmsdk::ThreadPool *thread_pool) {
+  thread_pool->Schedule(
+      [coordinator_client_pool, address = std::string(address),
```
yairgott, line 504: This should be written as: `[coordinator_client_pool, address,`

Aksha1812:
**Reviewers for this PR**

Frank-Gu-81:
Note from 1.3 release PR triage session: incomplete state (testing). Depending on #1287

src/commands/ft.hybrid.json
boda26: I think this refers to Redis version. But from the official documentation, they marked the version 8.4.0 https://redis.io/docs/latest/commands/ft.hybrid/

boda26, line 10: Can you please check should the version be 8.4.0 or 8.4.4 here? https://redis.io/docs/latest/commands/ft.hybrid/

src/commands/ft_hybrid.cc
```diff
@@ -0,0 +1,363 @@
...
+// Runs each arm inline on the main thread (no per-arm content resolution),
+// fuses, then resolves content for the fused list once — all on the main
+// thread, so it is inherently atomic — then sends the reply.
+absl::Status MultiSearchParameters::ExecuteSyncLocal(
+    ValkeyModuleCtx *ctx, std::unique_ptr<MultiSearchParameters> cmd) {
+  for (auto &arm : cmd->arms) {
+    // Run the index search only; defer the database content fetch until after
+    // fusion so the multi-arm result is validated as a unit.
+    arm->no_content = true;
+    auto s = query::Search(*arm, query::SearchMode::kLocal);
```
boda26, line 290: There is no null guard against a null cancellation token for arms here. The issue is actually documented at multi_search.cc 154-158. ft.hybrid.cc:298 also has a null-guard for the envelope token, but it is missing at the arms side. A null guard should be added here as well.

src/commands/ft_hybrid_parser.cc
```diff
@@ -0,0 +1,605 @@
...
+            "VSIM vector parameter must be a $-prefixed PARAMS reference");
+      }
+      param_name.remove_prefix(1);
+      auto it = arm->parse_vars.params.find(param_name);
+      if (it == arm->parse_vars.params.end()) {
+        return absl::InvalidArgumentError(
+            absl::StrCat("Parameter ", param_name, " not found."));
+      }
+      it->second.first++;  // bump usage refcount
+      arm->query = std::string(it->second.second);
```
boda26, line 579: This skips the vector blob size check that PostParseQueryString performs (search.cc:1154-1160). The else branch below gets it via PreParse/PostParseQueryString; this hand-rolled branch assigns arm->query directly and never validates it.
Recommend fix: route the pure-VSIM arm through the shared PostParseQueryString path.

src/commands/ft_hybrid.cc
```diff
@@ -0,0 +1,363 @@
...
+  void QueryCompleteBackground(
+      std::unique_ptr<query::SearchParameters> self) override {
+    DoComplete(std::move(self));
+  }
+
+ private:
+  void DoComplete(std::unique_ptr<query::SearchParameters> /*self*/) {
+    RestoreAliases(search_result.neighbors, saved_aliases);
+    envelope->search_result = std::move(search_result);
+    envelope->search_result.status = absl::OkStatus();
```
boda26, line 181: DoComplete() unconditionally returns OKStatus, discarding both the per-arm failures recorded by the tracker, and index-dropped errors from ResolveContent.

src/commands/ft_hybrid.cc
```diff
@@ -0,0 +1,363 @@
...
+          expr::Expression::EvalContext eval_ctx;
+          expr::Value v = fn->Evaluate(eval_ctx, rec);
+          auto d = v.AsDouble();
+          return d.has_value() ? *d : 0.0;
+        });
+  }
+  // The COMBINE WINDOW bounds the per-arm contribution AND the final result.
+  // Per-arm truncation already happened inside fusion; cap the final list too.
+  if (params.fusion.window > 0 && fused.size() > params.fusion.window) {
+    fused.resize(params.fusion.window);
```
boda26, line 93: fused is hard-truncated by the window size (default 20) before aggregate pipeline runs, so LIMIT can never return more rows than the LIMIT (20) unless user also raises window.

src/commands/ft_hybrid_parser.cc
```diff
@@ -0,0 +1,605 @@
...
+    }
+    // Pre-populate the two reserved record slots that AggregateParameters
+    // expects: __key at index 0 and the score alias at index 1. Mirrors the
+    // setup at ft_aggregate.cc:94-97. Without this, MakeReference fails when
+    // any APPLY/FILTER/SORTBY references @<score_alias>.
+    CHECK_EQ(env.agg->AddRecordAttribute("__key", "__key",
+                                         indexes::IndexerType::kNone),
+             0u);
+    auto score_sv = vmsdk::ToStringView(env.agg->score_as.get());
+    CHECK_EQ(env.agg->AddRecordAttribute(score_sv, score_sv,
```
boda26, line 457: line 453 already adds __key as record attribute. If the user input it again, it would cause the check to fail and abort the server. We probably do not want to abort here

src/coordinator/server.cc
```diff
@@ -198,48 +209,141 @@ void Service::EnqueueSearchRequest(
...
+    std::atomic<int> remaining;
+  };
+  auto completion = std::make_shared<Completion>();
+  completion->reactor = reactor;
+  completion->remaining.store(n, std::memory_order_relaxed);
+
+  // 4. Dispatch each arm in parallel via SearchOneArm.
+  for (int i = 0; i < n; ++i) {
+    auto* sub_resp = response->mutable_sub_responses(i);
+    SearchOneArm(
```
boda26, line 335: Each arm is dispatched through an individual SearchOneArm call, no coordination across different arms. This contradicts with comments "Per-shard atomicity across arms is handled shard-side" in ft_hybrid.cc:339-340.

Frank-Gu-81:

src/query/multi_search.cc
```diff
@@ -0,0 +1,220 @@
...
+    // atomically as a unit (see FuseThenResolveLocal). With no_content set,
+    // GetContentProcessing() returns kNoContent and the arm completes on the
+    // background thread without a per-arm ResolveContent.
+    arms[i]->no_content = true;
+    // Uncap each arm pre-fusion: fusion needs the full per-arm match set so
+    // a doc matched by both arms is contributed by both (the "both arms or
+    // neither" guarantee — see TestFtHybridParallelArmConsistency). The
+    // aggregate-pipeline LIMIT (post-fusion) bounds the final reply size.
+    arms[i]->limit.first_index = 0;
+    arms[i]->limit.number = std::numeric_limits<uint64_t>::max();
```
Frank-Gu-81, line 206: Maybe we can do some optimization here. Do we need to uncap the arm here when `fusion.window` is finite? I think capping it here based on window can improve some performance. 

Concretely: a SEARCH arm matching 500K docs with WINDOW 20 appears to execute with `limit = UINT64_MAX`, while only 20 candidates can contribute to fusion.

src/query/multi_search.cc
```diff
@@ -0,0 +1,220 @@
...
+  parameters->arms.resize(arm_count);  // keep size() == N for tracker init
+  // Acquire ONE reader lock on the index's time-sliced mutex BEFORE
+  // scheduling any arm. Each arm's own Search() will additionally take its
+  // own reader lock (recursive readers are fine — the mutex just counts), but
+  // this outer lock is what guarantees the mutex stays in read mode for the
+  // entire multi-arm operation. Without it, a writer could time-slice in
+  // between arm A releasing its inner lock and arm B acquiring its own,
+  // producing the "split-arm" inconsistency the user prohibited.
+  // Released in MultiSearchTracker::Finalize after every arm has reported.
+  vmsdk::TimeSlicedMRMWMutex* index_mutex = nullptr;
```
Frank-Gu-81, line 179: TL;DR: I know the comment said it's fine, but I'm a bit concerned the nested reader acquisition can deadlock because `TimeSlicedMRMWMutex::Lock()` is [not unconditionally recursive/reentrant](https://github.com/valkey-io/valkey-search/blob/main/vmsdk/src/time_sliced_mrmw_mutex.cc#L68). 


If the reader quota has expired and a write is waiting, the second `ReaderLock()` goes into `SwitchWithWaiy()` rather than succeeding. In that path, `SwitchWithWait()` deliberately waits for a switch to the inverse mode before allowing the reader to proceed. However, `ShouldSwitch()` only permits that mode switch once `active_lock_count_ == 0`. The outer reader lock held here keeps the active count nonzero until all arms finish, while the arm is blocked trying to acquire its inner reader lock.

src/coordinator/server.cc
```diff
@@ -244,14 +247,108 @@ grpc::ServerUnaryReactor *Service::SearchIndexPartition(
...
+    grpc::Status first_error ABSL_GUARDED_BY(mu);
+  };
+  auto completion = std::make_shared<Completion>();
+  completion->reactor = reactor;
+  completion->remaining.store(n, std::memory_order_relaxed);
+
+  // 4. Dispatch each arm in parallel via SearchOneArm.
+  for (int i = 0; i < n; ++i) {
+    auto *sub_resp = response->mutable_sub_responses(i);
+    SearchOneArm(
```
Frank-Gu-81, line 331: I agree with @boda26 here. 

`PerformMultiSearchLocalAsync()` holds one outer reader lock specifically so a writer cannot run between arm A and arm B, but `MultiSearchIndexPartition()` dispatches every arm independently through `SearchOneArm()`. Each underlying `Search()` acquires/releases its own reader lock, so if the arms don't execute concurrently a writer can time-slice between them and the shard can return SEARCH results from one index state and VSIM results from another. 

Perhaps we can establish one shared read snapshot across all of its sub-requests for shard-side multi-search?

src/commands/ft_hybrid_parser.cc
```diff
@@ -0,0 +1,1049 @@
...
+      itr.Next();
+      per_arm_alias = std::string(alias_sv);
+      arm->score_as = vmsdk::MakeUniqueValkeyString(alias_sv);
+    } else if (absl::EqualsIgnoreCase(next, kScorerKw)) {
+      itr.Next();
+      // Parse-but-no-op in V1.
+      VMSDK_ASSIGN_OR_RETURN(auto scorer_sv, itr.GetStringView());
+      (void)scorer_sv;
+      itr.Next();
+      // TODO(text-scoring): apply the scorer when text scoring lands.
```
Frank-Gu-81, line 215: Should we update the logic here now that text scorer is merged?

src/query/fanout.cc
```diff
@@ -419,6 +447,197 @@ absl::Status PerformSearchFanoutAsync(
...
+    auto &arm = *parameters->arms[i];
+    // ExecuteCommand populates the envelope's index fingerprint / slot
+    // fingerprint just before fanout — AFTER the arms were built at parse
+    // time. Propagate them now so the per-arm consistency check on each shard
+    // matches. (Mirror of the cancellation_token propagation below.)
+    arm.index_fingerprint_version = parameters->index_fingerprint_version;
+    arm.slot_fingerprint = parameters->slot_fingerprint;
+    arm.enable_partial_results = parameters->enable_partial_results;
+    arm.enable_consistency = parameters->enable_consistency;
+    auto req = coordinator::ParametersToGRPCSearchRequest(arm);
```
Frank-Gu-81, line 544: It looks like the aggregate `LOAD` projection is lost when building per-arm shard requests.

`ParametersToGRPCSearchRequest(arm)` sees the arm's empty `return_attributes` with `no_content == false`, which causes HASH shards to fetch all fields. Those fields can then survive fusion and show up as `extra_fields_` in the final response.

So in cluster mode, `LOAD 1 @title` may also return unrelated fields, while the local path correctly applies `agg->return_attributes`.

Could we propagate the aggregate LOAD semantics to the shard requests and add a few minimum tests for no `LOAD`, named `LOAD`, and `LOAD *`?

src/query/multi_search.cc
```diff
@@ -0,0 +1,220 @@
...
+      params->retained_arm_owners.push_back(std::move(owner));
+    }
+    arm_owners_.clear();
+    // Read first_error_ under mu_ (it is ABSL_GUARDED_BY(mu_)); use the local
+    // copy below, after the lock is released.
+    first_error = first_error_;
+  }
+  if (any_arm_failed_.load() && !params->enable_partial_results) {
+    params->search_result.status = first_error;
+  }
```
Frank-Gu-81, line 135: Is it intended to suppress consistent failures when partial results are enabled? 

`SearchPartitionResultsTracker` gives a slot-consistency `FailedPrecondition` precedence regardless of `enable_partial_results`, but `Finalize()` only propagates arm errors when partial results are disabled. `ReplyMulti()` also ignores cancellation when partial results are enabled.

So `enable-consistent-results=yes` + `enable-partial-results=yes` can return a fused partial result after a slot fingerprint mismatch.

Should consistency failures always propagate here?

src/commands/ft_hybrid.cc
```diff
@@ -0,0 +1,793 @@
...
+          ++dropped;
+          continue;
+        }
+        auto distance = vector_index->RecomputeDistance(
+            vmsdk::ToStringView(record->second.value.get()), arm->query);
+        if (distance.ok()) {
+          n.distance = *distance;
+          n.score = *distance;
+          rescored = true;
+        }
```
Frank-Gu-81, line 433: What if `RecomputeDistance()` returns an error because the new value is malformed? Looks like it would leave the neighbor's old distance untouched, so we would score on previous values while the current one cannot be scored at all.

src/commands/ft.hybrid.json
```diff
@@ -0,0 +1,228 @@
...
+      "SLOW",
+      "SEARCH"
+    ],
+    "history": [
+      [
+        "1.3.0",
+        "Hybrid search combining text/filter and vector arms"
+      ]
+    ],
+    "summary": "Performs hybrid search combining a text/filter arm with a vector arm, fuses results via RRF or LINEAR, and feeds the fused list through the FT.AGGREGATE post-pipeline.",
```
Frank-Gu-81, line 14: Nit: From `src/commands/ft_hybrid_parser.cc` L503, I see we also support `FUNCTION`.

Frank-Gu-81 (commented):
Can we also add user-facing documentation for `FT.HYBRID` (maybe a new file `ft.hybrid.md`) as part of this PR? A few other places we can add documentations: 
- README supported-command list
- `COMMANDS.md`
- or the Quick Start.

src/commands/ft_hybrid.cc
```diff
@@ -0,0 +1,365 @@
...
+    auto *raw = envelope.release();
+    raw->blocked_client->SetReplyPrivateData(raw);
+    raw->blocked_client->UnblockClient();
+  }
+};
+
+// Local async completion: all arms have delivered raw (content-free) results.
+// Fuse, then run the single atomic content resolution over the fused list.
+void FuseThenResolveLocal(std::unique_ptr<MultiSearchParameters> params) {
+  auto fused = BuildFusedNeighbors(*params);
```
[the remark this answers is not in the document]
allenss-amazon: stale.

src/commands/ft_hybrid_parser.cc
```diff
@@ -0,0 +1,599 @@
...
+    if (!next_or.ok()) {
+      break;
+    }
+    if (absl::EqualsIgnoreCase(next_or.value(), kPolicyKw)) {
+      itr.Next();  // consume POLICY
+      if (!itr.HasNext()) {
+        return absl::InvalidArgumentError("POLICY requires a value");
+      }
+      itr.Next();  // consume the value (silently discarded)
+      continue;
```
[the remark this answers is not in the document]
allenss-amazon: Ignoring policy is fine.

src/commands/ft.hybrid.json
```diff
@@ -0,0 +1,228 @@
...
+          },
+          {
+            "name": "combine_method",
+            "type": "string"
+          },
+          {
+            "name": "combine_count",
+            "type": "integer"
+          }
+        ]
```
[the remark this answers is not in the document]
allenss-amazon: Documentation issue, will fix the JSON.

src/commands/ft.hybrid.json
allenss-amazon: Fixed: This is now 1.3.0, which is correct for the search module.

src/commands/commands.cc
allenss-amazon: fixed.

src/commands/ft.hybrid.json
Frank-Gu-81: Resolving this as it has been addressed.

src/commands/ft_hybrid.cc
Frank-Gu-81: Fixed: https://github.com/allenss-amazon/valkey-search/blob/hybrid/src/commands/ft_hybrid.cc?#L434-L450

src/query/multi_search.cc
Frank-Gu-81: Tracked as P3: #1430

src/query/fanout.cc
Frank-Gu-81: Remaining work tracked in #1425

src/query/multi_search.cc
Frank-Gu-81: Fixed. https://github.com/allenss-amazon/valkey-search/blob/hybrid/src/query/multi_search.cc?#L171-L176

src/coordinator/server.cc
Frank-Gu-81: https://github.com/allenss-amazon/valkey-search/blob/hybrid/src/query/multi_search.cc?#L167

src/query/multi_search.cc
Frank-Gu-81: https://github.com/allenss-amazon/valkey-search/blob/hybrid/src/query/multi_search.cc?#L107-L125

Frank-Gu-81 (approved):
I took another look at the comments I raised earlier. Looks like they've all either been addressed or have corresponding tracking issues, and I didn't see any other blocking issues from my side. Approved this PR. Thanks for addressing them!

One last non-blocking nit: should we also update the top-level README supported-command list to include `FT.HYBRID`?

src/commands/ft_aggregate_exec.cc
allenss-amazon: I checked this again. At the call site we have only an IndexerType enum from AttributeRecordInfo, not an index object. Columns for fields that are LOADed but not indexed are registered as kNone (ft_aggregate_parser.cc:425) and have no index at all. A virtual on IndexBase would still need an enum-keyed fallback for those columns, so the switch would stay. ProcessFieldValue is also existing FT.AGGREGATE code that this PR only moved. If we later make record columns carry an index pointer, this would be a good follow-up.

src/commands/ft_aggregate_exec.cc
allenss-amazon: Agreed, "process neighbors for processing" said nothing. I renamed it to PrepareNeighborRecords. It sets up the key/score record columns and resolves neighbor content before the aggregation stages run. The rename is file-local and changes no behavior.

src/commands/ft_aggregate_exec.cc
allenss-amazon: The naming complaint is fair, but the file predates this PR: main already has ft_aggregate_exec.{h,cc} holding Record, RecordSet and the stages. This PR adds RunAggregatePipeline and moves in the helpers it needs from ft_aggregate.cc. Moving it into query/ today would create a static-library cycle: it is built on aggregate::AggregateParameters in commands/, and the commands library links the query libraries. FT.HYBRID's own pieces stay in commands/ for the same reason. I've filed #1428 with the split you describe and a sequence that fixes the dependency direction first, so the move is reviewable as a series of pure moves rather than growing this PR further.

src/commands/ft_aggregate_exec.cc
allenss-amazon: Yes -- moved verbatim, except for changes in comments. Diffed against the base-branch original in ft_aggregate.cc, the only difference is three dropped one-line comments ("Set key field if requested" and two others like it). It has since changed on purpose. When copying fetched fields into a vector query's record, it now skips the score column. That way a stored field with the same name as the score (e.g. `__score` under LOAD *) can no longer overwrite it. That fix applies to FT.AGGREGATE as well as FT.HYBRID.

src/commands/ft_aggregate_exec.cc
allenss-amazon: Yes -- moved verbatim, except for changes in comments. Diffed against the base-branch original in ft_aggregate.cc, the body is identical apart from the rename to PrepareNeighborRecords (per your other comment). The revision you reviewed differed only in pointer/reference formatting (`T *x` vs `T* x`).

src/commands/ft_aggregate_exec.cc
allenss-amazon: Yes -- moved verbatim, except for changes in comments. At the revision you reviewed, the body matches the base-branch original in ft_aggregate.cc apart from two dropped comments ("Check for timeout", "Testing purpose only"). One later change touches it: the cancellation message literal became a kCancelledMessage constant of the same text, and APPLY, FILTER, GROUPBY and SORTBY now poll for cancellation inside their loops. The function's own behaviour is unchanged.

src/commands/ft_aggregate_exec.h
allenss-amazon: I agree it isn't really command code, but it can't move cleanly yet. RunAggregatePipeline takes aggregate::AggregateParameters from commands/ft_aggregate_parser.h and runs the Record/stage machinery that lives in the commands library. The commands library already links the query libraries, so moving this into query/ creates a static-library cycle. FT.HYBRID's counterparts are in commands/ for the same reason. #1428 covers this together with the ft_aggregate_exec naming: it lifts the execution half of AggregateParameters into query/ first, which is what lets the files move without the cycle.

src/commands/ft_hybrid_parser.cc
allenss-amazon: Thanks. The copy is real, but it can't cause the failure you describe: FT.HYBRID never reaches VerifyQueryString. It enforces the FT.SEARCH limits on K, EF_RUNTIME, TIMEOUT and DIALECT at parse time instead, so nothing reads the per-arm counts afterwards. One side effect: FT.HYBRID doesn't report an unused PARAMS entry the way FT.SEARCH does. We're leaving use-count merging out on purpose, because the use-count machinery is being removed elsewhere, so sharing the map now would only add plumbing that's about to go away.

src/query/fusion.cc
allenss-amazon: Good catch, thanks. Fixed: LINEAR no longer normalizes per arm, so the unconditional 1.0 - x inversion is gone. The distance-vs-score direction is now handled once, before fusion: BuildFusedNeighbors calls ConvertVectorArmScoresToSimilarity (src/commands/ft_hybrid.cc) on vector arms only, turning each distance into a higher-is-better similarity for its metric (IP, COSINE, L2). Text/filter arms pass through untouched, so every arm reaching rank_fusion is higher-is-better. We dropped normalization rather than fix it because it made a document's fused score depend on which other documents came back in the same arm; ALPHA/BETA balance the arms instead.

src/commands/ft_hybrid.cc
allenss-amazon: Agreed, the score alias must win. We first switched to insert_or_assign, but that turned out to be a use-after-free: a RecordsMap key is a string_view into bytes its own mapped value owns, so replacing the value leaves the old key dangling (reproduced with LOAD * against a live server). Both RestoreAliases and AttachArmScore now erase then emplace. The aggregate read-back no longer overwrites the score column with the stored field either, and `__key` is rejected as an alias. Divergence from Redis on LOAD * collision precedence is tracked in #1408.

src/commands/ft_hybrid_parser.cc
allenss-amazon: Good catch, and the misparse was real: `KNN 2 K 5 POLICY BATCHES YIELD_SCORE_AS vs` sent YIELD_SCORE_AS to the aggregate parser. POLICY is now handled in the subclause scanners, as you suggested. VSIM now consumes POLICY and BATCH_SIZE with their value and keeps scanning. SEARCH rejects POLICY by name instead of failing later with a misleading "requires VSIM" error. POLICY stays in IsTopLevelKeyword because a VSIM with no mode block can be followed by a top-level POLICY. Tests: PolicyInsideVsimDoesNotSwallowTheNextToken, PolicyInsideVsimWithNoValueIsRejected, PolicyWithNoModeBlockStillParses and PolicyInsideTheSearchClauseIsRejected.

src/coordinator/server.cc
allenss-amazon: Agreed, and done: MultiSearchIndexPartition records the first non-OK arm status and finishes the reactor with it, so interceptors and node-failure counters see the failure. We went further: a multi-arm search failure is always hard on the shard, even with partial results enabled, so a shard answers every arm or none. The coordinator applies a non-OK RPC status to every arm, so a failing shard drops out entirely. The per-arm grpc_code stays as a diagnostic, but the RPC status is authoritative. A request to finish with OK instead (C34) is declined for this reason.

src/query/rank_fusion.cc
allenss-amazon: Done. I went with rank_fusion: the files are now src/query/rank_fusion.{h,cc} (test: rank_fusion_test.cc), and the namespace and CMake target are query::rank_fusion and rank_fusion.

src/query/fusion.h
allenss-amazon: rank_fusion.h (formerly fusion.h) uses #pragma once. I've left the PR's other two new headers, multi_search.h and ft_hybrid_combine.h, on #ifndef guards, which is what nearly all of the tree uses. Converting a handful of headers inside this PR would leave the codebase in two styles without settling anything. If #pragma once should be the rule for new headers, it's worth a style-guide note and a separate sweep that applies it consistently.

src/query/fusion.h
allenss-amazon: Done. The entry points are now rank_fusion::RRF, rank_fusion::Linear and rank_fusion::Function, and the callers in ft_hybrid.cc and rank_fusion_test.cc were updated to match.

src/query/fusion.cc
allenss-amazon: Done, thanks. FormatScore in src/query/rank_fusion.cc (renamed from fusion.cc) is now the one-liner `return absl::StrFormat("%.12g", v);`. The output format is unchanged: it is the same %.12g the reply path uses for other doubles.

src/query/rank_fusion.cc
```diff
@@ -0,0 +1,243 @@
...
+    n.external_id = entry.representative->external_id;
+    n.distance = static_cast<float>(entry.score);
+    // Carry over the representative's attribute_contents as the base. We
+    // shallow-copy via move-fresh; if we want to merge from multiple arms in
+    // the future, this is the place.
+    if (entry.representative->attribute_contents.has_value()) {
+      // Construct a fresh RecordsMap by emplacing copies of the keys/values;
+      // RecordsMapValue holds a UniqueValkeyString, so we re-create those.
+      n.attribute_contents.emplace();
+      for (const auto& [identifier, value] :
```
allenss-amazon: Thanks, I looked at this but it isn't safe here. RetainUniqueValkeyString calls VerifyMainThread(), which is a CHECK, and ValkeyModule_RetainString bumps a non-atomic refcount. Fusion does not run on the main thread: BuildFusedNeighbors is called from FuseAndResolveLocal and the cluster completion callback, on whichever worker or gRPC thread drops the last tracker reference, before the client is unblocked. Retaining would CHECK-fail, and without the check it would race the refcount. Copying via MakeUniqueValkeyString (ValkeyModule_CreateString with a null context) is thread-safe, and the cost is bounded by fused-result size times attribute count.

src/query/fusion.cc
allenss-amazon: Done. The intermediate std::string is gone, and all three accumulation loops in src/query/rank_fusion.cc (RRF, Linear, Function) now use `auto [it, inserted] = accum.try_emplace(n.external_id->Str());`. Because absl::flat_hash_map<std::string,...> supports heterogeneous lookup, a doc already seen from an earlier arm is found without building a std::string; the key is built only on first insert.

src/query/fanout.cc
allenss-amazon: Done. PerformRemoteMultiSearchRequestAsync now captures `[coordinator_client_pool, address, request = std::move(request),...]`. The same redundant `address = std::string(address)` init-capture is gone from the sibling sites in fanout.cc, so all four now read the same way. The enclosing functions take `const std::string&`, so a plain by-copy capture already stores an owned copy.

src/query/fanout.cc
```diff
@@ -414,6 +442,199 @@ absl::Status PerformSearchFanoutAsync(
...
+    uint64_t per_shard_limit;
+    if (arm.IsNonVectorQuery()) {
+      per_shard_limit = std::max<uint64_t>(window, 10);
+    } else {
```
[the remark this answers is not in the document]
allenss-amazon: Confirmed, thanks. The 10 in std::max(window, 10) is meant as a floor, but with window == 0 it became the ceiling, so asking for no cap produced a harder cap than any window a user could type. Rather than teach every consumer about the 0 sentinel, we resolve it at parse time: WINDOW 0 and COMBINE FUNCTION's default both become max-combine-window, a dev config added (default 1,000,000). Each shard is now asked for up to that many rows per non-vector arm. Parser unit tests and test_window_zero_means_unlimited pin the behavior.

src/commands/ft.hybrid.json
allenss-amazon: Thanks for checking. This field records the valkey-search module version rather than a Redis one, which is why the sibling command specs carry 1.2.0 and not a Redis release. The entry has been changed from 8.4.4 to 1.3.0, the valkey-search release that introduces FT.HYBRID, and 1.3.0 is the version anything new lands under. So the 8.4.0 versus 8.4.4 question doesn't apply here. If we want to record which Redis release first shipped FT.HYBRID, that belongs in the compatibility notes rather than in this spec.

src/commands/ft_hybrid.cc
allenss-amazon: Thanks, you were right: arms copied the token during parse, before ExecuteCommand created it, so the sync path passed a null token to query::Search. That path was fixed by copying the envelope token onto each arm, and the synchronous FT.HYBRID path has since been removed altogether. On every remaining path the token is non-null by the time an arm runs: the local async path assigns it at multi_search.cc:192, the fanout at fanout.cc:623, and shards build their own with cancel::Make. We fixed this by assigning the token rather than adding a null guard, and the guard in OnArmComplete remains only for unit tests that bypass dispatch.

src/commands/ft_hybrid_parser.cc
allenss-amazon: Agreed, the check was missing, and it is there now. The pure-VSIM branch now compares the blob size with the index's vector data size and returns FT.SEARCH's exact message. I didn't route it through PostParseQueryString: PostParseVectorParameters parses parse_vars.k_string/ef_string/score_as_string, which the VSIM clause parser never sets (it writes arm->k/ef/score_as directly), so it would fail on the empty K first. Tests: PureVsimArmRejectsOverLongVectorBlob, PureVsimArmRejectsUnderLongVectorBlob, PureVsimArmAcceptsCorrectlySizedVectorBlob, and VsimArmWithFilterRejectsWrongSizedVectorBlob for the path that does go through PostParseQueryString.

src/commands/ft_hybrid.cc
allenss-amazon: Agreed, and this is fixed: DoComplete no longer overwrites the status, so ResolveContent's result, including an index-dropped error, reaches ReplyMulti, which checks it first. Per-arm failures no longer get as far as DoComplete, because FuseThenResolveLocal returns the tracker's error without fusing. MultiSearchTracker::Finalize now sets that error whether or not partial results are enabled, and consistency failures take precedence. ArmGate also surfaces a cancellation or dropped index that shows up while the arms are parked. ResolveFusedContentInline was removed along with the synchronous path.

src/commands/ft_hybrid.cc
allenss-amazon: You were right about the code as it stood then. The resize of the fused list to WINDOW is gone. WINDOW now caps only how many of each arm's top results enter fusion (rank_fusion.cc), so with two arms the fused list can hold up to 2 x WINDOW documents, and LIMIT bounds the reply. test_window_bounds_the_arms_not_the_fused_list checks that WINDOW 6 with LIMIT 0 100 returns more than 6 rows. To page beyond the union of per-arm windows you still have to raise WINDOW, because WINDOW is by definition a per-arm candidate count.

src/commands/ft_hybrid_parser.cc
allenss-amazon: You're right. Before this fix, `YIELD_SCORE_AS __key` made AddRecordAttribute return kKeyColumn where kScoreColumn was expected, and the CHECK_EQ aborted the server with any LOAD clause. `__key` is now rejected as a score alias at all three YIELD_SCORE_AS sites, with an error saying it tries to redefine the reserved field. Both CHECK_EQs are now returned errors, so any alias path we haven't anticipated fails the command instead of the server. Tests: SearchArmYieldScoreAsKeyIsRejected, VsimArmYieldScoreAsKeyIsRejected, and the CombineYieldScoreAsKeyWith{NoLoad,LoadAll,LoadField}IsRejected tests.

src/coordinator/server.cc
allenss-amazon: You're right, the comment described a guarantee nothing implemented. It now says what actually happens: on every shard each arm runs through its own SearchOneArm -> SearchAsync -> ResolveContent and is mutation-checked on its own. That is deliberate. The arms don't need to share an index snapshot; each shard returns post-mutation-resolution, pre-fusion results, and we accept no cross-arm coordination between them. What must hold is failure atomicity, and remote shards now answer every arm or none. The coordinator's local shard doesn't have that yet, which is tracked in #1420.

integration/compatibility/generate_hybrid.py
```diff
@@ -0,0 +1,588 @@
...
+# that gap. FT.HYBRID has no key-type-specific code of its own.
+@pytest.mark.parametrize("key_type", ["hash"])
+class TestHybridCompatibility(BaseCompatibilityTest):
+    ANSWER_FILE_NAME = "hybrid-answers.pickle.gz"
+    # TODO(reference-image): temporary. FT.HYBRID needs the Redis 8.4+ query
+    # engine, which redis/redis-stack-server does not have. A separate PR moves
+    # BaseCompatibilityTest.DOCKER_IMAGE to redis:latest for every generator;
+    # once that lands this override can be dropped and this class can inherit
+    # the shared image again.
+    DOCKER_IMAGE = "redis:8"
```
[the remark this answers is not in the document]
allenss-amazon: This is now outdated, and the text that made it look current has been fixed. generate_hybrid.py no longer overrides the image: it inherits the shared redis:latest from BaseCompatibilityTest, like every other generator (#1366). Its docstring and unsupported_tests.md 5.8 still described the removed redis:8 override and its TODO(reference-image); both are corrected. Whether that shared reference should be pinned to an immutable digest, or recorded in the answer pickles, is a policy question for the whole compatibility suite rather than something specific to FT.HYBRID, so we're keeping it out of this PR.

src/coordinator/server.cc
```diff
@@ -244,14 +247,108 @@ grpc::ServerUnaryReactor *Service::SearchIndexPartition(
...
+  return reactor;
+}
+
+grpc::ServerUnaryReactor *Service::MultiSearchIndexPartition(
+    grpc::CallbackServerContext *context,
+    const MultiSearchIndexPartitionRequest *request,
+    MultiSearchIndexPartitionResponse *response) {
+  GRPCSuspensionGuard guard(GRPCSuspender::Instance());
+  grpc::ServerUnaryReactor *reactor = context->DefaultReactor();
+  const int n = request->sub_requests_size();
```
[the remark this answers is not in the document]
allenss-amazon: Thanks, but we're not adding a cap here. The coordinator gRPC service trusts its peers. We don't validate gRPC calls anywhere on this surface, and MultiSearchIndexPartitionRequest is only built by our own fanout, one sub-request per FT.HYBRID arm. An untrusted client that can reach the coordinator port could already send oversized SearchIndexPartition requests, so a limit on this one handler wouldn't change the exposure. Hardening the coordinator against untrusted peers is a separate, cross-cutting concern.

src/coordinator/server.cc
```diff
@@ -244,14 +247,108 @@ grpc::ServerUnaryReactor *Service::SearchIndexPartition(
...
+              1) {
+            grpc::Status final_status;
+            {
+              absl::MutexLock lock(&completion->mu);
+              final_status = completion->first_error;
+            }
+            completion->reactor->Finish(final_status);
+          }
+        });
   }
```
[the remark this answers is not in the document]
allenss-amazon: This is intentional, so we're declining it. If one arm fails on a shard, keeping the other arm's results would let that shard's documents enter fusion from a single arm, with wrongly low fused scores and nothing in the reply to flag it. The rule is that a shard answers every arm or none. A non-OK RPC status drops the whole shard, which the partial-results policy then tolerates or turns into an error. Finishing non-OK also lets interceptors and node-failure counters see the failure, as requested in C13. The per-arm grpc_code is diagnostic only.

src/query/rank_fusion.cc
```diff
@@ -0,0 +1,221 @@
...
+  }
+  auto value_str = vmsdk::MakeUniqueValkeyString(FormatScore(value));
+  auto identifier_str = vmsdk::MakeUniqueValkeyString(alias);
+  // attribute_contents is keyed by string_view; the view must point at the
+  // owned identifier_str's bytes for lifetime safety. The existing
+  // RecordsMapValue stores the identifier internally; we use that as the key.
+  auto identifier_view = vmsdk::ToStringView(identifier_str.get());
+  n.attribute_contents->emplace(
+      identifier_view,
+      RecordsMapValue(std::move(identifier_str), std::move(value_str)));
```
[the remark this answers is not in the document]
allenss-amazon: The collision is real, and the score now wins, but the suggested insert_or_assign is unsafe here; RestoreAliases already using it was itself a bug. A RecordsMap key is a string_view into the identifier owned by its own mapped value, so insert_or_assign frees those bytes when it replaces the value and keeps a dangling key; we reproduced this with LOAD * as a corrupted field name. Both sites now erase the old entry and then emplace the new one. RRFTest.ScoreAliasOverwritesCollidingAttribute checks the key pointers as well as the text. How LOAD * collision precedence should match Redis is tracked separately in #1408.

src/query/rank_fusion.cc
```diff
@@ -0,0 +1,221 @@
...
+  }
+  // Stable order: descending fused score, tie-break by external_id for
+  // determinism.
+  std::sort(out.begin(), out.end(),
+            [](const indexes::Neighbor& a, const indexes::Neighbor& b) {
+              if (a.score != b.score) {
+                return a.score > b.score;
+              }
+              return a.external_id->Str() < b.external_id->Str();
+            });
```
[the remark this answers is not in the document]
allenss-amazon: You're right, and the consequence is worse than an odd ordering. NaN answers false both ways, so the comparator made it equivalent to every score while those scores stayed ordered among themselves; that is not a strict weak ordering, and libstdc++'s unguarded partition and insertion loops have no bounds check, so the scan can run past the end of the range. The sort now keys on (is NaN, -score, external_id), ranking NaN last, which keeps the document in the reply with whatever score the caller's expression produced. It uses the bit-pattern IsNaN helper, because -ffast-math makes std::isnan unreliable here. The same comparator shape exists in FT.AGGREGATE's SORTBY, tracked in #1429.

src/query/fanout.cc
```diff
@@ -419,6 +447,197 @@ absl::Status PerformSearchFanoutAsync(
...
+      // Vector arm: fetch top-k from each shard (worst case all k from one).
+      per_shard_limit = std::max<uint64_t>(arm.k, window);
+    }
+    req->mutable_limit()->set_first_index(0);
+    req->mutable_limit()->set_number(per_shard_limit);
+    // Widen the COORDINATOR-side arm limit to effectively unlimited. The
+    // per-arm SearchPartitionResultsTracker trims its merged-across-shards
+    // result to arm.limit (times a buffer multiplier); leaving it bounded
```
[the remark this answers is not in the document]
allenss-amazon: Good catch. The flag can't be re-derived on the shard, because a VSIM arm with a FILTER is rewritten to the same query string a hybrid text=>[KNN] query produces. vector_score_only is now on SearchIndexPartitionRequest and round-trips through search_converter.cc. It also makes ranks_on_distance consult the flag, since serializing one without the other would leave a refreshed distance beside a stale score. is_match_all is serialized too, which had the same can't-re-derive problem, and adds search_converter_test.cc for the converter, which previously had no test coverage.

src/query/multi_search.cc
allenss-amazon: Agreed, this is a valid optimization: fusion reads only each arm's top window entries, so the rest of an uncapped arm is wasted work. The comment defending the uncap was wrong on both counts, and it has been rewritten -- fusion does not need the whole match set, and the "both arms or neither" guarantee comes from ArmGate and RevalidateArmsBeforeFusion re-checking every arm together. We're postponing the change itself, tracked in #1430. Capping needs care because pre-fusion revalidation can drop entries, so a cap of exactly window could leave an arm short. The cluster path already bounds the per-shard fetch by window.

src/query/fanout.cc
allenss-amazon: Confirmed and fixed. ApplyAggregateProjection now copies the aggregate's resolved LOAD attributes onto every per-arm shard request, so LOAD 1 @title returns only title. With no LOAD, we can't send no_content, because that also turns off the shard's content resolution and mutation revalidation. The shard still fetches, and the coordinator drops the fields before fusion. HASH LOAD * was already correct. JSON LOAD * depends on #1381 and is tracked in #1425. test_ft_hybrid_load_projection.py covers no LOAD, named LOAD and LOAD * in both standalone and cluster mode, as you asked.

integration/test_aggregate_hybrid_distance.py
```diff
@@ -0,0 +1,103 @@
...
+    return struct.pack(f"<{len(values)}f", *values)
+
+
+@pytest.mark.skip(reason=(
+    "https://github.com/valkey-io/valkey-search/issues/1410"
+    " -- FT.AGGREGATE reports the BM25 relevance in the `AS <alias>` column"
+    " for a hybrid `text=>[KNN ...]` query, where the equivalent FT.SEARCH and"
+    " the pure-vector aggregate both report the KNN distance. The test is"
+    " written to fail against that defect; unskip it when the issue is fixed."))
+class TestAggregateHybridDistance(ValkeySearchTestCaseBase):
```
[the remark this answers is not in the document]
allenss-amazon: will get fixed.

PR #1090, opened 2026-05-31
Fix AS parameter substitution for vector score aliases

### LINKED ISSUE #1089: [BUG] FT.SEARCH does not apply substituted AS alias to score_as in vector queries
**Describe the bug**

  `FT.SEARCH` does not fully apply parameter substitution for the `AS` clause in vector queries.

  When `AS` uses a substituted parameter such as `$alias`, the substitution is applied to a parse-time field, but the resolved value is not propagated to the final `score_as` field used later in execution.

  As a result, the literal parameter reference may be preserved instead of the substituted alias.

  **To Reproduce**
  Run a vector query where the score alias is parameterized:
  ```text
  FT.SEARCH idx "*=>[KNN 10 @vec $BLOB AS $ALIAS]" PARAMS 4 BLOB <binary-vector> ALIAS my_score DIALECT 2
```
  Observe that the returned score field alias may remain $ALIAS instead of my_score.

  **Expected behavior**

  The AS clause should honor parameter substitution in the same way as other vector-query parameters.
  For the example above, the final score field alias should be my_score, not the literal $ALIAS.

### PR DESCRIPTION
**Pull request by bandalgomsu**
Fix `FT.SEARCH` vector query parsing so parameter substitution is fully applied to the `AS` clause.

  Previously, queries like:

  `FT.SEARCH idx "*=>[KNN 10 @vec $BLOB AS $ALIAS]" PARAMS 4 BLOB <binary-vector> ALIAS my_score DIALECT 2`

  could preserve the literal `$ALIAS` in the final score field alias instead of the substituted value.

Resolves #1089

### CODE REVIEW

File: src/query/search.cc [hunk 1]
```diff
@@ -1096,8 +1096,10 @@ absl::Status PostParseVectorParameters(query::SearchParameters &parameters) {
 
   if (!parameters.parse_vars.score_as_string.empty()) {
     VMSDK_ASSIGN_OR_RETURN(
-        parameters.parse_vars.score_as_string,
+        auto score_as_string,
         SubstituteParam(parameters, parameters.parse_vars.score_as_string));
+    parameters.parse_vars.score_as_string = score_as_string;
```
Aksha1812: Is this line needed? looks like `parameters.parse_vars.score_as_string` is no longer used

Aksha1812:
AI suggested this . Can you verify if it is the better choice @bandalgomsu 

```
  
  ▎ FT.AGGREGATE isn't covered by this fix. In src/commands/ft_aggregate.cc (ParseCommand, ~line 100), score_as is read between 
  ▎ PreParseQueryString and PostParseQueryString:
  ▎ VMSDK_RETURN_IF_ERROR(PreParseQueryString());        // score_as = "$ALIAS"
  ▎ auto score_sv = vmsdk::ToStringView(score_as.get()); // reads "$ALIAS"
  ▎ CHECK(AddRecordAttribute(score_sv, score_sv, ...) == 1);
  ▎ ...
  ▎ VMSDK_RETURN_IF_ERROR(PostParseQueryString());       // score_as fixed here — too late
  ▎ So for FT.AGGREGATE ... [KNN ... AS $ALIAS], the score record-attribute gets registered under $ALIAS at parse time and 
  ▎ my_score at reply time. Could we resolve the alias substitution in PreParseQueryString (where params is already available) 
  ▎ instead of Pass 2? That fixes both commands with one path and removes the double-allocation. Did you get a chance to test the
  ▎ aggregate path with a parameterized alias?

```

Aksha1812:
any update here @bandalgomsu

bandalgomsu:
I think the aggregate issue is real.

  One nuance is that `FT.AGGREGATE` currently runs `PreParseQueryString()` before it parses `PARAMS`, so resolving `AS $ALIAS` in pre-parse would first require a small `PARAMS` pre-pass on the aggregate path.

  Would that approach make sense to you?

Aksha1812:
>in pre-parse would first require a small PARAMS pre-pass on the aggregate path. Would that approach make sense to you?

Sure lets see how that turns out, make sure you don't use the main iterator and make sure we don't increase any refcounts unnecessarily during the pre pass .

Aksha1812:
Can you address failing tests and new AI comments ?

src/commands/ft_aggregate_parser.cc
```diff
@@ -258,6 +302,20 @@ vmsdk::KeyValueParser<AggregateParameters> CreateAggregateParser() {
...
+  parser.AddParamParser(kDialectParam, ConstructSkipSingleValueParser());
+  parser.AddParamParser(kTimeoutParam, ConstructSkipSingleValueParser());
+  parser.AddParamParser(kAddScoresParam, ConstructSkipFlagParser());
+  parser.AddParamParser(kSlopParam, ConstructSkipSingleValueParser());
+  parser.AddParamParser(kInorder, ConstructSkipFlagParser());
+  parser.AddParamParser(kVerbatim, ConstructSkipFlagParser());
+  parser.AddParamParser(kLoadParam, ConstructSkipLoadParser());
+  parser.AddParamParser(kParamsParam, ConstructParamsParser());
+  return parser;
+}
```
Aksha1812: are we missing apply , filter , groupby , limit , sort by from here . will the pre parser exit early if any of these are encountered during pre parsing . ? is the keyword table needed at all, would scanning the raw token till we reach PARAMS  .

src/commands/ft_aggregate.cc
```diff
@@ -45,6 +45,23 @@ struct RealIndexInterface : public IndexInterface {
   RealIndexInterface(std::shared_ptr<IndexSchema> schema) : schema_(schema) {}
 };
 
+absl::StatusOr<absl::string_view> ResolveAggregateParam(
```
Aksha1812: possible redundancy . Is this exactly similar to `SubstituteParam` in `search.cc` ? .

Aksha1812:
Any alternate to pre parsing ?

bandalgomsu:
I dug into this, but I don’t have a clean minimal version of the refactor yet. The score alias substitution is tied to early record attribute registration, so removing the pre-pass cleanly is a bit larger than I expected.

Would it make sense to keep this PR focused on the FT.SEARCH bug fix, and handle the FT.AGGREGATE bug fix separately in a follow-up?

allenss-amazon:
we definitely want a solution to this that works for both FT.SEARCH and FT.AGGREGATE as well as the future FT.HYBRID.  this PR looks like a lot of work to support parameter substitution outside of the query string (where the multi-pass nature is already comprehended). Perhaps we might want to just fix the issue, i.e., substitution for the score AS field.

Aksha1812:
**Reviewers for this PR**

mohanrajendran (commented):
Could we add RediSearch compatibility coverage for parameterized vector score aliases? The existing aggregate test covers a `literal AS VECTORDISTANCE`, but it does not cover `AS $SCORE`.

Please add these cases to integration/compatibility/generate.py:

```text
   1. FT.SEARCH
      KNN ... AS $SCORE
      PARAMS ... SCORE distance
      → the score field is named distance
```
```text
   2. FT.AGGREGATE
      KNN ... AS $SCORE
      SORTBY 4 @n2 ASC @distance ASC
      PARAMS ... SCORE distance
      → @n2 remains a schema field
      → @distance binds to the generated score field
```

The second case also checks that schema fields bind before a deferred score alias.

Please run ./integration/compatibility/regenerate.sh after adding the cases.

mohanrajendran (changes requested):

allenss-amazon:
Ok, I've looked at this in more depth.

I don't believe that this actually fixes the FT.AGGREGATE and FT.HYBRID cases. I believe that you'll find that if you added a test case where the substituted vector store field name was used as an input to an aggregation stage that it won't work. This same problem occurs in other places yet to come (VECTOR_RANGE is another example of this).

```
FT.AGGREGATE xxx "*=>[KNN 10 @vec AS $score_field]" FILTER "@sf>1" PARAMS 2 score_field sf
```
I believe that the parse of the FILTER stage will fail as it won't know that @sf is the score field.

The fundamental problem is the dependency between the keywords on the command line. There are two dependencies between command line parameters. First is the creation of fields as part of the query, e.q., the creation of the ephemeral score field name in a KNN query. Second is the PARAMS clause is allowed to be anywhere on the command line. 

The current PreParse/PostParse machinery only solves this problem for the first part, i.e., creation of named fields as part of the query. Currently this is just the KNN score, but the future VECTOR_RANGE can do this also. 

To properly implement the command line argument dependencies requires a two-pass scheme over the command line. In the first pass any command argument that might contain a substitutable value ($xxxx) _or depend on a substituted value (score AS $xxxx)_ must be parsed as an opaque string -- because the PARAMS substitution values may not be available yet.

In the second pass, the opaque strings can themselves be properly parsed, because the PARAMS substitution is now available.

This will require the re-write of each of the FT.AGGREGATE stages, such that during the first pass they simply copy strings into their created stage and during the second pass they actually do the appropriate compilation.

I'd recommend that we split this resolution. We can simplify this PR for just FT.SEARCH which is simpler and then open an issue for FT.AGGREGATE as this is much more complex fix.

Frank-Gu-81:
Hi @bandalgomsu 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @mohanrajendran — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

mohanrajendran (approved):

allenss-amazon (approved):
This is much more complex than is needed. 

Currently, PreParseQuery parses the "AS xxxx" field and leaves the xxxx in the parameters

PR #1093, opened 2026-06-01
chore(multi-language support): add kRelease14

neerajr0:
## Motivation
Multi-language FTS needs to be gated to version 1.4.

## Summary of Changes
- Added new kRelease14 in version.h. Multi-language FTS will be introduced in version 1.4. Version 1.3 will be added later during release time.
- Added logic to return version 1.4 in `IndexSchema::GetMinVersion()` if the schema uses a text index and specifies a non-English language

## Testing
Added unit tests in `index_schema_test.cc` to assert that correct versions are returned for different index/language/has_db_num combinations.


neerajr0:
Closing - preferring merge into fork before direct merge into upstream for development.

PR #1103, opened 2026-06-04
Add CompactPostings: space-optimized posting list container for text index

Aksha1812:
Every term in the text index owns a map from document key to a `PostingValue`:
the term frequency, the document length, and a pointer to the term's positions in
that document. Today that map is an `absl::btree_map`, which pays a fixed price no
matter how little it holds: a 24-byte inline member plus a heap node on the first
insert. In real text most terms appear in only one or two documents, so the index
carries that btree overhead across millions of terms whose maps hold a single
entry.

`CompactPostings` replaces the per-term btree with a single 8-byte container that
chooses a heap representation to match how many documents a term has. It follows
the tagged-pointer idea from `BagOfInternedStringPtrs` (#1026), adapted from a
membership *set* to an ordered key to value *map*.

## How it works

The whole container is one `uintptr_t`. Heap allocations are 8-byte aligned, so
the low 2 bits of any pointer it holds are always zero; those two bits carry a
mode tag and are masked off before the pointer is dereferenced.

```
storage_ (8 bytes)
┌──────────────────────────────────────────────┬────┐
│  pointer to heap representation (top 62 bits)  │ TT │  <- 2-bit mode tag
└──────────────────────────────────────────────┴────┘

┌───────────────┬───────────┬──────────┬──────────────────────────────────┐
│ tag / value   │ mode      │ entries  │ heap allocation                  │
├───────────────┼───────────┼──────────┼──────────────────────────────────┤
│ storage_ == 0 │ empty     │ 0        │ none                             │
│ 00 (non-zero) │ Single    │ 1        │ one Entry = 24 B                 │
│ 10            │ Double    │ 2        │ one block = 8 + 2x24 = 56 B      │
│ 01            │ SmallVec  │ 3 to 4   │ one block = 8 + 4x24 = 104 B     │
│ 11            │ Map       │ 5 +      │ absl::btree_map                  │
└───────────────┴───────────┴──────────┴──────────────────────────────────┘
```

The first insert allocates one `Entry`; the second promotes to the two-slot
`Double`, the third to the four-slot `SmallVec`, the fifth to the btree, and
erases demote back down the same ladder. All four tag values are used. The
`Double` tier exists because the baseline `btree_map` packs two entries into a
single leaf node, so a two-document term promoted straight to the four-slot array
would cost *more* than the btree it replaces; an exact two-slot block keeps that
tier a saving (see the table below).

## Why every mode stays ordered

Keys are ordered by interned-pointer identity (`InternedStringPtrLess`) in every
mode, the same order the btree used. The term-merge join in `term.cc` walks two
posting lists in lockstep, and the membership probe in `predicate.cc` seeks a
key, both via `SkipForward`, an absolute `lower_bound` over the whole collection;
iteration and seeking must behave identically whatever representation a term
holds. That is why the array tiers (`Double`, `SmallVec`) are *sorted* arrays
(binary search) and the large tier is a *btree* rather than a hash table: the
membership set in #1026 has no ordering constraint and can use unsorted arrays
and a hash set. It is also why the single-element case still allocates, a
24-byte key+value cannot pack inline into 8 bytes the way #1026 packs its
8-byte key.

## Space savings

Measured with the module's own byte counter (`search_used_memory_bytes`), which
is exact and reproducible to the byte. Both builds were compiled from the same
workspace with identical dependencies — this PR's HEAD against its parent commit,
differing only by this change — and a `TEXT` index was loaded at controlled
document frequencies, waiting for background indexing to settle before reading.
Document key, `PostingValue`, position map and interned-string storage are
byte-identical between the two builds, so the delta below isolates the posting
container alone. 100,000 distinct terms per point; the counter is deterministic,
so each figure reproduces exactly across runs.

```
┌───────────┬──────────────┬────────────────┬──────────┐
│ docs/term │ Δ B/term     │ Δ %            │ Mode     │
├───────────┼──────────────┼────────────────┼──────────┤
│ 1         │ −32.00       │ −17.3%         │ Single   │
│ 2         │ −16.00       │  −6.0%         │ Double   │
│ 3         │ −16.00       │  −4.2%         │ SmallVec │
│ 4         │ −16.00       │  −3.6%         │ SmallVec │
│ 5         │ +16.00       │  +2.6%         │ Map      │
│ 6 – 100   │ +11 to +16   │ +2.4% → +0.1%  │ Map      │
│ 100+      │ ~+11 bounded │ <0.1%, → 0     │ Map      │
└───────────┴──────────────┴────────────────┴──────────┘
```

A negative Δ is a saving. The Δ column is the container difference alone: document
and position storage are byte-identical between the two builds and cancel. Rows 1–5
are exact per-document-frequency measurements; the 6–100 and 100+ rows are the
measured tail (anchors: df=10 → +16.00, df=50 → +10.88, df=100 → +10.88), presented
as a range because the constant does not vary with corpus content.

The win is concentrated at one to four documents, where — in natural-language text
— the large majority of distinct terms live. A single-document term drops from the
baseline's 24-byte inline btree member plus first heap node to an 8-byte container
plus one 24-byte `Entry`, a flat 32 B saved; the two-, three- and four-document
tiers save 16 B each.

At five or more documents both builds hold the same `absl::btree_map` and allocate
the same nodes, so those cancel; the container then differs only by its header — the
baseline's 24-byte inline member versus this container's 8-byte pointer plus the
heap-rounded btree object. That difference is a **small bounded constant, worst-case
+16 B around df 5–10, decaying to ~+11 B (+0.12%) by df=100** as the growing
per-term footprint dwarfs it. It never grows with document frequency. This is the
same structural top-tier cost `BagOfInternedStringPtrs` (#1026) carries when it
falls back to a heap set at nine-plus elements, and for the same reason: an 8-byte
container cannot inline a multi-word map object. It is bounded, shrinking as a
fraction, and paid only by the rare high-frequency terms — while the common
one-to-four-document mass, where the memory actually lives, is where this change
wins.

### Whole-index result on real corpora

The per-tier table above holds document frequency constant. To measure the *net*
effect on natural text, the same two builds indexed two whole corpora and the
`search_used_memory_bytes` counter was read after the background indexer settled.
Both builds indexed byte-identical documents, so the entire difference is the
posting container.

**Real corpus — 5,000 Simple English Wikipedia articles** (avg 818 tokens/doc,
4.09 M tokens, 143,189 distinct terms, ~54% hapax):

```
  baseline:         224,503,152 B
  CompactPostings:  223,644,720 B
  saving:               858,432 B  =  0.82 MiB  (−0.38%)  ≈  6.0 B / distinct term
```

**Synthetic corpus — Zipf-1.1, 20,000 docs × 100 tokens** (149,403 distinct terms,
~51% hapax), included to show the same measurement on a controlled distribution:

```
  baseline:         169,996,896 B
  CompactPostings:  167,622,032 B
  saving:             2,374,864 B  =  2.27 MiB  (−1.40%)  ≈ 15.9 B / distinct term
```

Both are real savings on identical corpora, but note the **gap between them**: the
synthetic figure is ~4× larger per term. Real full-length articles carry long
position lists (position and document storage — untouched by this PR — dominate a
text index's memory, ≈56 B per indexed token here), so the container-header slice
this change optimizes is a thinner fraction of the whole; and natural English has a
heavier high-frequency tail (more terms in the small-but-nonzero Map-tier cost).
The clean per-df deltas above do **not** sum to the whole-index saving — the honest,
directly-measured real-text number is ≈6 B per distinct term. It is a genuine saving
in pure container overhead that holds no user data, and it scales linearly with
vocabulary, but it is modest relative to total index size.

## Testing

`compact_postings_test.cc` adds white-box coverage of the container: all modes
(`Single`, `Double`, `SmallVec`, `Map`), both growth and shrink transitions
across every tier boundary, ordered iteration in each mode, transparent lookup
with a borrowed key, `SkipForward` hits and misses, the key ref-count lifecycle
through insert and erase, and move semantics. The container is exercised in situ
through the existing `indexes_test` and `text_index_test`, which pass across all
modes, and the whole suite is clean under ASAN with no leaks or errors.


Frank-Gu-81:
Hi @Aksha1812 👋 — this is tracked as a **P2** for **valkey-search 1.3**. P2s aren't RC1 blockers, but we'd love to land them for **GA**.

First-pass reviewer: @mohanrajendran — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

Aksha1812:
Hey @allenss-amazon  I tried applying the same tagged-pointer / tiered-container idea from #1026 to the postings object, replacing the per-term absl::btree_map with an 8-byte container that picks a heap layout based on how many docs a term has (Single/Double/SmallVec/Map) here. 

It works and it does save memory, but the practical benefit came out much smaller than in #1026, and I think the reason is structural. In BagOfInternedStringPtrs the elements are just 8-byte InternedStringPtr keys and it's a membership set, so your Array4/Array8 tiers could pack the elements inline in the object itself, no heap allocation at all for the small cases. That's where the real saving came from.

Postings can't do that. Each element is a key plus a value (a 24-byte Entry: key ptr + tf + doc_len + positions ptr), and the map has to stay ordered, so even the single-doc case has to heap-allocate, nothing fits inline in the 8-byte word. So all I'm really saving is the container header (btree's 24-byte inline member → 8-byte tagged pointer), not the elements. Measured with the module's own search_used_memory_bytes on two builds differing only by this change:

- Real corpus (5,000 Simple English Wikipedia articles, ~143K terms): −0.38%, ~6 B per distinct term.
- Synthetic Zipf: −1.40%, ~16 B per distinct term.

I added a table as well in the PR description for more info. The win is concentrated on single-doc terms, but position/document storage dominates a text index's memory and this change doesn't touch it.

Given the modest real-world gain, I am thinking of closing the PR as a <1% container-overhead reduction doesn't seem worth landing. Will close the PR soon and focus on if it is possible to improve the FlatPositionMap more instead. Will close PR after a few days if no further suggestions.

allenss-amazon:
Agree < 1% isn't worth landing.

I suspect this needs to get combined with another space optimization to replace InternStringPtr in indexes with a dense integer before it's effective.

Aksha1812:
> Agree < 1% isn't worth landing.
> 
> I suspect this needs to get combined with another space optimization to replace InternStringPtr in indexes with a dense integer before it's effective.

@allenss-amazon Ack . closing this PR for now

PR #1115, opened 2026-06-09
Fix SIGSEGV on FT.SEARCH query with empty OR right-hand side

### LINKED ISSUE #1105: FT.SEARCH/FT.AGGREGATE query with empty OR right-hand side ('@n:[0 10]|' or 'hello|the') crashes server (SIGSEGV)
### Description

An `FT.SEARCH` (or `FT.AGGREGATE`) query whose OR operator `|` has an empty right-hand side crashes the server with a SIGSEGV (null-pointer dereference) in the filter parser. No indexed data is required — the crash happens at query-parse time. Reachable by any client that can run `FT.SEARCH` (a `@read` command).

### Steps to reproduce

```
FT.CREATE idx ON HASH PREFIX 1 doc: SCHEMA n NUMERIC t TEXT
FT.SEARCH idx '@n:[0 10]|'      # trailing '|', empty OR right-hand side
FT.SEARCH idx 'hello|the'       # OR where the right operand is a stopword ('the')
```

Each drops the connection (`Error: Server closed the connection`). `docker inspect` ExitCode 139.

The general shape is any valid predicate immediately followed by an unescaped `|` whose right-hand side resolves to nothing: end-of-input, only whitespace, a closing `)`, or a stop word (the tokenizer drops stop words, yielding a null token). `'@n:[0 10]||'` does NOT crash — the inner recursion hits the second `|` with a null left operand and returns a graceful "Missing OR term" error, so the crash specifically requires the RHS to terminate empty.

### Crash output

```
valkey ... crashed by signal: 11, si_code: 1
libsearch.so(_ZN13valkey_search12FilterParser13WrapPredicate...+0x286)
  at src/commands/filter_parser.cc:536
  <- FilterParser::ParseExpression
  <- FilterParser::Parse
  <- query::ParsePreFilter at src/query/search.cc
```

### Root cause

`src/commands/filter_parser.cc`, `ParseExpression`, the `'|'` branch: it parses the right operand into `predicate`, then guards only the LEFT operand (`if (result.prev_predicate) ... else return "Missing OR term"`). It never checks that the right operand is non-null. A trailing `|` (RHS recursion returns null) or a stop-word RHS (`ParseUnquotedTextToken` returns a null token) passes a null `new_predicate` into `WrapPredicate`. There (around `filter_parser.cc:487`), the OR-flatten path dereferences `new_predicate->GetType()` on the null pointer, and a sibling path builds a `ComposedPredicate` owning a null child that crashes on destruction.

### Suggested fix

After obtaining the RHS in the `|` branch, return `InvalidArgumentError("Missing OR term")` if it is null (mirror the existing left-operand guard), and/or null-check `new_predicate` in `WrapPredicate`.

### Environment

- Confirmed live on current `main`: `filter_parser.cc` is untouched by the recent expr.cc crash fixes.

---

Aksha1812:
The ParseExpression function's '|' branch validated the left operand but never checked that the right operand (from the recursive call) was non-null. A trailing '|' or stopword-only RHS caused a null pointer to flow into WrapPredicate → MayNegatePredicate → GetType() dereference → SIGSEGV.

Add null-check for the right-hand predicate after the recursive ParseExpression call, returning "Missing OR term" error.

Add unit tests for trailing pipe variants:
- @field:[range]|       (end of input)
- @field:[range] |      (trailing with space)
- (@field:[range]|)     (inside parentheses)

Fixes #1105

### DIFF
--- a/src/commands/filter_parser.cc
+++ b/src/commands/filter_parser.cc
@@ -975,6 +975,9 @@ absl::StatusOr<FilterParser::ParseResult> FilterParser::ParseExpression(
       }
       VMSDK_ASSIGN_OR_RETURN(auto sub_result, ParseExpression(level));
       predicate = std::move(sub_result.prev_predicate);
+      if (!predicate) {
+        return absl::InvalidArgumentError("Missing OR term");
+      }
       if (result.prev_predicate) {
         node_count_++;
       } else {
--- a/testing/filter_test.cc
+++ b/testing/filter_test.cc
@@ -1627,6 +1627,25 @@ INSTANTIATE_TEST_SUITE_P(
             .create_success = false,
             .create_expected_error_message = "Missing OR term",
         },
+        {
+            .test_name = "or_with_missing_right_operand_trailing_pipe",
+            .filter = "@num_field_1.5:[1.0 2.0]|",
+            .create_success = false,
+            .create_expected_error_message = "Missing OR term",
+        },
+        {
+            .test_name = "or_with_missing_right_operand_trailing_pipe_space",
+            .filter = "@num_field_1.5:[1.0 2.0] |",
+            .create_success = false,
+            .create_expected_error_message = "Missing OR term",
+        },
+        {
+            .test_name =
+                "or_with_missing_right_operand_trailing_pipe_in_parens",
+            .filter = "(@num_field_1.5:[1.0 2.0]|)",
+            .create_success = false,
+            .create_expected_error_message = "Missing OR term",
+        },
         // =================================================================
         // Tests for escaped pipe separator in tag values
         // =================================================================


Aksha1812:
@allenss-amazon Compatibility question . Do we need to make the error message compatible as well with redis .

RedisSearch
```
127.0.0.1:6379> FT.SEARCH idx '@n:[0 10]|'
(error) Syntax error at offset 9 near 10
127.0.0.1:6379> FT.SEARCH idx '| @n:[0 10]'
(error) Syntax error at offset 0 near 
```

ValkeySearch
```
127.0.0.1:6399> FT.SEARCH idx '@n:[0 10]|'
(error) Invalid filter expression: `@n:[0 10]|`. Missing OR term
127.0.0.1:6399> FT.SEARCH idx '| @n:[0 10]'
(error) Invalid filter expression: `| @n:[0 10]`. Missing OR term
```

allenss-amazon (approved):

PR #1117, opened 2026-06-09
Fix GetTargetsForSlot to resolve slots in the middle/end of a range

allenss-amazon:
GetTargetsForSlot used a lower_bound() lookup against a map keyed by each range's start slot, so it only landed on the correct range when the slot equaled a range start. For any other slot it jumped to the next range and returned an empty target list, and it also used an exclusive end-slot check. As a result, single-slot queries for hash tags whose slot fell in the middle or end of a range failed with "No available nodes to execute the query" from every node, including the owner.

Replace the lookup with the same correct upper_bound()/decrement and inclusive [start, end] range check already used by GetShardBySlot. Add unit tests covering the start, middle, and end of the first, middle, and last ranges, plus gap handling; these fail on the old logic and pass now.

### DIFF
--- a/vmsdk/src/cluster_map.cc
+++ b/vmsdk/src/cluster_map.cc
@@ -140,15 +140,21 @@ std::vector<NodeInfo> ClusterMap::GetTargetsForSlot(FanoutTargetMode mode,
   if (slot_to_shard_map_.empty()) {
     return {};
   }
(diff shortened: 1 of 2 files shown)


allenss-amazon:
Fixes #1118

boda26 (approved):

PR #1120, opened 2026-06-10
Guard against disengaged optional in InTrackedMutationRecords

allenss-amazon:
ConsumeTrackedMutatedAttribute may leave a tracked_mutated_records_ entry in place with attributes = std::nullopt (the "else" branch at index_schema.cc:1997-2008 moves the attribute map out and disengages the optional but does not erase the entry). A subsequent call to InTrackedMutationRecords on the same key would then dereference the empty optional via `attributes->find(...)`, which is UB. Under ASAN this surfaced as a SIGSEGV from absl's SwissTable iterator-generation guard during the backfill scan path, observed in CI on a fulltext test (test_proximity_predicate).

The fix is a one-line has_value() check; the semantically correct return is false (no pending mutation for this key/identifier).

Adds InTrackedMutationRecordsAfterConsumeNoCrash to document and lock in the contract. The test exercises the exact code path that crashed in CI but does not, by itself, reliably trigger the crash on the unfixed code: after a single track/consume cycle the moved-from flat_hash_map inside the optional has its generation_ptr_ reset to absl's static EmptyGeneration() sentinel, so the dereference happens to read valid memory. The production crash needed a more complex sequence (multi- mutation cycles plus concurrent backfill). The test still asserts the correct behavior and catches future regressions in the has_value() check or the consume/disengage sequence.


Aksha1812 (approved):
Approved . change looks good . Can we add shorter comments replacing long AI ones

PR #1122, opened 2026-06-11
Fix FLAT BLOCK_SIZE 0 livelock and bound INITIAL_CAP/BLOCK_SIZE (#1107, #1110)

### LINKED ISSUE #1107: FT.CREATE HNSW INITIAL_CAP has no upper bound: single command OOM-kills server (SIGKILL)
### Description

`FT.CREATE` with an HNSW vector index accepts an arbitrarily large `INITIAL_CAP` with no upper bound. The value flows straight into the hnswlib constructor, which allocates structures sized to it up front, before any vector is indexed. A single `FT.CREATE` command can request tens of GB and get the server OOM-killed (SIGKILL).

### Steps to reproduce

```
FT.CREATE oomtest SCHEMA v VECTOR HNSW 8 DIM 4 TYPE FLOAT32 DISTANCE_METRIC L2 INITIAL_CAP 2147483647
```

Server dies; `docker inspect` ExitCode 137, `OOMKilled=true`. On a memory-capped instance it OOMs at far lower values too (e.g. `INITIAL_CAP 20000000` on a 512MB instance). FLAT with the same `INITIAL_CAP` returns OK (it does not preallocate the same way), isolating the cause to the HNSW constructor.

### Root cause

`FTCreateVectorParameters::Verify()` (`src/commands/ft_create_parser.cc`, ~line 786) only lower-bounds `INITIAL_CAP` (`initial_cap <= 0` rejected). There is no upper bound, unlike DIM/M/EF_CONSTRUCTION/EF_RUNTIME which are range-checked against configured maxima. The value flows to `HierarchicalNSW(space, max_elements=initial_cap, ...)` in `src/indexes/vector_hnsw.cc`. The constructor (`third_party/hnswlib/hnswalg.h`) immediately sizes `link_list_locks_` (a `vector<mutex>`, ~40B each), `element_levels_`, a `VisitedListPool`, and `ChunkedArray`s to `max_elements` — ~85GB for `link_list_locks_` alone at 2.1e9. The OS/cgroup OOM killer SIGKILLs the process before `std::bad_alloc` can be caught and turned into a graceful error.

### Suggested fix

Range-check `INITIAL_CAP` against a configured maximum in `FTCreateVectorParameters::Verify()` (mirror the DIM/M/EF validation).

### Environment

- Confirmed live on current `main`: `Verify()` still has only the `initial_cap <= 0` check.

---

### LINKED ISSUE #1110: FT.CREATE FLAT BLOCK_SIZE 0 accepted with no validation -> infinite ingestion resize loop (worker livelock DoS)
### Description

A FLAT vector index created with `BLOCK_SIZE 0` accepts the value with no validation, then enters an infinite resize loop on ingestion once capacity fills. This is not a server crash — the main thread stays responsive (PING works) — but a mutation/indexing worker thread is pinned at 100% CPU forever and the index never ingests past `INITIAL_CAP`. A resource-exhaustion / livelock DoS of the search ingestion subsystem.

### Steps to reproduce

```
FT.CREATE bs0 ON HASH PREFIX 1 doc: SCHEMA v VECTOR FLAT 10 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2 BLOCK_SIZE 0 INITIAL_CAP 1
# insert valid 8-byte FLOAT32 vectors past INITIAL_CAP (use a RESP client; the CLI shell mangles binary):
# HSET doc:1 v <8 bytes>, HSET doc:2 v <8 bytes>, ...
```

After the inserts: container CPU pinned at ~100%, `FT.INFO` `num_records` frozen, and the log repeats once per second:

```
src/indexes/vector_flat.cc:146: Resizing FLAT Index, current size: 1, expand by: 0
```

`BLOCK_SIZE 0` with the default `INITIAL_CAP 10240` also triggers it after 10240 inserts; `INITIAL_CAP 1` just makes it a 2-insert repro.

### Root cause

`FlatParameters` (`src/commands/ft_create_parser.cc` / `.h`) has no `Verify()` / range check on `block_size` (unlike HNSW M/EF and INITIAL_CAP). `BLOCK_SIZE 0` is accepted and stored. In `src/indexes/vector_flat.cc`, `AddRecordImpl` wraps `addPoint` in a `do { ... } while(true)` loop; on a full index hnswlib throws "number of elements exceeds the specified limit", the catch calls `ResizeIfFull()`, which computes `resizeIndex(GetCapacity() + block_size_)` = `resizeIndex(capacity + 0)` — a no-op — and `continue`s, so `addPoint` throws again forever.

### Suggested fix

Reject `block_size == 0` in `FlatParameters::Verify()` (and/or require a strictly positive growth in `ResizeIfFull`).

### Environment

- Confirmed live on current `main`: no recent commit touches `vector_flat.cc`.

Note: classified as a livelock/DoS, not a crash — the server process stays up and the main thread keeps answering, but the ingestion worker is wedged for the life of the process.

---

Aksha1812:
Two FT.CREATE vector-parameter validation gaps let a single command take the server down. Both are fixed here by adding the missing bounds in the parameter `Verify()` path.

## #1110 — FLAT `BLOCK_SIZE 0` livelock

`FlatParameters` had no `Verify()` override, so `block_size` was never validated. A FLAT index created with `BLOCK_SIZE 0` was accepted, and once it filled to capacity `ResizeIfFull()` computed `resizeIndex(capacity + 0)` — a no-op — so `AddRecordImpl`'s `do { addPoint(); } while(true)` loop spun forever on the mutation worker thread (100% CPU, ingestion permanently wedged) while the main thread stayed responsive.

- Add `FlatParameters::Verify()` that range-checks `block_size` (rejecting `0`).
- Add a defensive guard in `ResizeIfFull()` that returns an error instead of spinning if `block_size_` is ever `0`, so the loop terminates via `VMSDK_RETURN_IF_ERROR` even for an already-persisted index.

## #1107 — `INITIAL_CAP` has no upper bound (OOM)

`FTCreateVectorParameters::Verify()` only lower-bounded `INITIAL_CAP` (`<= 0`). The value flows straight into the hnswlib constructor, which preallocates per-element structures up front, so a single `FT.CREATE ... INITIAL_CAP 2147483647` could OOM-kill the server (SIGKILL).

- Range-check `INITIAL_CAP` against a fixed maximum (`kMaxInitialCap`), mirroring the existing `DIM`/`M`/`EF_CONSTRUCTION`/`EF_RUNTIME` validation.
- Apply the same fixed-max approach to `BLOCK_SIZE` (`kMaxBlockSize`).

## Approach note

This uses **fixed constant maximums** validated via `vmsdk::VerifyRange`, the same idiom already used for the other vector parameters in this file — not a new config or a memory-derived limit. Rationale:
- No new config, so it's safe to backport (e.g. to 1.2) without adding API surface.
- A fixed cap is type-agnostic, so the base-class check is correct for both FLAT and HNSW.

`kMaxInitialCap`/`kMaxBlockSize` are currently `10,000,000` — generous enough not to constrain real usage, low enough to prevent the OOM. Open to a different value if preferred.

## Tests

- `invalid_block_size_zero` — `BLOCK_SIZE 0` rejected.
- `invalid_block_size_too_big` — `BLOCK_SIZE` above max rejected.
- `invalid_initial_cap_too_big` — `INITIAL_CAP` above max rejected.

Fixes #1107
Fixes #1110


allenss-amazon (commented):
Would we accept a block_size of 1 ?? Seems like yes and this is foolish. Let's establish a minimum size that's like 512.

Aksha1812:
> Would we accept a block_size of 1 ?? Seems like yes and this is foolish. Let's establish a minimum size that's like 512.

@allenss-amazon Do we want to enforce that via a config or just in the Verify block ? 

Also worth noting here RedisSearch accepts 1 

RedisSearch output

```
127.0.0.1:6379> FT.CREATE bs0 ON HASH PREFIX 1 doc: SCHEMA v VECTOR FLAT 10 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2 BLOCK_SIZE 0 INITIAL_CAP 1
(error) Bad arguments for vector similarity FLAT index `BLOCK_SIZE`: Value is outside acceptable bounds
127.0.0.1:6379> FT.CREATE bs0 ON HASH PREFIX 1 doc: SCHEMA v VECTOR FLAT 10 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2 BLOCK_SIZE 1 INITIAL_CAP 1
OK
127.0.0.1:6379> ft.dropindex bs0
OK
127.0.0.1:6379> FT.CREATE bs0 ON HASH PREFIX 1 doc: SCHEMA v VECTOR FLAT 10 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2 BLOCK_SIZE 10000000 INITIAL_CAP 1
OK
127.0.0.1:6379> ft.dropindex bs0
OK
127.0.0.1:6379> FT.CREATE bs0 ON HASH PREFIX 1 doc: SCHEMA v VECTOR FLAT 10 DIM 2 TYPE FLOAT32 DISTANCE_METRIC L2 BLOCK_SIZE 100000000 INITIAL_CAP 1
(error) Vector index block size 100000000 exceeded server limit (17062946 with the given parameters)
127.0.0.1:6379> FT.CREATE bs0 ON HASH PREFIX 1 doc: SCHEMA v VECTOR FLAT 10 DIM 8 TYPE FLOAT32 DISTANCE_METRIC L2 BLOCK_SIZE 100000000 INITIAL_CAP 1
(error) Vector index block size 100000000 exceeded server limit (8531473 with the given parameters)
127.0.0.1:6379> FT.CREATE bs0 ON HASH PREFIX 1 doc: SCHEMA v VECTOR FLAT 10 DIM 9 TYPE FLOAT32 DISTANCE_METRIC L2 BLOCK_SIZE 100000000 INITIAL_CAP 1
(error) Vector index block size 100000000 exceeded server limit (7875205 with the given parameters)
127.0.0.1:6379> 
```

So i think the upper bound is a memory derived limit and tunes based on the dimension of vector  mentioned during ft.create

Aksha1812:
@allenss-amazon added a simpler fix for now. let me know if we need to modify the constant values.

allenss-amazon (approved):

PR #1138, opened 2026-06-13
    Fix ThreadPool suspend/resume hang from stale thread count

allenss-amazon:
SuspendWorkers/ResumeWorkers sized their BlockingCounter from the raw
    threads_.Size(), which after #1062 includes workers marked for shutdown
    but not yet reaped. Those workers exit without acknowledging
    suspension, so the counter never reaches zero and Wait() hangs.

    Use ThreadPool::Size(), which counts only active (non-shutdown,
    non-joinable) workers, matching the count already used by Resize().


allenss-amazon:
closed

PR #1146, opened 2026-06-15
Fix FT.INTERNAL_UPDATE crashes during AOF/RDB load (#1108, #1109)

### LINKED ISSUE #1108: FT.INTERNAL_UPDATE crashes server (SIGSEGV) on AOF load even with a valid entry; module writes these itself -> restart crash-loop
### Description

A node crashes with a SIGSEGV while loading its AOF if the AOF contains an `FT.INTERNAL_UPDATE` entry — even a perfectly valid (empty) one. The module itself writes `FT.INTERNAL_UPDATE` to the AOF for index-metadata persistence in coordinator mode (`MetadataManager::ReplicateFTInternalUpdate`), so a node that persisted index metadata can crash-loop on restart / AOF reload.

### Steps to reproduce

Build a multi-part AOF containing a valid (empty) `FT.INTERNAL_UPDATE` and load it:

```
mkdir -p /tmp/poisonaof/appendonlydir
printf 'file appendonly.aof.1.incr.aof seq 1 type i\n' > /tmp/poisonaof/appendonlydir/appendonly.aof.manifest
printf '*2\r\n$6\r\nSELECT\r\n$1\r\n0\r\n*4\r\n$18\r\nFT.INTERNAL_UPDATE\r\n$3\r\nidx\r\n$0\r\n\r\n$0\r\n\r\n' > /tmp/poisonaof/appendonlydir/appendonly.aof.1.incr.aof
docker run -d --name n -v /tmp/poisonaof:/data -e VALKEY_EXTRA_FLAGS="--appendonly yes" valkey/valkey-bundle:unstable
sleep 5
docker inspect n --format='{{.State.ExitCode}}'   # -> 139
```

The entry is a valid empty protobuf, so this is not a malformed-input problem.

### Crash output

```
valkey ... crashed by signal: 11, si_code: 1
  TriggerCallbacks at src/coordinator/metadata_manager.cc:168
  <- CreateEntryOnReplica at src/coordinator/metadata_manager.cc:925
  <- FTInternalUpdateCmd at src/commands/ft_internal_update.cc:86
  <- loadSingleAppendOnlyFile
```

### Root cause

Under `VALKEYMODULE_CTX_FLAGS_LOADING`, `FTInternalUpdateCmd` (`src/commands/ft_internal_update.cc:82-88`) calls `MetadataManager::CreateEntryOnReplica`, which calls `TriggerCallbacks` (`src/coordinator/metadata_manager.cc:165-168`). At metadata_manager.cc:168 the registered `update_callback` is invoked, but the callback / schema-manager state is not safely usable during early AOF replay, producing a null/invalid dereference. This is a load-path lifecycle bug, not input validation.

### Suggested fix

Guard the callback/schema-manager state before invoking `update_callback` during loading, or defer metadata-entry application until load completes.

### Environment

- Confirmed live on current `main`: no recent commit touches `ft_internal_update.cc` or `metadata_manager.cc`.

Related: see the companion issue on corrupt-AOF `FT.INTERNAL_UPDATE` handling.

---

### LINKED ISSUE #1109: Corrupt FT.INTERNAL_UPDATE AOF entry aborts load by default (SIGABRT); skip-corrupted flag is broken and still SIGSEGVs
### Description

Two related failures in how `FT.INTERNAL_UPDATE` handles a corrupt entry during AOF load:

1. **Default config aborts the load.** A malformed-protobuf `FT.INTERNAL_UPDATE` entry in the AOF hits `CHECK(false)` (SIGABRT) when `search.skip-corrupted-internal-update-entries=no` (the default). This crash-loops the node on every startup until the AOF is hand-edited.
2. **The skip flag does not actually skip.** Setting `search.skip-corrupted-internal-update-entries=yes` (the intended mitigation) logs "SKIPPING" and then still SIGSEGVs, because the skip branch returns `OkStatus()` and execution falls through into the same `CreateEntryOnReplica` -> `TriggerCallbacks` crash with a default-constructed entry.

So a corrupt AOF crashes the node by default, and the documented escape hatch is broken.

### Steps to reproduce

```
mkdir -p /tmp/poisonaof/appendonlydir
printf 'file appendonly.aof.1.incr.aof seq 1 type i\n' > /tmp/poisonaof/appendonlydir/appendonly.aof.manifest
# SELECT 0 + FT.INTERNAL_UPDATE poisonidx <0xFF*6> <0xFF*6> (argv[2]/argv[3] not valid protobuf)
printf '*2\r\n$6\r\nSELECT\r\n$1\r\n0\r\n*4\r\n$18\r\nFT.INTERNAL_UPDATE\r\n$9\r\npoisonidx\r\n$6\r\n\xff\xff\xff\xff\xff\xff\r\n$6\r\n\xff\xff\xff\xff\xff\xff\r\n' > /tmp/poisonaof/appendonlydir/appendonly.aof.1.incr.aof

# (1) default -> SIGABRT
docker run -d --name n1 -v /tmp/poisonaof:/data -e VALKEY_EXTRA_FLAGS="--appendonly yes" valkey/valkey-bundle:unstable
# (2) skip flag set -> still SIGSEGV
docker run -d --name n2 -v /tmp/poisonaof:/data -e VALKEY_EXTRA_FLAGS="--appendonly yes --search.skip-corrupted-internal-update-entries yes" valkey/valkey-bundle:unstable
```

### Crash output

```
# (1) default:
F0000 ... ft_internal_update.cc:46] Check failed: false Internal update failure during AOF loading - cannot continue
valkey ... crashed by signal: 6, si_code: -6

# (2) skip=yes:
ft_internal_update.cc:40: SKIPPING corrupted AOF entry due to configuration   (x2)
valkey ... crashed by signal: 11, si_code: 1
  TriggerCallbacks at src/coordinator/metadata_manager.cc:168 <- CreateEntryOnReplica
```

### Root cause

`src/commands/ft_internal_update.cc`, `HandleInternalUpdateFailure`:
- When `LOADING` and skip flag is false (default), it runs `CHECK(false)` at ft_internal_update.cc:46 -> abort.
- When skip flag is true, it logs "SKIPPING" and returns `absl::OkStatus()`. The caller wraps it in `VMSDK_RETURN_IF_ERROR`, which does NOT short-circuit on OK, so `FTInternalUpdateCmd` continues with the un-parsed default `metadata_entry` and (because `LOADING` is set) calls `CreateEntryOnReplica` -> `TriggerCallbacks` -> the same SIGSEGV as the companion issue. The skip path never actually skips the rest of the command body.

### Suggested fix

The skip branch must short-circuit the command (return early / signal "entry skipped"), not return `OkStatus` and fall through. Separately, reconsider aborting via `CHECK(false)` on corrupt persisted data — a corrupt AOF entry should not be a fatal assertion.

### Environment

- Confirmed live on current `main`: `ft_internal_update.cc` is unchanged.

Related: see the companion issue on `FT.INTERNAL_UPDATE` crashing on a *valid* AOF entry.

---

Aksha1812:
FT.INTERNAL_UPDATE is registered unconditionally, but MetadataManager is only initialized when the coordinator is enabled (cluster mode). A node that persisted FT.INTERNAL_UPDATE entries to its AOF and is later loaded without the coordinator (e.g. standalone) ran FTInternalUpdateCmd, which called MetadataManager::Instance() on an uninitialized (null) instance and dereferenced it via CreateEntryOnReplica -> TriggerCallbacks -> SIGSEGV. This happened even for a valid (empty) entry (#1108).

Corrupt entries (#1109) had two more problems:
- Default config hit CHECK(false) -> SIGABRT. A corrupt persisted entry must not be a fatal assertion.
- The skip-corrupted flag returned OkStatus from a helper wrapped in VMSDK_RETURN_IF_ERROR, which does not short-circuit on OK, so execution fell through into CreateEntryOnReplica with a default-constructed entry and hit the same SIGSEGV. The flag never actually skipped the command.

Fixes:
- Guard FTInternalUpdateCmd with MetadataManager::IsInitialized() (matching the existing pattern in server_events.cc); reply OK and return when there is no coordinator to apply the entry to.
- Make parse failures return immediately instead of falling through; replace CHECK(false) with a recoverable DataLossError that names the skip flag.
- Skip TriggerCallbacks in CreateEntryOnReplica during loading; the schema manager is not ready during early replay, and OnLoadingEnded already re-applies fingerprints/versions afterward (matching the staged-reconcile path used for RDB replication).

Verified by reproducing all three crashes on a standalone server with a crafted AOF, then confirming the fixed build survives each: valid entry loads, corrupt entry with skip=yes logs SKIPPING and starts, corrupt entry with default config returns a recoverable error instead of aborting.

Fixes #1108
Fixes #1109

### DIFF
--- a/src/commands/ft_internal_update.cc
+++ b/src/commands/ft_internal_update.cc
@@ -16,14 +16,10 @@ namespace valkey_search {
 
 constexpr int kFTInternalUpdateArgCount = 4;
 
-// Handles a parse/processing failure for an FT.INTERNAL_UPDATE entry. Returns a
-// status that the caller propagates by returning immediately (it never falls
-// through to applying a half-parsed entry).
-//
-// During AOF/RDB loading a failure is either skipped (when the operator opts in
-// via search.skip-corrupted-internal-update-entries) or surfaced as a
-// recoverable error. Corrupt persisted data must not abort the process, so this
-// never uses a fatal CHECK.
+// Handles an FT.INTERNAL_UPDATE failure. The caller must return its result
+// immediately (never fall through to applying a half-parsed entry). During
+// loading, a failure is either skipped (if the skip-corrupted config is set) or
+// returned as a recoverable error -- corrupt persisted data must not abort.
 absl::Status HandleInternalUpdateFailure(ValkeyModuleCtx *ctx,
                                          const std::string &operation_type,
                                          const std::string &id,
@@ -48,9 +44,8 @@ absl::Status HandleInternalUpdateFailure(ValkeyModuleCtx *ctx,
       return absl::OkStatus();
     }
     return absl::DataLossError(
-        "Corrupt FT.INTERNAL_UPDATE entry encountered during loading. Set "
-        "search.skip-corrupted-internal-update-entries=yes to skip such "
-        "entries, or repair/remove the offending AOF entry.");
+        "Corrupt FT.INTERNAL_UPDATE entry during loading; set "
+        "search.skip-corrupted-internal-update-entries=yes to skip it");
   }
 
   return error_status;
@@ -82,11 +77,10 @@ absl::Status FTInternalUpdateCmd(ValkeyModuleCtx *ctx,
             "Failed to parse GlobalMetadataVersionHeader"));
   }
 
-  // FT.INTERNAL_UPDATE only carries coordinator metadata. The MetadataManager
-  // is only initialized when the coordinator is enabled (cluster mode). If a
-  // node persisted these entries to its AOF and is later loaded without the
-  // coordinator (e.g. standalone), there is no MetadataManager to apply them
-  // to, so skip rather than dereference an uninitialized instance.
+  // The MetadataManager only exists when the coordinator is enabled (cluster
+  // mode). If these entries were persisted to the AOF and later loaded without
+  // the coordinator (e.g. standalone), skip rather than dereference the
+  // uninitialized instance.
   if (!coordinator::MetadataManager::IsInitialized()) {
     ValkeyModule_ReplyWithSimpleString(ctx, "OK");
     return absl::OkStatus();
--- a/src/commands/ft_internal_update.cc
+++ b/src/commands/ft_internal_update.cc
@@ -16,15 +16,18 @@ namespace valkey_search {
 
 constexpr int kFTInternalUpdateArgCount = 4;
 
-// Helper function to handle parse failures with poison pill recovery
-absl::Status HandleInternalUpdateFailure(
-    ValkeyModuleCtx *ctx, const std::string &operation_type,
-    const std::string &id,
-    const absl::Status &error_status = absl::OkStatus()) {
-  if (error_status.ok()) {
-    return absl::OkStatus();
-  }
-
+// Handles a parse/processing failure for an FT.INTERNAL_UPDATE entry. Returns a
+// status that the caller propagates by returning immediately (it never falls
+// through to applying a half-parsed entry).
+//
+// During AOF/RDB loading a failure is either skipped (when the operator opts in
+// via search.skip-corrupted-internal-update-entries) or surfaced as a
+// recoverable error. Corrupt persisted data must not abort the process, so this
+// never uses a fatal CHECK.
+absl::Status HandleInternalUpdateFailure(ValkeyModuleCtx *ctx,
+                                         const std::string &operation_type,
+                                         const std::string &id,
+                                         const absl::Status &error_status) {
   VMSDK_LOG(WARNING, ctx) << "CRITICAL: " << operation_type
                           << " failed in FT.INTERNAL_UPDATE. "
                           << "Index ID: " << vmsdk::config::RedactIfNeeded(id);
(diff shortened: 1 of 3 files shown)


src/commands/ft_internal_update.cc
```diff
@@ -65,27 +63,39 @@ absl::Status FTInternalUpdateCmd(ValkeyModuleCtx *ctx,
...
-            "Failed to parse GlobalMetadataVersionHeader")));
+            "Failed to parse GlobalMetadataVersionHeader"));
+  }
+
+  // The MetadataManager only exists when the coordinator is enabled (cluster
+  // mode). If these entries were persisted to the AOF and later loaded without
+  // the coordinator (e.g. standalone), skip rather than dereference the
+  // uninitialized instance.
+  if (!coordinator::MetadataManager::IsInitialized()) {
+    ValkeyModule_ReplyWithSimpleString(ctx, "OK");
```
allenss-amazon: Hmm, I don't like it when error conditions just vanish. At a minimum we should be logging this case -- it's solid evidence of a mis-configuration or some other kind of error. I would argue that it should return an error also. In theory that's skippable -- right?

src/coordinator/metadata_manager.cc
allenss-amazon: Let's refactor the knowledge of the CHECK above this with the test that you're adding down below.
I'd convert the if (!Loading) into:
```
if (.... SLAVE) {
  ... TriggerCallbacks....
} else {
   CHECK(...LOADING...)
}
```
Much cleaner.

allenss-amazon (commented):
I'm about to merge #868 which will create a merge conflict that'll be easy to fix.

allenss-amazon (approved):

PR #1148, opened 2026-06-15
Fix tag query parsing to support escaped closing brace (#979) + resolved merge conflicts

Aksha1812:
Duplicate of https://github.com/valkey-io/valkey-search/pull/979  + Resolved merge conflicts

  There was exactly one text file with conflicts — integration/compatibility/generate.py, with 3 conflict hunks:

 | Hunk | What clashed | Resolution |
  |------|--------------|------------|
  | 1 | PR disabled `test_bad_numeric_data` with a `'''` comment-block; main disabled it with `@pytest.mark.skip` | Took main (newer form), dropped PR's edit |
  | 2 | Same drift around `test_search_reverse` / `test_search` (PR's closing `'''`) | Took main |
  | 3 | PR adds new `test_tag_escaped_special_chars`; main has nothing there | Took PR |

PR which caused the conflicts  PR #1009 — "Fix compatibility test suite" (commit 66bbdce, Allen Samuels, May 26 2026)


integration/compatibility/generate.py
```diff
@@ -474,3 +474,38 @@ def test_search_sortby(self, key_type, dialect):
...
+                              "LIMIT", "0", "20", "DIALECT", str(dialect)])
+        self.execute_command(["FT.SEARCH", f"{key_type}_idx1",
+                              r"@tags:{ normal | a\|b | a\}b }",
+                              "LIMIT", "0", "20", "DIALECT", str(dialect)])
+        self.execute_command(["FT.SEARCH", f"{key_type}_idx1",
+                              r"@tags:{ tag-containing-\}-and-\} | a\\b }",
+                              "LIMIT", "0", "20", "DIALECT", str(dialect)])
+        self.execute_command(["FT.SEARCH", f"{key_type}_idx1",
+                              r"@tags:{ a\}b | a\|b | tag-containing-\}-and-\} | a\\b | normal }",
+                              "LIMIT", "0", "20", "DIALECT", str(dialect)])
```
allenss-amazon: These don't test anything. You need to follow the check/checkvec used in the other test cases. Once these are converted, you'll have to execute "integration/compatibility/regenerate.sh" and add those files to the commit.

Aksha1812: updated @Frank-Gu-81 's code

Aksha1812: converted to check() and regenerated the pickles via integration/compatibility/regenerate.sh (both aggregate-answers and text-search-answers).
  
Once the queries actually got compared, two cases failed, both the escaped-backslash ones (@tags:{ a\\b }). and only on JSON (HASH is fine): a stored value of a\b comes back from the JSON API still escaped (a\\b) and gets indexed with a doubled backslash, while the query side unescapes correctly to a\b so they never match. Looks like NormalizeJsonRecord strips the surrounding [ ]/" " but doesn't JSON-unescape the string body.

I removed those two backslash cases to keep this PR scoped to the }/| fix.

is the JSON backslash thing an actual issue we should fix (here or as a follow-up), or is it expected/deferred for some reason?

  Repro (returns 0; RediSearch returns 1):
```
  JSON.SET j:1 $ '{"tags":"a\\b"}'
  FT.CREATE jidx ON JSON PREFIX 1 j: SCHEMA $.tags AS tags TAG SEPARATOR ,
  FT.SEARCH jidx '@tags:{ a\\b }' 
 ```

allenss-amazon: We need to fix NormalizeJsonRecord. 

Long term, we need a different interface to the JSON module.

Aksha1812: created issue for long term fix here https://github.com/valkey-io/valkey-search/issues/1149 . Anything pending for this PR ?

allenss-amazon (approved):

PR #1154, opened 2026-06-18
Remove timeout settings in integration tests, defer to waiter default

allenss-amazon:
Root cause of a lot of ASAN failures is too short of a timeout. 

This PR removes most manual overrides of the timeout parameter.

In a couple of cases, manual loops are replaced with the waiter routines, which properly handle a longer timeout for ASAN.


integration/test_ft_dropindex_consistency.py
```diff
@@ -33,19 +33,18 @@ def counter_is_increasing():
             return count > 0
 
         # wait for reaching consistency check pausepoint
-        waiters.wait_for_true(wait_for_pausepoint, timeout=5)
+        waiters.wait_for_true(wait_for_pausepoint)
```
BCathcart, line 36: Wrapping `wait_for_pauspoint` in a `wait_for_true` seems misguided

BCathcart (approved):

integration/test_ft_dropindex_consistency.py
allenss-amazon: fixed.

Issue #190, opened 2025-06-19
[BUG] Incorrect handling of invalid values in a field


allenss-amazon:
If a key contains a field which can't be indexed as declared, then the entire key should not be indexed. 

For example, if a numeric field has a non-numeric value in it then today we index the remainder of the fields in the key, but this isn't compatible with Redisearch.

allenss-amazon:
This is still broken in 1.2 despite fixes that have been made.

Frank-Gu-81:
Looks like #1155 fixes this issue. I'll close this one.

PR #1155, opened 2026-06-18
Revise handling of keys with fields with invalid values to match Redisearch

allenss-amazon:
Uses the compatibility framework to revise the handling of fields with invalid values.

Current handling is that if a field has an invalid value, then that field is ignored, but the remainder of the fields in that key are still eligible for indexing.

Redisearch compatible behavior is to drop the entire key when any field has an invalid value. Keys dropped like this do not appear in negative queries.

This PR implements correct handling for Numeric and Vector fields. Tag and Text invalid handling (non-UTF8) remains.


Aksha1812:
@allenss-amazon this looks like change for 1.3 . can you verify and remove this from 1.2 project board ?

src/index_schema.cc
```diff
@@ -684,15 +753,15 @@ void IndexSchema::ProcessAttributeMutation(
     if (index->IsTracked(key)) {
       auto res = index->ModifyRecord(key, data_view);
       TrackResults(ctx, res, "Modify", stats_.subscription_modify);
-      if (res.ok() && res.value()) {
+      if (res.ok() && res.value() == indexes::RecordResult::kAdded) {
         ++Metrics::GetStats().time_slice_upserts;
       }
-      return;
+      return res.ok() && res.value() == indexes::RecordResult::kInvalidData;
```
Nivesh-01, line 760: Here if there is a case that res.ok() is false then we never get to evaluation of res.value and the kInvalidData is never set. If that happens I beleive that since invalid_data stays false when an error occurred, RemoveKeyFromAllIndexes is never called. The key remains in whatever partial state the indexes left it — some fields indexed, the failing field in an inconsistent state, this was not introduced here but maybe we should take a look at it if it is the case

src/index_schema.cc
```diff
@@ -657,9 +688,11 @@ void IndexSchema::SyncProcessMutation(ValkeyModuleCtx *ctx,
...
     }
-    ProcessAttributeMutation(ctx, itr->second, key,
-                             std::move(attribute_data_itr.second.data),
-                             attribute_data_itr.second.deletion_type);
+    if (ProcessAttributeMutation(ctx, itr->second, key,
+                                 std::move(attribute_data_itr.second.data),
+                                 attribute_data_itr.second.deletion_type)) {
+      invalid_data = true;
+    }
   }
```
Nivesh-01, line 696: Here all_deletes == true and invalid_data == true can never happen simultaneously so maybe we can add a check for future to make sure that this remains the case. Some thing like
DCHECK(!all_deletes || !invalid_data); 
This is just a defensive chech that we might not need but just bringing it up.

src/index_schema.cc
```diff
@@ -715,14 +784,15 @@ void IndexSchema::ProcessAttributeMutation(
           break;
       }
     }
-    return;
+    return res.ok() && res.value() == indexes::RecordResult::kInvalidData;
```
Nivesh-01, line 788: Same as before

src/index_schema.cc
```diff
@@ -672,9 +705,45 @@ void IndexSchema::SyncProcessMutation(ValkeyModuleCtx *ctx,
     // updates to all Text attributes in one operation for efficiency
     text_index_schema_->CommitKeyData(key);
   }
+
+  // Post-cleanup for invalid data. Redisearch drops the entire key from the
```
Nivesh-01, line 710: i understand this comment but wanted to just raise this. Over here if say we have a key with 5 fields and the last field is invalid then we are adding the record or mutating it for all the fields before and then removing them after the addition which is like a double loop. In high volume of such faulty key this could bloat up the queue. Instead if we gate everything before and see if they are valid or not and then add or modify the records we could perhaps make it faster. Though this is non-blocking just something I observed and wanted to bring up, might be a non-issue.

allenss-amazon: Later when we can rip out the old compatibility code, we can perform a validation check on each BEFORE we start the ingestion cycle like you suggest. That way if it's invalid it never gets ingested. 

However, queue bloat isn't really an issue, in case where there's a validation failure, the delete is done synchronously. SO the queue size is the same. It's just slower for an invalid value.

src/index_schema.cc
allenss-amazon: This is a ragged edge of philosophies. The original ModifyRecord call only failed when the indexes themselves are screwed up. Specifically "you told me to modify an existing value and I couldn't find it" was the only error -- and there wasn't an attempt to "cleanup" -- because that'll probably make things worse. I added the kInvalidData return code which is now the only valid failure mode for ModifyRecord. So here I'm informing the caller of whether there was invalid data that ought to result in deleting the key. If any other error occurrs, don't the delete the key....

src/index_schema.cc
allenss-amazon: Same story as above. AddRecord only used to fail when the indexes were screwed up -- here, I tried to add it but there was already a record there!. Thus same logic as above.

Nivesh-01 (approved):

Frank-Gu-81:
Fixes #190

PR #1163, opened 2026-06-20
Vector Search performance improvement

allenss-amazon:
USE_SIMSIMD wasn't enabled by default in the build logic causing substantial performance degradation.

The prefetch loop in the core base layer query path of HNSWLib wasn't updated when the "ChunkedArray" was used to avoid having the base layer data all in one memory allocation (which causes severe problems when attempting to resize it). This PR updates that logic with a "3-phase" software pipeline.

Here is some measured performance numbers with this PR on an AMD box vs baseline and Redisearch 8.8 container.

What's being measured here is single vector query latency. 

```
  ┌─────────────────────────────────────────────┬─────────────────┬──────┬──────┬──────┬──────┬───────────┐
  │                   Dataset                   │     Engine      │ p50  │ p90  │ p99  │ mean │ recall@10 │
  ├─────────────────────────────────────────────┼─────────────────┼──────┼──────┼──────┼──────┼───────────┤
  │ Cohere 1024-dim (Simple-Wiki, 250k, cosine) │ valkey baseline │ 1341 │ 1747 │ 2086 │ 1327 │ 0.9846    │
  ├─────────────────────────────────────────────┼─────────────────┼──────┼──────┼──────┼──────┼───────────┤
  │                                             │ valkey 3-phase  │ 1122 │ 1478 │ 1790 │ 1135 │ 0.9846    │
  ├─────────────────────────────────────────────┼─────────────────┼──────┼──────┼──────┼──────┼───────────┤
  │                                             │ redisearch      │ 1263 │ 1605 │ 2004 │ 1263 │ 0.9813    │
  ├─────────────────────────────────────────────┼─────────────────┼──────┼──────┼──────┼──────┼───────────┤
  │ lastfm 64-dim (292k, cosine-normalized)     │ valkey baseline │ 497  │ 657  │ 824  │ 514  │ 0.9858    │
  ├─────────────────────────────────────────────┼─────────────────┼──────┼──────┼──────┼──────┼───────────┤
  │                                             │ valkey 3-phase  │ 440  │ 609  │ 731  │ 463  │ 0.9858    │
  ├─────────────────────────────────────────────┼─────────────────┼──────┼──────┼──────┼──────┼───────────┤
  │                                             │ redisearch      │ 560  │ 791  │ 899  │ 602  │ 0.9857    │
  └─────────────────────────────────────────────┴─────────────────┴──────┴──────┴──────┴──────┴───────────┘

To summarize:


  ┌──────────┬─────────────────────────┬──────────────────────┐
  │          │ three-phase vs baseline │ three-phase vs redis │
  ├──────────┼─────────────────────────┼──────────────────────┤
  │ 1024-dim │ −16.3% (1341→1122)      │ −11.2%               │
  ├──────────┼─────────────────────────┼──────────────────────┤
  │ 64-dim   │ −11.5% (497→440)        │ −21.4%               │
  └──────────┴─────────────────────────┴──────────────────────┘
```


third_party/hnswlib/hnswalg.h
```diff
@@ -417,80 +417,123 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
         metric_distance_computations += size;
       }
 
+      // ---- Three-phase prefetch pipeline -----------------------------------
```
BCathcart, line 419: Did you look into optimizing `searchBaseLayer` with a similar pattern for insertions?

third_party/hnswlib/hnswalg.h
```diff
@@ -417,80 +417,123 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+      // A candidate's vector lives behind a pointer indirection at offsetData_,
+      // so reaching it is a two-deep pointer chase (slot -> vec) layered on top
+      // of a random visited_array probe. Fusing all of that into one loop forces
+      // the chase to serialize per candidate. Instead we fission the expansion
+      // into three passes, each hiding exactly one level of memory latency with
+      // its own lookahead:
+      //   Phase 1 - probe visited_array (random gather), collect unvisited ids.
```
BCathcart, line 428: "probe" and "random gather" make me think of random sampling. Maybe restructure slightly to be clearer that the randomness is with regards to spatial locality?

third_party/hnswlib/hnswalg.h
```diff
@@ -417,80 +417,123 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
 
+      // Phase 1: filter visited. Preserve neighbor-list order so the heap/
+      // lowerBound evolution in phase 3 is identical to the fused loop.
       for (size_t j = 1; j <= size; j++) {
-        int candidate_id = *(data + j);
-//                    if (candidate_id == 0) continue;
 #ifdef USE_PREFETCH
-        if (j + 1 < size) {
-          __builtin_prefetch((char *)(visited_array + *(data + j + 1)), 0, 3);
+        if (j + kVisitedLookahead <= size)
```
BCathcart, line 448: As someone who isn't too experienced with pre-fetching in practice, what makes 4 the magic number here?

And I imagine it's relatively inconsequential, but should there be a small benefit to pre-fetching elements 2 and 3 in the first iteration?

third_party/hnswlib/hnswalg.h
```diff
@@ -417,80 +417,123 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+          unvisited.push_back(candidate_id);
+        }
+      }
+      const size_t n_unvisited = unvisited.size();
 
-          char *currObj1 = (getDataByInternalId(candidate_id));
-          dist_t dist = fstdistfunc_(data_point, currObj1, dist_func_param_);
+      // Phase 2: resolve slot indirection. Prefetch the pointer slot ahead,
+      // then dereference the now-resident slot to the actual vector address.
+      vptrs.resize(n_unvisited);
```
BCathcart, line 462: Maybe just reserve maxM0_ entries from the start?

third_party/hnswlib/hnswalg.h
```diff
@@ -417,80 +417,123 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
-          }
+      // Phase 3: distance compute. Prefetch the head (~3 lines, two 128B
+      // sectors) of the next vector to cover the cold head and bridge the HW
+      // streamer's training window; the streamer covers the multi-line tail.
+      for (size_t k = 0; k < n_unvisited; k++) {
+#ifdef USE_PREFETCH
+        if (k + kVectorLookahead < n_unvisited) {
+          char *h = vptrs[k + kVectorLookahead];
+          __builtin_prefetch(h, 0, 0);
+          __builtin_prefetch(h + 128, 0, 0);
```
BCathcart, line 483: Small overfetching here for small vectors is negligible, right?

BCathcart:
Are the performance numbers in the description with SIMD enabled for both the baseline and pre-fetching runs?

third_party/hnswlib/hnswalg.h
```diff
@@ -417,80 +417,123 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+          } else {
+            flag_remove_extra = top_candidates.size() > ef;
+          }
+          while (flag_remove_extra) {
+            tableint id = top_candidates.top().second;
+            top_candidates.pop();
             if (!bare_bone_search && stop_condition) {
+              stop_condition->remove_point_from_result(
+                  getExternalLabel(id), getDataByInternalId(id), dist);
               flag_remove_extra = stop_condition->should_remove_extra();
```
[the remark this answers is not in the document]
allenss-amazon: Unused.

third_party/hnswlib/hnswalg.h
```diff
@@ -417,80 +417,123 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
-      __builtin_prefetch((*data_level0_memory_)[(*(data + 1))] + offsetData_, 0,
-                         3);
-      __builtin_prefetch((char *)(data + 2), 0, 3);
+      constexpr int kVisitedLookahead = 4;  // P1
+      constexpr int kSlotLookahead = 4;     // P2
+      constexpr int kVectorLookahead = 1;   // P3
 #endif
+      thread_local std::vector<tableint> unvisited;
+      thread_local std::vector<char *> vptrs;
+      unvisited.clear();
```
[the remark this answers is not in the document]
allenss-amazon: don't care.

third_party/hnswlib/hnswlib.h
```diff
@@ -17,6 +17,11 @@
 #endif
 
 #define USE_PREFETCH
+// Always enable SimSIMD distance kernels. Defined here (before space_l2.h /
+// space_ip.h / stop_condition.h test `#if defined(USE_SIMSIMD)`, all of which
+// include this header first) rather than via a build flag. SimSIMD does its
+// own runtime CPU-feature dispatch, so this stays portable.
+#define USE_SIMSIMD
```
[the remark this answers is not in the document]
allenss-amazon: don't care.

third_party/hnswlib/hnswalg.h
allenss-amazon: I did not. That can be done in a subsequent PR. Further research has identified substantial additional performance improvements most of which require changes in the core. So these will get address in the next major release.

third_party/hnswlib/hnswalg.h
allenss-amazon: With current hardware, yes.

third_party/hnswlib/hnswalg.h
allenss-amazon: Since it's thread_local, that makes it effectively static. Meaning that the first query will suffer some memory allocations but all subsequent queries won't have to do any allocations. So inserting a reserve would actually slow up all queries but the first one.

third_party/hnswlib/hnswalg.h
allenss-amazon: The key to prefetching is to "lead the duck" (in time). Here we have three loops. The first two loops are actually quite small and therefore to "lead the duck" you need to get far ahead of them. The last loop is fairly expensive, so you don't need to "lead the duck" by quite as much. This explains why the first two constants are substantially larger than the last one. The actual value of 4 was determined through hyper-parameter search.

allenss-amazon:
Yes, the perf #'s are SIMD for both. SIMD is now hardcoded on.

third_party/hnswlib/hnswalg.h
allenss-amazon: Dropped the confusing comment.

third_party/hnswlib/hnswalg.h
BCathcart: Yea I really meant resize() once up front to the max number of neighbours. Rather than potentially re-allocating if the first node we visit has fewer neighbours than the second node. We're treating vptrs like an array and not using size() so it should work fine. Just a minor detail though, not particularly important

BCathcart (approved):

PR #1166, opened 2026-06-22
Hnswlib templatized vectors

yairgott:
Refactors the vector indexing layer to allow third-party HNSWLib to directly own and manage vector lifecycle, preparing the codebase to allow utilizing the
 memory-sharing feature.

1. Simplification & Lifecycle Cleanliness:
   - By enabling HNSWLib to directly own and maintain the lifetime of vector data, we completely eliminate the need for VectorFlat and VectorHNSW to hold and track strong references to vectors. This allows us to deprecate and remove the complex `VectorBase::TrackVector` and `VectorTracker` abstractions.
2. Preparation for Engine Memory Sharing:
   - This architectural decoupling serves as a direct preparation for a follow-up change that will utilize memory-sharing capabilities between the Valkey-Search module and the Valkey engine, avoiding redundant data copying.

* Enhanced HNSWLib with additional templatized vector types:
  - `InputVectorT`: A generic parameter type used across all external HNSWLib interfaces (such as `addPoint`, `searchKnn`, etc.).
  - `SavedVectorT`: The internal storage type used to hold and represent point vectors inside the graph itself.
* Templatized Index Deserialization:
  - Templatized `LoadIndex` in `bruteforce.h` and `hnswalg.h` to accept an allocator-aware `FunctorT vector_constructor` callback. This allows the calling indexes to pass a lambda that utilizes their custom fixed-size allocator during RDB loading, preventing heap fragmentation.


cmake/Modules/valkey_search.cmake
```diff
@@ -151,6 +151,13 @@ function(valkey_search_target_update_compile_flags TARGET)
   if(HIGHWAY_HASH_INCLUDE_PATH)
     target_include_directories(${TARGET} PRIVATE ${HIGHWAY_HASH_INCLUDE_PATH})
   endif()
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND UNIX AND NOT APPLE)
+    include(CheckCompilerFlag)
+    check_compiler_flag(C "-Xassembler --gsframe=no" COMPILER_SUPPORTS_GSFRAME_NO)
+    if(COMPILER_SUPPORTS_GSFRAME_NO)
+      target_compile_options(${TARGET} PRIVATE -Xassembler --gsframe=no)
```
allenss-amazon, line 158: Won't this disable "perf" providing tracebacks?

src/commands/ft_aggregate.cc
```diff
@@ -68,7 +68,9 @@ absl::Status ManipulateReturnsClause(AggregateParameters &params) {
       content = true;
       VMSDK_ASSIGN_OR_RETURN(auto indexer, params.index_schema->GetIndex(load));
       auto indexer_type = indexer->GetIndexerType();
-      if (indexer->IsVectorIndex()) {
+      if (indexer_type == indexes::IndexerType::kVector ||
+          indexer_type == indexes::IndexerType::kHNSW ||
+          indexer_type == indexes::IndexerType::kFlat) {
```
allenss-amazon, line 73: why change this?

src/indexes/index_base.h
```diff
@@ -76,8 +76,6 @@ class IndexBase {
   /// Returns the mutation weight for this index type
   virtual uint32_t GetMutationWeight() const = 0;
 
-  virtual bool IsVectorIndex() const { return false; }
```
allenss-amazon, line 79: why remove this?

src/indexes/vector_base.cc
```diff
@@ -172,11 +172,17 @@ absl::StatusOr<bool> VectorBase::AddRecord(const InternedStringPtr &key,
   if (!interned_vector) {
```
allenss-amazon, line 172: InternVector can't fail, right? I think this should CHECK.

src/indexes/vector_base.cc
```diff
@@ -172,11 +172,17 @@ absl::StatusOr<bool> VectorBase::AddRecord(const InternedStringPtr &key,
...
+                         TrackKey(key, magnitude.value_or(kDefaultMagnitude)));
+  absl::Status add_result = AddRecordImpl(internal_id, interned_vector);
   if (!add_result.ok()) {
+    auto remove_result = RemoveRecordImpl(internal_id);
+    if (!remove_result.ok()) {
+      VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+          << "While processing error for AddRecord, encountered error in "
+             "RemoveRecordImpl: "
+          << remove_result.message();
+    }
```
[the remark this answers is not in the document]
allenss-amazon: I agree with Greptile. There's no legitimate failure case for TrackKey. So if it does fail, the system is in an invalid state. IMO, rather than attempt to cleanup when the system is in some invalid state (which stands a fair chance of making things worse AND obscuring what actually went wrong -- destroying any opportunity to fix it), it's better to either do nothing (log it of course) or crash.

src/indexes/vector_base.cc
allenss-amazon, line 234: Again, this is going to corrupt things even more. I'd prefer a nice clean crash here.

src/indexes/vector_base.cc
```diff
@@ -272,12 +279,13 @@ absl::StatusOr<std::vector<Neighbor>> VectorBase::CreateReply(
 
 absl::StatusOr<std::vector<char>> VectorBase::GetValue(
     const InternedStringPtr &key) const {
+  absl::WriterMutexLock lock(&key_to_metadata_mutex_);
```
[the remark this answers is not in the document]
allenss-amazon: This does seem like an error. But I'm not convinced that a ReaderLock is the correct answer. How is this used? What guarantees that nothing is changing for this key/vector combination??

I've searched the codebase, I haven't found where it's used. Same for Numeric::GetValue.

src/indexes/vector_base.cc
```diff
@@ -272,12 +279,13 @@ absl::StatusOr<std::vector<Neighbor>> VectorBase::CreateReply(
...
   auto it = tracked_metadata_by_key_.find(key);
   if (it == tracked_metadata_by_key_.end()) {
     return absl::NotFoundError("Record was not found");
   }
   std::vector<char> result;
-  char *value = GetValueImpl(it->second.internal_id);
+  const char *value = GetValueImpl(it->second.internal_id);
   if (normalize_) {
```
allenss-amazon, line 289: Maybe a better name for this. Is normalization required? Or has it already been done?

src/indexes/vector_base.cc
allenss-amazon, line 292: Maybe it should be std::optional?

src/indexes/vector_base.cc
```diff
@@ -324,16 +331,8 @@ absl::StatusOr<std::optional<uint64_t>> VectorBase::UnTrackKey(
...
-  return (char *)interned_vector->Str().data();
-}
-
 absl::StatusOr<uint64_t> VectorBase::TrackKey(const InternedStringPtr &key,
-                                              float magnitude,
-                                              const InternedStringPtr &vector) {
+                                              float magnitude) {
   if (key->Str().empty()) {
     return absl::InvalidArgumentError("key can't be empty");
   }
```
allenss-amazon, line 338: Incorrect. An empty key is a completely legal value.

src/indexes/vector_base.h
```diff
@@ -45,8 +44,25 @@ enum class QueryOperations : uint64_t;
...
 namespace valkey_search::indexes {
 
+class VectorRecord {
+ public:
+  explicit VectorRecord(const InternedStringPtr &vector)
+      : raw_vector_(vector) {}
+  explicit VectorRecord(absl::string_view vector)
+      : raw_vector_(StringInternStore::Intern(vector)) {}
+
+  inline const char *GetRawVector() const {
```
allenss-amazon, line 54: I'd much rather see a std::string_view here.

src/indexes/vector_base.h
```diff
@@ -45,8 +44,25 @@ enum class QueryOperations : uint64_t;
...
+    return const_cast<char *>(raw_vector_->Str().data());
+  }
+  inline operator const void *() const { return GetRawVector(); }
+  inline operator const char *() const { return GetRawVector(); }
```
allenss-amazon, line 58: same.

src/indexes/vector_flat.cc
```diff
@@ -72,28 +71,20 @@ absl::StatusOr<std::shared_ptr<VectorFlat<T>>> VectorFlat<T>::Create(
...
-}
+  const char *stored_vec =
+      reinterpret_cast<VectorRecord *>((*algo_->data_)[found->second])
+          ->GetRawVector();
 
-template <typename T>
-void VectorFlat<T>::UnTrackVector(uint64_t internal_id) {
-  absl::MutexLock lock(&tracked_vectors_mutex_);
-  tracked_vectors_.erase(internal_id);
+  absl::string_view record(stored_vec, dimensions_ * sizeof(T));
```
allenss-amazon, line 86: This is why I'd prefer GetRawVector to return a string_view.

allenss-amazon: If both vectors are interned, do we need a full compare? Wouldn't just comparing the start address be sufficient?

src/indexes/vector_flat.cc
allenss-amazon, line 97: make_shared

src/indexes/vector_base.cc
```diff
@@ -172,11 +172,17 @@ absl::StatusOr<bool> VectorBase::AddRecord(const InternedStringPtr &key,
   if (!interned_vector) {
     return false;
   }
-  VMSDK_ASSIGN_OR_RETURN(
-      auto internal_id,
-      TrackKey(key, magnitude.value_or(kDefaultMagnitude), interned_vector));
-  absl::Status add_result = AddRecordImpl(internal_id, interned_vector->Str());
+  VMSDK_ASSIGN_OR_RETURN(auto internal_id,
+                         TrackKey(key, magnitude.value_or(kDefaultMagnitude)));
+  absl::Status add_result = AddRecordImpl(internal_id, interned_vector);
```
allenss-amazon, line 177: Given that you later wrap the internedstringptr into VectorRecord. It seems like we should be making a VectorRecord here and letting all of the downstream routines have a VectorRecord -- which would be better documentation than an InternStringPtr type as the vector value.

src/indexes/vector_flat.cc
```diff
@@ -191,7 +186,10 @@ absl::Status VectorFlat<T>::ModifyRecordImpl(uint64_t internal_id,
 
   memcpy((*algo_->data_)[found->second] + algo_->data_ptr_size_, &internal_id,
          sizeof(hnswlib::labeltype));
-  *(char **)((*algo_->data_)[found->second]) = (char *)record.data();
+
+  VectorRecord *stored_vector =
+      reinterpret_cast<VectorRecord *>((*algo_->data_)[found->second]);
```
allenss-amazon, line 191: This is an example of why passing VectorRecord around is cleaner.

And, setting or accessing this slot should be a method in the algo.

src/indexes/vector_flat.cc
```diff
@@ -191,7 +186,10 @@ absl::Status VectorFlat<T>::ModifyRecordImpl(uint64_t internal_id,
 
   memcpy((*algo_->data_)[found->second] + algo_->data_ptr_size_, &internal_id,
```
allenss-amazon, line 187: Seems like this should be a method of the algo. Having external knowledge of this code seems dangerous.

src/indexes/vector_flat.h
```diff
@@ -58,51 +58,44 @@ class VectorFlat : public VectorBase {
...
+  int RespondWithInfoImpl(ValkeyModuleCtx *ctx) const override;
   absl::Status SaveIndexImpl(RDBChunkOutputStream chunked_out) const override;
   absl::StatusOr<std::pair<float, hnswlib::labeltype>>
   ComputeDistanceFromRecordImpl(uint64_t internal_id,
                                 absl::string_view query) const override;
-  char* GetValueImpl(uint64_t internal_id) const override
+  const char *GetValueImpl(uint64_t internal_id) const override
       ABSL_NO_THREAD_SAFETY_ANALYSIS {
-    return algo_->getPoint(internal_id);
+    return algo_->getPoint(internal_id)->GetRawVector();
```
allenss-amazon, line 84: Seems like this should be returning a VectorRecord ??

src/indexes/vector_hnsw.cc
```diff
@@ -64,15 +68,15 @@ std::optional<hnswlib::tableint> GetInternalIdLockFree(
 }
 
 template <typename T>
-std::optional<hnswlib::tableint> GetInternalId(
-    hnswlib::HierarchicalNSW<T> *algo, uint64_t internal_id) {
+std::optional<hnswlib::tableint> GetInternalId(HNSWIndexImpl<T> *algo,
```
allenss-amazon, line 71: is the <T> needed here?

src/indexes/vector_hnsw.cc
```diff
@@ -53,8 +52,13 @@
...
 
 template <typename T>
-std::optional<hnswlib::tableint> GetInternalIdLockFree(
-    hnswlib::HierarchicalNSW<T> *algo, uint64_t internal_id) {
+using HNSWIndexImpl =
+    hnswlib::HierarchicalNSW<T, valkey_search::indexes::VectorRecord,
+                             valkey_search::indexes::VectorRecord>;
+
+template <typename T>
+std::optional<hnswlib::tableint> GetInternalIdLockFree(HNSWIndexImpl<T> *algo,
```
allenss-amazon, line 60: Is the <T> needed here?

src/indexes/vector_hnsw.cc
allenss-amazon: Awful name, since internal_id is the input parameter.... SHouldn't this be something like GetSlotForInternalId?

src/indexes/vector_hnsw.cc
```diff
@@ -64,15 +68,15 @@ std::optional<hnswlib::tableint> GetInternalIdLockFree(
...
+std::optional<hnswlib::tableint> GetInternalId(HNSWIndexImpl<T> *algo,
+                                               uint64_t internal_id) {
   std::unique_lock<std::mutex> lock_table(algo->label_lookup_lock);
   return GetInternalIdLockFree(algo, internal_id);
 }
 
 template <typename T>
 std::optional<hnswlib::tableint> GetInternalIdDuringSearch(
-    hnswlib::HierarchicalNSW<T> *algo, uint64_t internal_id) {
+    HNSWIndexImpl<T> *algo, uint64_t internal_id) {
```
allenss-amazon, line 79: <T> ??

src/indexes/vector_hnsw.cc
```diff
@@ -123,16 +120,12 @@ bool VectorHNSW<T>::IsVectorMatch(uint64_t internal_id,
     if (!id.has_value()) {
       return false;
     }
-    char *data_ptrv = algo_->getDataByInternalId(*id);
+    const char *data_ptrv = algo_->getDataByInternalId(*id);
```
allenss-amazon, line 123: Doesn't the slot have a VectorRecord in it? If so, then returning that seems like it would clean up the code quite a bit, especially since we should be passing VectorRecord into this function -- right?

src/indexes/vector_hnsw.cc
```diff
@@ -177,12 +174,12 @@ VectorHNSW<T>::VectorHNSW(int dimensions,
...
 template <typename T>
 absl::Status VectorHNSW<T>::AddRecordImpl(uint64_t internal_id,
-                                          absl::string_view record) {
+                                          const InternedStringPtr &vector) {
   do {
     try {
       absl::ReaderMutexLock lock(&resize_mutex_);
 
-      algo_->addPoint((T *)record.data(), internal_id,
+      algo_->addPoint(VectorRecord(vector), internal_id,
```
allenss-amazon, line 182: Again. VectorRecord should be the input parameter here, right?

src/indexes/vector_hnsw.cc
```diff
@@ -274,14 +271,14 @@ absl::Status VectorHNSW<T>::ResizeIfFull() {
...
-                                             absl::string_view record) {
+                                             const InternedStringPtr &vector) {
   try {
     absl::ReaderMutexLock lock(&resize_mutex_);
     // TODO - an alternative approach is to call HierarchicalNSW::updatePoint.
     // The concern with calling updatePoint is that it might have implications
     // on the search accuracy. Need to revisit this in the future.
     algo_->markDelete(internal_id);
-    algo_->addPoint((T *)record.data(), internal_id,
+    algo_->addPoint(VectorRecord(vector), internal_id,
```
allenss-amazon, line 281: same

src/index_schema.cc
```diff
@@ -1217,6 +1218,12 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
   }
 }
 
+bool IsVectorIndex(std::shared_ptr<indexes::IndexBase> index) {
+  return index->GetIndexerType() == indexes::IndexerType::kVector ||
+         index->GetIndexerType() == indexes::IndexerType::kHNSW ||
+         index->GetIndexerType() == indexes::IndexerType::kFlat;
+}
```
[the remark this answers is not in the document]
allenss-amazon: This was already implemented in IndexBase as a virtual function, so if you're going to replace it at least make it be a member of IndexBase and avoid all of the cascading changes from that.

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
 
 #ifdef VMSDK_ENABLE_MEMORY_ALLOCATION_OVERRIDES
 #include "vmsdk/src/memory_allocation_overrides.h"  // IWYU pragma: keep
 #endif
 
 namespace hnswlib {
-template <typename dist_t>
-class BruteforceSearch : public AlgorithmInterface<dist_t> {
+
+template <typename dist_t, typename SavedVectorT, typename InputVectorT>
```
allenss-amazon, line 23: SavedVectorT is not very mnemonic. VectorAccessorT or VectorRefT would be much better.

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
 
-
-    BruteforceSearch(SpaceInterface <dist_t> *s)
-        : cur_element_count_(0), dist_func_param_(nullptr) {
+  BruteforceSearch(SpaceInterface<dist_t> *s)
+      : cur_element_count_(0), dist_func_param_(nullptr) {
+    vector_size_ = s->get_data_size();
+    fstdistfunc_ = s->get_dist_func();
+    dist_func_param_ = s->get_dist_func_param();
+  }
```
allenss-amazon, line 42: I doubt this is used, since data_ isn't set. It can probably be removed.

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
+  BruteforceSearch(SpaceInterface<dist_t> *s, size_t maxElements) {
+    cur_element_count_ = 0;
+    vector_size_ = s->get_data_size();
+    fstdistfunc_ = s->get_dist_func();
+    dist_func_param_ = s->get_dist_func_param();
+    data_ =
+        std::make_unique<ChunkedArray>(sizeof(SavedVectorT) + sizeof(labeltype),
```
allenss-amazon, line 49: Let's create a struct for each entry in the array. Then have accessor functions for the two pieces of information in that struct ... and get rid of all of the crazy places where we do address arithmetic and type-overrides to access those.

allenss-amazon: Why isn't ChunkedArray itself templated with this struct type?

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
+  void addPoint(const InputVectorT &datapoint, labeltype label,
+                bool replace_deleted = false) override {
+    int idx;
+    std::unique_lock<std::mutex> lock(index_lock);
+    auto search = dict_external_to_internal.find(label);
+    if (search != dict_external_to_internal.end()) {
+      idx = search->second;
+      SavedVectorT *stored_vector =
+          reinterpret_cast<SavedVectorT *>((*data_)[idx]);
+      *stored_vector = SavedVectorT(datapoint);
```
allenss-amazon, line 70: This gibberish gets cleaned up with a proper struct.

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
+      }
+      idx = cur_element_count_;
+      dict_external_to_internal[label] = idx;
+      cur_element_count_++;
+      SavedVectorT *stored_vector =
+          reinterpret_cast<SavedVectorT *>((*data_)[idx]);
+      new (stored_vector) SavedVectorT(datapoint);
     }
 
+    memcpy((*data_)[idx] + sizeof(SavedVectorT), &label, sizeof(labeltype));
```
allenss-amazon, line 84: Ditto.

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
-                data_ptr_size_ + sizeof(labeltype),
-                k_elements_per_chunk,
-                maxElements);
+  SavedVectorT *getPoint(labeltype cur_external) {
+    std::unique_lock<std::mutex> lock(index_lock);
+    auto found = dict_external_to_internal.find(cur_external);
+    if (found == dict_external_to_internal.end()) {
+      return nullptr;
     }
+    return reinterpret_cast<SavedVectorT *>((*data_)[found->second]);
```
allenss-amazon, line 93: more gibberish.

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
-        return nullptr;
+    labeltype label = *(
+        (labeltype *)((*data_)[cur_element_count_ - 1] + sizeof(SavedVectorT)));
+    dict_external_to_internal[label] = cur_c;
+
+    *reinterpret_cast<SavedVectorT *>((*data_)[cur_c]) = std::move(
+        *reinterpret_cast<SavedVectorT *>((*data_)[cur_element_count_ - 1]));
+    memcpy((*data_)[cur_c] + sizeof(SavedVectorT),
+           (*data_)[cur_element_count_ - 1] + sizeof(SavedVectorT),
+           sizeof(labeltype));
```
allenss-amazon, line 119: gibberish. We might as well be writing in assembler.

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
+  std::priority_queue<std::pair<dist_t, labeltype>> searchKnn(
+      const InputVectorT &query_data, size_t k,
+      BaseFilterFunctor *isIdAllowed = nullptr,
+      BaseCancellationFunctor *isCancelled = nullptr) const override {
+    assert(k <= cur_element_count_);
+    std::priority_queue<std::pair<dist_t, labeltype>> topResults;
+    if (cur_element_count_ == 0) return topResults;
+    for (int i = 0; i < k; i++) {
+      const SavedVectorT &stored_vector =
+          *reinterpret_cast<const SavedVectorT *>((*data_)[i]);
```
allenss-amazon, line 135: same

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
+      dist_t dist = fstdistfunc_(query_data, stored_vector, dist_func_param_);
+      labeltype label = *((labeltype *)((*data_)[i] + sizeof(SavedVectorT)));
```
allenss-amazon, line 137: same

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
-        if (cur_element_count_ - 1 == cur_c) {
-          cur_element_count_--;
-          return;
+    dist_t lastdist = topResults.empty() ? std::numeric_limits<dist_t>::max()
+                                         : topResults.top().first;
+    for (int i = k; i < cur_element_count_ &&
+                    (!isCancelled || !isCancelled->isCancelled());
+         i++) {
+      const SavedVectorT &stored_vector =
+          *reinterpret_cast<const SavedVectorT *>((*data_)[i]);
```
allenss-amazon, line 148: more

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
+    if (!header.SerializeToString(&serialized)) {
+      return absl::InternalError("Could not serialize bruteforce header");
+    }
+    VMSDK_RETURN_IF_ERROR(
+        output.SaveChunk(serialized.data(), serialized.size()));
+
+    // TODO: write in chunks to improve throughput
+    std::vector<char> buf(size_per_element);
+    for (int i = 0; i < cur_element_count_; i++) {
+      const SavedVectorT &stored_vector =
```
allenss-amazon, line 181: same.

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
-        VMSDK_ASSIGN_OR_RETURN(auto chunk, input.LoadChunk());
-        labeltype id;
-        memcpy((char *)&id, chunk->data() + vector_size_, sizeof(labeltype));
-        *(char **)(*data_)[i] =
-            vector_tracker->TrackVector(id, chunk->data(), vector_size_);
-        memcpy((*data_)[i] + data_ptr_size_, (char *)&id, sizeof(labeltype));
-        dict_external_to_internal[id] = i;
-      }
+    for (int i = 0; i < cur_element_count_; i++) {
+      VMSDK_ASSIGN_OR_RETURN(auto chunk, input.LoadChunk());
```
allenss-amazon, line 221: Need to check that chunk.size >= sizeof(labeltype)

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
+      labeltype id;
+      memcpy((char *)&id, chunk->data() + vector_size_, sizeof(labeltype));
```
allenss-amazon, line 223: same

third_party/hnswlib/bruteforce.h
```diff
@@ -11,203 +9,233 @@
...
 
-      return absl::OkStatus();
-    }
+      SavedVectorT *stored_vector =
+          reinterpret_cast<SavedVectorT *>((*data_)[i]);
+      new (stored_vector) SavedVectorT(
+          vector_constructor(absl::string_view(chunk->data(), vector_size_)));
```
allenss-amazon, line 228: Need to check the size of chunk->size()

third_party/hnswlib/hnswalg.h
```diff
@@ -40,8 +36,8 @@ namespace hnswlib {
 typedef unsigned int tableint;
 typedef unsigned int linklistsizeint;
 
-template <typename dist_t>
-class HierarchicalNSW : public AlgorithmInterface<dist_t> {
+template <typename dist_t, typename SavedVectorT, typename InputVectorT>
```
allenss-amazon, line 39: Same comment about SavedVectorT.

third_party/hnswlib/hnswalg.h
```diff
@@ -225,13 +221,31 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+    return reinterpret_cast<SavedVectorT *>(
+        (*data_level0_memory_)[internal_id] + offsetData_);
+  }
+
+  inline const SavedVectorT &getDataByInternalId(tableint internal_id) const {
+    return *getDataPtrByInternalId(internal_id);
+  }
+
+  inline void setDataByInternalId(tableint internal_id,
+                                  const InputVectorT &datapoint) {
```
allenss-amazon, line 234: SavedVectorT ??

third_party/hnswlib/hnswalg.h
```diff
@@ -225,13 +221,31 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+    SavedVectorT *stored_vector = getDataPtrByInternalId(internal_id);
+    *stored_vector = datapoint;
+  }
+
+  inline void initDataByInternalId(tableint internal_id,
+                                   const InputVectorT &datapoint) {
```
allenss-amazon, line 240: SavedVectorT

allenss-amazon (commented):
the formatting changes make this difficult to review. It would have been better to have split those into a separate PR.

I think we need to see some performance & recall measurement data to make sure that we haven't messed up either. Same with deletes.

yairgott:
Sorry for the confusion but this PR is abandon in favor of #1168.

cmake/Modules/valkey_search.cmake
yairgott: This is not present in the active PRs.

src/commands/ft_aggregate.cc
yairgott: This is present in the active PRs.

src/indexes/index_base.h
yairgott: This change is not present in the active PRs.

third_party/hnswlib/hnswalg.h
yairgott: Both initDataByInternalId and setDataByInternalId operate only on an InputVectorT.

third_party/hnswlib/hnswalg.h
yairgott: Both initDataByInternalId and setDataByInternalId operate only on an InputVectorT.

src/indexes/vector_base.cc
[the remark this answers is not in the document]
yairgott: I've added a proper cleanup [function](https://github.com/valkey-io/valkey-search/blob/memory_sharing/src/indexes/vector_base.cc#L246) into the followup PR.

src/indexes/vector_base.cc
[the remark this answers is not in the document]
yairgott: In the [actual](https://github.com/valkey-io/valkey-search/blob/0a37338743ad04ed402d55ff06fdcb3debffaf6f/src/indexes/vector_base.cc#L286) PR I'm using the write mutex lock. I believe that it's used in the prefilter path.

src/index_schema.cc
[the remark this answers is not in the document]
yairgott: `IsVectorIndex()` is not [deleted](https://github.com/valkey-io/valkey-search/blob/0a37338743ad04ed402d55ff06fdcb3debffaf6f/src/indexes/index_base.h#L29) in the actual PR.

src/indexes/vector_base.cc
yairgott: It may fail on a size check.

third_party/hnswlib/hnswalg.h
yairgott: Which is?

third_party/hnswlib/bruteforce.h
yairgott: I don't believe a size check is necessary here. The chunk size is inherently determined by the number of elements in the saved index. Furthermore, because the legacy RDB format embedded the raw vector data directly, the loaded chunk will always be larger than what we currently need (which is just enough space to hold a pointer to the vector).

third_party/hnswlib/bruteforce.h
yairgott: Yeah, I'm thinking to use VectorRecordT instead of SavedVectorT.

third_party/hnswlib/hnswalg.h
allenss-amazon: Are you saying that the input parameter to setdataByInternalID is a float reference???? I think not. The code itself clearly shows that it's a SavedVectorT.

third_party/hnswlib/bruteforce.h
yairgott: Yeah, I removed it in the subsequent [PR](https://github.com/valkey-io/valkey-search/blob/3b5bf00235952641926db7aa9656c7ba1228b5dc/third_party/hnswlib/bruteforce.h#L38).

third_party/hnswlib/bruteforce.h
yairgott: Yeah, this would be included in the a followup refactoring task I mentioned to you earlier. To make it a little easier on the reviewer, I'm pushing back on any non trivial change or that is required for this PR.

yairgott: You can look at the ChunkedArray as where the point data reside, kind of memory pool.

third_party/hnswlib/bruteforce.h
yairgott: Yeah, this is disgusting but it will all go away soon. Note that in the followup PR I've introduced a couple of methods to make it more readable. The changes in this PR are limited to the scope of what is needed to templetize the vector type without any logic changes.

third_party/hnswlib/bruteforce.h
yairgott: same as my previous replies.

third_party/hnswlib/bruteforce.h
yairgott: same as my previous replies.

third_party/hnswlib/bruteforce.h
yairgott: same as my previous replies.

third_party/hnswlib/bruteforce.h
yairgott: same as my previous replies.

third_party/hnswlib/hnswalg.h
yairgott: I'm not sure I follow what you mean by that. setDataByInternalId is a new function that I've introduced which is prototype is defined above which clearly show that it's using InputVectorT, rather than SavedVectorT.

src/indexes/vector_base.cc
yairgott: Yeah, that is exactly how it's done in the [last](https://github.com/valkey-io/valkey-search/blob/c33e8ce2299df2a8831fd35bb8fa4259bb022f11/src/indexes/vector_base.cc#L130) memory sharing stacked PRs.

src/indexes/vector_base.cc
yairgott: Again, see the [last](https://github.com/valkey-io/valkey-search/blob/c33e8ce2299df2a8831fd35bb8fa4259bb022f11/src/indexes/vector_base.cc#L176) stacked memory sharing PR.

src/indexes/vector_base.cc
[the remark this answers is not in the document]
yairgott: > So GetValue is not on the prefilter code path. Both the prefilter and inline-filter paths complete inside DoSearchVector; MaybeAddIndexedContent is a post-processing step.

Correct and as both of these functions are called while the time sliced read mutex is held, there is no need to hold another mutex as it ensures that no mutations would be made to any of the data structures held by the index until the time sliced read mutex is released.

src/indexes/vector_base.cc
yairgott: The vector is already normalized. Vectors remain normalized in this PR which is changed only in the followup PR.

src/indexes/vector_base.cc
yairgott: This is also going to be changed in the followup PRs.

src/indexes/vector_base.cc
yairgott: How come a key can be empty?

src/indexes/vector_base.h
yairgott: In the subsequent PRs, VectorRecord doesn't store the vector length

src/indexes/vector_base.h
yairgott: same reply

src/indexes/vector_flat.cc
yairgott: The dedup functionality is only introduced in the last stacked PR.

src/indexes/vector_flat.cc
yairgott: make_shared can't be used here as VectorFlat's constructor is declared as private.

src/indexes/vector_flat.cc
yairgott: Right, but again fixed by the last stacked PR.

src/indexes/vector_flat.h
yairgott: Refactored in the last stacked PR.

src/indexes/vector_hnsw.cc
yairgott: Addressed in the last stacked PR.

src/indexes/vector_hnsw.cc
yairgott: Right but this method got eliminated by the followup PRs.

src/indexes/vector_hnsw.cc
yairgott: In a followup PR, I've define the [following](https://github.com/valkey-io/valkey-search/blob/memory_sharing/src/indexes/vector_hnsw.h#L55) as a short name referencing.

src/indexes/vector_hnsw.cc
yairgott: I think we should not use the hnswlib internal id outside of the hnswlib code.

src/indexes/vector_hnsw.cc
yairgott: You are right but I've fixed this in the followup [PR](https://github.com/valkey-io/valkey-search/blob/3b5bf00235952641926db7aa9656c7ba1228b5dc/src/indexes/vector_hnsw.cc#L150).

PR #1168, opened 2026-06-22
Hnswlib templatized vectors

yairgott:
This change introduces a vector memory sharing between the module and the engine to reduce the memory footprint. Instead of duplicating vector data inside the index structure (FLAT and HNSW), valkey-search maintains a single copy of a vector which its reference is used by the indexes and also shared with the engine.

#### Key Changes:
1. **Vector Externalizer**:
   - Introduced `VectorExternalizer` (in [src/vector_externalizer.h](file:///usr/local/google/home/yairg/work/valkey-search4/src/vector_externalizer.h)) to coordinate shared/externalized vector allocations.
   - Added configuration options for memory sharing under `ValkeySearchOptions`.
2. **Index Refactoring**:
   - Integrated `VectorExternalizer` during vector addition/mutation and the RDB load paths.
   - Updated custom `hnswlib` space and distance metrics headers to support externalized vector structures.
3. **Valkey Module Integration**:
   - Updated schema manager and keyspace event manager to coordinate externalization lifecycles correctly.
   - Updated mocks to test Valkey Module APIs and prevent double-frees under memory-shared workloads.
4. **HNSWLib Modifications**:
   - Introduced a space function wrapper to optimize and support Cosine distance calculations efficiently.
5. **Testing**:
   - Extended [testing/vector_test.cc](file:///usr/local/google/home/yairg/work/valkey-search4/testing/vector_test.cc) to verify save/load logic under cosine/L2 metrics with memory sharing.
   - Added tests in [testing/vector_externalizer_test.cc](file:///usr/local/google/home/yairg/work/valkey-search4/testing/vector_externalizer_test.cc) to validate the externalizer lifecycle.


src/commands/ft_aggregate.cc
```diff
@@ -67,8 +67,7 @@ absl::Status ManipulateReturnsClause(AggregateParameters &params) {
       }
       content = true;
       VMSDK_ASSIGN_OR_RETURN(auto indexer, params.index_schema->GetIndex(load));
-      auto indexer_type = indexer->GetIndexerType();
-      if (indexer->IsVectorIndex()) {
+      if (indexes::IsVectorIndex(indexer)) {
```
allenss-amazon, line 70: Why change this? The replacement code is larger and slower.

src/indexes/vector_base.cc
```diff
@@ -111,11 +110,7 @@ query::EvaluationResult PrefilterEvaluator::EvaluateText(
 
 template <typename T>
 T CopyAndNormalizeEmbedding(T *dst, T *src, size_t size) {
-  T magnitude = 0.0f;
-  for (size_t i = 0; i < size; i++) {
-    magnitude += src[i] * src[i];
-  }
-  magnitude = std::sqrt(magnitude);
+  T magnitude = CalcMagnitude(src, size);
```
allenss-amazon, line 113: Both x86 and ARM offer scalar inverse square root instructions. But because you've split CalcMagnitude separate from the divide they can't be generated by the compiler. I'd rather see CalcInverseMagnitude and let it handle the 0 case, etc. It will enable optimization of inverse square root.

src/indexes/vector_base.cc
```diff
@@ -221,22 +226,27 @@ absl::StatusOr<bool> VectorBase::ModifyRecord(const InternedStringPtr &key,
...
-      bool res, UpdateMetadata(key, magnitude.value_or(kDefaultMagnitude),
-                               interned_vector));
+  VMSDK_ASSIGN_OR_RETURN(bool res, UpdateMetadata(key, magnitude, record));
   if (!res) {
     return false;
   }
 
-  auto modify_result = ModifyRecordImpl(internal_id, interned_vector->Str());
+  auto modify_result = ModifyRecordImpl(internal_id, record);
   if (!modify_result.ok()) {
```
allenss-amazon, line 251: same here, this is an internal logic error. would prefer a crash.

src/indexes/vector_base.cc
```diff
@@ -167,16 +162,26 @@ InternedStringPtr VectorBase::InternVector(absl::string_view record,
...
+  }
+  VMSDK_ASSIGN_OR_RETURN(auto internal_id, TrackKey(key, magnitude));
+  absl::Status add_result = AddRecordImpl(internal_id, record);
   if (!add_result.ok()) {
+    auto remove_result = RemoveRecordImpl(internal_id);
+    if (!remove_result.ok()) {
+      VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+          << "While processing error for AddRecord, encountered error in "
+             "RemoveRecordImpl: "
+          << remove_result.message();
```
allenss-amazon, line 183: this is really an internal error. IMO, attempting to recover from an internal error is a bad idea. It's likely to make the current problem worse and it will obscure what actually went wrong. IMO, for this kind of logic failure, you should crash as soon as you detect with as much local context as possible -- to maximize the ability to figure out what went wrong so you can fix it.

src/indexes/vector_base.cc
```diff
@@ -221,22 +226,27 @@ absl::StatusOr<bool> VectorBase::ModifyRecord(const InternedStringPtr &key,
                                               absl::string_view record) {
   // VectorExternalizer tracks added entries. We need to untrack mutations which
   // are processed as modified records.
-  std::optional<float> magnitude;
-  auto interned_vector = InternVector(record, magnitude);
-  if (!interned_vector) {
+
+  if (!IsValidSizeVector(record)) {
     [[maybe_unused]] auto res =
```
allenss-amazon, line 230: I think you sould CHECK(res.ok());

src/indexes/vector_base.cc
```diff
@@ -274,12 +283,13 @@ absl::StatusOr<std::vector<Neighbor>> VectorBase::CreateReply(
...
+  absl::WriterMutexLock lock(&key_to_metadata_mutex_);
   auto it = tracked_metadata_by_key_.find(key);
   if (it == tracked_metadata_by_key_.end()) {
     return absl::NotFoundError("Record was not found");
   }
   std::vector<char> result;
-  char *value = GetValueImpl(it->second.internal_id);
+  const char *value = GetValueImpl(it->second.internal_id);
   if (normalize_) {
     if (it->second.magnitude < 0) {
```
allenss-amazon, line 295: again, would prefer a crash.

src/indexes/vector_base.cc
```diff
@@ -348,16 +349,13 @@ absl::StatusOr<uint64_t> VectorBase::TrackKey(const InternedStringPtr &key,
...
-absl::StatusOr<bool> VectorBase::UpdateMetadata(
-    const InternedStringPtr &key, float magnitude,
-    const InternedStringPtr &vector) {
+
+absl::StatusOr<bool> VectorBase::UpdateMetadata(const InternedStringPtr &key,
+                                                float magnitude,
+                                                absl::string_view vector) {
   if (key->Str().empty()) {
     return absl::InvalidArgumentError("key can't be empty");
   }
```
allenss-amazon, line 362: Not true, empty key is legal key.

src/indexes/vector_base.h
```diff
@@ -43,8 +43,32 @@ enum class QueryOperations : uint64_t;
...
+
+template <typename T>
+T CalcMagnitude(const T *src, size_t size) {
+  T magnitude = static_cast<T>(0);
+  for (size_t i = 0; i < size; i++) {
+    magnitude += src[i] * src[i];
+  }
+  return (magnitude == static_cast<T>(0))
+             ? static_cast<T>(1)
+             : static_cast<T>(std::sqrt(magnitude));
```
allenss-amazon, line 70: as stated elsewhere. This should return 1/sqrt.

src/indexes/vector_base.h
```diff
@@ -43,8 +43,32 @@ enum class QueryOperations : uint64_t;
 
 namespace valkey_search::indexes {
 
+class VectorRecord {
+ public:
+  explicit VectorRecord(const InternedStringPtr &vector)
+      : raw_vector_(vector) {}
+
+  inline const char *GetRawVector() const {
```
allenss-amazon, line 51: Would like to see an accessor that returns a string_view, that'll be useful below.

src/indexes/vector_flat.cc
```diff
@@ -72,28 +68,18 @@ absl::StatusOr<std::shared_ptr<VectorFlat<T>>> VectorFlat<T>::Create(
...
-  return it->second->Str() == vector->Str();
-}
-
-template <typename T>
-void VectorFlat<T>::UnTrackVector(uint64_t internal_id) {
-  absl::MutexLock lock(&tracked_vectors_mutex_);
-  tracked_vectors_.erase(internal_id);
+  const char *stored_vec =
+      algo_->GetDataPtrByInternalId(found->second)->GetRawVector();
+  absl::string_view record(stored_vec, dimensions_ * sizeof(T));
```
allenss-amazon, line 81: Should use the string-view member of VectorRecord. If you want to be paranoid you can CHECK that hte length of the string-view matches dimensions_*sizeof(T)

src/indexes/vector_flat.cc
```diff
@@ -72,28 +68,18 @@ absl::StatusOr<std::shared_ptr<VectorFlat<T>>> VectorFlat<T>::Create(
...
+  return vector == record;
```
allenss-amazon, line 82: Do we really need a full compare? With interning, don't we just need to check the address itself?

src/indexes/vector_flat.cc
```diff
@@ -225,32 +219,38 @@ template <typename T>
...
-    return CreateReply(search_result);
+    norm_record = NormalizeEmbedding(query, GetDataTypeSize());
+    query =
+        absl::string_view((const char *)norm_record.data(), norm_record.size());
+  }
+
+  try {
+    CancelCondition canceler(cancellation_token);
+    VectorRecord embedding(
+        StringInternStore::Intern(query, GetVectorAllocator()));
```
allenss-amazon, line 238: This is copying the query vector. Probably ok now -- is there a plan to remove it?

src/indexes/vector_flat.cc
```diff
@@ -231,32 +225,31 @@ absl::StatusOr<std::vector<Neighbor>> VectorFlat<T>::Search(
...
+    // if (cancellation_token->IsCancelled()) {
+    //   return absl::CancelledError("Search operation cancelled due to
+    //   timeout");
+    // }
+    return CreateReply(res);
+  } catch (const std::exception &e) {
+    Metrics::GetStats().flat_search_exceptions_cnt.fetch_add(
+        1, std::memory_order_relaxed);
+    return absl::InternalError(e.what());
   }
```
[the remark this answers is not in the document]
allenss-amazon: Are we certain that all paths to this function already hold the reader lock?

src/indexes/vector_flat.cc
allenss-amazon, line 260: Isn't greptile correct which is that this lock is unneeded? Or is ComputeDistanceFromRecordImpl actually called by the insertion logic??

src/indexes/vector_hnsw.cc
```diff
@@ -104,31 +92,23 @@ absl::StatusOr<std::shared_ptr<VectorHNSW<T>>> VectorHNSW<T>::Create(
...
+                                  absl::string_view vector) {
+  absl::ReaderMutexLock lock(&resize_mutex_);
+  {
+    std::unique_lock<std::mutex> lock_label(
+        algo_->getLabelOpMutex(internal_id));
+    auto id = GetInternalId(algo_.get(), internal_id);
+    if (!id.has_value()) {
+      return false;
+    }
+    const auto &stored_record = algo_->GetDataByInternalId(*id);
```
allenss-amazon, line 107: Would prefer to use a string-view accessor in VectorRecord, ideally CHECK that its size matches GetVectorDataSize.

src/indexes/vector_hnsw.cc
```diff
@@ -230,7 +213,13 @@ template <typename T>
 absl::Status VectorHNSW<T>::SaveIndexImpl(
     RDBChunkOutputStream chunked_out) const {
   absl::ReaderMutexLock lock(&resize_mutex_);
-  return algo_->SaveIndex(chunked_out);
+  auto serializer = [vector_size =
+                         GetVectorDataSize()](const VectorRecord &record) {
+    return std::vector<char>(record.GetRawVector(),
+                             record.GetRawVector() + vector_size);
```
allenss-amazon, line 220: Again, another place for a string_view accessor.

src/indexes/vector_hnsw.h
```diff
@@ -23,30 +22,47 @@
...
+      ValkeyModuleCtx *ctx, const AttributeDataType *attribute_data_type,
+      const data_model::VectorIndex &vector_index_proto,
       absl::string_view attribute_identifier,
-      SupplementalContentChunkIter&& iter) ABSL_NO_THREAD_SAFETY_ANALYSIS;
+      SupplementalContentChunkIter &&iter) ABSL_NO_THREAD_SAFETY_ANALYSIS;
   ~VectorHNSW() override = default;
   size_t GetDataTypeSize() const override { return sizeof(T); }
 
-  const hnswlib::SpaceInterface<float>* GetSpace() const {
+  const hnswlib::SpaceInterface<float> *GetSpace() const {
```
allenss-amazon, line 65: <float> -> <T> ??

third_party/hnswlib/bruteforce.h
```diff
@@ -25,189 +20,246 @@
 #endif
 
 namespace hnswlib {
-template <typename dist_t>
-class BruteforceSearch : public AlgorithmInterface<dist_t> {
+template <typename dist_t, typename SavedVectorT>
```
allenss-amazon, line 23: Please see the extended comments on this file in #1166.

third_party/hnswlib/hnswalg.h
```diff
@@ -42,8 +31,9 @@ namespace hnswlib {
 typedef unsigned int tableint;
 typedef unsigned int linklistsizeint;
 
-template <typename dist_t>
-class HierarchicalNSW : public AlgorithmInterface<dist_t> {
+template <typename dist_t, typename InputVectorT, typename SavedVectorT>
```
allenss-amazon, line 34: I'm not finding InputVectorT and SavedVectorT to be very mnemonic. I **think** you mean QueryVectorT and DBVectorT or DBVectorRefT ?

allenss-amazon (commented):
I'd prefer to see the changes in the build environment broken out into a separate PR.

src/commands/ft_aggregate.cc
yairgott: It's a cleaner interface.

src/indexes/vector_base.cc
yairgott: This code is refactored in a latter PR.

src/indexes/vector_base.cc
yairgott: This code is refactored by a latter PR.

src/indexes/vector_base.cc
yairgott: same reply.

src/indexes/vector_base.cc
yairgott: same

src/indexes/vector_base.cc
yairgott: not true, how can an empty key be legal?

src/indexes/vector_base.h
yairgott: Refactored by a latter PR.

src/indexes/vector_base.h
yairgott: Not true, magnitude is calculated with sqrt

src/indexes/vector_flat.cc
yairgott: Refactored by a latter PR.

src/indexes/vector_flat.cc
yairgott: Refactored by a latter PR.

src/indexes/vector_flat.cc
yairgott: Refactored by a latter PR to avoid copy.

src/indexes/vector_flat.cc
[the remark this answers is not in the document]
yairgott: the multi reader/writer mutex ensures mutual exclusion between search and modification of the indexes.

src/indexes/vector_hnsw.cc
yairgott: By design VectorRecord does not store the vector size.

src/indexes/vector_base.cc
yairgott: I've addressed this in the followup PR, see: https://github.com/valkey-io/valkey-search/blob/62693bab52b9196293765dc1c5d5d56db86f26cd/src/indexes/vector_base.cc#L80

src/indexes/vector_flat.cc
yairgott: It's correct and it is addressed by a [followup PR change](https://github.com/valkey-io/valkey-search/blob/6c40d4e21e4ecafb20c2018864f597aa71576f8a/src/indexes/vector_base.cc#L525).

src/indexes/vector_hnsw.cc
yairgott: As mentioned in a previous comment, a VectorRecord doesn't hold the vector length.

src/indexes/vector_hnsw.h
yairgott: Please clarify

third_party/hnswlib/bruteforce.h
yairgott: I believe this is addressed in the followup PRs.

third_party/hnswlib/hnswalg.h
yairgott: The typenames have been changed to [QueryVectorT, StoredVectorT] in the followup PR.

Aksha1812:
**Reviewers for this PR**

allenss-amazon (approved):

PR #1184, opened 2026-06-25
feat: implement alias commands

rileydes-improving:
## Summary

Adds `FT.ALIASADD`, `FT.ALIASDEL`, and `FT.ALIASUPDATE` with RediSearch-compatible behavior. Also adds `FT.ALIASLIST` (not present in RediSearch) for listing all active aliases. Aliases live as a repeated field on the owning `IndexSchema` proto, persist in RDB as part of the index, and propagate cluster-wide through the existing `MetadataManager` / `FT.INTERNAL_UPDATE` path. No separate metadata type needed.

### Proto & IndexSchema

- Added `repeated string aliases = 15` to `IndexSchema` in `index_schema.proto`
- Extended `IndexSchema` class with `GetAliases()` / `SetAliases()` accessors and in-memory alias list

### SchemaManager — Alias CRUD & Resolution

- Added `db_to_aliases_` (`flat_hash_map<uint32_t, flat_hash_map<string, string>>`) as a forward alias map (alias → index name) per database
- Implemented `AddAlias`, `RemoveAlias`, `UpdateAlias`, and `GetAllAliases`
- In coordinator mode: each alias operation fetches the owning index's proto from `MetadataManager`, mutates the `aliases` repeated field, normalizes defaults, and re-commits via `CreateEntry`, which triggers `OnMetadataCallback` cluster-wide
- In standalone mode: directly mutates the in-memory alias map and `IndexSchema`'s alias list under the mutex
- `GetIndexSchema` falls through to alias resolution when direct name lookup misses
- `RemoveIndexSchemaInternal` purges all aliases for a dropped index via `EraseAliasesForIndex`
- `OnMetadataCallback` uses `MessageDifferencer` (ignoring `aliases` and `stats` fields) to detect alias-only changes: when structural fields are unchanged, it calls `RebuildAliasMapsForIndex` and `SetAliases` without tearing down the index
- `ComputeFingerprint` clears the `aliases` field before hashing so alias mutations don't destabilize search consistency checks
- `OnFlushDBEnded` in coordinator mode fetches the authoritative proto from `MetadataManager` (including aliases) when recreating indexes after FLUSHDB
- `OnSwapDB` swaps alias maps in lockstep with index maps

### FT.INTERNAL_UPDATE, Type Routing (Requested by Allen)

- Extended `FT.INTERNAL_UPDATE` to accept optional keyword/count/args-style parameters after the required 4 arguments. Currently recognizes `TYPE <1> <type_name>` (defaults to `kSchemaManagerMetadataTypeName` for AOF backward compatibility).
- `ReconcileMetadata` now skips replication for types not registered locally, preventing errors on nodes that don't recognize a metadata type.

### Commands

- `FTAliasAddCmd`, `FTAliasDelCmd`, `FTAliasUpdateCmd` in `ft_alias_add.cc`, `ft_alias_del.cc`, `ft_alias_update.cc`, registered in `module_loader.cc` with `@search @write @fast` ACL permissions
- `FTAliasListCmd` in `ft_alias_list.cc`, returns all alias->index pairs for the selected database
- In cluster (CME) mode, `ALIASADD`/`ALIASUPDATE` perform a consistency-check fanout (`AliasExistsConsistencyCheckFanoutOperation`). `ALIASDEL` uses `AliasRemovedConsistencyCheckFanoutOperation` to wait until the change is visible on all primaries.
- MULTI/EXEC and Lua are rejected in CME mode to avoid local-only mutations without fanout
- `ft_dropindex.cc` resolves an alias to its real index name before calling `RemoveIndexSchema`
- `ft_aggregate.cc` uses the alias field when constructing its response

### FT.INFO

- `ft_info_parser.cc` calls `GetAliases()` on the `IndexSchema` and reports a sorted `aliases` array in `FT.INFO` output

## Testing

- Unit tests in `ft_alias_commands_test.cc`: argument count errors, happy paths, duplicate rejection, alias-to-alias rejection, self-referential alias, and `ReplicateVerbatim` behavior in both coordinator modes
- Unit tests in `schema_manager_alias_property_test.cc`: property-based tests for add/remove/update round-trips, cross-database isolation, upsert semantics, FLUSHDB/SWAPDB behavior, alias-only metadata callback, and fingerprint stability
- Unit tests in `ft_info_test.cc`: sorted `aliases` array and querying via alias still reports the real index name
- Integration tests in `test_ft_alias.py`: all four commands, DROPINDEX cleanup, FT.SEARCH/FT.AGGREGATE via alias, FLUSHDB, SWAPDB, RDB persistence, wrong argument errors, and name-collision edge cases
- Integration tests in `test_ft_alias_cluster.py`: ALIASADD, ALIASDEL, ALIASUPDATE, ALIASLIST, and DROPINDEX tombstoning all propagate to every cluster primary
- Compatibility: `alias-answers.pickle.gz` added to the parametrized run in `compatibility_test.py`. Alias management commands are replayed inline to reconstruct alias state before dependent search/aggregate queries.

## Architecture Difference vs. PR #955

PR #955 introduced a separate metadata type (`vs_alias`) with its own `MetadataManager` registration, `OnAliasMetadataCallback`, and dedicated `RDB_SECTION_ALIAS_MAP` section. This PR stores aliases directly as a repeated field on `IndexSchema`, treating alias mutations as proto updates through the existing `vs_index_schema` metadata pathway. That removes the need for a separate metadata type, a separate RDB section, and the ordering constraint (indexes must load before aliases). Alias changes propagate via the same `CreateEntry` call used for index creation. `OnMetadataCallback` detects alias-only changes via `MessageDifferencer` and applies them without index teardown.

## Example

```
FT.CREATE my_idx ON HASH PREFIX 1 doc: SCHEMA category TAG
HSET doc:1 category books
FT.ALIASADD my_alias my_idx
FT.SEARCH my_alias "@category:{books}"
```

Returns 1 result (`doc:1`) via the alias without any change to the underlying index.


integration/compatibility/generate.py
```diff
@@ -307,6 +307,7 @@ def test_aggregate_numeric_dyadic_operators_sortable_numbers(self, key_type, dia
                 f"ft.aggregate {key_type}_idx1  * load 3 @__key @n1 @n2 apply @n1{op}@n2 as nn"
             )
 
+    @pytest.mark.skip(reason="Needs research")
```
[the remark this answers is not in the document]
rileydes-improving: This is matching the pattern in the rest of the file.

src/schema_manager.cc
```diff
@@ -865,4 +1056,429 @@ static vmsdk::info_field::Integer total_active_write_threads(
...
+    // Re-commit the modified proto.
+    auto any_proto = std::make_unique<google::protobuf::Any>();
+    any_proto->PackFrom(schema_proto);
+    auto result = coordinator::MetadataManager::Instance().CreateEntry(
+        kSchemaManagerMetadataTypeName,
+        coordinator::ObjName(db_num, index_name), std::move(any_proto));
+    if (!result.ok()) {
+      return result.status();
+    }
+    return absl::OkStatus();
```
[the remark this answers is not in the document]
rileydes-improving: It an eventual consistency tradeoff that is documented within the code.

docs/commands/ft.alias.md
```diff
@@ -0,0 +1,123 @@
+Aliases provide alternative names for indexes. Once created, an alias can be used in place of the real index name in `FT.SEARCH`, `FT.AGGREGATE`, `FT.INFO`, and `FT.DROPINDEX` commands. Multiple aliases can point to the same index. Aliases are per-database and are persisted across server restarts.
```
allenss-amazon: persisted across restarts isn't entirely correct. Rather say that alias definitions are global and part of the aliased index definition.

docs/commands/ft.alias.md
```diff
@@ -0,0 +1,123 @@
...
+
+Alias names must be non-empty strings that do not contain null bytes. An alias name must not match the name of an existing index (to avoid ambiguous resolution). The target index name must refer to a real index, not another alias (alias chaining is not supported).
```
allenss-amazon: Aliases of single slot indexes, must contain the same hashtag in their name. 
 
(This needs to get added and tested).

docs/commands/ft.alias.md
```diff
@@ -0,0 +1,123 @@
...
+- `<index-name>` (required): The name of the target index. Must be an existing index (not an alias).
+
+## Result
+
+Returns `OK` on success or an error.
+
+## Errors
+
+- `Alias name cannot be empty` — the alias argument is an empty string.
+- `Alias name must not contain null bytes` — the alias contains embedded null characters.
```
allenss-amazon: Why the prohibition around null bytes?

docs/commands/ft.alias.md
```diff
@@ -0,0 +1,123 @@
...
+- `Alias already exists` — an alias with this name is already registered.
+- `Alias collides with existing index name` — the alias matches a real index name.
```
allenss-amazon: Are we sure that Redisearch also disallows this case?

docs/commands/ft.alias.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+- `<alias>` (required): The alias name to create. Must be non-empty and must not already exist as an alias or match an existing index name.
+- `<index-name>` (required): The name of the target index. Must be an existing index (not an alias).
+
+## Result
+
+Returns `OK` on success or an error.
+
+## Errors
```
allenss-amazon, line 20: Need to add cluster inconsistency / timeout  to all of these sections.

integration/compatibility/data_sets.py
```diff
@@ -621,18 +624,63 @@ def generate_doc(doc_id):
...
+                ["FT.ALIASUPDATE", "alias_agg",    "json_idx1"],
+            ]
+    return data
+
+
 def load_data(client, data_set, key_type, data_source=None, schema_type="default"):
     # Auto-detect data source based on data_set name
     if data_source is None:
-        data_source = "text" if data_set in TEXT_DATASETS else "vector"
+        if data_set == "alias":
```
allenss-amazon: match data_set ?

integration/test_ft_alias_cluster.py
```diff
@@ -0,0 +1,1313 @@
...
+CREATE_TAG_INDEX_2 = [
+    "FT.CREATE", INDEX_NAME_2,
+    "ON", "HASH",
+    "PREFIX", "1", "doc2:",
+    "SCHEMA", "category", "TAG",
+]
+
+
+def _wait_for_alias_on_all_nodes(nodes, alias_name, expect_present=True,
+                                  timeout=10):
```
allenss-amazon: Is there any reason to override the default timeout machinery in the waiters? This is historically a source of flaky tests. Would prefer not to override.

integration/test_ft_alias_cluster.py
```diff
@@ -0,0 +1,1313 @@
...
+                client = self.new_client_for_primary(node_idx)
+                results[node_idx] = client.execute_command(
+                    "FT.ALIASADD", "contested_alias", INDEX_NAME
+                )
+            except ResponseError as e:
+                errors[node_idx] = e
+
+        threads = [
+            threading.Thread(target=add_alias, args=(i,))
+            for i in range(self.CLUSTER_SIZE)
```
allenss-amazon: The strategy here isn't guaranteeing collisions. You'll need to use pausepoints to choreograph the collisions and add some info fields to prove that the retry paths were actually exercised.

src/commands/ft_aggregate.cc
```diff
@@ -349,7 +349,7 @@ absl::Status GenerateResponse(ValkeyModuleCtx *ctx,
     for (size_t i = 0; i < rec->fields_.size(); ++i) {
       if (ReplyWithValue(
               ctx, parameters.index_schema->GetAttributeDataType().ToProto(),
-              parameters.record_info_by_index_[i].identifier_,
+              parameters.record_info_by_index_[i].alias_,
```
allenss-amazon: Should this be part of a separate PR ?

src/commands/ft_alias.cc
```diff
@@ -0,0 +1,301 @@
...
+  }
+}
+
+}  // namespace
+
+absl::Status FTAliasAddCmd(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
+                           int argc) {
+  if (argc != 3) {
+    return absl::InvalidArgumentError(vmsdk::WrongArity(kAliasAddCommand));
+  }
```
allenss-amazon: If you register with 3 command parameters you don't need this.

src/coordinator/metadata_manager.cc
```diff
@@ -969,15 +977,15 @@ absl::Status MetadataManager::CallFTInternalUpdateForReconciliation(
...
-  vmsdk::UniqueValkeyDetachedThreadSafeContext thread_safe_ctx;
 
-  reply =
-      ValkeyModule_Call(safe_context, "FT.INTERNAL_UPDATE", "!Kcbb", id.c_str(),
-                        metadata_binary.data(), metadata_binary.size(),
-                        header_binary.data(), header_binary.size());
+  reply = ValkeyModule_Call(
+      safe_context, "FT.INTERNAL_UPDATE", "!Kcbbccc", id.c_str(),
+      metadata_binary.data(), metadata_binary.size(), header_binary.data(),
+      header_binary.size(), "TYPE", "1", type_name_str.c_str());
```
allenss-amazon: Let's not include TYPE when it's the default value -- it improves inter-release compatibility.

src/coordinator/metadata_manager.cc
```diff
@@ -993,16 +1001,18 @@ absl::Status MetadataManager::CallFTInternalUpdateForReconciliation(
...
 
-  // Replicate FT.INTERNAL_UPDATE to replicas for AOF consistency
-  ValkeyModule_Replicate(detached_ctx_.get(), "FT.INTERNAL_UPDATE", "cbb",
+  // Always emit TYPE keyword for consistency
+  std::string type_name_str(type_name);
+  ValkeyModule_Replicate(detached_ctx_.get(), "FT.INTERNAL_UPDATE", "cbbccc",
                          std::string(encoded_id).c_str(),
                          metadata_binary.data(), metadata_binary.size(),
-                         header_binary.data(), header_binary.size());
+                         header_binary.data(), header_binary.size(), "TYPE",
```
allenss-amazon: same here.

allenss-amazon (commented):
1. Handling of aliases for single-slot indexes. An alias for a single-slot index must have the same hash-tag as the underlying index -- else it's an error. This must be enforced and tested for.
2. The change in the FT.INTERNAL_UPDATE to add the additional type should only be written when the type is different. When it's the current namespace, then suppress it for forward compatibility.
3. If an FT.ALIASxxxxx fails the consistency test, then it might end up with a stale entry in the forward map -- in this case the entire map forward map should get rebuilt. In fact, I suspect we should just always rebuild the entire forward map at the end of every FT.ALIASxxxxx command (and the FT.INTERNAL_UPDATE too). This allows the detection of duplicate entries -- though I'm not sure how to reconcile that, but at least we should have a log message.
4. in test_concurrent_aliasadd_same_alias_one_wins, need to assert that the loser doesn't have any cruft left lying around in the index table.
5. Are we testing alias updates colliding with FT.DROPINDeX ?
6. The integrations for consistency collisions are weak. They need to have pausepoints added to ensure the correct testing and info counters to prove that the conflict resolution paths were executed.

docs/commands/ft.alias.md
rileydes-improving: From my testing

Empty Name Alias - RediSearch accepts it silently (returns OK). Worth diverging because an empty string is never a valid identifier. Rejecting early and explicitly is more clear behavior.

Alias Name With Null Bytes - It appears that my implementation works with a minor tweak to properly handle null bytes to match Redisearch.

Alias name collides with existing index - RediSearch has inconsistent behavior: ALIASADD allows it (the alias shadows the index name), but ALIASUPDATE rejects it with an error. Worth partially diverging to reject on both commands, shadowing an index name makes that index unreachable by name, which would be unexpected and frustrating.

Alias already exists - RediSearch rejects the duplicate ALIASADD with an error. Same behavior we implement.

docs/commands/ft.alias.md
rileydes-improving: Fixed. Now reads: "Alias definitions are global and stored as part of the aliased index definition."

docs/commands/ft.alias.md
rileydes-improving: Done. `AddAlias`/`UpdateAlias` now validate via `vmsdk::ParseHashTag`. If the target index uses a `{hashtag}`, the alias must contain the same one or you get `InvalidArgument`.

unit & integration test (standalone + cluster) covers: matching hashtag (pass), no hashtag (reject), different hashtag (reject), non-hashtag index (anything goes). 

Docs updated. Added to `ft.alias.md`: "For single-slot indexes (those with a `{hashtag}` in the name), aliases must contain the same hashtag to ensure correct cluster slot routing."

docs/commands/ft.alias.md
rileydes-improving: Removed entirely, initially some of the alias processing was failing with null bytes but was able to resolve it. Code no longer rejects null-byte aliases. Test renamed to `NullByteAliasAcceptanceTest`. Compat tests in `generate_alias.py` exercise null-byte aliases.

docs/commands/ft.alias.md
rileydes-improving: After the discussion yesterday, both ALIASADD and ALIASUPDATE allow shadowing an existing index name. Collision rejection removed from both paths. There's no good reason to prevent it — the alias simply takes priority in name resolution, and the index remains accessible via its full name in internal operations. Tests updated to expect success on both commands.

docs/commands/ft.alias.md
rileydes-improving: Done. Each alias command's error section now documents the cluster inconsistency timeout scenario.

integration/compatibility/data_sets.py
rileydes-improving: Done. `load_data`/`load_data_cluster` converted from if/elif chains to `match` statements.

integration/test_ft_alias_cluster.py
rileydes-improving: Fixed. `_wait_for_alias_on_all_nodes` no longer takes a timeout param. It calls `waiters.wait_for_true(check)` with no override, so it uses `TEST_MAX_WAIT_TIME_SECONDS` (90s, 180s for ASAN) from the test framework.

integration/test_ft_alias_cluster.py
rileydes-improving: Done. `test_ft_alias_cluster.py` adds pausepoint based tests with guaranteed collisions to replace the previous threading-based collision tests. The new ones use `FT._DEBUG PAUSEPOINT` + `CONTROLLED_VARIABLE` to force propagation ordering, then assert on `search_info_fanout_retry_count` to prove retry paths fired.

src/coordinator/metadata_manager.cc
rileydes-improving: Done. Only conditionally emit TYPE only for non-default type names.

src/coordinator/metadata_manager.cc
rileydes-improving: Done. Only conditionally emit TYPE only for non-default type names.

src/commands/ft_alias.cc
rileydes-improving: Done. Added `arity` field to `CommandOptions` and wired `ValkeyModule_SetCommandInfo` into `RegisterCommands`. All alias commands now declare their arity at registration:

- `FT.ALIASADD` → 3
- `FT.ALIASDEL` → 2
- `FT.ALIASUPDATE` → 3
- `FT.ALIASLIST` → 1

The server rejects wrong-arity calls before they reach the handler, so the manual `argc` guards could be removed.

I didn't add modify any of the prexisting commands with manual arity checks, but with the wiring, should be very straightforward.

src/commands/ft_aggregate.cc
rileydes-improving: Likely worth tracking and fixing separately, I filed https://github.com/valkey-io/valkey-search/issues/1243. 

Reverted the ft_aggregate.cc change, the alias compatibility tests are patched to work around issue to match the existing aggregate and search workaround. The bug report tracks the underlying issue (JSON identifiers returned with $. prefix instead of the user-facing alias) for a dedicated fix in a different PR.

rileydes-improving:
> 1. Handling of aliases for single-slot indexes. An alias for a single-slot index must have the same hash-tag as the underlying index -- else it's an error. This must be enforced and tested for.
> 2. The change in the FT.INTERNAL_UPDATE to add the additional type should only be written when the type is different. When it's the current namespace, then suppress it for forward compatibility.
> 3. If an FT.ALIASxxxxx fails the consistency test, then it might end up with a stale entry in the forward map -- in this case the entire map forward map should get rebuilt. In fact, I suspect we should just always rebuild the entire forward map at the end of every FT.ALIASxxxxx command (and the FT.INTERNAL_UPDATE too). This allows the detection of duplicate entries -- though I'm not sure how to reconcile that, but at least we should have a log message.
> 4. in test_concurrent_aliasadd_same_alias_one_wins, need to assert that the loser doesn't have any cruft left lying around in the index table.
> 5. Are we testing alias updates colliding with FT.DROPINDeX ?
> 6. The integrations for consistency collisions are weak. They need to have pausepoints added to ensure the correct testing and info counters to prove that the conflict resolution paths were executed.
1 Tested
2 Done
For 3, in coordinator mode, AddAlias does not modify the forward map directly. It modifies the proto stored in MetadataManager via CreateEntry, which triggers OnMetadataCallback, which calls RebuildAliasMapsForIndex. The in-memory forward map is only updated as a result of the authoritative callback. In standalone, the operation is atomic under the mutex. The existing code already has RebuildAliasMapsForIndex doing a full erase+repopulate from the proto on every metadata callback. 
4 Fixed
5 & 6, added testing

integration/test_ft_alias_cluster.py
```diff
@@ -0,0 +1,2087 @@
...
+                assert idx is not None
+                assert info[idx + 1] != INDEX_NAME.encode(), (
+                    f"Alias still points to dropped index: {info[idx + 1]}")
+
+        # No cruft — FT.ALIASLIST consistent across all nodes.
+        alias_lists = set()
+        for node in self._all_primaries():
+            alias_lists.add(str(node.execute_command("FT.ALIASLIST")))
+        assert len(alias_lists) == 1, (
+            f"FT.ALIASLIST inconsistent across nodes: {alias_lists}")
```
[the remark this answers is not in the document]
rileydes-improving: Fixed with my pushed changes.

integration/test_ft_alias_cluster.py
```diff
@@ -0,0 +1,2087 @@
...
+        else:
+            # Safety fallback.
+            node0.execute_command(
+                "FT._DEBUG PAUSEPOINT RESET fanout_remote_pausepoint")
+            node1.execute_command(
+                "FT._DEBUG CONTROLLED_VARIABLE SET PauseHandleClusterMessage no")
+            exception = ValueError(f"Invalid order: {order}")
+    except Exception as e:
+        exception = e
+    return exception
```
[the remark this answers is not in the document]
rileydes-improving: Fixed with my pushed changes.

testing/ft_alias_commands_test.cc
```diff
@@ -0,0 +1,827 @@
...
+            .return_code = absl::StatusCode::kInvalidArgument,
+        },
+        {
+            // alias == index_name is allowed (shadows the index until
+            // the index is resolved by direct lookup).
+            .test_name = "alias_same_as_index_name",
+            .argv = {"FT.ALIASADD", "test_idx", "test_idx"},
+            .index_schema_pbtxt = std::string(kTestIndexSchemaPbtxt),
+            .pre_existing_alias = std::nullopt,
+            .return_code = absl::StatusCode::kOk,
```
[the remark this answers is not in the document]
rileydes-improving: Fixed with my pushed changes.

testing/schema_manager_alias_property_test.cc
```diff
@@ -0,0 +1,2243 @@
...
+    std::sort(proto->mutable_aliases()->begin(),
+              proto->mutable_aliases()->end());
+
+    auto packed = std::make_unique<google::protobuf::Any>();
+    packed->PackFrom(*proto);
+    auto entry_result = coordinator::MetadataManager::Instance().CreateEntry(
+        kSchemaManagerMetadataTypeName,
+        coordinator::ObjName(kDbNum, index_name), std::move(packed));
+    ASSERT_TRUE(entry_result.ok()) << entry_result.status();
+  }
```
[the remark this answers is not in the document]
rileydes-improving: Fixed with my pushed changes.

src/commands/ft_alias.cc
```diff
@@ -0,0 +1,301 @@
...
+    return ValkeyModule_ReplyWithSimpleString(ctx, "OK");
+  }
+
+  void ResetForRetry() override {}
+
+  bool ShouldRetry() override {
+    return !inconsistent_state_error_nodes.empty() ||
+           !communication_error_nodes.empty() ||
+           index_name_error_nodes.size() != targets_.size();
+  }
```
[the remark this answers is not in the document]
rileydes-improving: The ResetBaseForRetry() call in the retry loop clears inconsistent_state_error_nodes before each retry.

Aksha1812:
**Reviewers for this PR**

Frank-Gu-81:
Hi @rileydes-improving 👋 — this is tracked as a **P2** for **valkey-search 1.3**. P2s aren't RC1 blockers, but we'd love to land them for **GA**.

COMMANDS.md
VoletiRam, line 1: Do we need to update this legacy command file anymore? I see it doesn't have FT.AGGREGATE, TEXT fields, and many FT.SEARCH options that exist in `docs/commands/`

Not in this PR but may be we can remove this file to avoid confusion.

README.md
```diff
@@ -18,6 +18,10 @@ FT.INFO
 FT._LIST
 FT.SEARCH
 FT.AGGREGATE
+FT.ALIASADD
+FT.ALIASDEL
+FT.ALIASLIST
+FT.ALIASUPDATE
```
VoletiRam, line 24: Do we need to update here anymore?

src/schema_manager.h
```diff
@@ -158,6 +190,10 @@ class SchemaManager {
       uint32_t, absl::flat_hash_map<std::string, std::shared_ptr<IndexSchema>>>
       db_to_index_schemas_ ABSL_GUARDED_BY(db_to_index_schemas_mutex_);
 
+  // Forward alias map: db_num → {alias → index_name}
+  absl::flat_hash_map<uint32_t, absl::flat_hash_map<std::string, std::string>>
+      db_to_aliases_ ABSL_GUARDED_BY(db_to_index_schemas_mutex_);
+
```
VoletiRam, line 196: Aliases are stored in two places: the forward map here on `SchemaManager`, and a vector on `IndexSchema` via `SetAliases()`/`GetAliases()`. Every write path (AddAlias, RemoveAlias, UpdateAlias, CreateIndexSchemaInternal, RebuildAliasMapsForIndex, LoadIndex, OnLoadingEnded, RemoveIndexSchemaInternal) has to keep both in sync.

What's the reason for storing them on `IndexSchema` as well? Is it for `ToProto()` serialization and `RespondWithInfo()` output?

If so, could we keep `IndexSchema` unaware of aliases instead? `SaveIndexes()` already holds the lock and walks every schema, so it could inject aliases from `db_to_aliases_` at save time. `RespondWithInfo()` could take aliases as a parameter. That would give us one source of truth and remove the sync requirement from all those write paths.

src/schema_manager.cc
```diff
@@ -1029,4 +1278,432 @@ static vmsdk::info_field::Integer total_active_write_threads(
...
+          return absl::InvalidArgumentError(
+              "Unknown index name or name is an alias");
+        }
+        if (db_alias_it->second.contains(alias)) {
+          return absl::AlreadyExistsError("Alias already exists");
+        }
+      }
+    }
+
+    // Fetch the target index proto from MetadataManager.
```
VoletiRam, line 1318: The `GetEntryContent` → `UnpackTo` → `NormalizeIndexSchemaProtoDefaults` → `PackFrom` → `CreateEntry` sequence appears five times across `AddAlias`, `RemoveAlias`, and `UpdateAlias` (three in `UpdateAlias` for add/remove/rollback). The only thing that varies is the one-line proto mutation in the middle.

If the normalize step ever needs to change (e.g. a new default field gets added), all five sites need updating or `OnMetadataCallback`'s `MessageDifferencer` will misclassify alias-only changes as structural and trigger full index teardown/rebuild.

Could we extract something like `MutateIndexProtoInMetadata(db_num, index_name, mutate_fn)` that handles the boilerplate and takes a lambda for the proto mutation?

docs/commands/ft.alias.md
```diff
@@ -0,0 +1,123 @@
+Aliases provide alternative names for indexes. Once created, an alias can be used in place of the real index name in `FT.SEARCH`, `FT.AGGREGATE`, `FT.INFO`, and `FT.DROPINDEX` commands. Multiple aliases can point to the same index. Alias definitions are global and stored as part of the aliased index definition.
+
+Alias names must be non-empty strings. The target index name must refer to a real index, not another alias (alias chaining is not supported). For single-slot indexes (those with a `{hashtag}` in the name), aliases must contain the same hashtag to ensure correct cluster slot routing. An alias name may match an existing index name; in that case the real index takes precedence during name resolution until it is dropped, at which point the alias becomes effective.
```
VoletiRam, line 3: What happens if you create alias "foo" pointing to index "bar", then create a new index named "foo"? The `GetIndexSchema` resolution checks real indexes first, so "foo" would resolve to the new index, effectively shadowing the alias. Is that intentional? Is it documented?

src/schema_manager.h
rileydes-improving: Yeah, originally it was for handling serialization but I reworked it to not require as much syncing. Removed aliases_, SetAliases(), and GetAliases() from IndexSchema. SchemaManager::db_to_aliases_ is now the only place aliases live, passed into ToProto() and RespondWithInfo() as an argument at call time.

docs/commands/ft.alias.md
rileydes-improving: Intentional, undocumented until now. Real index always wins name resolution, that was the overall intention. Added a warning too: FT.CREATE warns for a name already claimed by an alias. Doc now states both rules explicitly.

src/schema_manager.cc
rileydes-improving: Sounds good, extracted MutateIndexProtoInMetadata() to cover fetch, unpack, normalize, pack, and commit in one place. AddAlias, RemoveAlias, and all three UpdateAlias sites (add, remove, rollback) now call it.

PR #1187, opened 2026-06-26
RemoveSubscription improper destruction handling fix

linbran5123:
# Summary
Fix replica crash in `KeyspaceEventManager::RemoveSubscription` when `IndexSchema::LoadFromRDB` fails partway through populating a new schema. Adds a deterministic regression test.

## Root cause
When `LoadFromRDB` errors out, the staged IndexSchema is dropped on the failure path on the main thread. Worker threads parked inside ScheduleMutation can hold the last strong refs via `weak_index_schema.lock()`. The last worker to exit ProcessSingleMutationAsync drops its ref, runs ~IndexSchema() on a worker thread, sees is_destructing_ == false, and calls MarkAsDestructing — which tries to access subscriptions_ (a main-thread-only resource) and aborts.

## Fix
Add an `absl::Cleanup` in LoadFromRDB that calls MarkAsDestructing on the main thread on every error path. The cleanup is cancelled at the success point so it only fires when the load aborts.

When LoadFromRDB errors out:

1. The cleanup runs on the main thread before LoadFromRDB returns. It removes the keyspace subscription (the only main-thread-only resource), notifies waiting queries, clears tracked_mutated_records_, and sets is_destructing_ = true.
2. The main thread's local shared_ptr<IndexSchema> drops as LoadFromRDB returns.
3. Multiple workers may still hold strong refs to the staged schema. Each one eventually finishes its ProcessSingleMutationAsync task and drops its ref. Workers that wake from the `mutation_processing` pause point (or backpressure) call ConsumeTrackedMutatedAttribute, find the now-empty tracked_mutated_records map, and exit cleanly.
4. The worker that happens to drop the last strong ref runs ~IndexSchema() on its thread. With the fix, `is_destructing_ == true`, so the destructor skips the call to MarkAsDestructing and never touches the main-thread-only subscriptions_. The rest of the destructor body is thread-safe on a worker: the log line, Collector::IndexDropped (which enqueues the main-thread-only metering cleanup via RunByMain and returns), and C++ member destructors.

**Net effect:** the main-thread-only work happens on the main thread proactively; whichever worker eventually runs the destructor only does thread-safe work.

## Test

`test_rdb_load_error_cleanup.py` reproduces the bug deterministically. If the bug window cannot be engaged (partial resync, workers drained too fast), the test reports diagnostics and skips rather than passing silently. A real test infrastructure failure is reported as a fail rather than a skip.

### Regression Testing

**Without the code fix, this is what the test returns:**

```
assert replica.is_alive(), (
            "Replica process died after RDB load error. IndexSchema "
            "destructor ran on a worker thread without MarkAsDestructing "
            "being called on the main thread - the bug in ELMO-119562 is "
            "unfixed.")
E       AssertionError: Replica process died after RDB load error. IndexSchema destructor ran on a worker thread without MarkAsDestructing being called on the main thread - the bug in ELMO-119562 is unfixed.
E       assert False
E        +  where False = <bound method RedisServerHandle.is_alive of RedisServerHandle(port=7617, bind_ip=0.0.0.0, server_id=3, multi_ip_mode=False, client_connect_ip=127.0.0.1)>()
E        +    where <bound method RedisServerHandle.is_alive of RedisServerHandle(port=7617, bind_ip=0.0.0.0, server_id=3, multi_ip_mode=False, client_connect_ip=127.0.0.1)> = RedisServerHandle(port=7617, bind_ip=0.0.0.0, server_id=3, multi_ip_mode=False, client_connect_ip=127.0.0.1).is_alive

test_rdb_load_error_cleanup.py:209: AssertionError
```

**[Local Stack Trace](https://paste.amazon.com/show/branylin/1779404208): (identical to [QA crash stack trace](https://sim.amazon.com/issues/ELMO-119562))**
```
25348 write-worker-12 *
/lib64/libpthread.so.0(+0x118e0)[0x7f75b6a6a8e0]
/lib64/libc.so.6(gsignal+0x110)[0x7f75b5d44ca0]
/lib64/libc.so.6(abort+0x148)[0x7f75b5d46148]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(+0xced352)[0x7f75ac9ef352]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(+0xced8b5)[0x7f75ac9ef8b5]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(+0xcede29)[0x7f75ac9efe29]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(+0x1a0926)[0x7f75abea2926]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(+0x1a0949)[0x7f75abea2949]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(_ZN13valkey_search11IndexSchema17MarkAsDestructingEv+0x39)[0x7f75ac73a0a3]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(_ZN13valkey_search11IndexSchemaD1Ev+0x7f8)[0x7f75ac73ad80]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(_ZNSt15_Sp_counted_ptrIPN13valkey_search11IndexSchemaELN9__gnu_cxx12_Lock_policyE2EE10_M_disposeEv+0x25)[0x7f75ac7561b5]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(+0x2d3a1a)[0x7f75abfd5a1a]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(_ZN5vmsdk10ThreadPool12WorkerThreadESt10shared_ptrINS0_6ThreadEE+0x669)[0x7f75ac090779]
/home/branylin/workplace/ECRedis/src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so(+0x390663)[0x7f75ac092663]
/lib64/libpthread.so.0(+0x744b)[0x7f75b6a6044b]
/lib64/libc.so.6(clone+0x3f)[0x7f75b5e0052f]

------------

/.../libsearch.so(+0x1a0926)[0x7f75abea2926]   ← this is RemoveSubscription
/.../libsearch.so(+0x1a0949)[0x7f75abea2949]   ← inside RemoveSubscription
```

```
addr2line -f -C -e src/ElastiCacheValkeySearch/build/private/.build-release/libsearch.so 0x1a0926 0x1a0949
valkey_search::KeyspaceEventManager::RemoveSubscription(valkey_search::KeyspaceEventSubscription*) [clone .cold]
keyspace_event_manager.cc:?
valkey_search::KeyspaceEventManager::RemoveSubscription(valkey_search::KeyspaceEventSubscription*) [clone .cold]
keyspace_event_manager.cc:?
```

### DIFF
--- a/integration/test_rdb_load_error_cleanup.py
+++ b/integration/test_rdb_load_error_cleanup.py
@@ -128,7 +128,7 @@ class TestRdbLoadErrorCleanup(ValkeySearchTestCaseDebugMode):
                 n = replica.client.execute_command(
                     "FT._DEBUG", "PAUSEPOINT", "TEST", self.PAUSEPOINT_NAME)
                 return int(n) >= 1
-            except Exception:
+            except (TypeError, ValueError):
                 return False
 
         try:
@@ -136,6 +136,17 @@ class TestRdbLoadErrorCleanup(ValkeySearchTestCaseDebugMode):
                 lambda: error_logged() and workers_paused(),
                 timeout=60)
         except Exception:
+            # If the replica died during the waiter loop, that IS the crash
+            # we're testing for. Detect it here so we fail loudly instead of
+            # falling through to pytest.skip below.
+            try:
+                replica.client.ping()
+            except Exception as exc:
+                pytest.fail(
+                    "Replica died before the bug window could be verified. "
+                    "This is the crash the test is designed to catch, but it "
+                    f"happened during setup instead of at the release point: "
+                    f"{exc!r}")
             err = error_logged()
             paused = workers_paused()
             load_ran = replica.does_logfile_contains("Loading Index Extension")
--- a/integration/test_rdb_load_error_cleanup.py
+++ b/integration/test_rdb_load_error_cleanup.py
@@ -1,0 +1,203 @@
+"""
+Regression test for a replica crash in
+KeyspaceEventManager::RemoveSubscription when LoadFromRDB fails mid-load.
+
+Without the fix, workers parked in ScheduleMutation can hold the last strong
+ref to the staged IndexSchema. When the load fails and the main thread drops
+its ref, the destructor runs on a worker thread, calls MarkAsDestructing,
+hits the main-thread CHECK in subscriptions_.Get(), and aborts.
+
+The fix (absl::Cleanup in LoadFromRDB) ensures MarkAsDestructing runs on the
+main thread on every error path, so the worker-thread destructor sees
+is_destructing_ == true and skips the main-thread-only cleanup.
+
+The test reproduces the bug window deterministically by:
+  - parking workers at the mutation_processing pause point (they hold strong
+    refs while parked),
+  - tripping the ForceRDBLoadFailure controlled variable mid-load,
+  - using diskless replica load (the only path where the engine survives an
+    aux_load failure - on disk-based load the engine exits before we can
+    observe the worker-thread crash).
+"""
+import struct
+import time
+
+import pytest
+
+from valkey_search_test_case import ValkeySearchTestCaseDebugMode
+from valkeytestframework.util import waiters
+from valkeytestframework.conftest import resource_port_tracker  # noqa: F401
+from ft_info_parser import FTInfoParser
+
+
+class TestRdbLoadErrorCleanup(ValkeySearchTestCaseDebugMode):
+    """Regression test for the worker-thread-destruction crash."""
+
+    INDEX_NAME = "idx"
+    PAUSEPOINT_NAME = "mutation_processing"
+    CONTROLLED_VAR = "ForceRDBLoadFailure"
+    NUM_VECTORS = 50
+    SIM_ERROR_LOG = "Simulated IO error during RDB load"
+
+    def get_config_file_lines(self, testdir, port):
+        # Diskless sync on both ends - required for the engine to survive
+        # an aux_load failure (cancelReplicationHandshake instead of exit(1)).
+        # `flush-before-load` (not `on-empty-db` or `swapdb`) is required so
+        # that the second sync also uses diskless: `on-empty-db` falls back
+        # to disk once the replica is populated, and `swapdb` requires every
+        # loaded module to opt into VALKEYMODULE_OPTIONS_HANDLE_REPL_ASYNC_LOAD
+        # (the JSON module doesn't).
+        return super().get_config_file_lines(testdir, port) + [
+            "repl-diskless-sync yes",
+            "repl-diskless-sync-delay 0",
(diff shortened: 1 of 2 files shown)


integration/test_rdb_load_error_cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+            primary.client.hset(f"vec:{i}", mapping={"vector": v})
+
+    def _trigger_full_sync(self, primary, replica):
+        """Force a full sync by changing the primary's replication id.
+
+        change-repl-id forces the next handshake to be a full sync rather
+        than a partial resync.
+        """
+        primary.client.execute_command("DEBUG", "change-repl-id")
+        replica.client.execute_command("REPLICAOF", "NO", "ONE")
```
KarthikSubbarao, line 85: Nice trick

integration/test_rdb_load_error_cleanup.py
```diff
@@ -0,0 +1,203 @@
...
+
+import pytest
+
+from valkey_search_test_case import ValkeySearchTestCaseDebugMode
+from valkeytestframework.util import waiters
+from valkeytestframework.conftest import resource_port_tracker  # noqa: F401
+from ft_info_parser import FTInfoParser
+
+
+class TestRdbLoadErrorCleanup(ValkeySearchTestCaseDebugMode):
```
KarthikSubbarao, line 33: How long does this test take to run?

linbran5123: Every time I've run this test it takes around 20 seconds to run and pass. The biggest time consumer is the 10 second loop that alive check retry window. Everything else shouldn't take much time

Aksha1812:
can we merge this ? Any blockers ?

allenss-amazon (approved):

yairgott:
I've addressed the same issue in the memory sharing [PR](https://github.com/valkey-io/valkey-search/pull/1213) by enforcing that the index schema destructor can only be issued on the main thread. See: https://github.com/valkey-io/valkey-search/blob/adbc03a90b4b919d2668d8f85d57b0311b034ac4/src/index_schema.cc#L255

PR #1191, opened 2026-06-27
Validate HNSW index data when loading from an external file

allenss-amazon:
Loading an HNSW index from RDB previously trusted the serialized data completely; a corrupt or malicious snapshot could drive out-of-bounds reads/writes and crash the server during load or on the next search.

Add defensive validation to HierarchicalNSW::LoadIndex, gated by a single choke-point (loadCheck) that throws on failure (converted to a rejected RDB load by VectorHNSW::LoadFromRDB):

- Header: cross-check M/maxM/maxM0/serialize-size/offsets against the authoritative index definition (expected_m, the space's vector size), bound enterpoint/max_level, guard every allocation against size_t overflow, and fix the capacity resolution so max_elements is always
  >= curr_element_count.
- Per record: level-0 chunk size, neighbor count <= 2M, neighbor ids < curr_element_count; upper-level size-chunk/multiple/chunk-size, level <= max_level, per-level count <= M, and duplicate-label detection.
- Global pass: enterpoint is at max_level, and every upper-level link targets a slot that exists at that level (closing the descent over-read that an id-range check alone cannot).

Validation scans are clamped to safe bounds so disabling the kill switch never introduces a new load-time OOB. Add the hnsw-validation-enable (Dev, default true) config as the kill switch.

Also fix a latent double-free: if the upper-level load loop returned/threw after setting element_levels_[i] nonzero but before storing the link-list pointer, clear() would free uninitialized ChunkedArray memory. Now the pointer is nulled first and the level published only after a valid alloc.

Tests (testing/vector_test.cc): a combined in-memory ChunkStream double, deterministic forced-level golden builder, empty/single/multi-layer happy-path round-trips (including save->load->save byte identity), and one corruption test per validation rule plus a kill-switch bypass test.

### DIFF
--- a/third_party/hnswlib/hnswalg.h
+++ b/third_party/hnswlib/hnswalg.h
@@ -25,6 +25,7 @@
 #include "src/metrics.h"
 #include "third_party/hnswlib/index.pb.h"
 #include "visited_list_pool.h"
+#include "vmsdk/src/log.h"
 #include "vmsdk/src/status/status_macros.h"
 
 #ifdef VMSDK_ENABLE_MEMORY_ALLOCATION_OVERRIDES
@@ -867,10 +868,18 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
   // disabled. On failure it throws; VectorHNSW::LoadFromRDB converts the
   // exception into a failed (rejected) RDB load.
   inline void loadCheck(bool ok, absl::string_view msg) const {
-    if (load_validation_enabled_ && !ok) {
+    if (ok) {
+      return;
+    }
+    if (load_validation_enabled_) {
       throw std::runtime_error(
           absl::StrCat("HNSW index load validation failed: ", msg));
     }
+    // Validation disabled: don't fail the load, but log (throttled) the defect.
+    VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+        << "HNSW load validation is disabled; ignoring detected index "
+           "corruption: "
+        << msg;
   }
 
   absl::Status LoadIndex(InputStream &input, SpaceInterface<dist_t> *s,
@@ -939,11 +948,14 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
                     size_links_level0_ + vector_size_ + sizeof(labeltype),
                 "serialized element size is inconsistent with the geometry");
       loadCheck(offsetLevel0_ == 0, "offset_level_0 must be 0");
-      // mult_ = 1/log(M_) is always in (0, ~1.45] for M_ >= 2. A range check
-      // (rather than std::isfinite, which is unreliable under finite-math
-      // build flags) rejects inf / out-of-range garbage. mult_ only affects
-      // future-insert level assignment, so this is a soft, non-safety check.
-      loadCheck(mult_ > 0.0 && mult_ <= 2.0, "mult is out of range");
+      // Recompute mult=1/log(M) and match within a guardband (cross-arch libm);
+      // skip M<2 where 1/log(M) is not finite. Soft, non-safety check.
+      if (M_ >= 2) {
+        const double expected_mult = 1.0 / std::log(static_cast<double>(M_));
+        loadCheck(mult_ > 0.0 &&
+                      std::fabs(mult_ - expected_mult) <= 1e-6 * expected_mult,
+                  "mult is inconsistent with M");
+      }
 
       // Element-count / level / entry-point consistency.
       loadCheck(cur_element_count_ <= max_elements_,
--- a/third_party/hnswlib/hnswalg.h
+++ b/third_party/hnswlib/hnswalg.h
@@ -4,6 +4,7 @@
 #include <stdlib.h>
 
 #include <atomic>
+#include <cmath>
 #include <cstdint>
 #include <cstring>
 #include <deque>
@@ -97,6 +98,10 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
 
   bool allow_replace_deleted_ = false;  // flag to replace deleted elements
                                         // (marked as deleted) during insertions
+
+  // Gate for load-time corruption validation, wired from the
+  // hnsw-validation-enable config via LoadIndex. Consulted only in loadCheck().
+  bool load_validation_enabled_ = false;
 
   std::mutex deleted_elements_lock;  // lock for deleted_elements
   std::unordered_set<tableint>
@@ -812,9 +817,23 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
     return absl::OkStatus();
   }
 
+  // Single choke-point for load-time validation. The kill switch
+  // (load_validation_enabled_, wired from the hnsw-validation-enable config) is
+  // consulted here and nowhere else, so a defective check can be globally
+  // disabled. On failure it throws; VectorHNSW::LoadFromRDB converts the
+  // exception into a failed (rejected) RDB load.
+  inline void loadCheck(bool ok, absl::string_view msg) const {
+    if (load_validation_enabled_ && !ok) {
+      throw std::runtime_error(
+          absl::StrCat("HNSW index load validation failed: ", msg));
+    }
+  }
+
   absl::Status LoadIndex(InputStream &input, SpaceInterface<dist_t> *s,
-                         size_t max_elements_i, VectorTracker *vector_tracker) {
+                         size_t max_elements_i, VectorTracker *vector_tracker,
+                         size_t expected_m, bool validate) {
     clear();
+    load_validation_enabled_ = validate;
 
     VMSDK_ASSIGN_OR_RETURN(auto serialized_header, input.LoadChunk());
     auto header = std::make_unique<data_model::HNSWIndexHeader>();
@@ -822,6 +841,7 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
       return absl::InternalError("Could not deserialize HNSW header");
     }
 
+    // --- Read raw header fields (untrusted) ---
     offsetLevel0_ = header->offset_level_0();
     max_elements_ = header->max_elements();
     cur_element_count_ = header->curr_element_count();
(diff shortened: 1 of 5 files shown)


Aksha1812:
Is this needed in 1.2.1 ? I think we can skip validation for 1.2 since it was already not there and keep the use after free fix ? @allenss-amazon

third_party/hnswlib/hnswalg.h
```diff
@@ -836,29 +856,84 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+    // Resolve capacity: at least the live element count, honoring the larger of
+    // the caller-requested cap and the file's recorded capacity.
+    size_t cur_count = cur_element_count_;
+    size_t max_elements =
+        std::max(cur_count, std::max(max_elements_i, max_elements_));
+    max_elements_ = max_elements;
+
+    // --- Header validation (the kill switch lives inside loadCheck) ---
+    {
+      const size_t exp_m = expected_m > 10000 ? 10000 : expected_m;
```
BCathcart, line 938: What's the purpose of this? Won't it just lead to an error here: `loadCheck(M_ == exp_m, "header M does not match index definition");` which will be a bit confusing because the header could match the index definition and we changed its value silently.

third_party/hnswlib/hnswalg.h
```diff
@@ -812,16 +817,31 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+  // Single choke-point for load-time validation. The kill switch
+  // (load_validation_enabled_, wired from the hnsw-validation-enable config) is
+  // consulted here and nowhere else, so a defective check can be globally
+  // disabled. On failure it throws; VectorHNSW::LoadFromRDB converts the
+  // exception into a failed (rejected) RDB load.
+  inline void loadCheck(bool ok, absl::string_view msg) const {
+    if (load_validation_enabled_ && !ok) {
+      throw std::runtime_error(
+          absl::StrCat("HNSW index load validation failed: ", msg));
+    }
```
BCathcart, line 877: Can we log the message to the engine logs regardless if `load_validation_enabled_` is enabled or not

third_party/hnswlib/hnswalg.h
```diff
@@ -836,29 +856,84 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+    // the caller-requested cap and the file's recorded capacity.
+    size_t cur_count = cur_element_count_;
+    size_t max_elements =
+        std::max(cur_count, std::max(max_elements_i, max_elements_));
+    max_elements_ = max_elements;
+
+    // --- Header validation (the kill switch lives inside loadCheck) ---
+    {
+      const size_t exp_m = expected_m > 10000 ? 10000 : expected_m;
+      // Layout parameters must match the index definition exactly.
```
BCathcart, line 939: What I would suggest here is to have a common function with the constructor to create all the constants based on the index definition. Then compare them to what's actually saved at the hnswlib level as a sanity check. Rather than taking the values from the hnswlib level and then computing the "real" values all over the place using the same equations in the constructor and cross referencing them.

third_party/hnswlib/hnswalg.h
```diff
@@ -836,29 +856,84 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+      loadCheck(vector_size_ > 0, "vector size must be > 0");
+      loadCheck(serialize_size_data_per_element_ ==
+                    size_links_level0_ + vector_size_ + sizeof(labeltype),
+                "serialized element size is inconsistent with the geometry");
+      loadCheck(offsetLevel0_ == 0, "offset_level_0 must be 0");
+      // mult_ = 1/log(M_) is always in (0, ~1.45] for M_ >= 2. A range check
+      // (rather than std::isfinite, which is unreliable under finite-math
+      // build flags) rejects inf / out-of-range garbage. mult_ only affects
+      // future-insert level assignment, so this is a soft, non-safety check.
+      loadCheck(mult_ > 0.0 && mult_ <= 2.0, "mult is out of range");
```
BCathcart: We're validating the range based on the equation `1/log(M_)` so we're already enforcing an expectation of how its calculated on the save side. Should we just recompute `1/log(M_)` and then sanity check it matches the `mult_` saved in the RDB similar to what we're doing for other variables?

third_party/hnswlib/hnswalg.h
```diff
@@ -836,29 +856,84 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+        maxM_ * sizeof(tableint) + sizeof(linklistsizeint);
 
-    fstdistfunc_ = s->get_dist_func();
-    dist_func_param_ = s->get_dist_func_param();
+    // Resolve capacity: at least the live element count, honoring the larger of
+    // the caller-requested cap and the file's recorded capacity.
+    size_t cur_count = cur_element_count_;
+    size_t max_elements =
+        std::max(cur_count, std::max(max_elements_i, max_elements_));
+    max_elements_ = max_elements;
```
BCathcart, line 934: nit: Can we just set `max_elements_` directly?

third_party/hnswlib/hnswalg.h
murphyjacob4, line 938: Why 10,000 and not 2,000,000? https://github.com/valkey-io/valkey-search/blob/ae132bea8db015c551ea6743e2e37ee34993c430/src/commands/ft_create_parser.cc#L71

allenss-amazon: There is pre-existing code elsewhere with the same fixed constant upper bound on M.

third_party/hnswlib/hnswalg.h
```diff
@@ -836,29 +856,84 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+      // Element-count / level / entry-point consistency.
+      loadCheck(cur_element_count_ <= max_elements_,
+                "curr_element_count exceeds max_elements");
+      if (cur_element_count_ == 0) {
+        loadCheck(maxlevel_ == -1 || maxlevel_ == 0,
+                  "empty index has a non-trivial max_level");
+      } else {
+        loadCheck(maxlevel_ >= 0, "non-empty index has a negative max_level");
+        loadCheck(maxlevel_ <= static_cast<int>(cur_element_count_),
+                  "max_level exceeds the element count");
```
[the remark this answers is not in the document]
allenss-amazon: Valid values of max-level will never exceed the range of like [1..100]

third_party/hnswlib/hnswalg.h
```diff
@@ -880,28 +900,84 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
 
     data_level0_memory_ = std::make_unique<ChunkedArray>(
         size_data_per_element_, k_elements_per_chunk, max_elements);
 
     for (size_t i = 0; i < cur_element_count_; i++) {
       VMSDK_ASSIGN_OR_RETURN(auto chunk, input.LoadChunk());
+      loadCheck(chunk->size() ==
+                    size_links_level0_ + vector_size_ + sizeof(labeltype),
+                "level-0 element chunk has the wrong size");
       memcpy((*data_level0_memory_)[i], chunk->data(), size_links_level0_);
```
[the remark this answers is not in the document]
allenss-amazon: With hnsw-validation-enable disabled, we are vulnerable to many different failures -- fixing this one doesn't change the vulnerability profile.. The disable is only provided as an emergency escape mechanism if somehow the validation itself had a bug in it.

third_party/hnswlib/hnswalg.h
allenss-amazon: expected_m comes from the index definition. The 10000 constant is a leftover from elsewhere in the HNSW lib and I didn't want to make any actual behavioral changes. This PR will either accept/reject the RDB. If it's accepted, then it will be 100% bit and behavior compatible to what we have today.

This 10000 limit might not make sense, but IMO this isn't the PR to question that behavior.

third_party/hnswlib/hnswalg.h
allenss-amazon: Good idea!

third_party/hnswlib/hnswalg.h
allenss-amazon: I looked at doing exactly that, it's a more intrusive change than I expected -- so I backed off from that.

third_party/hnswlib/hnswalg.h
allenss-amazon: I'll check into that.

third_party/hnswlib/hnswalg.h
allenss-amazon: again, I didn't want to change the existing flow.

third_party/hnswlib/hnswalg.h
```diff
@@ -928,21 +1027,85 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+                    "upper-level neighbor count exceeds M");
+        }
+      }
+    }
+
+    // --- Global graph-invariant pass (requires all element_levels_ loaded) ---
+    // The entry point must be one of the tallest nodes (proven invariant:
+    // maxlevel_ is only ever raised together with enterpoint_node_). The
+    // short-circuit also guards the array access if validation is disabled and
+    // enterpoint_node_ is out of range.
```
BCathcart, line 1081: >The
    // short-circuit also guards the array access if validation is disabled and
    // enterpoint_node_ is out of range.

We're guarding against OOB access here but we're really just pushing the problem down stream. A search starting at an invalid entry point will at best crash. Do we need two tiers of validation disablement? One where we overlook inconsistencies or strange values that will produce an unusual but still fundamentally valid index, and another where disabling it exposes the index to downstream memory errors? The latter probably shouldn't even be an option to disable. It feels like a dangerous switch to have. An operator might think they're bypassing a minor issue up top but once they turn off the validation it will go through regardless if there are any crucial problems like an OOB entry point.

third_party/hnswlib/hnswalg.h
BCathcart: It's not urgent but IMO we've modified enough code here to not worry about preserving existing code too much. We're using both `max_elements` and `max_elements_` below interchangeably and I think there are enough variables flying around here as is :)

third_party/hnswlib/hnswalg.h
BCathcart: Are you saying when a index is defined with M above 10,000, we'll maintain the original value in the index definition but successfully create the index with M adjusted down to 10,000? We're just using this for validation so I was more commenting on if `expected_m` and `M_` are both the same value above 10,000. We're decreasing `expected_m` and then returning an error saying it doesn't match `M_` even if it does in the RDB.

third_party/hnswlib/hnswalg.h
allenss-amazon: I left the old -- OOB susceptible -- code in place as this is going into a patch release and I wanted to be 100% we don't create an unrecoverable problem. With luck we will never need to disable validation.

third_party/hnswlib/hnswalg.h
allenss-amazon: Understood. Realistically, I'd love to refactor large sections of the HNSW code -- it's awful.

BCathcart (approved):

PR #1192, opened 2026-06-27
Validate query vector size against index dimensions at parse time

allenss-amazon:
The query reference vector supplied to FT.SEARCH / FT.AGGREGATE was only length-checked inside the index Search() methods. The pre-filtering path (DoSearchVector -> CalcBestMatchingPrefilteredKeys -> AddPrefilteredKey -> ComputeDistanceFromRecord) bypassed that check and cast the blob directly to the element type, reading dimensions_ elements from it -- an out-of-bounds read when an undersized blob is supplied via PARAMS.

Validate the substituted query vector once in PostParseVectorParameters, the common parse choke point for both commands, so every downstream search path receives a correctly sized blob. Remove the now-redundant IsValidSizeVector checks from VectorFlat::Search and VectorHNSW::Search (IsValidSizeVector is retained for the insert-time guard in InternVector). Add parser regression tests for too-small and too-large blobs.

### DIFF
--- a/testing/ft_search_parser_test.cc
+++ b/testing/ft_search_parser_test.cc
@@ -70,6 +70,7 @@ struct FTSearchParserTestCase {
   std::string search_parameters_str;
   uint64_t timeout_ms{query::kTimeoutMS};
   bool vector_query{true};
+  std::optional<size_t> query_blob_num_floats;
   bool verbatim{false};
   bool inorder{false};
   std::optional<unsigned> slop;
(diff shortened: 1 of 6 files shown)


testing/ft_search_parser_test.cc
```diff
@@ -529,6 +533,26 @@ INSTANTIATE_TEST_SUITE_P(
...
             .k = 10,
         },
+        {
+            .test_name = "vector_blob_size_too_small",
+            .success = false,
+            .params_str = " PARAMS 2",
+            .filter_str = " * => [KNN 10 @vec $BLOB]",
+            .expected_error_message =
+                "Error parsing vector similarity parameters: query vector "
+                "blob size",
```
BCathcart, line 543: I'd generally put the whole error message because why not

BCathcart (approved):

Aksha1812:
Do we do this check in the grpc path before fanning out search request  when cluster mode enabled  ?

allenss-amazon:
I think a check in that path is redundant, it would have already gone through this code.

PR #1196, opened 2026-06-29
Allow indexing the same field as both TEXT and TAG (#1195)

### LINKED ISSUE #1195: [BUG] Compat: Cannot index same field as both TEXT and TAG
RS allows this syntax:

```
127.0.0.1:26379> FT.CREATE idx ON HASH PREFIX 1 blog:post: SCHEMA sku AS sku_text TEXT sku AS sku_tag TAG SORTABLE
OK
```

But on Valkey Search, we get

```
127.0.0.1:16379> FT.CREATE idx ON HASH PREFIX 1 blog:post: SCHEMA sku AS sku_text TEXT sku AS sku_tag TAG SORTABLE
(error) Duplicate field in schema - sku
```

allenss-amazon:
## Summary

Fixes #1195. `FT.CREATE` rejected schemas that index the same source field under two distinct aliases:

```
FT.CREATE idx ON HASH PREFIX 1 blog:post: SCHEMA sku AS sku_text TEXT sku AS sku_tag TAG SORTABLE
(error) Duplicate field in schema - sku
```

RediSearch accepts this. The duplicate-field check in `ParseFTCreateArgs` keyed uniqueness on the source **identifier** (`sku`) rather than the attribute **alias** (the `AS` name). Since the engine already stores attributes in a map keyed by alias (`IndexSchema::AddIndex` / `attributes_`) and processes each attribute independently in `ProcessKeyspaceNotification` (reading by identifier, writing by alias), a single source field can legitimately be indexed multiple times under different aliases.

## Change

- Key the uniqueness check on `attribute->alias()` instead of `attribute->identifier()` in `src/commands/ft_create_parser.cc`. Genuine duplicate **aliases** are still rejected (the no-`AS` case is unchanged, since the alias defaults to the identifier).

## Testing

- New parser unit test `same_identifier_distinct_aliases_text_and_tag` (fails before the fix with `Duplicate field in schema - sku`).
- New integration tests in `test_ft_create.py`:
  - `test_same_field_as_text_and_tag` — runs the issue's exact `FT.CREATE`, then confirms both `@sku_tag:{...}` and `@sku_text:...` queries return the document.
  - `test_duplicate_alias_still_rejected` — confirms two attributes sharing an alias are still rejected.
- Full unit suite passes; full integration suite passes (238 passed, 8 skipped).


allenss-amazon:
Fix for #1195

Aksha1812:
even though this is bug fix . this looks like a functionality change. Should we remove this from 1.2 project board ? @allenss-amazon

allenss-amazon:
As I read the description, it's changing a rejected operation into a working operation. That's not really a "breaking change", so no reason it can't be backported. IDK if it makes the 1.2.1 train or not.

Aksha1812:
I believe we have a   `identifier_to_alias_`   https://github.com/valkey-io/valkey-search/blob/5a63e8269b8fd3b0fb98c47a150a7bab5f45e299/src/index_schema.cc#L496-L497 . what happens to that when we insert same identifier with different alias ? don't we lose the second entry when we populate the first one in this datra structure when we try to insert . Now that we allow duplicate identifiers , should we update this data structure  to handle that

integration/test_ft_create.py
```diff
@@ -84,3 +84,51 @@ def test_ft_create_fails_on_replica_cmd(self):
             self.get_primary_connection(),
             self.get_replica_connection(0),
         )
+
+    def test_same_field_as_text_and_tag(self):
```
Aksha1812, line 88: based on description redis search accepts this . do we need a compatibility test as well , so we don't break this again. ?

src/commands/ft_create_parser.cc
```diff
@@ -743,13 +747,12 @@ absl::StatusOr<data_model::IndexSchema> ParseFTCreateArgs(
                            schema_text_defaults),
         _.SetPrepend() << "Invalid field type for field `"
                        << attribute_identifier << "`: ");
-    if (identifier_names.find(attribute->identifier()) !=
-        identifier_names.end()) {
-      return absl::InvalidArgumentError(absl::StrCat(
-          "Duplicate field in schema - ", attribute->identifier()));
+    if (attribute_aliases.find(attribute->alias()) != attribute_aliases.end()) {
+      return absl::InvalidArgumentError(
+          absl::StrCat("Duplicate field in schema - ", attribute->alias()));
```
Aksha1812, line 752: is duplication case sensitive . how does redis handle it

testing/ft_create_parser_test.cc
```diff
@@ -674,6 +674,39 @@ INSTANTIATE_TEST_SUITE_P(
                               .indexer_type = indexes::IndexerType::kTag,
                           }}},
          },
+         {
```
Aksha1812, line 677: what happens when two tag attributes or two text attribute have duplicate aliases/ identifiers , should we include that in testing

Aksha1812 (changes requested):

src/commands/ft_create_parser.cc
allenss-amazon: We definitely haven't been treating field names as case-insensitive. If that's the case, it's a much bigger fix and definitely a potentially breaking-change. But it's a good question and we should answer it definitely.

BCathcart: Redis search is case-sensitive on the alias name:

```
127.0.0.1:6379> FT.CREATE idx ON HASH PREFIX 1 blog:post: SCHEMA sku AS sku_text TEXT sku AS sku_text TAG SORTABLE
(error) Duplicate field in schema - sku_text
127.0.0.1:6379> FT.CREATE idx ON HASH PREFIX 1 blog:post: SCHEMA sku AS sku_text TEXT sku AS sku_texT TAG SORTABLE
OK
```

integration/test_ft_create.py
allenss-amazon: We're adding a test here. The fact that it's a non-compatibility test doesn't matter.

testing/ft_create_parser_test.cc
allenss-amazon: That's fixed in another PR coming soon.

Aksha1812:
**Reviewers for this PR**

zackcam (commented):
Does this need to have changes in `FT.AGGREGATE` as well. From what I could see aggregate still assume identifier uniqueness. I see  `assert(identifier_itr == record_indexes_by_identifier_.end())` at line 119 of ft_aggregate_parser.h which would crash for debug builds and asan. Non debug builds would have the identifier map to first registered slot.

allenss-amazon:
@zackcam you are correct about Aggregation the fix is separate in #1084 which is ready for review.

BCathcart (approved):

PR #1202, opened 2026-06-30
Async Index Schema Destruction

KarthikSubbarao:
### Problem

FLUSHALL and FT.DROPINDEX commands results in index destruction today on the main thread which blocks it for a long duration when large indexes are destroyed because IndexSchema destruction (radix trees, posting lists, per-key indexes) runs synchronously in the event loop callback. Duration scales with index size — larger indexes block longer.

### Fix  - Async Index Schema Destruction

Move `shared_ptr<IndexSchema>` destruction to the utility thread pool via `ScheduleUtilityTask` in three places:
1. `RemoveIndexSchema` (FT.DROPINDEX path)
2. `OnMetadataCallback` (coordinated mode)
3. `OnFlushDBEnded` (FLUSHALL/FLUSHDB path)

`MarkAsDestructing()` is called on the main thread before scheduling, ensuring no new operations target the schema.

### Benchmark Setup

- Target: ~10GB index
- Schema: single TEXT field
- Measure: wall-clock time of `FLUSHALL` command from client perspective

### Results

| Version | FLUSHALL Duration | Memory (13GB index, 3.16M keys) |
|:--------|:-:|:------|
| Baseline (bc4bace) | **55,929 ms** (~56s) | Freed immediately (blocks main thread) |
| Fix (eb13f92) | **2 ms** | Freed in background over ~30s |

Background memory reclaim timeline (fix version):
- +2s: 8.1 GB
- +5s: 7.0 GB
- +10s: 4.8 GB
- +20s: 802 MB
- +30s: 41 MB (fully reclaimed)

**Improvement: 28,000x faster client response (56s → 2ms).**

The main thread is never blocked. Memory is reclaimed progressively on the utility thread pool.

### DIFF
--- a/src/schema_manager.cc
+++ b/src/schema_manager.cc
@@ -367,38 +367,33 @@ absl::Status SchemaManager::OnMetadataCallback(
   if (!old_schema.ok() && !absl::IsNotFound(old_schema.status())) {
     return old_schema.status();
   }
-
+  absl::Status result = absl::OkStatus();
   if (metadata == nullptr) {
-    if (old_schema.ok()) {
-      ValkeySearch::Instance().ScheduleIndexDestruction(
-          std::move(old_schema.value()));
+    // Nothing to create — just clean up the old schema below.
+  } else {
+    auto proposed_schema = std::make_unique<data_model::IndexSchema>();
+    if (!metadata->UnpackTo(proposed_schema.get())) {
+      result = absl::InternalError(absl::StrCat(
+          "Unable to unpack metadata for index schema ", obj_name));
+    } else {
+      auto create_status =
+          CreateIndexSchemaInternal(detached_ctx_.get(), *proposed_schema);
+      if (!create_status.ok()) {
+        result = create_status;
+      } else {
+        auto created_schema =
+            LookupInternal(obj_name.GetDbNum(), obj_name.GetName()).value();
+        CHECK(created_schema != nullptr);
+        created_schema->SetFingerprint(fingerprint);
+        created_schema->SetVersion(version);
+      }
     }
-    return absl::OkStatus();
   }
-  auto proposed_schema = std::make_unique<data_model::IndexSchema>();
-  if (!metadata->UnpackTo(proposed_schema.get())) {
-    if (old_schema.ok()) {
-      ValkeySearch::Instance().ScheduleIndexDestruction(
-          std::move(old_schema.value()));
-    }
-    return absl::InternalError(
-        absl::StrCat("Unable to unpack metadata for index schema ", obj_name));
-  }
-
-  VMSDK_RETURN_IF_ERROR(
-      CreateIndexSchemaInternal(detached_ctx_.get(), *proposed_schema));
-
-  auto created_schema =
-      LookupInternal(obj_name.GetDbNum(), obj_name.GetName()).value();
-  CHECK(created_schema != nullptr);
-  created_schema->SetFingerprint(fingerprint);
-  created_schema->SetVersion(version);
-
   if (old_schema.ok()) {
     ValkeySearch::Instance().ScheduleIndexDestruction(
         std::move(old_schema.value()));
   }
-  return absl::OkStatus();
+  return result;
 }
 
 uint64_t SchemaManager::GetNumberOfIndexSchemas() const {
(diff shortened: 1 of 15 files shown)


src/valkey_search_options.cc
```diff
@@ -223,6 +223,13 @@ constexpr absl::string_view kSearchResultBackgroundCleanup{
 static config::Boolean search_result_background_cleanup(
     kSearchResultBackgroundCleanup, false);
 
+/// Enable async index destruction
+/// If set to true (default), index schema destruction is deferred to a
+/// background thread, avoiding main-thread blocking during FLUSHALL or
+/// FT.DROPINDEX on large indexes.
+constexpr absl::string_view kAsyncIndexDestruction{"async-index-destruction"};
+static config::Boolean async_index_destruction(kAsyncIndexDestruction, true);
```
KarthikSubbarao: I am considering removing this config

I had added it earlier for my own testing of flushall on mainthread vs background. But after testing is done, I plan to delete this config, FYA.

src/schema_manager.cc
```diff
@@ -300,7 +300,13 @@ absl::Status SchemaManager::RemoveIndexSchema(uint32_t db_num,
...
   // In non-coordinated mode, apply the update inline.
   absl::MutexLock lock(&db_to_index_schemas_mutex_);
-  return RemoveIndexSchemaInternal(db_num, name).status();
+  auto old_schema = RemoveIndexSchemaInternal(db_num, name);
+  if (!old_schema.ok()) {
+    return old_schema.status();
+  }
+  ValkeySearch::Instance().ScheduleUtilityTask(
+      [s = std::move(old_schema.value())]() mutable { s.reset(); });
+  return absl::OkStatus();
```
Aksha1812, line 309: should we add a `CHECK(schema->IsMarkedDestructing())`  to be safe after `RemoveIndexSchemaInternal ` or inside it.

src/schema_manager.cc
```diff
@@ -300,7 +300,13 @@ absl::Status SchemaManager::RemoveIndexSchema(uint32_t db_num,
 
   // In non-coordinated mode, apply the update inline.
   absl::MutexLock lock(&db_to_index_schemas_mutex_);
-  return RemoveIndexSchemaInternal(db_num, name).status();
+  auto old_schema = RemoveIndexSchemaInternal(db_num, name);
+  if (!old_schema.ok()) {
+    return old_schema.status();
+  }
+  ValkeySearch::Instance().ScheduleUtilityTask(
```
Aksha1812, line 307: during the destruction 
```
IndexSchema::~IndexSchema() {
  VMSDK_LOG_EVERY_N_SEC(NOTICE, detached_ctx_.get(), 1)
      << "Index schema " << vmsdk::config::RedactIfNeeded(name_)
      << " dropped from DB " << db_num_;
``` 
does this work the same way when it is run from the background thread instead of main thread and still get the `detached_ctx_ `

Aksha1812:
Will you add a test that inits a real utility pool, defers a destruction, joins the pool, and runs under TSan/ASan to actually cover the new path and prove no off-thread crash?

KarthikSubbarao:
> Will you add a test that inits a real utility pool, defers a destruction, joins the pool, and runs under TSan/ASan to actually cover the new path and prove no off-thread crash?

Utility pool is always initiated and set to 1, now we are just using it. I was going to add some integration tests, but saw we already have a good number of tests related to flushall that already cover the functionality we need to validate

src/schema_manager.cc
KarthikSubbarao: Yes, `detached_ctx_` is intended to be used from background threads for most part. If we are on the main thread, we should have access to the context itself

src/schema_manager.cc
KarthikSubbarao: MarkAsDestructing is called within RemoveIndexSchemaInternal itself, so it did not make sense to add an assert right there

src/schema_manager.cc
```diff
@@ -356,31 +362,38 @@ absl::Status SchemaManager::OnMetadataCallback(
...
+    }
   }
-  auto proposed_schema = std::make_unique<data_model::IndexSchema>();
-  if (!metadata->UnpackTo(proposed_schema.get())) {
-    return absl::InternalError(
-        absl::StrCat("Unable to unpack metadata for index schema ", obj_name));
+  if (old_schema.ok()) {
+    ValkeySearch::Instance().ScheduleIndexDestruction(
+        std::move(old_schema.value()));
   }
```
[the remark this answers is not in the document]
allenss-amazon: I don't see this as a problem that we want to worry about now. It's mostly theoretical and self-correcting.

allenss-amazon (approved):

PR #1217, opened 2026-07-02
Fix SORTBY with NOCONTENT returning unsorted results (#1215)

### LINKED ISSUE #1215: [BUG] Query on TAG returns wrong order when SORTBY ... NOCONTENT
Perhaps numeric sorting is still done on the fly, and NOCONTENT bypasses loading the hash?

```
127.0.0.1:16379> FT.CREATE idx SCHEMA t TAG n NUMERIC SORTABLE
OK
127.0.0.1:16379> HSET k1 t 1 n 1
(integer) 2
127.0.0.1:16379> HSET k2 t 2 n 2
(integer) 2
127.0.0.1:16379> HSET k3 t 3 n 3
(integer) 2
127.0.0.1:16379> FT.SEARCH idx @t:{1|2|3}
1) (integer) 3
2) "k2"
3) 1) "n"
   2) "2"
   3) "t"
   4) "2"
4) "k1"
5) 1) "t"
   2) "1"
   3) "n"
   4) "1"
6) "k3"
7) 1) "t"
   2) "3"
   3) "n"
   4) "3"
127.0.0.1:16379> FT.SEARCH idx @t:{1|2|3} SORTBY n
1) (integer) 3
2) "k1"
3) 1) "n"
   2) "1"
   3) "t"
   4) "1"
4) "k2"
5) 1) "n"
   2) "2"
   3) "t"
   4) "2"
6) "k3"
7) 1) "t"
   2) "3"
   3) "n"
   4) "3"
127.0.0.1:16379> FT.SEARCH idx @t:{1|2|3} SORTBY n NOCONTENT
1) (integer) 3
2) "k2"
3) "k1"
4) "k3"
```

Aksha1812:
## What & why

Fixes #1215. `FT.SEARCH ... SORTBY <field> NOCONTENT` returned rows in
insertion order instead of sorted order.

`NOCONTENT` forced `kNoContent` content processing, which skips loading
attribute values. But `SORTBY` on an attribute needs those values to order
the result, so sorting was silently bypassed:

- **Standalone:** the `NOCONTENT` reply was emitted before `ApplySorting` ran.
- **Cluster:** remote shards never fetched the sort field, so the coordinator
  had nothing to order by.

## The change

- `GetContentProcessing()` no longer returns `kNoContent` when the query
  requires complete results (`SORTBY` present); the sort field is loaded.
- `SendReply` runs `ApplySorting` first, then, under `NOCONTENT`, replies with
  ids only (`SendReplyNoContent`) — sorted, without leaking content.
- `HandleEarlyReplyScenarios` only takes the early `NOCONTENT` path when there
  is no `SORTBY`.
- Relax the coordinator remote-responder invariant: `NOCONTENT + SORTBY` now
  legitimately completes on the main thread.

## Behavior — valkey-search vs RediSearch

Verified live against RediSearch (redis-stack 2.10.2, DIALECT 2) on an
identical index and dataset. Index `sidx ON HASH PREFIX 1 s: SCHEMA t TAG
n NUMERIC SORTABLE`; keys inserted in id order with `n = 50, 10, 40, 20, 30`
for `s:1 … s:5` (so id order ≠ sorted order).

| Query | valkey **before** | valkey **after** (this PR) | RediSearch |
|---|---|---|---|
| `SORTBY n ASC NOCONTENT LIMIT 0 10` | s:1,s:2,s:3,s:4,s:5 | s:2,s:4,s:5,s:3,s:1 | s:2,s:4,s:5,s:3,s:1 |
| `SORTBY n DESC NOCONTENT LIMIT 0 10` | s:1,s:2,s:3,s:4,s:5 | s:1,s:3,s:5,s:4,s:2 | s:1,s:3,s:5,s:4,s:2 |
| `SORTBY n ASC NOCONTENT LIMIT 0 3` | s:1,s:2,s:3 | s:2,s:4,s:5 | s:2,s:4,s:5 |
| `SORTBY n DESC NOCONTENT LIMIT 0 3` | s:1,s:2,s:3 | s:1,s:3,s:5 | s:1,s:3,s:5 |

Before the fix valkey returned insertion order regardless of `SORTBY`; after,
every result matches RediSearch exactly.

## Testing

- `commands_test` — 444/444 pass.
- Integration `test_sortby_nocontent_ordering` (standalone) and a new
  `test_sortby_nocontent_cross_shard` (cluster) — both pass. The cross-shard
  test seeds 30 keys whose values permute their key order, asserts the keys
  span more than one shard, and checks globally sorted output for both a full
  scan and a top-k slice, ASC and DESC.
- Live cross-check against RediSearch (table above).


src/commands/ft_search.cc
```diff
@@ -261,7 +261,9 @@ bool HandleEarlyReplyScenarios(ValkeyModuleCtx *ctx,
     return true;  // Early reply sent, stop processing
   }
 
-  if (command.no_content) {
+  // SORTBY needs content loaded and sorted first, so only take the fast
+  // NOCONTENT path when there is no sort.
+  if (command.no_content && !command.sortby_parameter.has_value()) {
```
allenss-amazon: This says that if SORTBY is present, then go through the slow path.

But for a NOCONTENT query, there are only two things that you CAN sort on: key and distance. We know that for a vector query, the output is already sorted by distance (_I think_). So, I think we want to detect not just the presence of the sortby parameter, but whether the sortby parameter points to the vector distance -> in which case we can still short-circuit.

Apparently, Redisearch doesn't always yield stuff in distance order for NOCONTENT, so I've seen client code where they do NOCONTENT and SORTBY <distance-field>. So we want to catch that case and ensure that it's correct. 

We should make sure that our testing includes all 6 combinations of NOCONTENT with SORTBY {__key, __distance}

Aksha1812: had claude do a analysis 

```

  ┌────────────────────┬───────────────────────────────┬────────────────────────────────┬─────────────────────────┬────────────────────────────┐
  │       SORTBY       │     WITHOUT NOCONTENT —       │ WITHOUT NOCONTENT — valkey (my │    WITH NOCONTENT —     │  WITH NOCONTENT — valkey   │
  │                    │          RediSearch           │               PR)              │       RediSearch        │          (my PR)           │
  ├────────────────────┼───────────────────────────────┼────────────────────────────────┼─────────────────────────┼────────────────────────────┤
  │ none               │ v1,v2,v3 (insertion, not      │ v1,v3,v2 (distance)            │ v1,v2,v3 (insertion)    │ v1,v3,v2 (distance)        │
  │                    │ distance)                     │                                │                         │                            │
  ├────────────────────┼───────────────────────────────┼────────────────────────────────┼─────────────────────────┼────────────────────────────┤
  │ vscore ASC         │ v1,v3,v2 ✅                   │ ❌ Index field vscore does not │ v1,v3,v2 ✅             │ ❌ same error              │
  │ (distance)         │                               │  exist                         │                         │                            │
  ├────────────────────┼───────────────────────────────┼────────────────────────────────┼─────────────────────────┼────────────────────────────┤
  │ vscore DESC        │ v2,v3,v1 ✅                   │ ❌ error                       │ v2,v3,v1 ✅             │ ❌ error                   │
  ├────────────────────┼───────────────────────────────┼────────────────────────────────┼─────────────────────────┼────────────────────────────┤
  │ n ASC (numeric)    │ v2,v1,v3 ✅                   │ v2,v1,v3 ✅                    │ v2,v1,v3 ✅             │ v2,v1,v3 ✅                │
  ├────────────────────┼───────────────────────────────┼────────────────────────────────┼─────────────────────────┼────────────────────────────┤
  │ n DESC             │ v3,v1,v2 ✅                   │ v3,v1,v2 ✅                    │ v3,v1,v2 ✅             │ v3,v1,v2 ✅                │
  ├────────────────────┼───────────────────────────────┼────────────────────────────────┼─────────────────────────┼────────────────────────────┤
  │ __key              │ ❌ Property __key not loaded  │ ❌ Index field __key does not  │ ❌ same                 │ ❌ same                    │
  │                    │                               │ exist                          │                         │                            │
  └────────────────────┴───────────────────────────────┴────────────────────────────────┴─────────────────────────┴────────────────────────────┘

```

Aksha1812: > But for a NOCONTENT query, there are only two things that you CAN sort on: key and distance.  

the bug this PR is trying to solve is when `SORTBY field_name` is provided with NOCONTENT . Redisearch sorts by that field_name and the removes the content .  we are trying to match that.

allenss-amazon: Ok, I didn't know that. Then this PR is closer, but where is the content removed??

Maybe what we want to do is to reject the SORTBY clause that's on any field other than the distance clause. Enhancing SORTBY to solve the other situations can be a future feature.

Aksha1812: do ytou mean just remove sorting on distance when no SORTBY clause is provided and fix sorting for `vscore ASC` and `vscore DESC` for `NOCONTENT` and `CONTENT` which is diverging in above table . and not add changes for `n ASC` and `n DESC` which was fixed in this PR . I can move this PR to draft since you said it can be a future feature . and add a new PR for fixing distance SORTBY, I am confused though . do you want me to fix just the `CONTENT` case since we do not support SORTBY with NOCONTENT at all currently

Aksha1812: > Ok, I didn't know that. Then this PR is closer, but where is the content removed??

i think it is not technically "removed" from what i understand . we just don't load it . when sortby clause is also provided we would load it for sorting and not write it into reply . no change if sortby not provided

allenss-amazon: I'm saying that for NOCONTENT searches that also specify SORTBY, we treat the command as a "not implemented" unless the SORTBY clause specifies the vector-distance in ascending order (which is what you get with no sort).

Aksha1812: https://github.com/valkey-io/valkey-search/pull/1267 i think this is what you meant . And since you said we can implement the changes here as a future feature , i will change this PR to draft

Aksha1812:
**Reviewers for this PR**

Aksha1812:
Not too be reviewed right now . some changes moved to #1267 as discussed in this PR . Will make this draft or update it soon

boda26 (commented):
I suggest also add cluster-mode integration tests to validate the remote shards correctly fetch the sorting field and returns correct results.

Frank-Gu-81:
Hi @Aksha1812 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

Aksha1812:
I drafted compatibility changes here https://github.com/Aksha1812/valkey-search/compare/dev-fix-issue-1215-sortby-nocontent...dev-1217-compat-sortby-nocontent  as well if you want to include them with this PR

src/coordinator/server.cc
```diff
@@ -133,7 +133,7 @@ class RemoteResponderSearch : public query::SearchParameters {
   void QueryCompleteMainThread(
       std::unique_ptr<SearchParameters> self) override {
     CHECK(vmsdk::IsMainThread());
-    CHECK(!no_content);  // Shouldn't be here!
+    CHECK(!no_content || RequiresCompleteResults());
```
KarthikSubbarao: This is adding more complexity if we keep adding more conditions here.

Context: no_content was meant to be the function / state that informed us that there is NO work needed after a background search. If this is changing and there are cases where nocontent still requires main thread processing or complete results, we need a new synonym for noprocessing/nowork.

1. Can we please use a single function every where for this? (e.g. noprocessing)?
2. Can we also check every usage of nocontent and double check if it should be changed to use this new function (e.g. noprocessing)?

boda26 (approved):

src/coordinator/server.cc
Aksha1812: Added a single predicate on SearchParameters:
```
  bool NoProcessingRequired() const {
    return no_content && !RequiresCompleteResults();
  }
```

Every no_content && !RequiresCompleteResults() condition now routes through it: GetContentProcessing(), HandleEarlyReplyScenarios, and the two coordinator completion checks. I also audited every no_content usage. The rest keep the raw flag on purpose

Aksha1812:
@KarthikSubbarao Addressed your comment above. should i add below compatibility tests mentioned here https://github.com/valkey-io/valkey-search/pull/1217#issuecomment-5623845299 as well to this PR ?

> I drafted compatibility changes here [Aksha1812/valkey-search@dev-fix-issue-1215-sortby-nocontent...dev-1217-compat-sortby-nocontent](https://github.com/Aksha1812/valkey-search/compare/dev-fix-issue-1215-sortby-nocontent...dev-1217-compat-sortby-nocontent) as well if you want to include them with this PR

KarthikSubbarao (approved):

testing/integration/vector_search_integration_test.py
```diff
@@ -687,6 +687,71 @@ def test_vector_search(self, config):
             )
             self.assertEqual(want, got)
             
+    def test_sortby_nocontent_ordering(self):
```
KarthikSubbarao, line 690: Please don't use time.sleep in tests. this creates flaky tests

Aksha1812: fixed here https://github.com/valkey-io/valkey-search/pull/1217/commits/e46350690e60f7b4e3873f80e7e9f3f8729cfb59 , test was failing , fixed here https://github.com/valkey-io/valkey-search/pull/1217/commits/851bedc1c5f74a454b76ca4e41a61fc78996c887

PR #1224, opened 2026-07-04
enable memory sharing for vectors

yairgott:
This change introduces a vector memory sharing between the module and the engine to reduce the memory footprint. Instead of duplicating vector data inside the index structure (FLAT and HNSW), valkey-search maintains a single copy of a vector which its reference is used by the indexes and also shared with the engine.

#### Key Changes:
1. **Vector Externalizer**:
   - Introduced `VectorExternalizer` (in [src/vector_externalizer.h](file:///usr/local/google/home/yairg/work/valkey-search4/src/vector_externalizer.h)) to coordinate shared/externalized vector allocations.
   - Added configuration options for memory sharing under `ValkeySearchOptions`.
2. **Index Refactoring**:
   - Integrated `VectorExternalizer` during vector addition/mutation and the RDB load paths.
   - Updated custom `hnswlib` space and distance metrics headers to support externalized vector structures.
3. **Valkey Module Integration**:
   - Updated schema manager and keyspace event manager to coordinate externalization lifecycles correctly.
   - Updated mocks to test Valkey Module APIs and prevent double-frees under memory-shared workloads.
4. **HNSWLib Modifications**:
   - Introduced a space function wrapper to optimize and support Cosine distance calculations efficiently.
5. **Testing**:
   - Extended [testing/vector_test.cc](file:///usr/local/google/home/yairg/work/valkey-search4/testing/vector_test.cc) to verify save/load logic under cosine/L2 metrics with memory sharing.
   - Added tests in [testing/vector_externalizer_test.cc](file:///usr/local/google/home/yairg/work/valkey-search4/testing/vector_externalizer_test.cc) to validate the externalizer lifecycle.


yairgott:
Sorry for the mix up, the updated PR that should be looked at is #1213

Issue #918, opened 2026-03-18
[BUG] version 1.2.0 bulid failure


remicollet:
**Describe the bug**

Build failure
```
/usr/bin/ld: .deps/install/lib64/libbenchmark.a(benchmark.cc.o): relocation R_X86_64_32 against hidden symbol `_ZN9benchmark8internal14global_contextB5cxx11E' can not be used when making a PIE object
/usr/bin/ld: failed to set dynamic section sizes: bad value
collect2: error: ld returned 1 exit status

```

**To Reproduce**
./build.sh --verbose

**Expected behavior**

No error

**Environment**

- Fedora 43
- Valkey Search 1.2.0
- Valkey Core Version 9.1.0-rc1 

**Additional context**

Valkey search 1.1.0 was OK.

allenss-amazon:
no movement. If this is still a problem, please re-open or submit a new issue

PR #1225, opened 2026-07-06
Modernize and standardize build system

mnunberg1:
This refactors the build system to use more idiomatic CMake

- Flags not strictly necessary for compilation are removed from the outer layer (they can/should be implemented internally) in build.sh
- Remove ICU from our own repository root
- Build internal components as OBJECT or SHARED libraries
- Allow using system OR vendored dependencies
- Tests can now be run in CTest

The overall approach removes as much valkey-search-specific customization in CMakeLists.txt and so on, and attempts to let build system defaults do most of the heavy lifting.

- Removed most of the custom CMake functions, preferring standard flags, e.g. `finalize_test_flags` or `valkey_search_add_static_library`
- Removed intermediate static libraries. All intermediate components are built as CMake `OBJECT` libraries. Object libraries behave like static libraries for the most part, but they are a CMake-specific logical collection of objects rather than a physical archive created by `ar`
- Internal components as well as external libraries can be packaged/built as shared libraries. To this effect, there is a `libsearch_dso.so` and a `libvmsdk_dso.so` which are shared objects that can be linked against for tests. This makes build and link times substantially quicker, allowing for quicker iteration during development

While doing all these changes, care was taken to retain compatibility with the existing package/vendoring approach, i.e. static dependencies, specific compile-time flags; however, the the gist and direction of the change itself is to allow for an easy migration away from this model in the future. Open Source communities and most Linux distributions do not look kindly on an all-static approach


KarthikSubbarao (commented):
@mnunberg1 - Thanks, can we work on getting the checks passing?

cmake/Modules/valkey_search.cmake
```diff
@@ -1,212 +0,0 @@
...
-  target_compile_options(${TARGET} PRIVATE -mtune=generic)
-  target_compile_options(${TARGET} PRIVATE -gdwarf-5)
-  target_compile_options(${TARGET} PRIVATE -gz=zlib)
-  if(NOT VALKEY_SEARCH_DEBUG_BUILD)
-    target_compile_options(${TARGET}
-                           PRIVATE -ffile-prefix-map=${CMAKE_SOURCE_DIR}=)
-  endif()
-  target_compile_options(${TARGET} PRIVATE -ffast-math)
-  target_compile_options(${TARGET} PRIVATE -funroll-loops)
-  target_compile_options(${TARGET} PRIVATE -ftree-vectorize)
```
Nivesh-01, line 123: Removing all of these performance flags will make search a lot slower speacially for vector search

CMakeLists.txt
```diff
@@ -1,72 +1,91 @@
...
-    CACHE STRING "Sanitizer to enable: 'address' or 'thread'")
+# @TODO: these are only here for testing purposes.. should be removed.
+add_compile_options(-Wno-defaulted-function-deleted)
+add_compile_options(-Wno-deprecated-declarations)
+add_compile_options(-Wno-unknown-pragmas)
+add_compile_definitions(-DTESTING_TMP_DISABLED)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules/")
 
-if(APPLE)
```
Nivesh-01, line 32: Removing OS specific code will and making everything general will cause many problems for those platforms. No macOS guardrails can be dangerous.

CMakeLists.txt
```diff
@@ -1,72 +1,91 @@
...
-  set(CMAKE_C_COMPILER_LAUNCHER ccache)
-  set(CMAKE_CXX_COMPILER_LAUNCHER ccache)
-endif()
+option(BUILD_UNIT_TESTS "Build valkey-search unit tests" ON)
+option(VALKEY_VENDORED_DEPS "Use static vendored dependencies" OFF)
+set(SAN_BUILD "OFF" CACHE STRING "Build with sanitizers (OFF, address, thread, memory, undefined)")
 
-# This project can only compile macOS / Linux, we can use the UNIX predefined
-# variable to test this
-if(NOT UNIX)
```
Nivesh-01, line 20: same as macOS platform removal problem

CMakeLists.txt
```diff
@@ -1,72 +1,91 @@
...
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-set(SAN_BUILD
-    ""
-    CACHE STRING "Sanitizer to enable: 'address' or 'thread'")
+# @TODO: these are only here for testing purposes.. should be removed.
+add_compile_options(-Wno-defaulted-function-deleted)
+add_compile_options(-Wno-deprecated-declarations)
+add_compile_options(-Wno-unknown-pragmas)
+add_compile_definitions(-DTESTING_TMP_DISABLED)
```
Nivesh-01, line 44: why are we disabling the tests over here

vmsdk/src/testing_infra/utils.h
```diff
@@ -16,8 +16,8 @@
 #include "absl/strings/string_view.h"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
-#include "vmsdk/src/testing_infra/module.h"
-#include "vmsdk/src/valkey_module_api/valkey_module.h"
+#include "testing_infra/module.h"
+#include <valkey/valkey_module.h>
```
Nivesh-01: is there a reason for changing from direct paths to relative paths. I see that there are files where this refactoring has been done and parts where it is not, could you please explain this.

vmsdk/testing/CMakeLists.txt
```diff
@@ -1,106 +1,80 @@
...
+target_link_libraries(testing_common INTERFACE GTest::gtest_main GTest::gmock)
+target_include_directories(testing_common INTERFACE .)
+target_include_directories(testing_common INTERFACE ${PROJECT_SOURCE_DIR})
+
+set(SRCS_BLOCKED_CLIENT_TEST blocked_client_test.cc)
 add_executable(blocked_client_test ${SRCS_BLOCKED_CLIENT_TEST})
-target_include_directories(blocked_client_test PUBLIC ${CMAKE_CURRENT_LIST_DIR})
-target_link_libraries(blocked_client_test PUBLIC vmsdklib)
-target_link_libraries(blocked_client_test PUBLIC vmsdk_testing_infra)
-finalize_test_flags(blocked_client_test)
```
Nivesh-01, line 6: Removing finalize_test_flags removes all sanitizers, optimisations and linker flags. The output directory is also no longer set.

src/commands/CMakeLists.txt
```diff
@@ -1,73 +1,50 @@
...
-target_include_directories(ft_search_parser PUBLIC ${CMAKE_CURRENT_LIST_DIR})
-target_link_libraries(ft_search_parser PUBLIC filter_parser)
-target_link_libraries(ft_search_parser PUBLIC index_schema)
-target_link_libraries(ft_search_parser PUBLIC metrics)
-target_link_libraries(ft_search_parser PUBLIC schema_manager)
-target_link_libraries(ft_search_parser PUBLIC index_base)
-target_link_libraries(ft_search_parser PUBLIC search)
-target_link_libraries(ft_search_parser PUBLIC vmsdklib)
-target_link_libraries(ft_search_parser PUBLIC valkey_module)
+add_library(ft_search_parser OBJECT
```
Nivesh-01, line 25: We will lose all of the compiler flags from valkey_search_add_static_library

third_party/hnswlib/CMakeLists.txt
```diff
@@ -1,34 +1,14 @@
...
-add_library(hnswlib_vmsdk INTERFACE ${SRCS_HNSWLIB_VMSDK})
-target_include_directories(hnswlib_vmsdk INTERFACE ${CMAKE_CURRENT_LIST_DIR})
-target_link_libraries(hnswlib_vmsdk INTERFACE iostream)
-# Link the compiled SimSIMD dispatch unit (simsimd_c builds c/lib.c exactly
-# once and PUBLIC-links the simsimd headers). simsimd.h now includes only
-# declarations, so the simsimd_* definitions must come from here -- otherwise
-# they'd be emitted into every TU and collide at link time (macOS hard error).
-target_link_libraries(hnswlib_vmsdk INTERFACE simsimd_c)
-target_link_libraries(hnswlib_vmsdk INTERFACE vmsdklib)
-target_compile_definitions(hnswlib_vmsdk
```
Nivesh-01, line 22: This (target_compile_definitions) is important for production without it, hnswlib allocates memory using the system allocator malloc instead of Valkey's tracked allocator.

CMakeLists.txt
```diff
@@ -1,72 +1,91 @@
...
+else()
+    find_package(gRPC REQUIRED)
+    find_package(Protobuf REQUIRED)
 endif()
 
-message(STATUS "GRPC_CPP_PLUGIN_PATH is set to ${GRPC_CPP_PLUGIN_PATH}")
-message(STATUS "protoc_EXE is set to ${protoc_EXE}")
-message(STATUS "highwayhash include path: ${HIGHWAY_HASH_INCLUDE_PATH}")
+find_package(absl REQUIRED)
+find_package(ICU REQUIRED COMPONENTS uc i18n data)
```
Nivesh-01, line 82: The find_package requires that these packages are installed in the system if they are not it will not build. This is one of the reason for the test failures since build is never passing.

CMakeLists.txt
```diff
@@ -1,72 +1,91 @@
...
-if(APPLE)
-  add_compile_options(-Wno-defaulted-function-deleted)
+# Shared root for generated protobuf/grpc sources consumed across subdirectories.
+set(VALKEY_PROTOGEN_BINARY_DIR "${CMAKE_BINARY_DIR}/protogen")
+
+if (VALKEY_VENDORED_DEPS)
+    set(gRPC_USE_STATIC_LIBS ON)
+    set(Protobuf_USE_STATIC_LIBS ON)
+    set(absl_USE_STATIC_LIBS ON)
+    set(ICU_USE_STATIC_LIBS ON)
```
Nivesh-01, line 55: CMake's built-in `FindICU.cmake` does not have any `ICU_USE_STATIC_LIBS` input variable from my research. This should be a silient failure if I am not mistaken. Some of the ones above have a similar problem.

cmake/Modules/valkey_search.cmake
mnunberg1: They should be added out-of-band, they don't belong inside CMake

CMakeLists.txt
mnunberg1: This preprocessor definition was already present. This disables certain sections of the tests which won't compile otherwise.

vmsdk/src/testing_infra/utils.h
mnunberg1: Because it's clearer to read and more resilient should the path structure ever change. A file doesn't need to know the entire directory structure; just the files that are relevant to it.

vmsdk/testing/CMakeLists.txt
mnunberg1: You can and should do all of this outside CMake. CMake is a generic build system, not an optimization platform. You can configure global optimizer flags (what CMake was doing anyway) outside of the build system

src/commands/CMakeLists.txt
mnunberg1: see above. You can add those flags out-of-band

third_party/hnswlib/CMakeLists.txt
mnunberg1: Good catch

CMakeLists.txt
mnunberg1: Putting random OS specific code (is Mac even a production platform?) just makes things more confusing. Don't put "OS guardrails" unless there is a good and documented reason to do so.

CMakeLists.txt
mnunberg1: This is completely redundant. The file will fail to compile on e.g. Windows

vmsdk/src/testing_infra/utils.h
mnunberg1: I don't want to refactor the entire project to keep the delta relatively small; but I did touch parts where there were build issues

CMakeLists.txt
mnunberg1: Yes.. that's kind of the point. You can make a vendored/static version. I don't see a reason to include ICU (a well-defined external library) as part of our repository.

CMakeLists.txt
mnunberg1:
```
[mbp-work] vendored $ otool -L src/libsearch.so 
src/libsearch.so:
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.12)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 5026.5.4)
	/opt/homebrew/opt/openssl@3/lib/libssl.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/opt/homebrew/opt/openssl@3/lib/libcrypto.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1356.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 2100.43.0)
```

This is a standard CMake variable syntax. It's not set by FindXXX.cmake

cmake/Modules/valkey_search.cmake
mnunberg1: These flags have been readded via build.sh

third_party/icu/VERSION
```diff
@@ -1 +0,0 @@
-76.1
```
KarthikSubbarao, line 1: Are we checking that the new approach of pinning this is downloading the same version?

CMakeLists.txt
```diff
@@ -1,72 +1,131 @@
 cmake_minimum_required(VERSION 3.16)
 
-# Project definition
-
 project(
-  ValkeySearch
-  LANGUAGES C CXX)
+    ValkeySearch
+    VERSION 1.0.0
```
Aksha1812: Maybe this was not rebased cleanly. If you see latest commit in this file on main https://github.com/valkey-io/valkey-search/commit/b8663b30ef46c94834e643bf6df955feba798935  i removed versioning info from this line as it was ill maintained and never used anywhere.

third_party/icu/VERSION
mnunberg1: No. I'm actually not sure who is responsible for bundling and/or packaging, or whether ICU versions even matter?

CMakeLists.txt
mnunberg1: Rebased onto main and removed the VERSION line in b82a64f.

KarthikSubbarao:
Can we update the PR description with details on what was removed and what was changed?

I am pulling the PR and trying to build the package the way we do today and see this:
```
X% ./build.sh 
Root directory: X
Checking if configure is required...yes
Checking for cmake...ok
Checking for ctest...failed.
ERROR - could not locate tool 'ctest'.
```

third_party/icu/VERSION
KarthikSubbarao: > or whether ICU versions even matter?

Distros can be running on different versions? From what I read, that could be different Unicode / normalization logic. ie, we may see subtle Unicode behavior differences across distros.

>  I'm actually not sure who is responsible for bundling and/or packaging,

For the bundled version of all modules + core, we, valkey-bundle is responsible.

See the image types we need to support: https://hub.docker.com/r/valkey/valkey-bundle

```
Image Variants
valkey/valkey-bundle:<version>
This is the primary image, which includes Valkey along with common modules like valkey-json, valkey-bloom, valkey-search, and valkey-ldap preloaded. It is suitable for development, testing, and production environments where these modules are needed out of the box.

Some of the tags may include names like bookworm, which refer to [Debian release codenames](https://wiki.debian.org/DebianReleases). These indicate the base image used and help ensure compatibility if your container needs additional packages. Specifying these explicitly is recommended to avoid unexpected changes when base image versions update.

If you want a minimal yet functional Valkey container with built-in modules, this image is a great place to start.

valkey/valkey-bundle:<version>-alpine
This image is based on the popular [Alpine Linux project](https://alpinelinux.org/), available in [the alpine official image](https://hub.docker.com/_/alpine). Alpine Linux is much smaller than most distribution base images (~5MB), and thus leads to much slimmer images in general.

This variant is useful when final image size being as small as possible is your primary concern. The main caveat to note is that it does use [musl libc](https://musl.libc.org/) instead of [glibc and friends](https://www.etalabs.net/compare_libcs.html), so software will often run into issues depending on the depth of their libc requirements/assumptions. See [this Hacker News comment thread](https://news.ycombinator.com/item?id=10782897) for more discussion of the issues that might arise and some pro/con comparisons of using Alpine-based images.
```

mnunberg1: We are not the only user of ICU. Plenty of libraries, databases, servers, and so on use and depend on ICU, usualy dynamically.

We don't need to be bug-for-bug compatible with a given version of ICU.. though newer is usually better.

By "who is responsible" I mean and I repeat, versioning is not the job of a build system. If a specific distribution of search depends on a specific version of ICU, then the coupling should take place as late as possible, usually in that distribution's packaging scripts.

mnunberg1:
> Can we update the PR description with details on what was removed and what was changed?
> 
> I am pulling the PR and trying to build the package the way we do today and see this:
> 
> ```
> X% ./build.sh 
> Root directory: X
> Checking if configure is required...yes
> Checking for cmake...ok
> Checking for ctest...failed.
> ERROR - could not locate tool 'ctest'.
> ```


ctest is usually bundled with cmake. What kind of distribution are you running that lacks it?

vmsdk/include/valkey/module_sdk/cluster_map.h
```diff
@@ -0,0 +1 @@
+#include "../../../src/cluster_map.h"
```
yairgott, line 1: Why does the include path specify relative path?

cmake/Modules/valkey_search.cmake
yairgott: In one way or the other, these flags are required to be automatically turned on when following our standard defined build process. From the user perspective, the build process should be very smooth and the common expectation is that the optimal compilation flags are already turned on by the build system.

vmsdk/include/valkey/module_sdk/cluster_map.h
mnunberg1: This is just a placeholder until we can get/move the actual file into its proper location. The path is relative because the include file shouldn't assume anything else about `-I` flags being passed that aren't other files in `include` directories.

cmake/Modules/valkey_search.cmake
mnunberg1: Most of these optimizations could be had by selecting a specific `-march=` option [e.g. `-march=native` or `-march=x86-64-v3`, in addition to `-ffast-math`.

For users compiling from scratch, it should be simple enough to place this hint within a README.. but rather than complicating the flag matrix (as it was previously), I decided to leave the CMake flags strictly to what is necessary for functionality.

The build.sh script, which is still the "official way" (as of this commit) still effectively applies the same optimization flags.

yairgott: SG! Pointing out that my AI code assistant flagged one inconsistency: "-Wno-missing-requires: Removed from global build flags and isolated only to submodules (submodules/CMakeLists.txt)."

src/commands/ft_aggregate_parser.h
```diff
@@ -109,23 +109,35 @@ struct AggregateParameters : public expr::Expression::CompileContext,
   size_t AddRecordAttribute(absl::string_view identifier,
                             absl::string_view alias,
                             indexes::IndexerType data_type) {
+    // Reserve the alias slot up front (defaulting to index 0) so we can tell
+    // "brand new alias" apart from "alias already pointed elsewhere" below.
+    std::string alias_str(alias);
+    auto [alias_itr, alias_is_new] =
+        record_indexes_by_alias_.try_emplace(alias_str);
```
yairgott, line 116: There are two issues with using `try_emplace` and checking `if (alias_itr->second != index)`:
1.  If the identifier is already registered record_info_by_index_[index].alias_ is never updated to the new  alias_str, leaving the slot with its old metadata.
2. try_emplace(alias_str) value-initializes ,alias_itr->second,  to 0. If the index is equal to 0, `alias_itr->second != index` (`0 != 0`) evaluates to `false`, bypassing the rebinding block completely.

We can make this simpler and robust by explicitly managing the identifier slot and alias lookup

Gemini suggests to address this by:

```cpp
size_t AddRecordAttribute(absl::string_view identifier,
                          absl::string_view alias,
                          indexes::IndexerType data_type) {
  std::string alias_str(alias);
  // 1. Find or allocate the slot for this identifier
  size_t index;
  auto id_itr = record_indexes_by_identifier_.find(identifier);
  if (id_itr != record_indexes_by_identifier_.end()) {
    index = id_itr->second;
    record_info_by_index_[index].alias_ = alias_str;
  } else {
    std::string id_str(identifier);
    index = record_info_by_index_.size();
    record_indexes_by_identifier_.emplace(id_str, index);
    record_info_by_index_.push_back(AttributeRecordInfo{
        .identifier_ = id_str,
        .alias_ = alias_str,
        .data_type_ = data_type});
  }
  // 2. Rebind the alias map to this slot; if previously bound elsewhere, clear old display alias
  auto alias_itr = record_indexes_by_alias_.find(alias);
  if (alias_itr != record_indexes_by_alias_.end()) {
    if (alias_itr->second != index) {
      record_info_by_index_[alias_itr->second].alias_.clear();
      alias_itr->second = index;
    }
  } else {
    record_indexes_by_alias_.emplace(alias_str, index);
  }
  return index;
}
```

ci/build_macos.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash -e
+# Generated with Claude Code (https://claude.com/claude-code)
+
+CI_DIR=$(readlink -f $(dirname $0))
+ROOT_DIR=$(readlink -f ${CI_DIR}/..)
+BUILD_SH_ARGS=$@
```
yairgott, line 5: Gemini feedback:

Native BSD readlink on macOS does not support the -f flag (readlink: illegal option -- f), causing the script to fail when executed on standard macOS environments without GNU coreutils. Fix: Use standard portable directory resolution:

```
CI_DIR="$(cd "$(dirname "$0")" && pwd -P)"
ROOT_DIR="$(cd "${CI_DIR}/.." && pwd -P)"
BUILD_SH_ARGS=("$@")

```

ci/build_macos.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash -e
+# Generated with Claude Code (https://claude.com/claude-code)
```
yairgott: Such comments should be removed.

ci/summarize_test_results.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+# generated with Claude Code
```
yairgott: Such comments should be removed.

testing/CMakeLists.txt
```diff
@@ -1,182 +1,130 @@
...
-target_link_libraries(testing_common_coordinator INTERFACE coordinator_cc_proto)
-target_link_libraries(testing_common_coordinator INTERFACE util)
-target_link_libraries(testing_common_coordinator INTERFACE client)
-target_link_libraries(testing_common_coordinator INTERFACE client_pool)
+set(SRCS_COMMON common.cc common.h)
+
+if (VALKEY_VENDORED_DEPS)
+    # when using vendored (static deps) mode, tests are not linked against the intermediate search_dso object
+    # (which itself is also not built), because we end up with the same static deps being linked in twice.
+    # I asked Claude to manually enumerate the dependencies of the _dso variants instead.
```
yairgott: Such comments should be removed.

CMakeLists.txt
```diff
@@ -1,72 +1,130 @@
...
+endif ()
+
+# this is a built-in CMake variable.. but let's prefer SHARED_LIBS unless specifically disabled..
+set(BUILD_SHARED_LIBS ON CACHE BOOL "Build shared libraries by default")
+
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+# @TODO: these are only here for testing purposes.. should be removed.
```
yairgott, line 40: The TODO seems outdated. `add_compile_definitions` should be updated to accept macro names directly. E.g. change -DTESTING_TMP_DISABLED to TESTING_TMP_DISABLED.

src/CMakeLists.txt
```diff
@@ -1,192 +1,99 @@
...
+        vector_externalizer.h
+        version.h)
+
+# search_base is the collection of compiled objects that forms the final library/module
+add_library(search_base OBJECT ${SEARCH_SRCS})
+add_generated_proto_sources(search_base index_schema)
+add_generated_proto_sources(search_base rdb_section)
+
+# mydeps is the dependency bundle; it contains no objects, only include paths
+add_library(mydeps INTERFACE)
```
yairgott, line 34: rename `mydeps` to `search_dependencies`

yairgott:
One concern I have with the current changes is that adding default include search paths (like `src/`, `include/`, and `vmsdk/`) is currently only leveraged in a small handful of files (e.g., in `src/server_events.cc`, `vmsdk/src/testing_infra/module.h`, `vmsdk/testing/*`), where leading prefixes like `src/` or `vmsdk/src/` were dropped.
Since the vast majority of the codebase still uses root-relative include paths (`#include "src/..."` and `#include "vmsdk/src/..."`), having a few files use short/relative paths introduces inconsistency across the repository.
To keep the codebase consistent and make this PR easier to review and land:
1. Could we maintain the full root-relative include paths (`#include "src/..."` and `#include "vmsdk/src/..."`) across the board in this PR?
2. We can then do a dedicated, repository-wide follow-up PR to transition all includes uniformly to take advantage of the new include path capabilities.

ci/build_macos.sh
mnunberg1: I use Mac natively, and this works on the CI; I'll ask Claude to fix this script

ci/build_macos.sh
mnunberg1: Why? I find it helpful to know if I'm looking at code written by a human being vs code written by AI

CMakeLists.txt
mnunberg1: Did you remove the `TESTING_TMP_DISABLED` in the past month?

src/CMakeLists.txt
mnunberg1: I don't like `search_dependencies`, but if `my` is too cryptic (it's a Perl convention for things that are purely local in scope) then I'll figure something else out

mnunberg1:
> One concern I have with the current changes is that adding default include search paths (like `src/`, `include/`, and `vmsdk/`) is currently only leveraged in a small handful of files (e.g., in `src/server_events.cc`, `vmsdk/src/testing_infra/module.h`, `vmsdk/testing/*`), where leading prefixes like `src/` or `vmsdk/src/` were dropped. Since the vast majority of the codebase still uses root-relative include paths (`#include "src/..."` and `#include "vmsdk/src/..."`), having a few files use short/relative paths introduces inconsistency across the repository. To keep the codebase consistent and make this PR easier to review and land:
> 
> 1. Could we maintain the full root-relative include paths (`#include "src/..."` and `#include "vmsdk/src/..."`) across the board in this PR?
> 2. We can then do a dedicated, repository-wide follow-up PR to transition all includes uniformly to take advantage of the new include path capabilities.

The reason I specifically changed those paths is that they are used across components. Otherwise it would have involved exposing/adding another `target_include_directories` dependency. I agree with your approach though, and will see if this is easily rectified in this PR

src/commands/ft_aggregate_parser.h
mnunberg1: 1. We don't need to update the identifier's old alias information. The old alias is still valid, it just gains a _new_ alias. We should probably verify this behavior to see what RS does, but it seems sensible. It is also quite possible that in an ORM or similar, different layers of code might "add" the same identifier but use the same alias. We only update `record_info_by_index` where it would be clearly wrong (e.g. with a stale alias)

2. The condition at the end, `alias_itr->second != index` is only important in order to unbind or remove the stale alias. If both `alias_itr->second == 0` and `index == 0`, then one of the following must be true:
  - A new identifier was created with index 0. In this case, we know that no possible identifiers (and no possible aliases) existed beforehand, so there is no need for an unbind.
  - A *new* alias was added for identifier at index 0; in this case, there is no unbind either, because unbind only happens if `alias_is_new`
  - The same alias was added for the same identifier twice, e.g. `LOAD 2 n1 AS first n1 AS first`, in which case this remains a glorified no-op.

The condition should probably read as:

```
if (!alias_is_new) {
  if (alias_itr->second != index) { // sanity check for duplicate alias declaration
    record_info_by_index_[alias_itr->second].alias_.clear();
  }
}

alias_itr->second = index; // we should always assign here
```

build.sh
```diff
@@ -422,19 +449,29 @@ if [[ "${FORMAT}" == "yes" ]]; then
...
 fi
 
-BUILD_DIR=${ROOT_DIR}/.build-${BUILD_CONFIG}
 if [[ "${SAN_BUILD}" != "no" ]]; then
     printf "${BOLD_PINK}${SAN_BUILD} sanitizer build is enabled${RESET}\n"
+    # Needed before the build step, not just the test-run step: CMake's default
+    # gtest_discover_tests() runs each test binary's --gtest_list_tests as part
+    # of the build itself, which would otherwise hit an ASan ODR-violation
+    # false positive.
+    export ASAN_OPTIONS="detect_odr_violation=0"
```
KarthikSubbarao, line 458: This will replace any options set by the CI or environment. 

Can we add on rather than replacing?

.github/workflows/macos.yml
```diff
@@ -30,6 +30,19 @@ jobs:
...
+
+    - name: Restore highwayhash cache
+      uses: actions/cache@v4
+      with:
+        path: ~/valkey-search-deps
+        key: macos-highwayhash-${{ runner.os }}-${{ env.IMAGE_OS }}-${{ env.IMAGE_VERSION }}-${{ hashFiles('ci/build_macos.sh') }}
+        # No restore-keys deliberately: an exact-match-only key means a cache miss
+        # (OS bump, image refresh, or a ref change in the script) always triggers a
+        # clean rebuild rather than risking a stale/incompatible binary being reused.
+
```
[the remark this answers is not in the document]
KarthikSubbarao: @mnunberg1 - Any thoughts on this?

build.sh
```diff
@@ -391,7 +416,9 @@ function is_configure_required() {
 
     local build_file_lastmodified=$(get_file_last_modified "${top_level_build_file}")
     local IFS=$'\n'
-    local cmake_files=$(find "${ROOT_DIR}" -name "CMakeLists.txt" -o -name "*.cmake" | grep -v ".build-release" | grep -v ".build-debug")
+    local cmake_files=$(find "${ROOT_DIR}" \
+        \( -path "${BUILD_DIR}" -o -path "${BUILD_DIR}/*" \) -prune -o \
+        \( -name "CMakeLists.txt" -o -name "*.cmake" \) -print)
```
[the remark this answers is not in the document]
KarthikSubbarao: @mnunberg1 - Any thoughts on this?

build.sh
```diff
@@ -170,114 +202,110 @@ else
...
+    case "${UNAME_S}" in
+    Darwin)
+        openmp_flags=""
+        ;;
+    esac
+
+    # Always-on flags, merged by CMake with the _RELEASE/_DEBUG variants below.
+    # SAN_COMPILE_FLAGS/SAN_LINKER_FLAGS are set directly at the --asan/--tsan
+    # argument-parsing sites (empty when no sanitizer was requested).
+    local common_flags="-falign-functions=5 -fmath-errno -ffp-contract=off -fno-rounding-math ${arch_simd_flags} -mtune=generic -gdwarf-5 -gz=zlib -ffast-math -funroll-loops -ftree-vectorize ${openmp_flags} -flax-vector-conversions -Wno-unknown-pragmas -Wno-sign-compare -Wno-uninitialized -DTESTING_TMP_DISABLED ${SAN_COMPILE_FLAGS}"
```
KarthikSubbarao, line 235: Question about performance: Are we really sure if we want these flags globally enabled for every target? previously, we had per target flags.

There are certain targets within the package which will observe different performance based on the flags used. Examples of this I have seen first hand are snowball stemmer and HNSW.

Something like -funroll-loops can slow down snowball stemmer execution. I did some microbenchmarks on snowball individually in the past.... https://github.com/valkey-io/valkey-search/pull/826

mnunberg1: I'm not a fan of -funroll-loops in general,  but I wanted to preserve the existing system.

The thing is, specifying individual flags for individual components while still keeping those components together just makes things more fragile.

If a specific bit of code depends on a specific compiler or processor behavior, then such behavior should be as close to the code as possible.

https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gcc/Loop-Specific-Pragmas.html

build.sh
mnunberg1: 👍

Sure -- we can generate the script as you wish. I don't care what happens in build.sh, because that's a packaging script

src/CMakeLists.txt
yairgott: Sure, keep in mind also the naming schema.

ci/build_macos.sh
yairgott: 1. Over time, more and more code would be written by AI. 
2. We shouldn't treat AI written code differently than human written code. It should follow the same standards and coding guidances.

mnunberg1: We don't know how and which code AI will write. Currently we are human beings writing our own code. Certainly we should hold AI code to even higher standards than human code, but on the other hand, AI code is harder to read and understand. Signalling that code is AI indicates to the reader that this code was generated and is not intentional, neat, elegant, or consistent in its design in the way we humans understand it.

Who exactly is hurt by adding this comment?

yairgott: I stand by my earlier comment, but I'd like to address your points:

1. I don't believe AI code is universally harder to read but rather it goes both ways and depends heavily on the context.
2. At the end of the day, the PR author owns the code. We expect all merged code to be neat, high quality and maintainable. If the AI's output isn't up to that standard, we shouldn't use a comment to excuse it. Instead, the author should work with the AI to refine the code or rewrite it themselves until it aligns with our coding guidelines.

mnunberg1:
1) AI is definitely harder to read vs human-directed "curated" code. If
code is generic/glue/copy-paste, then yes, AI might do a better job, but
even this is debatable. Code that is written by AI is poorly changed and/or
significantly modified by humans, whereas code written by humans can be
edited by either humans or AI.
2) The comments were not excusing sub-par AI output, such output could have
very well been written by a human being. The specific reason I mentioned "I
asked Claude to" is because some of the code is repetitive, and explaining
why it's repetitive; this can be removed if you don't find it helpful, but
it seems your objection is pointing out that some segments of code are
AI-generated

I certainly think the quality of the code proposing to be merged is
substantially higher than the code it's replacing, and most would tend to
agree.

third_party/icu/VERSION
KarthikSubbarao: > We don't need to be bug-for-bug compatible with a given version of ICU.. though newer is usually better.

Yes, this seems ok.


I wanted to hear your thoughts on my second question in the message above.


We build valkey-search into the Valkey Bundle docker image.

See the image variants: https://hub.docker.com/r/valkey/valkey-bundle#image-variants

Are there any extra steps we will have to take for the Alpine Image?

> valkey/valkey-bundle:<version>-alpine
> This image is based on the popular [Alpine Linux project](https://alpinelinux.org/), available in [the alpine official image](https://hub.docker.com/_/alpine). Alpine Linux is much smaller than most distribution base images (~5MB), and thus leads to much slimmer images in general.
> To minimize image size, it's uncommon for additional related tools (such as git or bash) to be included in Alpine-based images. Using this image as a base, add the things you need in your own Dockerfile (see the [alpine image description](https://hub.docker.com/_/alpine/) for examples of how to install packages if you are unfamiliar).

KarthikSubbarao:
Any response to these comments?

https://github.com/valkey-io/valkey-search/pull/1225#discussion_r3738843466

https://github.com/valkey-io/valkey-search/pull/1225#discussion_r3738846510

Aksha1812:
**Reviewers for this PR**

BCathcart:
We'll need to update the "Build Instructions" section of the README. The flow right now is installing some basic tools and then `build.sh` pulls in everything else needed at build time. We'll need to incorporate running the scripts that install the third party dependencies that `build.sh` used to (e.g. `build_macos.sh`) right? Maybe we shouldn't hide these in the `ci/` directory?

Better yet, how hard do you think it would be to make it a single step for everyone with an `environment_setup.sh` entry point that detects the environment and calls the correct setup script to install all dependencies?

BCathcart:
The `./build.sh --run-tests` output isn't too pleasant. I would vote for continuing to log the detailed output to a file and just providing pass/fail in the terminal

mnunberg1:
> The `./build.sh --run-tests` output isn't too pleasant. I would vote for continuing to log the detailed output to a file and just providing pass/fail in the terminal

I would rather build.sh not be used at all; the idea is to use CMake directly and configure everything from there. But there is existing tooling still using build.sh, as well existing project contributors relying on it, so I'd rather not break it.

There are also, supposedly, magic flags that are used that significantly impact performance. These are placed in build.sh as well

That being said, build.sh is a second class citizen with this commit

build.sh
```diff
@@ -30,15 +33,19 @@ Usage: build.sh [options...]
...
+    --build-dir[=PATH]                Use an alternate build directory.
+    --run-tests                       Run all unit tests via CTest. Optionally, pass a ctest -R regex to
+                                       filter: "--run-tests=<pattern>" (e.g. "SchemaManagerTest" or
+                                       "SchemaManagerTest.SomeCase").
     --no-build                        By default, build.sh always triggers a build. This option disables this behavior.
-    --test-errors-stdout              When a test fails, dump the captured tests output to stdout.
+    --test-errors-stdout              When a test fails, dump the captured tests output to stdout (ctest --output-on-failure).
     --run-integration-tests[=pattern] Run integration tests.
     --use-system-modules              Use system's installed gRPC, Protobuf & Abseil dependencies.
+    --vendored                        Statically link gRPC, Protobuf & Abseil (VALKEY_VENDORED_DEPS=ON).
```
BCathcart, line 44: This is needed to run unit tests?

BCathcart:
> build.sh is a second class citizen with this commit

If important flags are in `build.sh` and developers still use it and the README still indicates it should be used, I don't think we can call it a second class citizen yet. Why place the flags in build.sh and not at the cmake level? That sounds like it's adding more importance to it

BCathcart:
I had Claude whip up a setup script for amazon linux:

```
#!/bin/bash -e

# Amazon Linux CI/local build entrypoint for valkey-search.
#
# Amazon Linux has neither a prebuilt valkey-search-deps .deb (Ubuntu CI) nor
# Homebrew (macOS CI), so this script builds gRPC, Protobuf, Abseil, GoogleTest,
# google-benchmark, highwayhash, and ICU from source into a private prefix, then
# points CMAKE_PREFIX_PATH at that prefix before invoking ./build.sh.
#
# Only OpenSSL and systemd-devel come from the distro (via yum/dnf).

CI_DIR=$(readlink -f $(dirname $0))
ROOT_DIR=$(readlink -f ${CI_DIR}/..)
BUILD_SH_ARGS=$@

RESET='\e[0m'
GREEN='\e[32;1m'
RED='\e[31;1m'

# Where the from-source third-party dependencies get installed. Override with
# VALKEY_SEARCH_DEPS_DIR to reuse an existing prefix (same knob as
# build_macos.sh). Set SKIP_SYSTEM_DEPS=1 to skip the yum/dnf step (e.g. when you
# lack sudo but the headers are already installed).
DEPS_DIR="${VALKEY_SEARCH_DEPS_DIR:-$HOME/valkey-search-deps}"

# Pinned dependency versions (override via environment). These match
# submodules/package-submodules.sh which produces the Ubuntu deps deb.
GRPC_REF="${GRPC_REF:-v1.70.1}"
GTEST_REF="${GTEST_REF:-main}"
BENCHMARK_REF="${BENCHMARK_REF:-v1.8.3}"
HIGHWAYHASH_REF="${HIGHWAYHASH_REF:-master}"
ICU_VERSION="${ICU_VERSION:-76.1}"

# Sanitizer builds need deps compiled with the same sanitizer, so we install
# them into a per-sanitizer prefix (e.g. ~/valkey-search-deps-asan).
SAN_BUILD="no"
san_suffix=""
SAN_C_FLAGS=""
SAN_CXX_FLAGS=""
SAN_LD_FLAGS=""
for arg in "$@"; do
    case $arg in
    --asan)
        SAN_BUILD="address"
        san_suffix="-asan"
        SAN_C_FLAGS="-fsanitize=address"
        SAN_CXX_FLAGS="-fsanitize=address"
        SAN_LD_FLAGS="-fsanitize=address"
        ;;
    --tsan)
        SAN_BUILD="thread"
        san_suffix="-tsan"
        SAN_C_FLAGS="-fsanitize=thread"
        SAN_CXX_FLAGS="-fsanitize=thread"
        SAN_LD_FLAGS="-fsanitize=thread"
        ;;
    esac
done
DEPS_DIR="${DEPS_DIR}${san_suffix}"
SRC_DIR="${DEPS_DIR}/.src"

function LOG_INFO() {
    printf "${GREEN}INFO ${RESET} $1\n"
}

function LOG_ERROR() {
    printf "${RED}ERROR${RESET} $1\n"
}

# dnf on Amazon Linux 2023 / recent RHEL/Fedora, yum on Amazon Linux 2.
function pkg_mgr() {
    if command -v dnf >/dev/null 2>&1; then
        echo dnf
    else
        echo yum
    fi
}

function install_system_deps() {
    if [[ "${SKIP_SYSTEM_DEPS}" == "1" ]]; then
        LOG_INFO "SKIP_SYSTEM_DEPS=1 set, skipping yum/dnf install"
        return
    fi
    local PKG=$(pkg_mgr)
    LOG_INFO "Installing system packages via ${PKG}"

    # systemd headers and basic build tools. gcc/g++, cmake and ninja are
    # intentionally NOT installed here: the base repos ship versions too old for
    # valkey-search (which needs gcc>=12 / clang>=16), and a newer toolchain is
    # expected to already be on PATH (verified by select_compiler below). ICU is
    # built from source (build_icu), not taken from the distro. The install is
    # best-effort (|| true) so a single unavailable package -- e.g. ninja-build,
    # which lives in EPEL on some releases -- doesn't abort the run when the tool
    # is already present.
    sudo ${PKG} install -y \
        git wget tar gzip make \
        systemd-devel \
        ninja-build || true

    # gRPC's "package" SSL provider does a find_package(OpenSSL); gRPC ${GRPC_REF}
    # needs OpenSSL >= 1.1.1. On Amazon Linux 2 the default openssl-devel is
    # 1.0.2, so prefer the openssl11-devel package there. On AL2023/RHEL9+
    # openssl-devel is already 3.x.
    if [[ "${PKG}" == "yum" ]]; then
        sudo ${PKG} install -y openssl11-devel || sudo ${PKG} install -y openssl-devel || true
    else
        sudo ${PKG} install -y openssl-devel || true
    fi
}

# Pick a C/C++ compiler new enough for valkey-search (gcc>=12 or clang>=16) and
# export CC/CXX so CMake -- both here and in build.sh -- doesn't fall back to the
# distro default `cc`, which on Amazon Linux 2 is GCC 7.3.
function select_compiler() {
    if [[ -n "${CC}" && -n "${CXX}" ]]; then
        LOG_INFO "Using compiler from environment: CC=${CC} CXX=${CXX}"
        return
    fi

    local gcc_major=0
    if command -v g++ >/dev/null 2>&1; then
        gcc_major=$(g++ -dumpversion | cut -d. -f1)
    fi
    local clang_major=0
    if command -v clang++ >/dev/null 2>&1; then
        clang_major=$(clang++ -dumpversion | cut -d. -f1)
    fi

    if [[ "${gcc_major}" -ge 12 ]]; then
        export CC=$(command -v gcc)
        export CXX=$(command -v g++)
    elif [[ "${clang_major}" -ge 16 ]]; then
        export CC=$(command -v clang)
        export CXX=$(command -v clang++)
    else
        LOG_ERROR "No suitable compiler found (need GCC >= 12 or Clang >= 16; found gcc ${gcc_major}, clang ${clang_major})."
        LOG_ERROR "Install a newer toolchain (or set CC and CXX to point at one) and re-run."
        exit 1
    fi
    LOG_INFO "Using compiler CC=${CC} CXX=${CXX}"
}

# Shallow-clone a tag/branch, including submodules (gRPC needs its submodules).
function clone_tag() {
    local url=$1 ref=$2 dest=$3
    rm -rf "${dest}"
    git clone --branch "${ref}" "${url}" "${dest}" \
        --recurse-submodules --shallow-submodules --depth=1 --single-branch
}

# Configure + build + install a CMake dependency into DEPS_DIR.
# $1: source dir  $2: extra cmake args  $3: ninja target (default "install";
# pass "" to only build, e.g. highwayhash which has no install rule).
function build_cmake_dep() {
    local src=$1 extra_args=$2 install_target=${3-install}
    local base_args="-Wno-dev -DCMAKE_BUILD_TYPE=Release -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_INSTALL_PREFIX=${DEPS_DIR} -DCMAKE_CXX_STANDARD=20"
    local build_dir="${src}/.build-release"
    mkdir -p "${build_dir}"
    (
        cd "${build_dir}"
        export CXXFLAGS="-Wno-missing-requires -Wno-attributes -Wno-deprecated -Wno-return-type -Wno-stringop-overflow -Wno-deprecated-declarations ${SAN_CXX_FLAGS}"
        export CFLAGS="${SAN_C_FLAGS}"
        export LDFLAGS="${SAN_LD_FLAGS}"
        cmake .. -GNinja ${base_args} ${extra_args}
        ninja ${install_target}
    )
}

# Build ICU from source (static libraries + static data packaging). Autotools-
# based, so it doesn't go through build_cmake_dep. find_package(ICU) in the
# main build picks it up via ICU_ROOT.
function build_icu() {
    if [ -f "${DEPS_DIR}/lib/libicuuc.a" ] && [ -f "${DEPS_DIR}/include/unicode/utypes.h" ]; then
        LOG_INFO "ICU already present in ${DEPS_DIR}, skipping"
        return
    fi
    local icu_major=${ICU_VERSION%%.*}
    local icu_minor=${ICU_VERSION#*.}
    local tarball="icu4c-${icu_major}_${icu_minor}-src.tgz"
    local url="https://github.com/unicode-org/icu/releases/download/release-${icu_major}-${icu_minor}/${tarball}"

    LOG_INFO "Building ICU ${ICU_VERSION} from source into ${DEPS_DIR}"
    rm -rf "${SRC_DIR}/icu"
    (cd "${SRC_DIR}" && wget -q "${url}" -O "${tarball}" && tar xzf "${tarball}")

    (
        cd "${SRC_DIR}/icu/source"
        CFLAGS="-O2 -fPIC ${SAN_C_FLAGS}" \
        CXXFLAGS="-O2 -fPIC ${SAN_CXX_FLAGS}" \
        LDFLAGS="${SAN_LD_FLAGS}" \
        ./configure \
            --enable-static \
            --disable-shared \
            --with-data-packaging=static \
            --disable-extras \
            --disable-icuio \
            --disable-layout \
            --disable-tests \
            --disable-samples \
            --enable-tools \
            --prefix="${DEPS_DIR}"
        make PKGDATA_MODE=static -j"$(nproc)"
        make install PKGDATA_MODE=static
    )
}

# Build all third-party deps. Each is guarded by a sentinel file so re-runs skip
# already-built libraries.
function build_deps() {
    mkdir -p "${DEPS_DIR}" "${SRC_DIR}"

    build_icu

    # gRPC brings Protobuf, Abseil, utf8_range, re2, c-ares and friends into the
    # same prefix via -DgRPC_INSTALL=ON. -DSKIP_INSTALL_ALL=ON stops its bundled
    # zlib from installing; -DgRPC_SSL_PROVIDER=package uses the system OpenSSL.
    if [ ! -f "${DEPS_DIR}/include/absl/base/options.h" ]; then
        LOG_INFO "Building gRPC ${GRPC_REF} (+ Protobuf/Abseil) into ${DEPS_DIR}"
        clone_tag "https://github.com/grpc/grpc" "${GRPC_REF}" "${SRC_DIR}/grpc"
        build_cmake_dep "${SRC_DIR}/grpc" "-DSKIP_INSTALL_ALL=ON -DgRPC_INSTALL=ON -DgRPC_BUILD_TESTS=OFF -DgRPC_SSL_PROVIDER=package"
    else
        LOG_INFO "gRPC/Protobuf/Abseil already present in ${DEPS_DIR}, skipping"
    fi

    if [ ! -f "${DEPS_DIR}/include/gtest/gtest.h" ]; then
        LOG_INFO "Building GoogleTest ${GTEST_REF}"
        clone_tag "https://github.com/google/googletest" "${GTEST_REF}" "${SRC_DIR}/googletest"
        build_cmake_dep "${SRC_DIR}/googletest" ""
    else
        LOG_INFO "GoogleTest already present, skipping"
    fi

    if [ ! -f "${DEPS_DIR}/include/benchmark/benchmark.h" ]; then
        LOG_INFO "Building google-benchmark ${BENCHMARK_REF}"
        clone_tag "https://github.com/google/benchmark" "${BENCHMARK_REF}" "${SRC_DIR}/benchmark"
        # BENCHMARK_ENABLE_WERROR=OFF: benchmark defaults to -Werror, and GCC 14's
        # <regex>/std_function headers trip a -Wmaybe-uninitialized false positive
        # in benchmark_register.cc (surfaces only under --asan, which alters inlining).
        build_cmake_dep "${SRC_DIR}/benchmark" "-DBENCHMARK_ENABLE_TESTING=OFF -DBENCHMARK_DOWNLOAD_DEPENDENCIES=OFF -DBENCHMARK_ENABLE_WERROR=OFF"
    else
        LOG_INFO "google-benchmark already present, skipping"
    fi

    # highwayhash has no CMake install rule; build then copy artifacts manually.
    if [ ! -f "${DEPS_DIR}/include/highwayhash/highwayhash.h" ]; then
        LOG_INFO "Building highwayhash ${HIGHWAYHASH_REF}"
        clone_tag "https://github.com/google/highwayhash" "${HIGHWAYHASH_REF}" "${SRC_DIR}/highwayhash"
        build_cmake_dep "${SRC_DIR}/highwayhash" "" ""
        mkdir -p "${DEPS_DIR}/lib" "${DEPS_DIR}/include"
        cp "${SRC_DIR}/highwayhash/.build-release/libhighwayhash.a" "${DEPS_DIR}/lib/"
        cp -fr "${SRC_DIR}/highwayhash/highwayhash" "${DEPS_DIR}/include/"
    else
        LOG_INFO "highwayhash already present, skipping"
    fi
}

function prepare_env() {
    install_system_deps
    select_compiler
    build_deps
}

function build_and_run() {
    export CMAKE_PREFIX_PATH="${DEPS_DIR}:${CMAKE_PREFIX_PATH}"
    export ICU_ROOT="${DEPS_DIR}"
    LOG_INFO "CC=${CC} CXX=${CXX}"
    LOG_INFO "CMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}"
    (cd ${ROOT_DIR} && ./build.sh ${BUILD_SH_ARGS})
}

cd ${CI_DIR}
prepare_env
build_and_run
```

What I tested could run with this:

- ci/build_amazonlinux.sh --run-tests --vendored
- ci/build_amazonlinux.sh --run-tests --vendored --asan 
- ci/build_amazonlinux.sh --run-integration-tests
- ci/build_amazonlinux.sh --run-integration-tests --asan


(unit tests need to be built with --vendored)

BCathcart:
#1268 will allow us to easily perf test this

mnunberg1:
Because the flags occupy some kind of mysterious dead zone. There's too
much logic in determining how and when they're applied, and much of this
existed in fragile and difficult-to-trace logic in the previous build
system. I would like to know which flags belong to which component and why
they're being used, but it seems nobody knows; "that's the way it's always
been" is the answer I received when I asked last.

I don't want to break things, but I also don't want to keep a complex
system in tact that nobody can even justify (though Yair has provided some
insight wrt HNSW's flags, which I will look into).

BCathcart:
The project has been around long enough now and had enough different contributors that I don't think it's surprising the motivation for each flag isn't readily known. The build system was a combination of presumably AI generation and then human refinement over time. I agree the foundation wasn't good so it's great it's being addressed, but I think the onus still needs to be on this PR to prove why it's okay to change the flags and not for others to justify why they shouldn't be changed. Based on the analysis below, it looks like it's mostly unchanged except for Snowball. I know @KarthikSubbarao manually tested and selected flags for snowball to gain a notable performance improvement, so I think that's motivation enough to not change those flags as well. I'm totally open to simplifying the flags, but I'd suggest this is better done in follow-ups after we have the automated perf regression testing up and running.

### Claude's per-component compile flags delta analysis

| Component | Delta? | Flags **added** | Flags **removed** |
|---|---|---|---|
| **Module `src/` + `vmsdk/src/`** | **YES** | — | `-Wthread-safety` (Clang) — moved to clang-tidy CI only |
| **snowball** | **YES — gains the full opt set** | SIMD (`-msse4.2 -mavx -mavx2 -maes -mfma -mpclmul -mcx16 -mprfchw`), `-ffast-math`, `-funroll-loops`, `-ftree-vectorize`, `-falign-functions=5`, `-fmath-errno`, `-ffp-contract=off`, `-fno-rounding-math`, `-mtune=generic`, `-gdwarf-5`, `-gz=zlib` | — (keeps its own `-O3 -w` + IPO) |
| **hnswlib** | **NONE** | — | — |
| **simsimd_c** | **NONE** | — | — |
| **hdrhistogram_c** | **NONE** | — | — |
| **generated protobuf/gRPC** | **NONE** | — | — |

pnbrown:
Hey @mnunberg1 — I tried building this branch on AL2023/aarch64 (Graviton) because the memory footprint on `main` sounds like a real barrier for contributors and I want to help get this over the line. I ran into a few issues that should be useful data for you.

## The problem this PR solves (with numbers)

I profiled a clean `main` build on a 32-core, 124 GB Graviton box:

- **Peak system-wide RAM: 80.4 GB** (sampled every 2s)
- **Wall clock: 9 min 1 sec**
- **CPU: 1933%**

That works out to roughly **2.5 GB per parallel job** (80.4 GB ÷ 32 cores). Anyone with less than ~85 GB of RAM gets OOM-killed at full parallelism, and even a contributor on a 16 GB machine would need to limit to `-j6` or so. This PR's direction would make a real difference for contributors.

## What I hit on the `build-fixes` branch (AL2023/aarch64, clean state)

I installed system deps (`grpc-devel`, `grpc-plugins`, `abseil-cpp-devel`, `protobuf-devel`, `libicu-devel`, `c-ares-devel`, `re2-devel`, `zlib-devel`) and then ran into a cascade of issues. In order:

1. **Protobuf config-mode CMake files don't exist on AL2023** — `protobuf-devel` only ships the legacy `FindProtobuf.cmake`, not `protobuf-config.cmake`. `protobuf_generate()` is unavailable. Had to build protobuf v25.6 from source (with `--recurse-submodules` for abseil).

2. **HighwayHash not packaged anywhere + broken on ARM** — upstream Makefile passes `-mavx2` unconditionally. Had to manually compile individual source files with `-DHH_DISABLE_TARGET_SPECIFIC`, skipping x86/test/macOS files.

3. **Google Benchmark not packaged on AL2023** — built v1.8.3 from source.

4. **Abseil version mismatch** — after a truly clean build (`git clean -fdx`), compilation fails because AL2023's `abseil-cpp-devel` is too old. The code uses `absl::SetGlobalVLogLevel` and `absl/base/no_destructor.h` which aren't in the system package.

5. **Double-define bug in `build.sh`** — `-DTESTING_TMP_DISABLED` is passed inside `CMAKE_CXX_FLAGS`, but CMake prepends another `-D`, producing `-D-DTESTING_TMP_DISABLED` which fails with `error: macro names must be identifiers`. Every compile unit hits this.

An earlier run (before `git clean -fdx`) appeared to succeed, but it was accidentally picking up vendored abseil/gRPC headers left over from a previous `main` build in `.build-release/.deps/`. From a truly clean state, it does not build. More concerning was that it still consumed similar amounts of memory in the linking stage.

## Suggestion: early dependency validation

One thing that would help contributors a lot is failing fast with actionable messages before the build starts. Something like:

```cmake
# Early in CMakeLists.txt, before doing any real work:
include(CheckDependencies)  # or inline

find_package(Protobuf CONFIG QUIET)
if(NOT Protobuf_FOUND)
  message(FATAL_ERROR
    "Protobuf (config-mode) not found.\n"
    "On AL2023/Fedora: must build from source (system package lacks CMake config files).\n"
    "On Ubuntu 22.04+: sudo apt install libprotobuf-dev protobuf-compiler\n"
    "Or set Protobuf_DIR to your install prefix.")
endif()

find_program(GRPC_CPP_PLUGIN grpc_cpp_plugin)
if(NOT GRPC_CPP_PLUGIN)
  message(FATAL_ERROR
    "grpc_cpp_plugin not found.\n"
    "On AL2023: sudo dnf install grpc-plugins\n"
    "On Ubuntu: sudo apt install protobuf-compiler-grpc")
endif()
```

This way contributors get a clear error with the fix on the first try instead of discovering dependencies one-at-a-time across 8 sequential cmake failures.

## Summary

|  | `main` | `build-fixes` |
|---|---|---|
| System packages | ~5 | ~12+ plus 3 from source |
| Peak RAM | **80.4 GB** | Does not build from clean state |
| Wall clock | 9m01s | N/A |
| Blockers hit | 2 (easy) | 8+ |

I'm happy to help test fixes, write a dependency install script, or put together a `BUILDING.md` for AL2023/aarch64. Let me know how I can be useful.

BCathcart:
#918 shows a current Fedora build problem fyi

PR #1226, opened 2026-07-06
Handle kOneReplicaPerShard in GetTargetsForSlot for single-slot indexes

KarthikSubbarao:
Single-slot indexes call GetTargetsForSlot which did not have a case for kOneReplicaPerShard. When ForceReplicasOnly is enabled (or in the future when we want the client mode based routing. e.g. READONLY routing to replicas), this mode is selected and the missing case causes CHECK(false) → SIGABRT, crashing the node.

Add the kOneReplicaPerShard case to GetTargetsForSlot: pick a random replica from the owning shard, preferring local if under low utilization — mirroring what GetTargets (multi-slot) already does.

### DIFF
--- a/integration/test_singleslot.py
+++ b/integration/test_singleslot.py
@@ -88,3 +88,60 @@ class TestSingleSlot(ValkeySearchClusterTestCaseDebugMode):
             assert(requests() == expected_requests)
             assert(rpcs() == expected_rpcs)
             assert result[0] == 10  # number of results
+
+    @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


yairgott (approved):

baswanth09:
Can we backport this to 1.2?

Aksha1812:
> Can we backport this to 1.2?

@baswanth09 you can ask any maintainer to mark this to be backported in 1.2 project board for auto backport or since this is just one small change you can manually backport too

Issue #1216, opened 2026-07-02
Valkey Search 1.2.1 patch release checklist


Aksha1812:
#### Prerequisite
- [x] 1. Make a list of all the commits (especially the ones with label `bug`) there were added to the main branch after last patch release for this branch (e.g. https://github.com/valkey-io/valkey-search/compare/1.2...main) and audit which changes need to be backported for this patch release. Run this by a valkey-search maintainer.
- [x] 2. Make a list of all the open issues and open PRs that need to be closed and added to this patch (especially the ones with label `bug`, and which are part of the project board , e.g https://github.com/orgs/valkey-io/projects/58/views/1) and communicate a code cutoff date in the OSS weekly meeting.
- [x] 3. Whichever change needs to be backported, ask a maintainer to add those items to respective release branch's project board and mark the status "To be backported" so that valkey-ci-agent can generate auto backport PRs. To understand how to the backporting agent works see `BACKPORTING.md` https://github.com/valkey-io/valkey-search/blob/main/BACKPORTING.md. You will need to keep an eye on backport PRs opened by the agent and review them and ask a mainatiner to merge them e.g https://github.com/valkey-io/valkey-search/pull/1212. (For merging backport PRs you can ask the maintainer to `Set DCO to pass` to fix failing DCO check as agent does not signoff) ( Also please don't squash and merge backport PRs) 
- [x] 4. Once all changes backported. Verify we didn't introduce a breaking change in the patch release. e.g . new functionality, new info/configs, changes that affect memory profile , RDB changes . etc...

#### Release Checklist (MAINTAINER ONLY)
- [x] 1. Confirm all fixes for this patch are merged and backported onto the release branch (check the project board).
- [x] 2. Move released PRs from "To be backported" to "Done" on the project board.
- [x] 3. Submit a PR to the release branch updating `src/version.h` and the release notes / `00-RELEASENOTES`: (This can be done by a non maintainer)
   - [x] set `kModuleVersion` to the new version (e.g. 1.2.0 -> 1.2.1)
   - [x] set `MODULE_RELEASE_STAGE` to `ga`
   - [x] if this patch adds a new metadata encoding (rare), add the corresponding `kReleaseXY` constant
   - [x] confirm `kMinimumServerVersion` is still correct
   - [x] (1.0 branch only) version is inline in `src/module_loader.cc` (`MODULE_VERSION`), not `version.h`
   - [x] Add patch release notes to `00-RELEASENOTES`
[...]

PR #1222, opened 2026-07-03
The query vector is not normalized on the prefiltering query execution path

yairgott:
Fixing issue #1221

PR #1227, opened 2026-07-06
Bump version to 1.2.1 and add release notes

Aksha1812:
## Summary

Release step 3 of the 1.2.1 patch checklist (#1216): bump the module version and add release notes on the `1.2` branch.

- `src/version.h`: `kModuleVersion` 1.2.0 → 1.2.1
- `00-RELEASENOTES`: add the `1.2.1 GA` section

`MODULE_RELEASE_STAGE` is already `"ga"` and `kMinimumServerVersion` remains `9.0.1` (correct). No new metadata encoding is introduced in this patch, so no `kReleaseXY` constant was added.

### DIFF
--- a/00-RELEASENOTES
+++ b/00-RELEASENOTES
@@ -48,10 +48,6 @@ Optimizations
 * Replaced a deque with a flat hash map for tracked_vectors (#1015)
 * Added memory reservation based on the mutation queue (#934)
 
-Contributors
-============
-@Aksha1812, @allenss-amazon, @KarthikSubbarao, @boda26, @baswanth09, @bandalgomsu, @Frank-Gu-81, @hjkwon0, @smkher, @BCathcart, @Nivesh-01, @daddaman-amz, @yairgott, @linbran5123
-
 ================================================================================
 Valkey Search 1.2.0 GA - Released Tue 17 March 2026
 ================================================================================
@@ -140,3 +136,7 @@ Major New Features / Updates
 We appreciate the efforts of all who contributed code to this release!
 
 @Aksha1812, @AlexFilipImproving, @allenss-amazon, @baswanth09, @BCathcart, @daddaman-amz, @eifrah-aws, @eliastam, @eric-musliner, @murphyjacob4, @KarthikSubbarao, @otherscase, @boda26, @Arunsarma07, @Nivesh-01, @VoletiRam, @yairgott, @yanamolo, @zvi-code, @rlunar, @bandalgomsu
+
+We appreciate the efforts of all who contributed code to the 1.2.1 release!
+
+@Aksha1812, @allenss-amazon, @KarthikSubbarao, @boda26, @baswanth09, @bandalgomsu, @Frank-Gu-81, @hjkwon0, @smkher, @BCathcart, @Nivesh-01, @daddaman-amz, @yairgott, @linbran5123
--- a/00-RELEASENOTES
+++ b/00-RELEASENOTES
@@ -34,6 +34,7 @@ Bug fixes
 * Corrected memory and INFO reporting: accurate memory reporting when reader/writer thread counts change at runtime, a reclaimable_memory underflow, and coordinator bytes-in counting before response mutation (#1119, #1131, #1046)
 * search_number_of_indexes is now updated correctly during restore, and aggregate recursion depth is handled correctly (#974, #973)
 * Included a set of thread-sanitizer (TSAN) fixes (#860)
+* Improvements to tests and test infrastructure with no corresponding module code changes.
 
 Optimizations
 =============
@@ -46,10 +47,6 @@ Optimizations
 * Optimized tag search to avoid relying on the root iterator in common cases (#994)
 * Replaced a deque with a flat hash map for tracked_vectors (#1015)
 * Added memory reservation based on the mutation queue (#934)
-
-Tests, CI, and build
-====================
-* Improvements to tests and test infrastructure with no corresponding module code changes.
 
 Contributors
 ============
--- a/00-RELEASENOTES
+++ b/00-RELEASENOTES
@@ -53,7 +53,7 @@ Tests, CI, and build
 
 Contributors
 ============
-@Aksha1812, @allenss-amazon, @KarthikSubbarao, @boda26, @baswanth09, @bandalgomsu, @Frank-Gu-81, @hjkwon0, @smkher, @BCathcart, @Nivesh-01, @daddaman-amz
+@Aksha1812, @allenss-amazon, @KarthikSubbarao, @boda26, @baswanth09, @bandalgomsu, @Frank-Gu-81, @hjkwon0, @smkher, @BCathcart, @Nivesh-01, @daddaman-amz, @yairgott, @linbran5123
 
 ================================================================================
 Valkey Search 1.2.0 GA - Released Tue 17 March 2026
--- a/00-RELEASENOTES
+++ b/00-RELEASENOTES
@@ -133,6 +133,8 @@ Major New Features / Updates
 * Replica nodes no longer participate in metadata consistency protocol. All mutation information comes from the Primary
 * Enhanced FT.INFO command with additional information in Cluster mode
 
-We appreciate the efforts of all who contributed code to this release!
+================================================================================
+Contributors
+================================================================================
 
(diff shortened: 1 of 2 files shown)


Aksha1812:
Added maintainers for review. PLEASE DO NOT MERGE THIS YET . still #1220 to be merged  and #1222 to be backported.

00-RELEASENOTES
```diff
@@ -1,6 +1,60 @@
...
+* Space savings for tag and numeric indexes (#1026)
+* Reduced reference counting in text iterators with short-circuited field-containment checks, backed by borrowed interned-string pointers (#1044, #1043)
+* Optimized content fetch to eliminate a double key open and use HashGet for specific field access (#1045)
+* Removed a reader-thread scaling bottleneck caused by the PausePoint mutex (#1037)
+* Used a protobuf arena for fanout response deserialization (#1040)
+* Optimized tag search to avoid relying on the root iterator in common cases (#994)
+* Replaced a deque with a flat hash map for tracked_vectors (#1015)
+* Added memory reservation based on the mutation queue (#934)
+
+Tests, CI, and build
```
BCathcart: Compared to past release notes this is quite verbose, which overall I don't mind. I like the calling out of the specific PRs in the Bug Fixes and Optimzations sections. I would vote against including this section though since it's related to development processes and not directly relevant for customers using the product.

We should also consider adding a "release-notes" label like the valkey repo has to tag specific PRs to be included in the release notes.

Aksha1812: > We should also consider adding a "release-notes" label like the valkey repo has to tag specific PRs to be included in the release notes.

yes, there is some work in the valkey-ci-agent, which would automate the release notes generation if we add release-notes label to the PRs included in the release. 

> Compared to past release notes this is quite verbose, which overall I don't mind. I like the calling out of the specific PRs in the Bug Fixes and Optimzations sections. I would vote against including this section though since it's related to development processes and not directly relevant for customers using the product.

Yeah, it is a little verbose but we have merged many fixes / optimizations after 1.2 and backported them . I have neglected very small changes and grouped relevant changes together but wasn't able to make it more concise. Sure if everyone agrees , i can remove that section .

00-RELEASENOTES
```diff
@@ -1,6 +1,60 @@
...
+* Fixed the query vector not being normalized on the prefiltering execution path for cosine-distance vector indexes (#1222)
+* Tag queries now handle an escaped closing brace correctly (#1148)
+* JSON string values containing backslash escapes are now decoded before indexing rather than stored raw (#1174)
+* Numeric range searches now accept values in scientific notation (#937)
+* Field serialization now uses a fixed ordering, which matters for text indexes (#1003)
+* Corrected memory and INFO reporting: accurate memory reporting when reader/writer thread counts change at runtime, a reclaimable_memory underflow, and coordinator bytes-in counting before response mutation (#1119, #1131, #1046)
+* search_number_of_indexes is now updated correctly during restore, and aggregate recursion depth is handled correctly (#974, #973)
+* Included a set of thread-sanitizer (TSAN) fixes (#860)
+
+Optimizations
```
allenss-amazon, line 39: How is "Optimizations" not an "Improvement"? We either need to remove "Improvements" from the first section or skip having sections.

I'd vote to avoid classifying the PRs into sections.

00-RELEASENOTES
```diff
@@ -1,6 +1,60 @@
...
+* Optimized content fetch to eliminate a double key open and use HashGet for specific field access (#1045)
+* Removed a reader-thread scaling bottleneck caused by the PausePoint mutex (#1037)
+* Used a protobuf arena for fanout response deserialization (#1040)
+* Optimized tag search to avoid relying on the root iterator in common cases (#994)
+* Replaced a deque with a flat hash map for tracked_vectors (#1015)
+* Added memory reservation based on the mutation queue (#934)
+
+Tests, CI, and build
+====================
+* A round of test, CI, and build stabilization: deflaked several ASAN and integration tests, suppressed spurious core dumps in death tests, fixed a use-after-free in a drop-index test, added macOS test-runner compatibility and ASAN build-script fixes, corrected an ASAN module config type mismatch, and enabled CI to run across all release branches (#868, #863, #1067, #950, #933, #929, #961, #1014, #777, #1027, #1007, #943, #1058, #1130, #942)
```
allenss-amazon: I think less said here, the better. Maybe just say something like: "Improvements to tests and test infrastructure with no corresponding module code changes", or something like that....

00-RELEASENOTES
Aksha1812: I saw previous valkey-search release notes https://github.com/valkey-io/valkey-search/blob/1.2/00-RELEASENOTES and core release notes having similar sections . I am going to remove "Improvements" for now .

00-RELEASENOTES
```diff
@@ -1,6 +1,53 @@
 Valkey Search 1.2 release notes
 ===============================
 
+================================================================================
+Valkey Search 1.2.1 GA - Released Mon 06 July 2026
```
allenss-amazon: Need to fix the date ;-)

allenss-amazon (approved):

BCathcart (approved):

00-RELEASENOTES
```diff
@@ -1,6 +1,53 @@
...
+* Field serialization now uses a fixed ordering, which matters for text indexes (#1003)
+* Corrected memory and INFO reporting: accurate memory reporting when reader/writer thread counts change at runtime, a reclaimable_memory underflow, and coordinator bytes-in counting before response mutation (#1119, #1131, #1046)
+* search_number_of_indexes is now updated correctly during restore, and aggregate recursion depth is handled correctly (#974, #973)
+* Included a set of thread-sanitizer (TSAN) fixes (#860)
+* Improvements to tests and test infrastructure with no corresponding module code changes.
+
+Optimizations
+=============
+* Vector search performance improvement (#1163)
+* Space savings for tag and numeric indexes (#1026)
```
KarthikSubbarao, line 42: The only question I had was about space savings whether a rolling patch upgrade can pose any issue. If the primary is on a newer patch (1.2.1) and replica is on an older patch (1.2.0), then the primary can accept more writes which will fail on a replica node if the replica is close to max-memory and primary is not.

But validated offline (and discussed) that this is not fatal since replicas will not reject due to maxmemory 
```
replica-ignore-maxmemory yes
```

00-RELEASENOTES
KarthikSubbarao: July 7

KarthikSubbarao (approved):

00-RELEASENOTES
```diff
@@ -1,6 +1,53 @@
...
 
+================================================================================
+Valkey Search 1.2.1 GA - Released Mon 06 July 2026
+================================================================================
+
+A patch release on top of 1.2.0 focused on stability, correctness, and
+efficiency. It hardens command and load paths against crashes found through
+fuzzing, fixes concurrency and shutdown issues, tightens cluster and query
+correctness, and lands vector, tag, numeric, and text index optimizations.
+There are no breaking changes. Like the 1.2 release, this version requires
```
KarthikSubbarao: > There are no breaking changes.

This should be implied since it is a patch

00-RELEASENOTES
Aksha1812: Yes , discussed this with @allenss-amazon . We don't think this should cause any issues.

Aksha1812:
there is a flaky test https://github.com/valkey-io/valkey-search/actions/runs/28889085369/job/85696764993?pr=1227 
```
E                   valkeytestframework.util.waiters.WaitTimeout: Expected value to be: True, actual value is: False

../../../integration/valkeytestframework/util/waiters.py:80: WaitTimeout
---------------------------- Captured log teardown -----------------------------
WARNING  root:valkey_test_case.py:109 SHUTDOWN was unsuccessful
=========================== short test summary info ============================
FAILED ../../../integration/test_fulltext_inflight_blocking.py::TestFullTextInFlightBlockingCMD::test_fulltext_inflight_blocking_with_pausepoint - valkeytestframework.util.waiters.WaitTimeout: Expected value to be: True, a...
============ 1 failed, 221 passed, 16 skipped in 684.71s (0:11:24) =============
```

this has been failing on main as well . I don't believe we have a fix . do we want to skip this test for now ?

I reran the workflow. it usually passes on the rerun 

opened a an issue https://github.com/valkey-io/valkey-search/issues/1229 to address it

PR #1230, opened 2026-07-07
Fix SIGSEGV in LatencySampler when hdr_init fails

VoletiRam:
`LatencySampler::SubmitSample()` calls `hdr_init()` to lazily allocate a histogram, but never checks the return value. If `hdr_init()` fails (e.g., `calloc` returns NULL under memory pressure), `histogram_` remains NULL. The code then sets `initialized_ = true` and calls `hdr_record_value(NULL, ...)`, which dereferences the NULL pointer and crashes with SIGSEGV. 

Fix:
Check `hdr_init()` return value. If it fails, return early (drop the sample). The histogram will be retried on the next call since `initialized_` remains false.


vmsdk/src/latency_sampler.h
```diff
@@ -54,7 +54,9 @@ class LatencySampler {
   void SubmitSample(absl::Duration latency) {
     absl::MutexLock lock(&histogram_lock_);
     if (!initialized_) {
-      hdr_init(min_value_, max_value_, precision_, &histogram_);
+      if (hdr_init(min_value_, max_value_, precision_, &histogram_) != 0) {
+        return;  // Drop sample on allocation failure (OOM or invalid params)
```
KarthikSubbarao: We could use the LOG_EVERY_N_SECONDS fn here to know this sample failed

Aksha1812 (changes requested):
PR description says if hdr_init() fails . `histogram_` remains NULL . I see `histogram_`  https://github.com/valkey-io/valkey-search/blob/a9a8b9ad2183e59c33b76dd6dd4f210990f42506/vmsdk/src/latency_sampler.h#L92 is not initialized to be null . Can we update that to something like? 

```
hdr_histogram *histogram_ ABSL_GUARDED_BY(histogram_lock_) = nullptr;
```

AI says it is not initialized to nullptr by default

VoletiRam:
Can you please add the PR to the respective project boards to be backported? @allenss-amazon

Aksha1812 (approved):

vmsdk/src/latency_sampler.h
```diff
@@ -54,7 +55,11 @@ class LatencySampler {
   void SubmitSample(absl::Duration latency) {
     absl::MutexLock lock(&histogram_lock_);
     if (!initialized_) {
-      hdr_init(min_value_, max_value_, precision_, &histogram_);
+      if (hdr_init(min_value_, max_value_, precision_, &histogram_) != 0) {
```
yairgott, line 58: my concern is that a `hdr_init` failure may result in successful internal resource allocation for only a subset of the resources. Ideally, we should ensure that `hdr_init` frees all the allocated internal resources  before returning with an error.

Aksha1812: @VoletiRam can you respond to above and help close this ? thanks

VoletiRam: The fix is a one-line free(counts) in third_party/hdrhistogram_c before the second return ENOMEM. Just want to check on the right approach before I revise. Would you prefer we patch our local copy of hdrhistogram_c directly, or file the fix with the library maintainers first? @yairgott

VoletiRam: Fixed it inline at hdr_histogram.c by freeing the counts buffer before the second return ENOMEM, so the three failure paths are all leak-free. @yairgott

yairgott (approved):

Aksha1812:
Can we add this to 1.2 project board @allenss-amazon , @BCathcart and mark it "To be backported" . we have seen this crash in 1.2 .

PR #1232, opened 2026-07-08
Changing HNSWLib to store raw vectors

yairgott:
This change refactors the HNSW and Flat vector indexes to store original (denormalized) vectors, while tracking and utilizing their magnitudes during distance evaluation. Storing the raw vectors is required a followup PR to utilize the engine memory sharing capabilities.

Key Changes:
- Modified the space functions to support Cosine distance calculations efficiently.


testing/vector_externalizer_test.cc
```diff
@@ -30,7 +30,7 @@
 
 namespace valkey_search {
 // Vector externalizer is not used and planned to be implemented differently
-#ifndef SAN_BUILD
+#if defined(ENABLE_VECTOR_EXTERNALIZER_TEST)
```
allenss-amazon, line 33: claude says this isn't enabled in a non-SAN build. It should be -- right?

third_party/hnswlib/bruteforce.h
allenss-amazon, line 34: This is what confuses me about being based on 1188, doesn't it have the same reformatting??

allenss-amazon, line 71: else CHECK(cur_element_count_ == 0) ?

allenss-amazon, line 127: why both a legacy cast and a reinterpret_cast? Makes this hard to read.

allenss-amazon, line 131: Shouldn't this be a std::move and then you can skip the last destroy_at?

third_party/hnswlib/hnswalg.h
allenss-amazon, line 881: can't delete load validation.

allenss-amazon, line 928: need to restore load validation

allenss-amazon, line 955: load validation.

allenss-amazon, line 972: load validation.

allenss-amazon, line 1015: load validation

third_party/hnswlib/bruteforce.h
allenss-amazon, line 58: I'd be much happier if this had a name that didn't overload the same function in the HNSW file. That can be a subtle source of errors. If we rename it, then there's no chance.

allenss-amazon (commented):
1. The load validation code and associated unit tests have to be restored at some point. Given the fixup strategy associated with the late fixup of the VectorRecord shared pointers, there will also need to be a validation path that ensures 100% of the nullptr's are also fixed up. 
2. There doesn't appear to a test that deletes from a cosine index. The only delete tests use L2 which so the tombstone code doesn't get exercised.
3. Testing in this area -- HNSW inser/delete (w &w/o overwrite-delete)  & query plus save/restore across different distance metrics that validate recall -- has always been light, which was based on assumption that HNSWalg was correct. That's no longer true, we need need to fully populate this area. We need a set of tests for this.

third_party/hnswlib/bruteforce.h
yairgott: Yeah, I might just make sense to maintain this function in their base class.

yairgott: Makes sense

yairgott: Right, it might make sense to maintain a function to get the label by internal id.

yairgott: Destroy at can't be avoided as any C++ object is needed to be destructed regardless if it was moved or not.

testing/vector_externalizer_test.cc
yairgott: This file is removed in the following PR.

third_party/hnswlib/bruteforce.h
yairgott: I'll address this in the hnsw refactoring task.

src/indexes/vector_base.cc
```diff
@@ -467,8 +471,23 @@ absl::Status VectorBase::LoadTrackedKeys(
...
-                      attribute_identifier_);
+
+    auto key = vmsdk::MakeUniqueValkeyString(interned_key->Str());
+    auto key_obj = vmsdk::MakeUniqueValkeyOpenKey(
+        ctx, key.get(), VALKEYMODULE_OPEN_KEY_NOEFFECTS | VALKEYMODULE_READ);
+    CHECK(key_obj) << "Failed to open key during LoadTrackedKeys: "
+                   << interned_key->Str();
+    auto record = attribute_data_type->GetRecord(
+        ctx, key_obj.get(), interned_key->Str(), attribute_identifier_);
+    CHECK(record.ok());
```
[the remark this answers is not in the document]
yairgott: It's by design as it's not legal to miss data.

src/indexes/vector_hnsw.cc
```diff
@@ -122,21 +98,30 @@ absl::StatusOr<std::shared_ptr<VectorHNSW<T>>> VectorHNSW<T>::LoadFromRDB(
...
     RDBChunkInputStream input(std::move(iter));
+    index->algo_->normalized_ = index->normalize_;
+
     auto generator = [allocator = index->GetVectorAllocator()](
-                         absl::string_view vector_data) {
-      return VectorRecord(StringInternStore::Intern(vector_data, allocator));
+                         absl::string_view vector_data,
+                         bool is_marked_deleted) {
+      if (!is_marked_deleted) {
+        return std::shared_ptr<VectorRecord>(nullptr);
```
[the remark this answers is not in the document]
yairgott: The correct value is set during `VectorBase::LoadTrackedKeys` before HNSW distance evaluation.

.devcontainer/socket_proxy.py
```diff
@@ -56,8 +57,8 @@ def main():
             pass
             
     server.bind(listen_path)
-    # Restrict access to the owner; the container should run with a matching UID.
-    os.chmod(listen_path, 0o600)
+    # Allow read/write access to the socket for the container user
```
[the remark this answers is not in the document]
yairgott: I believe this is not relevant anymore.

yairgott:
> 1. The load validation code and associated unit tests have to be restored at some point. Given the fixup strategy associated with the late fixup of the VectorRecord shared pointers, there will also need to be a validation path that ensures 100% of the nullptr's are also fixed up.
> 2. There doesn't appear to a test that deletes from a cosine index. The only delete tests use L2 which so the tombstone code doesn't get exercised.
> 3. Testing in this area -- HNSW inser/delete (w &w/o overwrite-delete)  & query plus save/restore across different distance metrics that validate recall -- has always been light, which was based on assumption that HNSWalg was correct. That's no longer true, we need need to fully populate this area. We need a set of tests for this.

I ported the index validation changes from the latter PR.

src/indexes/vector_base.cc
```diff
@@ -282,24 +296,19 @@ absl::StatusOr<std::vector<Neighbor>> VectorBase::CreateReply(
...
-  if (normalize_) {
-    if (it->second.magnitude < 0) {
-      return absl::InternalError("Magnitude is not initialized");
-    }
-    result = DenormalizeVector(absl::string_view(value, GetVectorDataSize()),
-                               GetDataTypeSize(), it->second.magnitude);
-  } else {
-    result.assign(value, value + GetVectorDataSize());
+  auto &vector_record = GetVectorLockFree(it->second.internal_id);
+  if (!vector_record) {
```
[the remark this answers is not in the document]
yairgott: the multi-read-write-mutex ensures mutual exclusion between reads/search and index mutation.

Aksha1812:
**Reviewers for this PR**

src/indexes/vector_base.h
```diff
@@ -252,18 +273,20 @@ class VectorBase : public IndexBase {
                          const AttributeDataType *attribute_data_type,
                          absl::string_view key_cstr,
                          absl::string_view attribute_identifier);
-  virtual const char *GetValueImpl(uint64_t internal_id) const = 0;
+  virtual std::shared_ptr<VectorRecord> &GetVectorLockFree(
+      uint64_t internal_id) const = 0;
```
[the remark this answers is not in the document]
yairgott: This is fixed in the following PR of the branch memory_sharing

allenss-amazon (approved):

PR #1233, opened 2026-07-08
introduce VectorRegistry for memory sharing

yairgott:
This commit enables memory sharing between the module and the engine to reduce overall memory footprint.

Changes:
- Implement VectorRegistry to maintain a single copy of the vector payload within the module and between the module and the engine.
- Modify the load index flow to ignore the serialized hnswlib RDB during load keys, avoiding OOMs by discarding the RDB. Instead, once the mapping to the keys data is loaded, by VectorBase::LoadTrackedKeys, the vectors are loaded from the engine, shared with the engine and stored in hnswlib.


src/indexes/vector_base.cc
```diff
@@ -308,16 +233,38 @@ absl::StatusOr<std::vector<char>> VectorBase::GetVectorDuringSearch(
 }
 
 absl::StatusOr<bool> VectorBase::RemoveRecord(
-    const InternedStringPtr &key,
-    [[maybe_unused]] indexes::DeletionType deletion_type) {
+    const InternedStringPtr &key, indexes::DeletionType deletion_type) {
   VMSDK_ASSIGN_OR_RETURN(auto res, UnTrackKey(key));
+  VectorRegistry::Instance().Untrack(key, attribute_identifier_);
   if (!res.has_value()) {
```
allenss-amazon, line 286: Why would UnTrackKey fail?

src/indexes/vector_base.h
```diff
@@ -87,6 +87,36 @@ T CalcMagnitude(const T *src, size_t size) {
              : static_cast<T>(std::sqrt(magnitude));
 }
 
+template <typename T>
+std::vector<char> NormalizeVector(absl::string_view record,
+                                  T reciprocal_magnitude) {
```
allenss-amazon: let's make the type of the reciprocal_magnitude be not the same as the underlying vector type. I suspect we will end up storing the magntinude in FP32 regardless of the format of the vector itself.

allenss-amazon: Also, simsimD provide this function(scalar times vector)

allenss-amazon: Same for inverse Magnitude computation

src/indexes/vector_base.h
```diff
@@ -167,8 +197,18 @@ absl::string_view LookupKeyByValue(
   }
 }
 
+struct TrackedKeyMetadata {
+  uint64_t internal_id;
+  // If normalize_ is false, this will be 1.0f. Otherwise, it will be the
+  // magnitude of the vector. If the magnitude is not initialized, it will
+  // be -inf (this is an intermediate state during backfill when
```
allenss-amazon: Let's use a const for the uninitialized value, not -inf

src/indexes/vector_flat.cc
```diff
@@ -85,18 +85,19 @@ absl::StatusOr<std::shared_ptr<VectorFlat<T>>> VectorFlat<T>::LoadFromRDB(
     absl::string_view attribute_identifier,
     SupplementalContentChunkIter &&iter) {
   try {
-    auto index = std::shared_ptr<VectorFlat<T>>(new VectorFlat<T>(
-        vector_index_proto.dimension_count(),
-        vector_index_proto.distance_metric(),
-        vector_index_proto.flat_algorithm().block_size(), attribute_identifier,
-        attribute_data_type->ToProto()));
+    auto index = std::shared_ptr<VectorFlat<T>>(
```
allenss-amazon, line 89: make_shared

src/indexes/vector_hnsw.cc
```diff
@@ -92,9 +92,10 @@ absl::StatusOr<std::shared_ptr<VectorHNSW<T>>> VectorHNSW<T>::LoadFromRDB(
     absl::string_view attribute_identifier,
     SupplementalContentChunkIter &&iter) {
   try {
-    auto index = std::shared_ptr<VectorHNSW<T>>(new VectorHNSW<T>(
-        vector_index_proto.dimension_count(), attribute_identifier,
-        attribute_data_type->ToProto()));
+    auto index = std::shared_ptr<VectorHNSW<T>>(
```
allenss-amazon, line 97: make_shared

src/vector_registry.cc
```diff
@@ -0,0 +1,282 @@
...
+       VALKEYMODULE_OK) &&
+      options::GetEnableVectorSharing().GetValue();
+  ctx_ = vmsdk::MakeUniqueValkeyDetachedThreadSafeContext(ctx);
+}
+
+std::shared_ptr<VectorRecord> VectorRegistry::DedupConstruct(
+    const InternedStringPtr &key, absl::string_view attribute_identifier,
+    absl::string_view vector, float magnitude, Allocator *allocator) {
+  RegistryKey search_key =
+      std::make_pair(key, std::string(attribute_identifier));
```
allenss-amazon: Why not intern the attribute_identifier? Seems like it would be a big space savings, one per vector, right?

Actually, seems like you could use the IndexBase pointer itself, right?

src/vector_registry.cc
```diff
@@ -0,0 +1,282 @@
...
+  }
+  RegistryKey search_key =
+      std::make_pair(key, std::string(attribute_identifier));
+  bool schedule_run_by_main = false;
+  {
+    absl::MutexLock lock(&mutex_);
+    auto it = tracked_vectors_.find(search_key);
+    if ((it != tracked_vectors_.end() && it->second.shared_externally &&
+         it->second.shared_externally.get() == vector_record.get())) {
+      // Already shared externally or pending to be shared externally, no need
```
allenss-amazon: comment incorrect? pending shared check is the next section, right?

allenss-amazon (commented):
I'm publishing these comments so far. Per our discussion, I stopped reviewing at the Registry and will pick up once it's revised.

src/indexes/vector_base.cc
```diff
@@ -151,54 +119,34 @@ void VectorBase::Init(int dimensions,
...
-  if (!IsValidSizeVector(record)) {
-    return {};
+std::shared_ptr<const VectorRecord> VectorBase::GetOrConstructVectorRecord(
+    const InternedStringPtr &key, absl::string_view record,
+    float magnitude) const {
+  auto [vector_record, vector_record_size] =
+      VectorRegistry::Instance().LookupRecord(key,
+                                              interned_attribute_identifier_);
+  if (vector_record && vector_record_size == record.size() &&
+      std::memcmp(vector_record->GetRawVector(), record.data(),
```
allenss-amazon, line 163: Can't you just compare the raw pointer, not the contents?

src/vector_registry.cc
```diff
@@ -0,0 +1,198 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
```
allenss-amazon: 2026

src/vector_registry.cc
```diff
@@ -0,0 +1,198 @@
...
+    if (!vector) {
+      tracked_vectors_.erase(search_key);
+      return nullptr;
+    }
+    auto vector_str = vmsdk::ToStringView(vector);
+    auto it = tracked_vectors_.find(search_key);
+    if (it != tracked_vectors_.end()) {
+      if (it->second.vector_record_size == vector_str.size() &&
+          std::memcmp(it->second.vector_record->GetRawVector(),
+                      vector_str.data(), vector_str.size()) == 0) {
```
allenss-amazon: WHy the full compare? Won't it be sufficient to compare the addresses?

Actually, claude pointed out that if the application does an HSET on a vector field with _exactly_ the same bytes, then we'll do the wrong thing. Because we'll think we already have the vector shared because we're doing a full value compare here -- not an actual test to see if the vector was already shared. Thus this test needs to be an address compare. We can certainly do a full check of the value too -- but if that check fails then we should CHECK because we have memory corruption....

src/indexes/vector_base.cc
```diff
@@ -151,54 +119,34 @@ void VectorBase::Init(int dimensions,
...
-  } else {
-    magnitude = CalcMagnitude(reinterpret_cast<const float *>(record.data()),
-                              dimensions_);
-  }
-  auto vector_record =
-      VectorRecord::Construct(record, magnitude, vector_allocator_.get());
+  float magnitude = CalcMagnitude(
+      reinterpret_cast<const float *>(record.data()), dimensions_);
+
+  auto vector_record = GetOrConstructVectorRecord(key, record, magnitude);
```
allenss-amazon: This leads to a redundant computation of magnitude when the vector is already registered. Can't we push the magnitude computation down into this routine so that it's skipped when the vector is already registered?

src/indexes/vector_base.cc
```diff
@@ -234,41 +182,25 @@ absl::StatusOr<InternedStringPtr> VectorBase::GetKeyDuringSearch(
...
-  } else {
-    magnitude = CalcMagnitude(reinterpret_cast<const float *>(record.data()),
-                              dimensions_);
-  }
-  auto vector_record =
-      VectorRecord::Construct(record, magnitude, vector_allocator_.get());
+  float magnitude = CalcMagnitude(
+      reinterpret_cast<const float *>(record.data()), dimensions_);
+  auto vector_record = GetOrConstructVectorRecord(key, record, magnitude);
   VMSDK_ASSIGN_OR_RETURN(auto internal_id, GetInternalId(key));
```
allenss-amazon, line 229: same here, let's move the magnitude down into the routine to avoid recomputing it when it's already registered.

src/indexes/vector_base.cc
```diff
@@ -318,6 +249,29 @@ absl::StatusOr<bool> VectorBase::RemoveRecord(
...
+void VectorBase::RemoveRecordDueToError(const InternedStringPtr &key,
+                                        std::optional<uint64_t> internal_id) {
+  auto res = UnTrackKey(key);
+  if (!res.ok()) {
+    VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+        << "While processing error, failed to untrack the key "
+           "with id: "
+        << (internal_id.has_value() ? std::to_string(internal_id.value())
+                                    : "unknown")
+        << ": " << res.status().message();
```
allenss-amazon, line 302: In the case where the incoming vector length is wrong, you're going to log that as an error, which I think is wrong. 

Regardless, this should have an INFO counter.

src/indexes/vector_base.cc
```diff
@@ -318,6 +249,29 @@ absl::StatusOr<bool> VectorBase::RemoveRecord(
...
+  }
+  if (internal_id.has_value()) {
+    auto remove_vector_res = RemoveRecordImpl(internal_id.value());
+    if (!remove_vector_res.ok()) {
+      VMSDK_LOG_EVERY_N_SEC(WARNING, nullptr, 1)
+          << "While processing error, failed to remove vector with id: "
+          << internal_id.value() << ": " << remove_vector_res.message();
+    }
```
allenss-amazon, line 310: This is always an internal bookkeeping failure, right?

src/indexes/vector_base.h
```diff
@@ -87,6 +87,36 @@ T CalcMagnitude(const T *src, size_t size) {
...
+  T *dst = (T *)&ret[0];
+  for (size_t i = 0; i < dimensions; i++) {
+    dst[i] = reciprocal_magnitude * src[i];
+  }
+  return ret;
+}
+
+template <typename T>
+std::vector<char> NormalizeVector(absl::string_view record,
+                                  T *magnitude = nullptr) {
```
allenss-amazon: In what cases would you normalize a vector and discard the magnitude?

src/vector_registry.cc
```diff
@@ -0,0 +1,198 @@
...
+namespace valkey_search {
+
+void VectorRegistry::Init(ValkeyModuleCtx *ctx) {
+  hash_registration_supported_ =
+      (ValkeyModule_GetApi("ValkeyModule_HashSetStringRef",
+                           (void **)&ValkeyModule_HashSetStringRef) ==
+       VALKEYMODULE_OK) &&
+      (ValkeyModule_GetApi("ValkeyModule_HashHasStringRef",
+                           (void **)&ValkeyModule_HashHasStringRef) ==
+       VALKEYMODULE_OK) &&
```
allenss-amazon: Given that 1.2 requires 9.0.1 I don't think that making the hash_registration_supported optional makes sense. I'd rather CHECK here and avoid the complexity of testing for it later -- needless complexity.

src/vector_registry.cc
```diff
@@ -0,0 +1,198 @@
...
+  ShareWithValkeyHash(key, attribute_identifier, vector_record.get(),
+                      vector_size, attribute_data_type);
+  return vector_record;
+}
+
+bool VectorRegistry::ShareWithValkeyHash(
+    const InternedStringPtr &key, absl::string_view attribute_identifier,
+    const indexes::VectorRecord *vector_record, size_t vector_size,
+    const data_model::AttributeDataType &attribute_data_type) {
+  if (!hash_registration_supported_ ||
```
allenss-amazon: I think HASH data type check should be a CHECK. Plus I think throwing in a VerifyMainThread wouldn't be a bad thing either.

src/vector_registry.cc
```diff
@@ -0,0 +1,198 @@
...
+  }
+  ++stats_.lookup_record_misses;
+  return {nullptr, 0};
+}
+
+std::shared_ptr<indexes::VectorRecord> VectorRegistry::Track(
+    const InternedStringPtr &key, absl::string_view attribute_identifier,
+    ValkeyModuleString *vector, Allocator *allocator,
+    const data_model::AttributeDataType &attribute_data_type) {
+  auto interned_attribute_identifier =
```
allenss-amazon: Why can't we have an interned string stored in IndexBase ?

src/vector_registry.cc
```diff
@@ -0,0 +1,198 @@
...
+
+    float magnitude = indexes::CalcMagnitude(
+        reinterpret_cast<const float *>(vector_str.data()),
+        vector_str.size() / sizeof(float));
+    vector_record =
+        indexes::VectorRecord::Construct(vector_str, magnitude, allocator);
+    vector_size = vector_str.size();
+    tracked_vectors_[search_key] = {vector_record, vector_size};
+  }
+  ShareWithValkeyHash(key, attribute_identifier, vector_record.get(),
```
allenss-amazon: ShareWithValkeyHash is declared as needing mutex_ yet here is called without it. Something is rotten in the lock checking code.

src/vector_registry.cc
```diff
@@ -0,0 +1,198 @@
...
+    absl::string_view attribute_identifier,
+    InternedStringHashMap<indexes::TrackedKeyMetadata>
+        &&tracked_metadata_by_key) {
+  absl::MutexLock lock(&mutex_);
+  for (auto &&[key, _] : tracked_metadata_by_key) {
+    auto interned_attribute_identifier =
+        StringInternStore::Intern(attribute_identifier);
+    RegistryKey search_key = std::make_pair(key, interned_attribute_identifier);
+    LockFreeUntrackIfUnused(search_key);
+  }
```
allenss-amazon, line 160: We drop indexes from a background thread. I think that'll result in HashSet being called from a background thread -> bad.

src/vector_registry.cc
```diff
@@ -0,0 +1,198 @@
...
+}
+
+void VectorRegistry::DetachFromValkeyHash(const RegistryKey &search_key) {
+  auto key_str = vmsdk::MakeUniqueValkeyString(search_key.first->Str());
```
allenss-amazon: This needs a VerifyMainThread in it.

allenss-amazon (commented):
claude believes that key deletion isn't tested. Specifically that it doesn't unTrack the key to recover the vector registry entry.

src/indexes/vector_base.cc
yairgott: To clarify, `RemoveRecordDueToError` is called if the vector length is wrong just from ModifyRecord, not AddRecord, and the error message would be emitted just if UnTrackKey was unsuccessful. I don't see an issue here.

src/vector_registry.cc
```diff
@@ -0,0 +1,203 @@
...
+    ValkeyModuleString *vector, Allocator *allocator,
+    const data_model::AttributeDataType &attribute_data_type) {
+  RegistryKey search_key = std::make_pair(key, attribute_identifier);
+
+  std::shared_ptr<indexes::VectorRecord> vector_record;
+  size_t vector_size;
+  {
+    absl::MutexLock lock(&mutex_);
+    if (!vector) {
+      tracked_vectors_.erase(search_key);
```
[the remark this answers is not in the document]
yairgott: A vector is nullptr means that either the hash record was deleted or the vector hash field was deleted from the engine, therefore there is no need to update the engine.

src/vector_registry.cc
```diff
@@ -0,0 +1,205 @@
...
```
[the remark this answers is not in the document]
yairgott: A vector is nullptr means that either the hash record was deleted or the vector hash field was deleted from the engine, therefore there is no need to update the engine.

src/vector_registry.cc
```diff
@@ -0,0 +1,203 @@
...
+      return nullptr;
```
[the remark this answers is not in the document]
yairgott: A vector is nullptr means that either the hash record was deleted or the vector hash field was deleted from the engine, therefore there is no need to update the engine.

src/vector_registry.cc
```diff
@@ -0,0 +1,205 @@
...
```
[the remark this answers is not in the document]
yairgott: A vector is nullptr means that either the hash record was deleted or the vector hash field was deleted from the engine, therefore there is no need to update the engine.

src/vector_registry.cc
```diff
@@ -0,0 +1,205 @@
...
+    }
```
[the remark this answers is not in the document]
yairgott: A vector is nullptr means that either the hash record was deleted or the vector hash field was deleted from the engine, therefore there is no need to update the engine.

src/vector_registry.cc
```diff
@@ -0,0 +1,205 @@
...
+      vector_record = it->second.vector_record;
+      vector_size = it->second.vector_record_size;
+    } else {
+      float magnitude = indexes::CalcMagnitude(
+          reinterpret_cast<const float *>(vector_str.data()),
+          vector_str.size() / sizeof(float));
+      vector_record =
+          indexes::VectorRecord::Construct(vector_str, magnitude, allocator);
+      vector_size = vector_str.size();
+      tracked_vectors_[search_key] = {vector_record, vector_size};
```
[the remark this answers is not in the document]
yairgott: When HSET writes different vector bytes for a shared field, the engine set a non-string-ref value to the hash record field.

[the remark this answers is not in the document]
yairgott: When HSET writes different vector bytes for a shared field, the engine set a non-string-ref value to the hash record field.

src/vector_registry.cc
```diff
@@ -0,0 +1,205 @@
...
+      vector_record = it->second.vector_record;
+      vector_size = it->second.vector_record_size;
+    } else {
+      float reciprocal_magnitude = indexes::CalcReciprocalMagnitude(
+          reinterpret_cast<const float *>(vector_str.data()),
+          vector_str.size() / sizeof(float));
+      vector_record = indexes::VectorRecord::Construct(
+          vector_str, reciprocal_magnitude, allocator);
+      vector_size = vector_str.size();
+      tracked_vectors_[search_key] = {vector_record, vector_size};
```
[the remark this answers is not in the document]
yairgott: When HSET writes different vector bytes for a shared field, the engine set a non-string-ref value to the hash record field.

src/vector_registry.cc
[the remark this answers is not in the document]
yairgott: A vector is nullptr means that either the hash record was deleted or the vector hash field was deleted from the engine, therefore there is no need to update the engine.

src/vector_registry.cc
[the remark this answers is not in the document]
yairgott: A vector is nullptr means that either the hash record was deleted or the vector hash field was deleted from the engine, therefore there is no need to update the engine.

vmsdk/src/valkey_module_api/valkey_module.h
```diff
@@ -1987,6 +1992,8 @@ static int ValkeyModule_Init(ValkeyModuleCtx *ctx, const char *name, int ver,
   VALKEYMODULE_GET_API(ZsetRangeEndReached);
   VALKEYMODULE_GET_API(HashSet);
   VALKEYMODULE_GET_API(HashGet);
+  VALKEYMODULE_GET_API(HashSetStringRef);
```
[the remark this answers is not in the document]
yairgott: The release notes will clarify that the required valkey engine is at least 9.0.1 where HashSetStringRef is available.

vmsdk/src/valkey_module_api/valkey_module.h
```diff
@@ -1987,6 +1990,8 @@ static int ValkeyModule_Init(ValkeyModuleCtx *ctx, const char *name, int ver,
...
+  VALKEYMODULE_GET_API(HashHasStringRef);
```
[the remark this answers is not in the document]
yairgott: The release notes will clarify that the required valkey engine is at least 9.0.1 where HashHasStringRef is available.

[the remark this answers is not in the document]
yairgott: The release notes will clarify that the required valkey engine is at least 9.0.1 where the new valkey module interfaces are available.

vmsdk/src/valkey_module_api/valkey_module.h
```diff
@@ -1987,6 +1992,8 @@ static int ValkeyModule_Init(ValkeyModuleCtx *ctx, const char *name, int ver,
...
```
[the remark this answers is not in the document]
yairgott: The release notes will clarify that the required valkey engine is at least 9.0.1 where the new valkey module interfaces are available.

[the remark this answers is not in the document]
yairgott: The release notes will clarify that the required valkey engine is at least 9.0.1 where the new valkey module interfaces are available.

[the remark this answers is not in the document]
yairgott: The release notes will clarify that the required valkey engine is at least 9.0.1 where the new valkey module interfaces are available.

src/indexes/vector_hnsw.cc
```diff
@@ -114,11 +115,12 @@ absl::StatusOr<std::shared_ptr<VectorHNSW<T>>> VectorHNSW<T>::LoadFromRDB(
       if (!is_marked_deleted) {
         return std::shared_ptr<VectorRecord>(nullptr);
       }
```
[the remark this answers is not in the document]
yairgott: The flow relies on VectorBase::LoadTrackedKeys which sets the correct vector values, after denormalizing it for cosine distance metric using the loaded magnitude, and is called before the vector index is used.

src/indexes/vector_hnsw.cc
```diff
@@ -115,11 +116,12 @@ absl::StatusOr<std::shared_ptr<VectorHNSW<T>>> VectorHNSW<T>::LoadFromRDB(
...
```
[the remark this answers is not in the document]
yairgott: The flow relies on VectorBase::LoadTrackedKeys which sets the correct vector values, after denormalizing it for cosine distance metric using the loaded magnitude, and is called before the vector index is used.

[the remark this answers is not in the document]
yairgott: The flow relies on VectorBase::LoadTrackedKeys which sets the correct vector values, after denormalizing it for cosine distance metric using the loaded magnitude, and is called before the vector index is used.

src/indexes/vector_base.cc
```diff
@@ -323,6 +290,29 @@ absl::StatusOr<bool> VectorBase::RemoveRecord(
   return true;
 }
```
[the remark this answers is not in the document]
yairgott: That is not true because if the hash record or the hash attribute is deleted an empty vector will be used while calling VectorRegistry::Track which would remove the entry.

Aksha1812:
**Reviewers for this PR**

allenss-amazon (approved):

PR #1234, opened 2026-07-10
Fix flaky test_fulltext_inflight_blocking: actually wait for backfill

allenss-amazon:
TestFullTextInFlightBlockingCMD called
IndexingTestHelper.is_indexing_complete_on_node() as a bare statement. That function is a predicate returning bool, not a waiter, so its result was discarded and the tests never waited for the initial backfill.

These tests run with search.writer-threads=2 and depend on exactly one writer thread being parked at the mutation_processing pausepoint. When the FT.CREATE backfill scan was still in flight, its low-priority mutation tasks occupied both writer threads and parked at mutation_processing. The tests' own HSET tasks then never got a worker, so they never reached the block_mutation_queue pausepoint at the top of ScheduleMutation, and the test timed out waiting for it.

Wrap the five call sites in waiters.wait_for_true(). FT.INFO's backfill_in_progress is driven by IsBackfillInProgress(), which stays true while backfill_inqueue_tasks > 0, so waiting on it guarantees the backfill mutation tasks have drained and not merely that the scan finished. Every other caller in the repo, including the cluster test in this same file, already wraps the predicate in a waiter.

### DIFF
--- a/integration/test_fulltext_inflight_blocking.py
+++ b/integration/test_fulltext_inflight_blocking.py
@@ -31,7 +31,9 @@ class TestFullTextInFlightBlockingCMD(ValkeySearchTestCaseDebugMode):
         )
         client.execute_command("HSET", "doc:1", "content", "hello world")
         client.execute_command("HSET", "doc:2", "content", "hello there")
-        IndexingTestHelper.is_indexing_complete_on_node(client, "idx")
+        waiters.wait_for_true(
+            lambda: IndexingTestHelper.is_indexing_complete_on_node(client, "idx")
+        )
         assert client.execute_command("FT.SEARCH", "idx", "@content:hello")[0] == 2
 
         # Pause mutation processing to keep key in-flight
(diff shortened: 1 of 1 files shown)


integration/test_fulltext_inflight_blocking.py
```diff
@@ -31,7 +31,9 @@ def test_fulltext_inflight_blocking_with_pausepoint(self):
         )
         client.execute_command("HSET", "doc:1", "content", "hello world")
         client.execute_command("HSET", "doc:2", "content", "hello there")
-        IndexingTestHelper.is_indexing_complete_on_node(client, "idx")
+        waiters.wait_for_true(
```
mohanrajendran, line 34: Should we extract these instances to a common `IndexingTestHelper.wait_for_indexing_complete_on_node()`? Similar to the existing example: https://github.com/valkey-io/valkey-search/blob/53ae957fd739beb409b90571efba135fc353cbca/integration/utils.py#L144

mohanrajendran (changes requested):

Aksha1812:
**Reviewers for this PR**

integration/test_fulltext_inflight_blocking.py
```diff
@@ -31,7 +31,9 @@ def test_fulltext_inflight_blocking_with_pausepoint(self):
...
+            lambda: IndexingTestHelper.is_indexing_complete_on_node(client, "idx")
+        )
```
BCathcart, line 36: Any reason the HSET response isn't blocked until indexing is complete here? I wouldn't think this changes anything

integration/test_fulltext_inflight_blocking.py
allenss-amazon: added.

allenss-amazon:
Somebody already fixed this.

PR #1249, opened 2026-07-13
Scoring arguments updates

boda26:
Adds basic scoring support to search, enabling both FT.SEARCH and FT.AGGREGATE to accept a SCORER parameter and FT.SEARCH to support WITHSCORES.

Commit 1: Adding scorer and withscores argument to ft.search command
Parse WITHSCORES and SCORER parameter in FT.SEARCH. Add score field to the Neighbor struct (initialized from distance for vector queries, computed by scorer for text queries). Emit score in the response: as a top-level element for non-vector queries (WITHSCORES), and as a named attribute for vector queries (via the AS alias). Add parser tests covering WITHSCORES, SCORER, and their combinations

Commit 2: Adding scorer field to ft.aggregate
Add SCORER parameter parsing to FT.AGGREGATE. Move scorer field from SearchCommand to the shared SearchParameters base class so both commands inherit it. Use n.score instead of n.distance for vector query results in aggregate

### DIFF
--- a/docs/commands/ft.search.md
+++ b/docs/commands/ft.search.md
@@ -13,8 +13,10 @@ FT.SEARCH <index> <query>
   [SLOP <slop>]
   [SORTBY <field> [ ASC | DESC]]
   [TIMEOUT <timeout>]
+  [SCORER <scorer>] 
   [VERBATIM]
   [WITHSORTKEYS]
+  [WITHSCORES]
 ```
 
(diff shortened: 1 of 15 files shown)


docs/commands/ft.search.md
zackcam, line 1: Lets remove this file for now until it is fully useable

Issue #365, opened 2025-09-03
[BUG] Crash when running JSON.SET


zvi-code:
### Description
Valkey server crashes when executing `JSON.SET` commands with both valkey-search and valkey-json modules loaded. The crash occurs due to cross-module memory allocator interference caused by valkey-search exposing all internal symbols globally.

### Suspected Root Cause
PR #329 introduced a version script that exports all symbols from valkey-search:
```lds
{
  global: *; # Expose all symbols so that crash dumps have them.
};
```

This causes valkey-search's internal memory management functions (`__wrap_free`, `__wrap_malloc`, and C++ operator new/delete overrides) to become globally visible. When valkey-json is loaded after valkey-search, its memory operations incorrectly resolve to valkey-search's allocator wrappers, leading to a crash.

### Crash Backtrace
```
valkey-1-1  | Backtrace:
valkey-1-1  | /lib/x86_64-linux-gnu/libc.so.6(+0x3c050)[0x7f8afac2a050]
valkey-1-1  | /lib/x86_64-linux-gnu/libc.so.6(+0x16d67d)[0x7f8afad5b67d]
valkey-1-1  | valkey-server *:6379 [cluster](+0x244b84)[0x55f52936cb84]
valkey-1-1  | valkey-server *:6379 [cluster](memtest_test_linux_anonymous_maps+0x201)[0x55f529277431]
valkey-1-1  | valkey-server *:6379 [cluster](printCrashReport+0x109)[0x55f529283399]
valkey-1-1  | valkey-server *:6379 [cluster](+0x159b9d)[0x55f529281b9d]
valkey-1-1  | /lib/x86_64-linux-gnu/libc.so.6(+0x3c050)[0x7f8afac2a050]
valkey-1-1  | valkey-server *:6379 [cluster](je_malloc_usable_size+0x71)[0x55f5293a9d61]
valkey-1-1  | /opt/libsearch.so(_ZN5vmsdk19PerformAndTrackFreeEPvPFvS0_EPFmS0_E+0xe)[0x7f8af8c7e9f8]
valkey-1-1  | /opt/libsearch.so(__wrap_free+0xe5)[0x7f8af8c7ee05]
valkey-1-1  | /opt/libjson.so(_ZN7JParser5ParseEPKcm+0x2be)[0x7f8afa8d982e]
[...]

Issue #305, opened 2025-08-11
[BUG]


hwware:
**Describe the bug**
If any CI job fails, I can not find the rerun the button as Valkey repo provide

**To Reproduce**
The Job https://github.com/valkey-io/valkey-search/actions/runs/16888618932/job/47844239987?pr=290 fails, but there is no Re-run jobs button on the top-right corner.

<img width="1526" height="926" alt="Image" src="https://github.com/user-attachments/assets/fb36bf16-435c-4bb4-a66d-7340c62956a4" />

The Job https://github.com/valkey-io/valkey-search/actions/runs/16885221498/job/47833032819?pr=286 fails, but there is no Re-run jobs button on the top-right corner

**Expected behavior**
The Job https://github.com/valkey-io/valkey/actions/runs/16880217224/job/47814019388?pr=2465 in valkey repo fails, but it provides the Re-run jobs button on the top-right corner

<img width="1570" height="907" alt="Image" src="https://github.com/user-attachments/assets/c646f6ad-0d3b-4ee5-aab3-0957ef512efa" />

**Environment**

Additional information to reproduce the failure

- Valkey Search Module Version
- Valkey Core Version 

**Additional context**

PR #1256, opened 2026-07-16
RFC Proposal for SVS_VAMANA compression algorithms to valkey-search

eshenayo:
This PR is for review of the approach for inclusion of the SVS_VAMANA compression algorithms to valkey-search. 

# References

* [Minutes from 06‐25‐2026 meeting](https://github.com/valkey-io/valkey-search/wiki/Minutes-from-06%E2%80%9025%E2%80%902026-meeting)
* **Dev Repo:** https://github.com/eshenayo/valkey-search/tree/svs-hnsw-mem-parity

# Reviewers

/reviewer allenss-amazon
/reviewer yairgott


allenss-amazon (commented):
Two high-level questions on first reading.

1. W.r.t. the conversion scheme from non-compressed to compressed. There wasn't a discussion about how to handle write operations that happen during the conversion. Since you also spec'ed the conversion to take less than 10mSec is the plan to simply block write operations while this is on-going? (FWIW, I don't see the temporary duplicate memory as an issue). Also, I assume that when the index is uncompressed, you are retaining the raw vectors and that those are discarded after the transition to compressed -- right?

2. The integration of memory allocators needs further treatment. There is a large semantic mismatch between C++ static initializers and the Valkey memory allocation model. We have solved this problem in the Search module, but that solution is fragile and is incomplete (incomplete as in their are valid code sequences that must be avoided as they will trigger an allocator mismatch corruption of memory). I don't think that this solution can be extended to an externally vended *.so file.

eshenayo:
@allenss-amazon - regarding the questions:

> 1. W.r.t. the conversion scheme from non-compressed to compressed. There wasn't a discussion about how to handle write operations that happen during the conversion. Since you also spec'ed the conversion to take less than 10mSec is the plan to simply block write operations while this is on-going? (FWIW, I don't see the temporary duplicate memory as an issue). Also, I assume that when the index is uncompressed, you are retaining the raw vectors and that those are discarded after the transition to compressed -- right?

From my SVS colleague: In case if algorithm (Vamana) and it's key parameters (metric, alpha) are not changed, graph reconstruction is not needed - just copy as is. In case of 'explicit' conversion, the cloning SVS routines may decide whether graph have to be reconstructed of copied. Or svs_index_clone() can return with error if it is called with a special flag (svs_no_rebuild), but algorithm configurations are incompatible.

> 2. The integration of memory allocators needs further treatment. There is a large semantic mismatch between C++ static initializers and the Valkey memory allocation model. We have solved this problem in the Search module, but that solution is fragile and is incomplete (incomplete as in their are valid code sequences that must be avoided as they will trigger an allocator mismatch corruption of memory). I don't think that this solution can be extended to an externally vended *.so file.

What I have requested is that with transition the to the C API, a feature request to use a function-pointer struct pattern, set on the builder before index construction -- I think we should have a more in depth conversation about this w/one of the SVS developers.

rfc/svs-vamana.md
```diff
@@ -0,0 +1,414 @@
...
+| 4 | Partial results on timeout | Medium | valkey-search | Return best results found so far when search times out |
+| 5 | C API migration | Blocked (on Intel) | valkey-search | Migrate core ops to `libsvs_c_api.so`; Intel delivers the C API |
+| 6 | SharedAPI pro module | Medium | valkey-search | Implement `libsearch_svs_pro.so` with LVQ/LeanVec via SharedAPI |
+| 7 | Deferred compression | Medium | valkey-search | Copy-semantic transition orchestration with non-blocking swap |
+
+### Known Gaps
+
+| Gap | Impact | Mitigation |
+|-----|--------|------------|
+| Filtered search (C API) | Hybrid queries (vector + tag/numeric predicates) cannot use native SVS filtering | Over-fetch with inflated K + local post-filter in valkey-search; pursuing upstream contribution |
```
KarthikSubbarao, line 403: I did not understand the limitation . Today, we use pre-filtering and inline filtering based on "number of qualified entries" in search.cc. 

Will be able to integrate to work with both search modes?
This will allow us to support hybrid queries

rfc/svs-vamana.md
```diff
@@ -0,0 +1,414 @@
...
+- Thread-safe concurrent `add()` operations
+- `save()` / `load()` for persistence
+- `reconstruct_at()` for exact vector retrieval
+- `get_distance()` for pairwise distance computation
+
+#### C API (Core Operations)
+
+The stable C API (`svs_c.h`, implemented by Intel SVS team) provides the foundation for all index operations:
+- Stable ABI (C linkage, opaque handles) replacing C++ vtable interface
+- Custom threadpool callback interface for integration with valkey-search's reader thread pools
```
KarthikSubbarao, line 319: Any details on how this works and which operations are handled by multi threading and which remain single threaded?
In valkey-search, reader and writer threads are controlled based on number of vCPUs available, and also by configs.

Will this be controllable by us?

rfc/svs-vamana.md
```diff
@@ -0,0 +1,414 @@
...
+- All-or-nothing: cannot ship open-source-only `libsearch.so` that later gains LVQ
+- C++ vtable ABI is fragile across compiler versions
+- Memory accounting is opaque — `malloc` interposition cannot intercept SVS allocations because the runtime has its own PLT entries
+
+#### Target Model: C API + SharedAPI Hybrid (Recommended)
+
+```
+valkey-server
+  └── MODULE LOAD libsearch.so
+         ├── links: libsvs_c_api.so (open-source, built from Apache-2.0 source)
```
KarthikSubbarao, line 62: How is `libsvs_c_api.so` built and generated? When packaging, do we / valkey-bundle, need to build some source code in order to generate this?

yairgott, line 62: The proposed approach to support SVS by a separate shared object (`libsvs_c_api.so`) introduces many challenges and I have high concerns about this approach. I would like to point out that a substantial amount of engineering effort has been invested into `valkey-search` to establish a robust, high-performance, and unified execution model across all index types (HNSW, Flat, Text, Tag, Numeric). Introducing an external dynamic shared library breaks several of these core guarantees.

Here is a list of concrete challenges:
1. Memory Accounting & Allocator Bypass (PLT Boundaries): This approach leads to bypassing `vmsdk`'s memory wrappers which would result in SVS memory usage being opaque to Valkey: `used_memory` and `FT.INFO` cannot accurately track index sizes, and critical engine safeguards like `maxmemory` limits and key evictions will fail to trigger, risking kernel OOM kills. Pointing out that globally exporting allocator wrapper symbols to force interposition across `.so` boundaries is known to cause fatal cross-module allocator mismatch crashes (as seen in [#365](https://github.com/valkey-io/valkey-search/issues/365) with `valkey-json`).

2. Threading & Concurrency Impedance Mismatch:
   * `valkey-search` relies on a strict concurrency model combining the Valkey Main Thread guarantee, dedicated Reader/Writer worker pools, and Time-Sliced Phase Mutexes.
   * An external `.so` running its own uncoordinated thread pools (e.g., OpenMP / internal workers) creates thread contention, CPU over-subscription, and unpredictable P99 latency spikes that bypass Valkey's scheduling.

3. Deployment Fragility:
   * Requiring an external `.so` breaks the single self-contained binary model (`libsearch.so`). It introduces operational friction around `LD_LIBRARY_PATH`, container distribution, and library version drift.

4. Loss of Inlining and Link-Time Optimization (LTO):
   * Function calls crossing `.so` boundaries cannot be inlined and cannot benefit from whole-program LTO across graph traversal and distance kernel loops.
   * Another benefit of LTO is that it substantially reduces the binary size as it removes unused code.

Rather than dynamically loading an external shared library, we should adopt the established `valkey-search` pattern used for `hnswlib`, `highwayhash`, and our other dependencies. Instead of embed a copy of the SVS source code, we can integrate SVS as a git submodule and link it statically/as an object library directly into `libsearch.so`. While this requires an upfront integration effort the scope of work is well-bounded and well-understood, while the ROI in terms of memory safety, single-binary distribution, and performance is substantial.

Advantages of the Submodule Approach:
* Automatic Memory Tracking: By compiling statically into `libsearch.so`, all SVS heap allocations automatically route through `vmsdk::__wrap_malloc` and Valkey's allocator (`jemalloc`) with zero extra wrapper boilerplate, making memory immediately visible to `used_memory`.
* Low-Effort, High-Performance ARM Support: Following our current implementation that leverages the `simsimd` library for cross-architecture distance calculations, statically integrating SVS allows us to easily plug into this SIMD layer, enabling efficient ARM (NEON / SVE / DotProd) support with very low effort—overcoming the x86-only limitation of the pre-built binary runtime.
* Single Self-Contained Binary: Distributes cleanly as a single `libsearch.so` without external path dependencies or runtime `.so` discovery issues.
* Hermetic Symbol Isolation: Statically linked symbols remain hidden via `-Wl,--version-script`, eliminating any risk of symbol collision with other Valkey modules.
* Unified Concurrency: SVS indexing and search routines can be cleanly scheduled on `valkey-search`'s existing thread pools and synchronized via index phase locks.
* Full Compiler Optimization: Enables full Link-Time Optimization (LTO), SIMD vectorization, and compiler inlining across the vector search pipeline.

I strongly recommend pivoting the integration model to embed SVS statically as a git submodule to maintain architectural consistency, safety, and operational simplicity.

rfc/svs-vamana.md
eshenayo: yes, hybrid queries are supported. The filter_rate() hint maps to the same qualified-entry count that search.cc already
  computes to choose between pre-filter and inline-filter for HNSW -- SVS uses that value to size its adaptive candidate batches. The one gap that remains is true inline-filter (where the predicate gates which graph edges are explored during traversal, rather than being applied to batches after retrieval) -- that is noted in Known Gaps as a planned upstream contribution to SVS. The pre-filter path works today.

rfc/svs-vamana.md
eshenayo: The RFC has been updated with a dedicated Thread Ownership section that answers this directly in detail. It will be controllable. The SVS C API (PR #305) exposes a svs_threadpool_interface C struct that valkey-search registers at index construction via svs_index_builder_set_threadpool_custom(). 

One note on current state: the PoC build uses OpenMP (--svs-omp-threads) which does not coordinate with the  reader/writer pools -- this is a known gap flagged in Known Gaps and is resolved by Phase 5 (submodule + C API threadpool integration).  Until then, --svs-omp-threads 1 is the mitigation.

rfc/svs-vamana.md
eshenayo: The architecture has been updated in the RFC -- libsvs_c_api.so as a separate shared library no longer exists in the new model, which directly answers the concern behind this question.

eshenayo: The updated RFC proposal has pivoted to use the git submodule approach in the initial integration for the open source compression, with a separate section for future work to integrate the proprietary algorithms that requires additional discussion.

rfc/svs-vamana.md
```diff
@@ -0,0 +1,633 @@
...
+| 4 | Partial results on timeout | Medium | Intel (contributor) | Return best results found so far when search times out |
+| 5 | SVS submodule integration | Blocked (on Intel SVS C API) | Intel (contributor) | Replace pre-built runtime with static submodule under `third_party/svs/`; remove OpenMP; integrate C API threadpool |
+| 6 | Proprietary compression path | Low (follow-on) | Intel + community decision | LVQ/LeanVec distribution model (build flag, dynamic .so, or standalone build); see Future Considerations |
+| 7 | Deferred compression | Medium | Intel (contributor) | Copy-semantic transition orchestration with non-blocking swap |
+
+### Known Gaps
+
+| Gap | Impact | Status | Mitigation |
+|-----|--------|--------|------------|
+| Inline graph-traversal filter | True inline-filter (predicate gates which edges are explored during traversal) is not yet available in SVS | Planned upstream contribution | Hybrid queries supported via `svs_index_search_topk` adaptive batching (SVS PRs #352/#363): `svs_id_filter_t` passes `is_member()` callback and `filter_rate()` function pointer; filter applied post-traversal per batch |
```
yairgott: Hopefully, the SVS API will support true inline filtering soon so we can utilize it for the initial release.

rfc/svs-vamana.md
```diff
@@ -0,0 +1,633 @@
...
+
+### Comparison with Existing Algorithms
+
+| Property | FLAT | HNSW | SVS_VAMANA |
+|----------|------|------|------------|
+| Search complexity | O(n) exact | O(log n) approximate | O(log n) approximate |
+| Graph structure | None | Multi-layer skip-list graph | Single-level Vamana graph with alpha-pruning |
+| Compression | None | None | FP16, SQ8, LVQ (4/8-bit), LeanVec |
+| Platform | Any | Any | x86_64 Linux (primary); ARM64: no planned Intel SVS optimization; submodule approach removes hard x86 binary constraint |
+| Dynamic updates | N/A | Supported | In progress (thread-safe add/remove) |
```
yairgott: To clarify, FLAT does support dynamic updates today.

rfc/svs-vamana.md
```diff
@@ -0,0 +1,633 @@
...
+
+1. Threshold crossed -- valkey-search detects the live vector count exceeds `LEANVEC_TRAINING_THRESHOLD`.
+2. Capability check -- valkey-search verifies the target compression type is built in:
+   - FP16/SQ8: always available in the open-source submodule build (this RFC)
+   - LVQ/LeanVec: require proprietary backends outside the scope of this RFC; `FT.CREATE` with these types returns an error unless a future proprietary build is loaded
+3. Clone with new storage -- a new index is built with compressed storage, copying the graph structure from the existing index. Searches continue against the original uncompressed index during this phase.
+4. Atomic swap -- once the compressed index is ready, valkey-search atomically swaps the index pointer. The old uncompressed storage is freed.
+5. Memory accounting update -- the freed memory is reflected in `FT.INFO` and per-index byte counters.
+
+**Hard constraint:** Searches must never block for more than ~10ms during the transition. The copy-then-swap approach ensures this -- 2x peak memory during the overlap window is acceptable.
```
yairgott: I wonder if it would be possible to perform this transition incrementally, such that the old and new instances coexist during the transition window. In this approach:
- Mutations are applied exclusively to the new instance.
- Search queries are executed against both instances, followed by a merge of the two result sets.

For this to work, queries against the old instance must ignore entries that have already been mutated in the new instance. We could achieve this if the SVS KNN search API supports a callback function evaluated before an entry is added to the result set, which happens to be the exact same mechanism required to support the inline filtering.

yairgott: Another option is an incremental catch-up approach:
1. Route all search queries and mutations to the old index while the new index is built incrementally in the background.
2. Once the new index is nearly caught up, route incoming mutations to the new index.
3. Briefly suspend searches until the new index is fully synchronized, then swap the pointers and resume serving queries.

allenss-amazon: The vast majority of vector search algorithms do not handle incremental edits (delete, replace, etc.) perfectly in all cases, resulting in degradation of recall, query rates and/or stranded memory. Generally, there is no better repair procedure than to generate a new index to replace the old one. 

Further, many algorithms rely on a global view of the corpus that is immutable, i.e., overall average, centroids, partitioning, code books, etc. These algorithms want to make two passes over the data, using the first pass to generate these global information and then in the second pass generating the final index. Clearly, the first pass over the data can be accelerated through some sub-sampling technique at the expense of potentially degrading the quality of the global data. That's what's being proposed here, i.e., that the sub-sample be the first N vectors. This creates a tension around the value of N, larger N means higher quality global data (i.e., if the data set isn't temporally uniform) but smaller N improves query performance before conversion of the index (from FLAT to xxxx). 

IMO, we need to solve this problem above the level of the individual algorithm level. We need to build into the internal platform of Valkey Search the ability to rebuild an index in the background. The technique that @yairgott proposes is a part of that, but we also need to address issues like CPU and memory management.

Thus, I would propose that a separate RFC get created to make a framework for this. That RFC should propose both the user interface (keywords on the FT.CREATE command) as well as the internal machinery. It should address both learning-free algorithms (like HNSW) as well as learning-based algorithms (most quantization and compression algorithms) -- allowing the rebuild of a learning-based algorithm to access as much of the corpus as is necessary to obtain the desired quality. It should provide a framework to manage CPU and memory resources.

rfc/svs-vamana.md
```diff
@@ -0,0 +1,633 @@
...
+
+**Fallback behavior:** If the target compression is unavailable at `FT.CREATE` time (LVQ/LeanVec requested), the command returns an error immediately. Deferred compression transitions within the open-source build target FP16 or SQ8 only.
+
+### Platform Requirements
+
+- **x86_64 Linux** (current): submodule build. Optimal with AVX-512; functional with AVX2 at reduced throughput. All open-source compression backends available.
+- **ARM64**: Intel SVS does not have planned ARM-specific optimization work. The submodule approach removes the hard x86 constraint of the prior pre-built binary -- ARM64 compatibility would depend on valkey-search's existing `simsimd` layer (NEON/SVE/DotProd) rather than SVS-specific ARM work, which is an improvement over the prior model where the pre-built binary was x86-only.
+
+The `ENABLE_SVS` CMake flag (currently defaults to OFF) controls whether SVS is compiled into valkey-search. Phase 1 will default it to ON on x86_64 Linux.
```
allenss-amazon: Do we need a compile flag? I vote no.

rfc/svs-vamana.md
```diff
@@ -0,0 +1,633 @@
...
+**Thread pool breakdown by operation:**
+
+| Operation | Thread pool | Notes |
+|-----------|-------------|-------|
+| KNN search / graph traversal | Reader thread pool | Per-query; dispatched by valkey-search's search dispatcher |
+| Add / graph mutation (flush) | Writer thread pool | Batched; under exclusive index lock |
+| Save / Load (RDB) | Caller thread | Single-threaded stream I/O |
+| Deferred compression transition (clone phase) | Writer thread pool | Atomic swap is single-threaded |
+
+**Trigger semantics:** Search tasks are dispatched to the reader pool by valkey-search's search dispatcher, exactly as with HNSW. Graph mutations (add/remove during flush) run under the writer thread pool's exclusive lock. SVS receives work items through `parallel_for` callbacks; it does not observe or react to Valkey phase transitions directly.
```
allenss-amazon: It's not clear to me that this is entirely aligned with Valkey.

Valkey's current model for query operations is:

1. Query operation can assume that no index mutations (writer operations) are running. Generally, this means no locks on index structures are required (potential exception is stats counters, which are typically atomic increments).
2. No sub-threading. Only the thread that invoked the query operation is used.

The mutation operations assume:
1. No queries are running while mutations are happening.
2. Multiple mutations can be running simultaneously, but not on the same key. If the application queues multiple mutations of a key, then these are serialized at a level above the index.
3. No sub-threading. Only the thread that invoked the query operation is used.

rfc/svs-vamana.md
```diff
@@ -0,0 +1,633 @@
...
+
+`svs_index_get_memory_usage()` supplements automatic tracking by providing per-index byte counts for `FT.INFO` detail reporting.
+
+Note: the prior runtime model (`libsvs_runtime.so.0.4.0`) had genuinely opaque memory accounting because it predates the `get_memory_usage()` API and has its own PLT entries. The submodule approach does not share this limitation.
+
+#### Proprietary Compression (LVQ, LeanVec)
+
+LVQ and LeanVec backends are out of scope for this RFC. Three distribution approaches -- a valkey-bundle build flag (`SVS_PRO`), a separate Intel-optimized image, and a standalone user build -- are evaluated in the Future Considerations section, including the community concerns and engineering trade-offs for each.
```
allenss-amazon, line 399: Does this means that only scalar quantization is available in the first version?

eshenayo: Yes, this does mean that scalar quantization will be available in the first version

rfc/svs-vamana.md
eshenayo: That can be removed to have SVS just be built, no flag.

rfc/svs-vamana.md
eshenayo: I'm checking to see where this falls in the plans

rfc/svs-vamana.md
eshenayo: Thank you, I'll update the table so this is accurately reflected

rfc/svs-vamana.md
eshenayo: @yairgott, regarding:

> I wonder if it would be possible to perform this transition incrementally, such that the old and new instances coexist during the transition window. In this approach:
> 
> * Mutations are applied exclusively to the new instance.
> * Search queries are executed against both instances, followed by a merge of the two result sets.
> For this to work, queries against the old instance must ignore entries that have already been mutated in the new instance. We could achieve this if the SVS KNN search API supports a callback function evaluated before an entry is added to the result set, which happens to be the exact same mechanism required to support the inline filtering.

Once the inline filtering is fully supported, the merge approach could be implemented as you describe. As for the catch up/near synchronized approach:

> Another option is an incremental catch-up approach:
> 
> 1. Route all search queries and mutations to the old index while the new index is built incrementally in the background.
> 2. Once the new index is nearly caught up, route incoming mutations to the new index.
> 3. Briefly suspend searches until the new index is fully synchronized, then swap the pointers and resume serving queries.

This aligns to **Phase 7** in the table and can be incorporated, which I'll capture in the scope.

To @allenss-amazon's point on the platform-level framework:

> propose that a separate RFC get created to make a framework for this. That RFC should propose both the user interface (keywords on the FT.CREATE command) as well as the internal machinery. It should address both learning-free algorithms (like HNSW) as well as learning-based algorithms (most quantization and compression algorithms) -- allowing the rebuild of a learning-based algorithm to access as much of the corpus as is necessary to obtain the desired quality. It should provide a framework to manage CPU and memory resources.

Agreed. The scope of the deferred compression for this RFC proposal is intentionally minimal, so let's open a separate RFC and work on a proposal for that.

rfc/svs-vamana.md
eshenayo: I'll update the model table so that it is correctly aligned to Valkey as noted.

rfc/svs-vamana.md
```diff
@@ -191,7 +191,7 @@ FT.CREATE <index> ... SCHEMA <field> VECTOR SVS_VAMANA <num_params>
...
-| RAW_VECTOR_STORAGE | enum | KEEP | KEEP, DROP | **KEEP**: retain original FP32 vectors; `reconstruct_at()` returns full precision; exact vector matching in `ModifyRecord`. **DROP**: do not retain original vectors; reconstruction uses SVS's stored representation (lossy for FP16/SQ8/LVQ/LeanVec); `ModifyRecord` uses `get_distance()` comparison with a per-dimension epsilon for match detection; reduces memory footprint at the cost of reconstruction precision. |
+| RAW_VECTOR_STORAGE | enum | KEEP | KEEP, DROP | **KEEP**: retain original FP32 vectors; `reconstruct_at()` returns full precision; `IsVectorMatch` compares the stored vector exactly; `ModifyRecord` performs remove+add using the supplied data. **DROP**: do not retain original vectors; reduces memory footprint; `ModifyRecord` still performs remove+add using the supplied data unchanged; `reconstruct_at()` returns the lossy stored representation (precision depends on compression type); `IsVectorMatch` uses `get_distance()` with a per-dimension epsilon to approximate identity -- this may misclassify near-threshold vectors under IP and COSINE metrics where the self-distance is not zero. |
```
allenss-amazon: I believe IsVectorMatch is removed from the codebase (now that shared memory is merged).

Is there a use planned for reconstruct_at internal to SVS, I don't think we use it externally today? If not, then I'd propose that we not implement this as user visible surface area.

It's conceivable that in the future we would want to exploit the ability of some vector indexes to be able to disgorge approximate values of a vector, at which point this becomes a useful feature. This will take some significant changes in the core and likely also the JSON module. At which time this capability would be something that applies to all vector indexes.

rfc/svs-vamana.md
allenss-amazon: The inability to perform inline vector filtering is a significant limitation (our experience is that a high percentage of vector queries are hybrid queries, requiring this).

rfc/svs-vamana.md
eshenayo: I see. the reconstruct_at is used by SVS, updating the language to clarify this point, but agreed, it is not user-facing

rfc/svs-vamana.md
eshenayo: I've updated this section with a more accurate representation of the filtered search implementation differences between SVS and HNSW. Please see the updated in the most recent commit.

PR #1259, opened 2026-07-20
fix hardcoded db_num in metadata_manager.cc

boda26:
fix hardcoded db_num in metadata_manager.cc

Bug trigger (without the fix)

Requires all three: cluster mode + an index in a non-zero DB + an RDB load (restart, failover, or replica sync).

On RDB load, the schema is rebuilt with fingerprint/version = 0 (they aren't serialized). OnLoadingEnded should backfill them from metadata, but the buggy loop hardcodes db_num = 0 and uses the encoded key as the name, so the lookup misses and the values stay 0. A later cluster query then hits FAILED_PRECONDITION "Slot fingerprint mismatch" and returns an error.

Not affected: DB 0 indexes, or non-DB-0 indexes that never reload (the live OnMetadataCallback path sets them correctly).

### DIFF
--- a/src/coordinator/metadata_manager.cc
+++ b/src/coordinator/metadata_manager.cc
@@ -757,8 +757,7 @@ void MetadataManager::OnLoadingEnded(ValkeyModuleCtx *ctx) {
                               .at(kSchemaManagerMetadataTypeName)
                               .entries();
     for (const auto &[name, entry] : entries) {
-      // In cluster mode, only support DB 0 for now
-      uint32_t db_num = 0;
+      uint32_t db_num = ValkeyModule_GetSelectedDb(ctx);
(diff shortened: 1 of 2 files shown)


BCathcart:
Do we know the impact of this bug?

boda26:
> Do we know the impact of this bug?

Chi just found this today, no impact for now

BCathcart (approved):

BCathcart:
The fix LGTM but let's also understand the potential customer impact this bug can have

BCathcart:
Let's also add a regression test

boda26:
Added two tests to validate correct db_num is used in (1) normal index creation (2) index created from loading RDB

PR #1263, opened 2026-07-22
feat: introduce multi-language support for 12 snowball stemmer languages 

neerajr0:
# Overview

Introduces FTS for 11 new languages in addition to English that are supported by the Snowball Stemmer: French, German, Spanish, Italian, Portuguese, Russian, Swedish, Turkish, Dutch, Indonesian and Arabic. Language-specific behavior (segmentation, normalization, stemming, stop-word filtering) is encapsulated behind a polymorphic `Language` interface, so each language is a first-class, self-contained object. Adding a new language is one new class plus one registry entry — no modifications to the core ingestion or query paths.

The feature ships behind the MetadataVersion bump to 1.3. Non-English languages are gated by `Language::IsSupported()` (which compares the module version against the language's `MinRequiredVersion()`), so mixed-version clusters fail closed — nodes running < 1.3 will not accept index metadata carrying a non-English `LANGUAGE`.

# Summary of Changes

## High-level architecture

Text handling was originally a monolithic `Lexer` class that hard-coded English tokenization, ASCII-only punctuation, and English Snowball stemming. This PR replaces it with a polymorphic **`Language`** interface. Callers program to `Language*`; the concrete type is selected once at index-creation time and shared, immutably, across every index using that language.

The design follows standard OO patterns:

- **`Language`** is the strategy interface. It exposes both data accessors (`GetDefaultPunctuation`, `GetDefaultStopWords`, `GetNormalizationForm`, `CaseFoldLocale`, `Name`, `Id`) and behavioral methods (`Tokenize`, `TokenizeWithStemMap`, `IsQueryDelimiter`, `NormalizeInPlace`, `IsStopWord`, `GetStemmer`, `GetPunctuationSet`, `IsSupported`, `MinRequiredVersion`).
- **`SnowballLanguage`** is the shared base for all punctuation-segmented, Snowball-stemmed European languages. It implements the entire tokenization/query algorithm once. Concrete subclasses only supply data (punctuation string, stop-word list, normalization form, locale, stemmer algorithm name). At construction it normalizes each stop word through the *same* `NormalizeCaseFoldFilter` used for tokens (NFC/NFKC + locale-aware casefold), so stop-word matching is Unicode-consistent with the tokens it filters — not a plain ASCII lowercase.
- **`LanguageRegistry`** is a process-wide singleton and the single point of `Language` creation. It holds one shared, immutable `Language` instance per enum value; callers never construct concrete languages directly. `LANGUAGE_UNSPECIFIED` maps to English. Unrecognized language enum values are rejected loudly (no silent fallback), preventing older module versions from incorrectly indexing data they cannot tokenize.
- **`CustomizedLanguage`** decorates a base language with FT.CREATE `PUNCTUATION` / `STOPWORDS` overrides, delegating normalization, locale, version gating, and stemming to the base. It holds a `std::shared_ptr<const SnowballLanguage>` (not `const Language`), so it reads the base's stemmer algorithm directly via `GetStemmerAlgorithm()` — no downcast.
- **`SnowballStemFilter`** adapts the external `libstemmer` C API to the `Stemmer` interface, using a thread-local `sb_stemmer` cache keyed by language enum to avoid mutex contention on ingestion threads.

Language instances are immutable, stateless, and shared — heavyweight members (`PunctuationSet`, stop-word hash set, normalizer) exist once per language, not once per index. Two indexes using the same language literally share the same `Language*`; there is no per-index mutable language state, so cross-contamination is impossible by construction.

Stemming is deliberately **not** part of the ingestion segmentation pass — tokens are indexed in their normalized-but-unstemmed form; the stemmer is invoked separately when a stem-map (ingestion) or query expansion is needed. During ingestion, `TokenizeWithStemMap()` first collects all tokens via `SegmentInternal`, then invokes `SnowballStemFilter::BuildStemMap()` as a second pass over the collected token vector. This two-pass design resolves the stemmer once (avoiding per-token thread-local hash lookups) and allocates only for unique stems rather than every token.

## Ingestion pipeline

```
Input text
    │
    ▼
┌────────────────────────────┐
│ SnowballLanguage::Tokenize │  UTF-8 validation, then:
│  (SegmentInternal loop)    │   • NormalizeInPlace on the full input text
│                            │     (NFC/NFKC + casefold) — ensures that
│                            │     compatibility mappings (e.g. NFKC:
│                            │     U+FF0C fullwidth comma → U+002C) are
│                            │     visible to the delimiter scanner
│                            │   • codepoint-aware segmentation loop:
│                            │     - ASCII fast path (lead < 0x80): byte-level
│                            │       bitset punctuation check, no UTF-8 decode
│                            │     - non-ASCII (lead >= 0x80): decode one
│                            │       codepoint via Scanner, check against the
│                            │       per-language non-ASCII punctuation set
│                            │       (includes Unicode whitespace: U+00A0,
│                            │       U+2000–200A, U+202F, U+3000, etc.)
│                            │     - stop-word filter per token
└────────────────────────────┘
    │
    ▼
Indexed tokens  ──┐
                  │  (ingestion-with-stemming path)
                  ▼
        TokenizeWithStemMap(): BuildStemMap() second pass
                  ▼
        stem_root → {surface_forms}   (one stemmer resolution,
                                       allocates only for unique stems)
```

## Query pipeline

Query grammar characters (`@`, `(`, `)`, `|`, `"`, `*`, `%`, `-`, `\`) are handled directly by `FilterParser`, whose overall flow is kept as close to upstream `main` as possible. The only language-aware hooks are:

- punctuation checks via a cached `PunctuationSet&` obtained once at function entry (`language.GetPunctuationSet()`), used directly in the byte-level inner loop — no virtual dispatch per character;
- non-ASCII word-boundary detection via `Language::IsQueryDelimiter(codepoint)`, invoked only when a lead byte `>= 0x80` is encountered;
- normalization via `language.NormalizeInPlace()`;
- stop-word checks via `language.IsStopWord()`.

```
Query expression
    │
    ▼
┌────────────────────────────────────────────────────┐
│ FilterParser walks query syntax chars (byte-level)  │
│   ASCII punctuation → cached PunctuationSet.Contains │
│   non-ASCII lead    → Language::IsQueryDelimiter(cp) │
│   escape handling   → \<char>                        │
└────────────────────────────────────────────────────┘
    │
    ├── Regular term ──────────────────────┐
    │   language.NormalizeInPlace()         │
    │   language.IsStopWord() → drop        │
    │   Stemmer.GetStemRoot()               │→ TermPredicate
    │                                       │
    ├── Wildcard / Fuzzy ───────────────────┤
    │   language.NormalizeInPlace() only     │→ Prefix/Suffix/FuzzyPredicate
    │                                       │
    └── Exact phrase (quoted) ──────────────┘
        language.NormalizeInPlace() only     → TermPredicate(exact=true)
```

Fuzzy DP is codepoint-aware so `min_stem_size` and edit distance work correctly for multi-byte scripts (Arabic, Russian, Turkish); the change lives in `fuzzy.h` via `Scanner::NextUtf8()`, with buffering across Rax edge boundaries so a codepoint split by radix-tree edge compression decodes correctly. Proximity/OrProximity operate on word-level positions in an already-tokenized stream and are unchanged.

## Per-language data ownership

Each language's punctuation set and stop-word list live in that language's own header (`languages/french.h` defines French punctuation and French stop words). There is **no** shared composition with common constants — each language defines its complete, flat data set independently. Duplication across languages is intentional and reviewer-approved: each language is self-contained and independently modifiable, which is also what future non-Snowball languages (CJK) require.

The `PunctuationSet` struct (ASCII `std::bitset<128>` + non-ASCII `flat_hash_set<uint32_t>`) and the `BuildPunctuationSet()` helper live in `language.h` — they are part of the `Language` interface contract (`GetPunctuationSet()` returns a `PunctuationSet&`). The ASCII bitset preserves the `main`-branch O(1) punctuation lookup for the common (English/ASCII) case. `BuildPunctuationSet` also seeds Unicode whitespace codepoints (U+00A0, U+2000–200A, U+202F, U+3000, etc.) as word boundaries, ensuring that non-breaking spaces and typographic whitespace in French, Arabic, and other scripts are correctly treated as delimiters. Stop-word sets are not built by a shared helper: `SnowballLanguage` normalizes each stop word through its own `NormalizeCaseFoldFilter` at construction, so the set matches the language's normalization form and casefold locale.

## File layout principle

One file per major class. Snowball is isolated into `snowball_stem.*` so `libstemmer` linkage stays localized. Future CJK support (Chinese/Japanese/Korean/Polish/Vietnamese) plugs in as new `Language` subclasses that call their own segmenter library (Jieba/MeCab/Morfeusz/ICU), return `nullptr` from `GetStemmer()`, and provide their own `Tokenize` — without touching existing code.

# Per-file breakdown

## New files

| File | Purpose |
| --- | --- |
| `src/indexes/text/language.h` | The `Language` abstract interface and the `Stemmer` interface. Also owns the `PunctuationSet` struct, the `BuildPunctuationSet()` helper (which seeds Unicode whitespace codepoints as word boundaries), `kAsciiPunctuation` (base ASCII punctuation, also used for attribute-alias validation), and the `InProgressStemMap` type alias. |
| `src/indexes/text/snowball_language.h` / `.cc` | `SnowballLanguage` base class. Implements the shared segmentation algorithm once (`SegmentInternal` parameterized by `handle_escapes` / `filter_stop_words`, driving `Tokenize` and `TokenizeWithStemMap`), plus `IsQueryDelimiter`, `NormalizeInPlace`, `IsStopWord`, `GetStemmer`, `GetPunctuationSet`, `IsSupported`. Query tokenization is handled directly by `FilterParser`. |
| `src/indexes/text/language_registry.h` / `.cc` | `LanguageRegistry` singleton (one shared instance per language enum) and the `CreateLanguage(language, punctuation, stop_words)` factory used for FT.CREATE overrides. |
| `src/indexes/text/customized_language.h` / `.cc` | `CustomizedLanguage` decorator: wraps a base `Language` with FT.CREATE punctuation/stop-word overrides; delegates everything else to the base. |
| `src/indexes/text/languages/{english,french,german,spanish,italian,portuguese,russian,swedish,turkish,dutch,indonesian,arabic}.h` | One trivial file per language. Each defines its own flat punctuation string and stop-word vector inline and wires them into the `SnowballLanguage` base, along with `Id()`, `Name()`, normalization form, casefold locale, and `MinRequiredVersion()`. |
| `src/indexes/text/snowball_stem.h` / `.cc` | `SnowballStemFilter` (Snowball adapter): `GetStemRoot()` for single-word stemming and `BuildStemMap()` for ingestion-time stem→surface-form maps. Thread-local `sb_stemmer` per language. Isolates the `libstemmer` dependency. |
| `src/indexes/text/unicode_normalizer.h` / `.cc` | `NormalizeCaseFoldFilter` and the ICU-backed `UnicodeNormalizer` (NFC by default; NFKC for Arabic presentation-form folding; ASCII fast-path lowercasing; locale-aware casefold for Turkish dotless-i). |
| `third_party/snowball/src_c/stem_UTF_8_{LANGUAGE}.{c,h}` | 11 generated UTF-8 stemmer implementations from upstream Snowball. |

## Modified files

| File | Change |
| --- | --- |
| `src/index_schema.proto` | Adds the `Language` enum (12 values) plus `language`, `punctuation`, `stop_words`, `min_stem_size`, `with_offsets` fields on `IndexSchema`. |
| `src/index_schema.cc` | Wires `Language` through create/serialize/deserialize. FT.INFO reports the language name via `LanguageRegistry::Instance().Get(language_)->Name()` (no switch on the enum) — including for indexes without text fields. Computes the metadata min-version dynamically from `Language::MinRequiredVersion()` (non-English text → `kRelease13`, English text → `kRelease12`). Rejects unrecognized language enum values at index creation rather than falling back to English. |
| `src/commands/ft_create_parser.{h,cc}` | Parses `LANGUAGE`, `STOPWORDS`, `PUNCTUATION`, `MINSTEMSIZE` clauses. Validates the language via `LanguageRegistry::Instance().Get(language)->IsSupported()`. Defaults punctuation/stop words via `Get(language)->GetDefaultPunctuation()` / `->GetDefaultStopWords()`. The set of accepted `LANGUAGE` tokens is gated at compile time on `kModuleVersion >= kRelease13`. Attribute-alias validation uses `kAsciiPunctuation` from `language.h`. |
| `src/commands/filter_parser.{h,cc}` | Query parsing flow restored to match `main`. Language behavior enters only through a cached `PunctuationSet&` (`GetPunctuationSet()`), `IsQueryDelimiter(cp)` for non-ASCII bytes, `NormalizeInPlace()`, and `IsStopWord()`. Malformed-UTF-8 rejection of the whole expression is gated at `>= 1.3.0` (legacy behavior below). |
| `src/indexes/text/text_index.{h,cc}` | `TextIndexSchema` holds a `std::shared_ptr<const Language>` and exposes `GetLanguage()`. Ingestion calls `language_->Tokenize()` / `TokenizeWithStemMap()`; delete path and query-time stem expansion call `language_->GetStemmer()->GetStemRoot()`. The in-progress stem-map mutex is held only long enough to fetch a stable per-key entry pointer (the `node_hash_map` guarantees pointer stability); tokenization then runs **outside** the lock, matching `main`'s ingestion concurrency pattern and avoiding lock contention on ingestion threads. |
| `src/indexes/text/textinfocmd.cc` | `FT._TEXTINFO` tokenization goes through `GetLanguage().Tokenize()`. |
| `src/coordinator/search_converter.{cc,h}` | Coordinator-side malformed-UTF-8 gate for inter-node predicate transmission. Rejects malformed bytes at emulate-release `>= 1.3`; under legacy emulation substitutes U+FFFD in text tokens while leaving tag/numeric values raw so exact match still works (belt-and-suspenders alongside the originating-node validation, plus backward compat for versions < 1.3). |
| `src/indexes/text/fuzzy.h` | Fuzzy DP sized in codepoints, not bytes; multi-byte codepoints split across radix-tree edges are buffered and decoded across edges. |
| `src/version.h` | Sets `kModuleVersion` to `1.3.0`; adds the `kRelease13` constant used by the metadata min-version machinery for non-English text indexes. |
| `src/utils/scanner.h` | Shared UTF-8 `Scanner` utility: strict decode (`NextUtf8`), `IsValidUtf8`, `ReplaceInvalidUtf8`, codepoint counting, ASCII fast-path helpers. Used by the tokenizer, fuzzy DP, and UTF-8 gates. |
| `THIRD_PARTY_NOTICES` | Licensing notices for Unicode CLDR (Unicode License V3) covering the per-language punctuation exemplars, and Apache Lucene (Apache 2.0) covering the per-language stop-word lists (now embedded in the per-language headers). |
| `.config/typos.toml` | Non-English/multi-language word allowances (`caf`, `café`, `mak`, `continuent`, …) and generated-source exclusions so the new multilingual content passes the typo check. |
| `src/indexes/CMakeLists.txt`, `testing/CMakeLists.txt` | Build wiring for the new pipeline sources (`language.h`, `snowball_language.*`, `language_registry.*`, `customized_language.*`, per-language headers) and the new/renamed test binaries; drops the old lexer sources. |
| `third_party/snowball/CMakeLists.txt`, `third_party/snowball/libstemmer/modules.h`, `third_party/snowball/add_language.sh`, `third_party/snowball/OWNERS` | Adds and registers the 11 new UTF-8 stemmer translation units; `add_language.sh` rewritten to be idempotent and UTF-8-only. |
| `third_party/icu/README.md` | Documents the ICU static-library dependency used by `UnicodeNormalizer` (NFC/NFKC) and locale-aware casefold. |

## Deleted files

| File | Reason |
| --- | --- |
| `src/indexes/text/lexer.h` / `lexer.cc` | Monolithic English-only lexer replaced by the `Language` interface + `SnowballLanguage`. |
| `src/indexes/text/language_processor.h` / `.cc` | Intermediate processor abstraction folded into `Language` (interface), `SnowballLanguage` (algorithm), and the standalone utility classes. |
| `src/multi_language.h` | Version-guard helpers absorbed into the `Language` interface (`IsSupported()` / `MinRequiredVersion()`) and the registry. |
| `src/indexes/text/punctuation.h` | Per-language punctuation data moved into each `languages/*.h`; `PunctuationSet` + `BuildPunctuationSet()` moved into `language.h`. |
| `src/indexes/text/stop_words.h` | Per-language stop-word data moved into each `languages/*.h`; each `SnowballLanguage` normalizes its stop words through its own `NormalizeCaseFoldFilter` at construction (no shared build helper). |

# Unit & Integration Testing

The refactor moves behavior from a single monolithic class into a small set of composable, polymorphic types, so the test strategy is: (a) test the shared `SnowballLanguage` algorithm through the interface, (b) test per-language behavior where code paths actually diverge (locale, punctuation set, NFC vs NFKC), and (c) integration-test the whole feature against a running server.

## Unit tests

| Test file | Coverage |
| --- | --- |
| `testing/snowball_language_test.cc` (new; replaces `language_processor_test.cc`) | The comprehensive behavioral surface of `SnowballLanguage` through the `Language` interface: tokenization, normalization, stop-word filtering, punctuation handling, query tokenization, and version gating. Uses `TEST_F` for shared-fixture cases and parameterized tests for the cases that hit different code paths per language (non-ASCII punctuation splitting per language, locale/normalization-form differences). Includes cross-language isolation checks and the `NonEnglishRequiresRelease13` version-gate assertion. |
| `testing/snowball_stem_test.cc` (new) | `SnowballStemFilter` + `BuildStemMap` + `GetStemRoot` across all 12 languages; German compound words; French apostrophe elision; Turkish locale-aware stem length; cross-language stemmer independence. |
| `testing/unicode_normalizer_test.cc` (new) | NFC round-trips, NFKC Arabic presentation-form folding, casefold behavior (parameterized). |
| `testing/filter_test.cc` (modified) | Query-parser word extraction through the `Language` hooks, escape handling, wildcard/fuzzy normalization, quoted phrases, and the malformed-UTF-8 gate (`RejectsWhenEmulatingCurrentRelease` at 1.3, `ToleratesWhenEmulatingLegacyRelease` at 1.2). |
| `testing/ft_create_parser_test.cc` (modified) | `LANGUAGE`, `STOPWORDS`, `PUNCTUATION`, `MINSTEMSIZE` clause parsing and validation; pulls per-language defaults directly from `languages/french.h` / `languages/german.h`. |
| `testing/index_schema_test.cc` (modified) | `Language` round-trips through serialize/deserialize; metadata min-version (non-English text → `kRelease13`). |
| `testing/search_test.cc` (modified) | Search paths through the `Language` interface, including coordinator-side malformed-UTF-8 gating at 1.3. |
| `testing/text_test.cc`, `testing/text_index_schema_test.cc` (modified) | `TextIndexSchema` tokenization/stem-map behavior through the interface; per-index punctuation/stop-word defaults via the registry. |
| `testing/utils/scanner_test.cc` (modified) | Multi-byte codepoint decoding, punctuation-driven delimiters, ASCII fast-path branch. |
| `testing/lexer_test.cc` (removed) | Superseded by `snowball_language_test.cc` + `snowball_stem_test.cc`. |

## Integration tests

| Test file | Coverage |
| --- | --- |
| `integration/test_multi_language_search.py` | `TestMultiLanguageAllAccepted` (all 12 languages accepted on FT.CREATE); `TestPerLanguageStemming` (parametrized create→ingest→search roundtrip); `TestMultiLanguageSearch` (stop words, NFC, Arabic NFKC, non-ASCII punctuation); `TestLanguageInFTInfo` (FT.INFO reports the configured language); `TestLanguageSaveRestore` (RDB persistence across restart); `TestLanguageCrossContamination` (French + German indexes on the same server stay isolated); `TestGermanCompoundWord`; `TestLanguageClusterMetadata` (language consistency across cluster nodes). |
| `integration/test_rdb_cross_version_compat.py` | Forward/backward RDB compatibility. The module stamps each RDB with the minimum version required to load it (English text → `kRelease12`, non-English text → `kRelease13`); tests verify older modules reject RDBs whose stamp exceeds their own and that older RDBs load on the current module (1.3+). |
| `integration/compatibility/*` (`generate_text.py`, `data_sets.py`, regenerated answer pickles) | Multi-language text-search compatibility corpus and expected-answer regeneration. |
| `integration/test_fulltext.py`, `integration/utils.py` (modified) | Existing FTS suites extended for multi-byte input, non-ASCII punctuation, and cross-version RDB helpers. |

# Notes for reviewers

- **Version:** multi-language ships in **1.3** (`kModuleVersion = 1.3.0`, `kRelease13`). All version gates (FT.CREATE language acceptance, coordinator/filter UTF-8 rejection, RDB metadata stamping, and the corresponding unit/integration assertions) reference 1.3.
- **No enum switches:** there are no `switch`/`case` statements on the language enum outside protobuf-generated code. Language-varying behavior is resolved through virtual dispatch on `Language`.
- **Performance parity with `main`:** ASCII text keeps the byte-level, single-index `std::bitset` punctuation path with no UTF-8 decoding; codepoint decoding is triggered only by a non-ASCII lead byte. The thread-local per-language stemmer cache, the `word.reserve(64)` build-in-place tokenization, and the two-pass stem-map build (one stemmer resolution for the entire token vector via `BuildStemMap`) are all preserved.

## Performance Testing and Benchmarking

The following report covers in detail compatibility divergences introduced from RediSearch, information retrieval benchmarking results, and performance testing results:

[snowball_language_benchmarking.md](https://github.com/user-attachments/files/30328048/snowball_language_benchmarking.md)


src/indexes/text/stop_words.h
```diff
@@ -0,0 +1,493 @@
...
+#include <vector>
+
+#include "absl/container/flat_hash_set.h"
+#include "absl/strings/ascii.h"
+#include "src/index_schema.pb.h"
+
+namespace valkey_search::indexes::text {
+
+// English stop words (33 words)
+inline const std::vector<std::string> kEnglishStopWords{
```
KarthikSubbarao: Are we keeping this the same as current release (1.2)?

src/coordinator/search_converter.cc
```diff
@@ -56,9 +60,88 @@ std::optional<query::SortByParameter> SortByFromGRPC(
   return sortby;
 }
 
+namespace {
+
+// Text content of a leaf predicate that must be valid UTF-8. Empty for
+// composite/numeric predicates. Tag content is included so the reject gate is
+// comprehensive; callers derive tag-ness from predicate_case() to apply the
+// legacy policy (substitute text, keep tag raw).
```
KarthikSubbarao, line 68: Why was this needed? We should be validating this on the query side on the originating node, right?

src/indexes/text/punctuation.h
```diff
@@ -0,0 +1,192 @@
...
+// Indonesian: ASCII + common typographic
+inline const std::string kIndonesianPunctuation =
+    kAsciiPunctuation + kCommonMultiBytePunctuation;
+
+// Arabic: ASCII + Arabic-specific (no guillemets — Arabic uses its own marks)
+inline const std::string kArabicPunctuation = kAsciiPunctuation + kArabicExtra;
+
+// Returns the default punctuation string for the given language.
+// Returns English (ASCII-only) punctuation for LANGUAGE_UNSPECIFIED.
+inline const std::string& GetDefaultPunctuation(data_model::Language language) {
```
KarthikSubbarao: One thing which is not great in the current PR is we have massive switch statements throughout several files for very specific functions based on the language.

It would be nicer if we created a first class interface for a Language which implements each of these APIs as virtual functions. As it is right now, the complexity is spread out across the entire package

src/indexes/text/text_index.cc
```diff
@@ -382,7 +385,9 @@ std::string TextIndexSchema::GetAllStemVariants(
     uint64_t stem_enabled_mask, bool lock_needed) {
   // Stem the search term
   std::string stemmed(search_term);
-  lexer_.StemWordInPlace(stemmed, lexer_.GetStemmer());
+  if (auto *stem_filter = processor_->GetStemmer()) {
+    stemmed = stem_filter->GetStemRoot(search_term);
```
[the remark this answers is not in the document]
KarthikSubbarao: @BCathcart / @Aksha1812 - You worked on some aspects of this for 1.2. Can you check if this does not break our logic from 1.2?

src/commands/filter_parser.cc
```diff
@@ -544,47 +586,6 @@ absl::StatusOr<std::unique_ptr<query::Predicate>> FilterParser::WrapPredicate(
       logical_operator, std::move(children), options_.slop, options_.inorder);
 };
 
-// Handles backslash escaping for both quoted and unquoted text
-// Escape Syntax:
-// \\ -> \
-// \<punctuation> -> <punctuation>
-// \<non-punctuation> -> (break to new token)<non-punctuation>...
-// \<EOL> -> Return error
-absl::StatusOr<bool> FilterParser::HandleBackslashEscape(
```
KarthikSubbarao, line 553: Any reason why this was changed? Are we ensuring compatibility wrt to backslashes? What has changed?

src/commands/filter_parser.cc
```diff
@@ -638,93 +643,106 @@ absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseQuotedTextToken(
 //   { } [ ] : ; $ -> error
 absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseUnquotedTextToken(
     std::shared_ptr<indexes::text::TextIndexSchema> text_index_schema,
```
KarthikSubbarao, line 701: This file has some nuanced behavior which we clearly documented in comments. Let's please retain the comments and not remove them unless it has changed. If it has changed, let's document the new behavior

src/commands/filter_parser.cc
```diff
@@ -593,34 +594,38 @@ absl::StatusOr<bool> FilterParser::HandleBackslashEscape(
 // Token boundaries (separated by space): " <punctuation> \<non-punctuation>
 absl::StatusOr<FilterParser::TokenResult> FilterParser::ParseQuotedTextToken(
     std::shared_ptr<indexes::text::TextIndexSchema> text_index_schema,
```
KarthikSubbarao, line 643: There seems to be major changes to the flow here, I would strongly suggest we retain the code flow we had originally unless it is explicitly required. Using AI for refactors here  and fns such as ParseUnquotedTextToken is not safe since it can impact correctness.

For the sake of the review, can we try to retain the original logic? I am fine with calls to new functions where the same logic exists but with  uint32_t instead. However, re-writing this logic unnecessarily is unsafe

integration/module/1.2.0-libsearch.so.zip
KarthikSubbarao, line 1: Why is this needed? Compat testing? Could you describe what it does?

src/multi_language.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2025, valkey-search contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ *
+ */
+
+#ifndef VALKEYSEARCH_SRC_MULTI_LANGUAGE_H_
+#define VALKEYSEARCH_SRC_MULTI_LANGUAGE_H_
```
KarthikSubbarao: This file seems out of place and it is unclear when someone should extend it. Can these functions be moved somewhere better?

KarthikSubbarao: See the comment about having Language as a first class citizen

src/version.h
```diff
@@ -98,6 +98,11 @@ constexpr vmsdk::ValkeyVersion kRelease11(1, 1, 0);
 //
 constexpr vmsdk::ValkeyVersion kRelease12(1, 2, 0);
 
+//
+// Release 1.4, added support for multi-language full text search.
+//
+constexpr vmsdk::ValkeyVersion kRelease14(1, 4, 0);
```
KarthikSubbarao: Likely 1.3. But we need to bump this separately in main, not in a multi language PR

THIRD_PARTY_NOTICES
```diff
@@ -35,3 +35,24 @@ License: ICU License
 License URL: https://spdx.org/licenses/ICU.html
 License File: third_party/icu/LICENSE
 Details: Provides Unicode normalization and case folding for international text processing.
+
+Package: Unicode CLDR (Punctuation Exemplars)
+License: Unicode License V3
```
KarthikSubbarao, line 40: Did we have to bring in anything here?

src/indexes/text/text_index.cc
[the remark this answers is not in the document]
Aksha1812: I think this is correct it is during tokenizing we need to specify the min stem size so that we only add stem roots, stem variants which should be added to the stem map . while during the search path, while getting all stem variants even though we don't want to look for the search_term because it was never stemmed in the first place due to min_stem_size we still want other stem variants that may have come out of the root for that search_term. I beilieve this is fine . will try to verify later .

src/coordinator/search_converter.cc
neerajr0: We do run a validation on the originating node in `FilterParser::Parse`, however this additional check on the receiving side is a belt-and-suspenders to ensure that if a peer ever sends a malformed predicate, we handle it gracefully and don't crash. It also maintains legacy behavior since the receiver won't re-parse, so we want to ensure in versions < 1.4.0 we substitute malformed UTF-8 bytes with `U+FFFD`, instead of rejecting.

src/indexes/text/punctuation.h
neerajr0: This is feasible - the `Language` interface you suggest could define virtual functions for the stop word and punctuation sets; we can also bake in normalization form and locale selection in these, since they can vary by language (e.g. for Arabic and Turkish). `LanguageProcessor::Create` can then fetch from a registry of languages and pass them in to the concrete factory. One thing to consider – languages should theoretically be able to be wired into multiple processors (e.g. if we create a new processor that optimizes on a specific language in the future). We can achieve this by adding a `Pipeline` dimension when creating the `LanguageProcessor`. Something like this for Chinese:

```
class ChineseLanguage : public Language {
 public:
  data_model::Language Id() const override { return data_model::LANGUAGE_CHINESE; }
  Pipeline DefaultPipeline() const override { return Pipeline::kJieba; }

  const PunctuationSet& DefaultPunctuation() const override { /*…CJK punct…*/ }
  const std::vector<std::string>& DefaultStopWords() const override {
    return kChineseStopWords;  // shared across pipelines
  }
  NormalizationForm PreferredNormalizationForm() const override { return NormalizationForm::NFKC; }
  absl::string_view CaseFoldLocale() const override { return ""; }

  std::shared_ptr<Stemmer> CreateStemmer() const override {
    return nullptr;   // neither pipeline stems Chinese
  }
};
```

with two processors:

```
std::shared_ptr<LanguageProcessor> CreateJiebaProcessor(
    const Language& lang, const TextIndexOptions& opts) {
  auto segmenter = std::make_shared<JiebaSegmenter>(/* dict for lang */);
  auto normalizer = std::make_shared<NormalizeCaseFoldFilter>(
      lang.PreferredNormalizationForm(), std::string(lang.CaseFoldLocale()));
  auto stop_filter = std::make_shared<StopWordFilter>(
      opts.stop_words_override.value_or(lang.DefaultStopWords()));
  // No stemmer — lang.CreateStemmer() returns nullptr.

  return LanguageProcessor::Builder()
      .AddSegmenter(std::move(segmenter))
      .SetQueryTokenizer(std::make_shared<JiebaQueryTokenizer>(/*…*/))
      .SetNormalizer(std::move(normalizer))
      .SetStopWordFilter(std::move(stop_filter))
      .Build();
}

std::shared_ptr<LanguageProcessor> CreateIcuBreakProcessor(
    const Language& lang, const TextIndexOptions& opts) {
  auto segmenter = std::make_shared<IcuBreakSegmenter>(/* icu locale */);
  // ...same normalizer / stop-word wiring, different segmenter + query tokenizer...
}
```

and two instantiations:

```
// Default: jieba (language's default pipeline)
LanguageProcessor::Create({.language = LANGUAGE_CHINESE});

// Explicit: ICU
LanguageProcessor::Create({.language = LANGUAGE_CHINESE,
                           .pipeline = Pipeline::kIcuBreak});
```

src/indexes/text/stop_words.h
neerajr0: We want to keep English stop words the same for backwards compatibility with the existing implementation which matches RediSearch. Ideally we update these to match Lucene in the future as with the new stop words we've introduced. Since this is change in behavior, we can introduce it later acknowledging it as a compatibility divergence.

src/commands/filter_parser.cc
neerajr0: Addressing https://github.com/valkey-io/valkey-search/pull/1263#discussion_r3669448912 and https://github.com/valkey-io/valkey-search/pull/1263#discussion_r3669464774 here as well. The goal behind the changes to `filter_parser.cc` is to maintain all existing query logic, but apply the principles behind the multi-language refactor to create a logical boundary between what will be universal across the languages we support and what may vary. 

At query time, this means all handling of query grammar, including backslash and reserved char handling and quotes, still lives in the filter parser and remains logically unchanged. Logic that can vary by language, specifically language-specific tokenization, should be delegated to the `LanguageProcessor`. This allows new languages to be added (with potentially different tokenization implementations or third-party library invocations) without modifying the query path in the future by adding branching or duplicate logic. 

The tradeoff we make here is a one-time refactor of `filter_parser.cc` so the query path can remain untouched for future languages. I also did go through the changes to ensure logical equivalence when introducing them in https://github.com/VoletiRam/valkey-search/pull/16, but can audit again to confirm we don't have potential coverage gaps. I'll also make sure we restore comments we've removed. If we feel like we don't want to make this optimization here that's fair, but we should be aware that we'll have to update the query path when we introduce languages with different tokenization behavior (e.g. CJK).

integration/module/1.2.0-libsearch.so.zip
neerajr0: This was introduced in https://github.com/VoletiRam/valkey-search/pull/19. We added a backward compatibility test to confirm that the current version can load an RDB file created on version 1.2 with English indexes. Forward compatibility tests existed already confirming that an old version will reject an RDB file created on the current version, but those used the 1.0 binary (which we can't use for backward compatibility because 1.0 didn't support FTS).

src/multi_language.h
neerajr0: Fair point, this file had more logic when we were using a feature flag for multi-language support at development time, see https://github.com/VoletiRam/valkey-search/pull/17. With the `Language` interface proposed, we can add these checks as attributes, then remove this file and update the call sites, something like:

```diff
// ft_create_parser.cc — support check
- if (!IsLanguageSupported(language)) { ... }
+ if (!LanguageRegistry::Get(language).IsSupportedByCurrentModule()) { ... }

// index_schema.cc — RDB stamp decision
- if (IsNonEnglishLanguage(unpacked->language())) {
-   min_version = kRelease14;
- }
+ min_version = std::max(min_version,
+     LanguageRegistry::Get(unpacked->language()).MinRequiredModuleVersion());
```

`IsMultiLanguageSupported` can be inlined since it's just a check on the module version. I can make these changes along with the `Language` interface refactor.

src/version.h
neerajr0: I can change this, will have to temporarily skip some unit tests that set module version to `kRelease14` and integ tests in `test_multi_language_search.py` and `test_multi_language_guard.py` that expect a higher module version in order for this to get merged in on its own.

THIRD_PARTY_NOTICES
neerajr0: No libraries, it's just an attribution of where we sourced default punctuation from.

Aksha1812:
**Reviewers for this PR**

src/indexes/text/text_index.cc
```diff
@@ -382,7 +381,9 @@ std::string TextIndexSchema::GetAllStemVariants(
     uint64_t stem_enabled_mask, bool lock_needed) {
   // Stem the search term
   std::string stemmed(search_term);
-  lexer_.StemWordInPlace(stemmed, lexer_.GetStemmer());
+  if (auto *stem_filter = language_->GetStemmer()) {
+    stemmed = stem_filter->GetStemRoot(search_term);
```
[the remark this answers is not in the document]
VoletiRam: this is intentional behavior. Aksha confirmed this in one of the above comments.

src/version.h
```diff
@@ -99,9 +99,9 @@ constexpr vmsdk::ValkeyVersion kRelease11(1, 1, 0);
 constexpr vmsdk::ValkeyVersion kRelease12(1, 2, 0);
 
 //
-// Release 1.4, added support for multi-language full text search.
+// Release 1.3, added support for multi-language full text search.
 //
-constexpr vmsdk::ValkeyVersion kRelease14(1, 4, 0);
+constexpr vmsdk::ValkeyVersion kRelease13(1, 3, 0);
```
neerajr0, line 104: I think we should address https://github.com/valkey-io/valkey-search/pull/1263#discussion_r3669638790 and increment the actual module version in a separate change.

VoletiRam: I agree, will create the new PR updating the module version. Leaving this upgraded for now to allow tests to work but will chain this PR to queue behind module version upgrade PR.

VoletiRam: here is the PR bumping the module version: https://github.com/valkey-io/valkey-search/pull/1302

neerajr0:
Generally the refactor looks good and addresses the comments as intended. I want to align though on the choices we're making in this implementation especially as we consider adding new non-Snowball languages in the future.  

With this implementation we lose some of the composability that we had by defining `Segmenter` and `TokenFilter` primitives. For example, the implementation of `SegmentInternal` now handles normalization and stop word filtering inline. This allows us to keep the logic of `filter_parser.cc` intact as is, but it requires future languages to implement the block of text → tokens transformation as a monolith that cannot be broken into smaller reusable parts. Moreover, it removes the object-oriented advantage of naturally extending the family of 1 → many (e.g. dictionary-based segmentation for Chinese or German decompounding) and 1 → 1/0  (e.g. tashkeel normalization for Arabic) transformation functions when we add new languages or improve existing ones.

On the one hand creating one class per language allows this level of per-language customization already. But at the same time, we will eventually have to face a more significant refactor of the filter parser as soon as we start integrating non-Snowball languages with differing approaches to tokenization. I see `QueryTokenize` indicating the intent to do this in the future (currently unused). The benefit we had from the previous approach that refactored the filter parser is that as long as a new language introduced functions that fit within the 1 → many or 1 → 1/0 transforms, the language would be supported on the read path with no additional changes.

Overall I'm okay with moving forward with these changes as we discussed. I just want to make sure we're on the same page that with this implementation integrating non-Snowball or non-punctuation-based tokenizing languages in the future will require more changes than just adding a new language header file.

src/indexes/text/language.h
```diff
@@ -0,0 +1,174 @@
...
+    return non_ascii.contains(cp);
+  }
+};
+
+// Build a PunctuationSet from a punctuation string. Iterates as code points
+// (not bytes) so multi-byte chars like U+060C are stored correctly.
+// ASCII whitespace and control characters are always included as boundaries.
+inline PunctuationSet BuildPunctuationSet(const std::string& punctuation) {
+  PunctuationSet result;
+  for (int i = 0; i < 128; ++i) {
```
boda26, line 54: Unicode whitespace is not a word boundary — breaks French, the flagship new language. This adds only 0 to 127. No language's punctuation string contains U+00A0, U+202F, U+2000–200A or U+3000. There is no Unicode-whitespace handling and also no test covers it. Right now it seems only ASCII whitespaces are supported.
From AI: "A query for dit or bonjour never matches. Fix belongs in BuildPunctuationSet — seed the non-ASCII set from Unicode White_Space/Zs (ICU is already linked)." Please check by running these queries in French to confirm.

src/indexes/text/snowball_language.cc
```diff
@@ -0,0 +1,213 @@
...
+}
+
+absl::StatusOr<std::vector<std::string>> SnowballLanguage::Tokenize(
+    absl::string_view text) const {
+  if (!utils::Scanner::IsValidUtf8(text)) {
+    return absl::InvalidArgumentError("Invalid UTF-8");
+  }
+  std::vector<std::string> tokens;
+  SegmentInternal(
+      text, /*handle_escapes=*/true, /*filter_stop_words=*/true,
```
boda26, line 153: Ingestion now interprets backslash escapes. There is a silent index-content change against main, and ungated. Tokenize/TokenizeWithStemMap pass handle_escapes=true (snowball_language.cc:145,159); QueryTokenize passes false (:202). That seems backwards, as documents are opaque bytes; queries are the thing with escape syntax, and filter_parser.cc already has its own HandleBackslashEscape.

main's Lexer::Tokenize had no escape handling at all — \ is just a punctuation byte — so "hello\,world" gave ["hello","world"]. The new code gives ["hello,world"], asserted at testing/snowball_language_test.cc:250. text_index.cc:162 feeds it the raw attribute value, so a HASH field containing he said \"hi\" now indexes the token "hi" with the quotes embedded instead of hi, and tokens can now begin with punctuation. Every other behavior change in this PR sits behind VALKEY_SEARCH_COMPATIBILITY_FIX / emulate-release; this one doesn't, and it isn't in COMPATIBILITY.md. It might be helpful to check against these examples.

Potential solutions are either removing handle_escapes from the ingestion path or gating it.

src/indexes/text/language.h
```diff
@@ -0,0 +1,174 @@
...
+      result.ascii.set(cp);
+    } else {
+      result.non_ascii.insert(cp);
+    }
+  }
+  return result;
+}
+
+// Build a stop words hash set from a vector. Lowercases all entries.
+inline absl::flat_hash_set<std::string> BuildStopWordsSet(
```
boda26: This function seems to be dead code and not called elsewhere. It was used for the plain ASCII lowercase which description says already replaced. Recommend to remove this if no longer being used.

src/indexes/text/snowball_language.cc
```diff
@@ -0,0 +1,213 @@
...
+        }
+        if (punct_set_.Contains(cp)) break;
+        uint8_t len = s.LastUtf8ByteLen();
+        word.append(text.data() + pos, len);
+        pos += len;
+      }
+    }
+
+    if (!word.empty()) {
+      normalizer_.NormalizeInPlace(word);
```
boda26: Arabic only: Tokens are split on punctuation before they get normalized (snowball_language.cc:129), and NFKC can turn a harmless character into a delimiter — so a delimiter ends up inside a finished token. 
Example: on an Arabic index (the only one using NFKC), ابحرم + non-breaking space + ملاعلاب is stored as the single token ملاعلاب ابحرم, because U+00A0 isn't punctuation but NFKC converts it to a plain space; the query side always splits on spaces, so neither word can be found. Adding U+00A0 to the punctuation list won't cover it — abc�delimiterdef with a fullwidth comma has the same problem — so the fix is to normalize first, then split.

src/indexes/text/language.h
boda26: ## 1. Unicode whitespace is not a word boundary (French)

  `BuildPunctuationSet` (`language.h:48-55`) only seeds boundaries for code points `0..127`, and no language's
  punctuation string lists U+00A0, U+202F, U+2000–200A or U+3000. French typography puts NBSP before
  `:` `;` `!` `?` and inside `« »`.

  ```
  FT.CREATE idx_fr ON HASH PREFIX 1 fr: LANGUAGE FRENCH SCHEMA body TEXT
  FT._DEBUG TEXTINFO idx_fr LEXER "Il a dit\xc2\xa0: \xc2\xab\xc2\xa0Bonjour\xc2\xa0!\xc2\xa0\xc2\xbb"
  ```

  | | |
  |---|---|
  | **expected** | `"a"  "dit"  "bonjour"` |
  | **actual** | `"a"  "dit\xc2\xa0"  "\xc2\xa0bonjour\xc2\xa0"  "\xc2\xa0"` |

  NBSP glues onto tokens, plus a junk token that is only a NBSP.

  ```
  HSET fr:nbsp  body "Il a dit\xc2\xa0: \xc2\xab\xc2\xa0Bonjour\xc2\xa0!\xc2\xa0\xc2\xbb"
  HSET fr:plain body "Il a dit : \xc2\xab Bonjour ! \xc2\xbb"
  FT.SEARCH idx_fr @body:bonjour NOCONTENT
  ```

  | | |
  |---|---|
  | **expected** | `fr:plain` and `fr:nbsp` |
  | **actual** | `1) (integer) 1`  `2) "fr:plain"` |

src/indexes/text/snowball_language.cc
boda26: ## 2. Ingestion interprets backslash escapes — diverges from `main`, not gated

  `Tokenize` / `TokenizeWithStemMap` pass `handle_escapes=true` (`snowball_language.cc:145,159`) while
  `QueryTokenize` passes `false` (`:202`). Documents are raw bytes; escapes are query syntax, and
  `filter_parser` already has its own `HandleBackslashEscape`.

  ```
  FT.CREATE idx_en ON HASH PREFIX 1 en: SCHEMA body TEXT
  FT._DEBUG TEXTINFO idx_en LEXER 'hello\,world'
  FT._DEBUG TEXTINFO idx_en LEXER 'he said \"hi\"'
  FT._DEBUG TEXTINFO idx_en LEXER 'price 1\.5 usd'
  ```

  | input | expected (`main` / 1.2) | actual (this PR) |
  |---|---|---|
  | `hello\,world` | `"hello" "world"` | `"hello,world"` |
  | `he said \"hi\"` | `"he" "said" "hi"` | `"he" "said" "\"hi\""` |
  | `price 1\.5 usd` | `"price" "1" "5" "usd"` | `"price" "1.5" "usd"` |

  ```
  HSET en:esc body 'he said \"hi\"'
  FT.SEARCH idx_en @body:hi NOCONTENT
  ```

  | | |
  |---|---|
  | **expected** | `en:plain` and `en:esc` |
  | **actual** | `1) (integer) 1`  `2) "en:plain"` |

  Not behind the compatibility gate:

  ```
  CONFIG SET search.emulate-release 1.2.0
  FT._DEBUG TEXTINFO idx_en LEXER 'hello\,world'
  ```

  | | |
  |---|---|
  | **expected** | `"hello" "world"` (revert to 1.2 behavior) |
  | **actual** | `"hello,world"` (unchanged) |

src/indexes/text/snowball_language.cc
boda26: ## 3. NFKC injects delimiters into already-closed tokens (Arabic)

  `SegmentInternal` decides boundaries on raw code points (lines 83–126) but normalizes only after the token
  is closed (line 129). Nothing re-checks the result.

  ```
  FT.CREATE idx_ar ON HASH PREFIX 1 ar: LANGUAGE ARABIC SCHEMA body TEXT
  FT._DEBUG TEXTINFO idx_ar LEXER "ابحرم\xc2\xa0ملاعلاب"
  ```

  | | |
  |---|---|
  | **expected** | two tokens: `"\xd9\x85..."` `"\xd8\xa8..."` |
  | **actual** | one token: `"\xd9\x85\xd8\xb1\xd8\xad\xd8\xa8\xd8\xa7 \xd8\xa8\xd8\xa7\xd9\x84..."` — ASCII `0x20` embedded |

  The fullwidth-comma case matters most: it is **not** whitespace, so adding NBSP to the punctuation lists
  would not fix it.

  ```
  FT._DEBUG TEXTINFO idx_ar LEXER "abc\xef\xbc\x8cdef"   # U+FF0C, Arabic = NFKC
  FT._DEBUG TEXTINFO idx_fr LEXER "abc\xef\xbc\x8cdef"   # French = NFC, contrast
  ```

  | | |
  |---|---|
  | **expected** | `"abc" "def"` |
  | **actual** | Arabic (NFKC): `"abc,def"` — French (NFC): `"abc\xef\xbc\x8cdef"` |

  Blast radius today is Arabic only — it is the sole language configured NFKC.

  ```
  HSET ar:nbsp body "ابحرم\xc2\xa0ملاعلاب"
  HSET ar:plain body "ملاعلاب ابحرم"
  FT.SEARCH idx_ar @body:ابحرم NOCONTENT
  FT.SEARCH idx_ar '@body:"ملاعلاب ابحرم"' NOCONTENT
  FT.SEARCH idx_ar @body:ابحرم* NOCONTENT
  ```

  | query | expected | actual |
  |---|---|---|
  | exact term | both docs | `ar:plain` only |
  | quoted phrase | both docs | `ar:plain` only |
  | prefix `*` | both docs | both ✅ |

  Prefix/wildcard still reaches the token; exact-term and phrase queries miss it.

src/indexes/text/snowball_stem.cc
```diff
@@ -0,0 +1,103 @@
...
+  // verb suffix chains. This is a valid outcome, not an error — return the
+  // original word unchanged when it happens.
+  if (stemmed_length <= 0) {
+    return word;
+  }
+  return {reinterpret_cast<const char*>(stemmed),
+          static_cast<std::string_view::size_type>(stemmed_length)};
+}
+
+std::string SnowballStemFilter::GetStemRoot(absl::string_view token,
```
boda26, line 73: GetStemRoot calls GetStemmer(), a thread-local flat_hash_map lookup, and then copies the stemmer's internal buffer into a std::string, per token. So for a T-token document this adds T hash lookups and up to T heap allocations.

A more efficient implementation already exists and is correct: SnowballStemFilter::BuildStemMap (snowball_stem.cc:83-101) does exactly this with string_view and one stemmer resolution, but it is dead in production: grep finds callers only in testing/snowball_stem_test.cc. 

Simplest fix: after the segmentation loop, call stemmer_->BuildStemMap(tokens, min_stem_size, stem_mappings). That costs a second cheap pass over the token vector and eliminates every per-token allocation, reusing already-tested code. (If the single-pass property is worth keeping, add a string_view-returning stem entry point to the Stemmer interface instead.)

src/index_schema.cc
```diff
@@ -2207,6 +2206,12 @@ absl::StatusOr<vmsdk::ValkeyVersion> IndexSchema::GetMinVersion(
     }
   }
   if (has_text_index) {
+    auto min_lang_version = indexes::text::LanguageRegistry::Instance()
+                                .Get(unpacked->language())
```
boda26: LanguageRegistry::Get() falls back to English for any unrecognized enum, and English's MinRequiredVersion() is ValkeyVersion(0,0,0), so the stamp comes out kRelease12. As a result, A 1.2 module then accepts an RDB carrying a language it cannot tokenize. Recommend to fail loudly rather than silently fallback to English.

src/indexes/text/fuzzy.h
```diff
@@ -31,33 +38,81 @@ struct FuzzySearch {
                         kWordExpansionInlineCapacity>
         key_iterators;
 
+    // Decode pattern to code points so the DP matrix is indexed per character.
+    // The pattern reaches here already well-formed UTF-8. Both entry points
+    // resolve malformed bytes upstream, compat-gated (>= 1.4.0 rejects with
```
boda26: Should the version here be 1.3.0?

src/index_schema.cc
```diff
@@ -1281,13 +1282,11 @@ void IndexSchema::RespondWithInfo(ValkeyModuleCtx *ctx) const {
...
-      ValkeyModule_ReplyWithSimpleString(ctx, "english");
-      break;
-    default:
-      ValkeyModule_ReplyWithSimpleString(ctx, "english");
-      break;
+  if (text_index_schema_) {
+    ValkeyModule_ReplyWithSimpleString(
+        ctx, std::string(text_index_schema_->GetLanguage().Name()).c_str());
+  } else {
+    ValkeyModule_ReplyWithSimpleString(ctx, "english");
```
boda26:
```
A) FT.CREATE idx_novec ON HASH PREFIX 1 doc: LANGUAGE FRENCH \
       SCHEMA emb VECTOR HNSW 6 TYPE FLOAT32 DIM 4 DISTANCE_METRIC L2
     -> OK                          # accepted; no TEXT attribute required
     FT.INFO idx_novec  ->  language: english      ← wrong

B) FT.CREATE idx_text  ON HASH PREFIX 1 t: LANGUAGE FRENCH SCHEMA body TEXT
   FT.INFO idx_text    ->  language: french      ← control, correct

C) FT.CREATE idx_tag   ON HASH PREFIX 1 g: LANGUAGE TURKISH SCHEMA tags TAG
     FT.INFO idx_tag     ->  language: english     ← same bug, TAG-only
```
With an index created with LANGUAGE French but no text field, FT.INFO hits the else branch and reports the hardcoded English.

The PR summary says FT.INFO reports the language via LanguageRegistry::Instance().Get(language_)->Name(). Written that way the bug wouldn't exist, since language_ is always populated correctly from the proto. This should fix the bug here.

src/indexes/text/snowball_language.cc
neerajr0: Took a look at this and I don't think we are actually changing behavior by handling backslashes as escape characters during ingestion. References:

1. Code reference handling backslash as escape during ingestion in upstream main path in `Lexer::Tokenize`: https://github.com/valkey-io/valkey-search/blob/main/src/indexes/text/lexer.cc#L104-L135
2. Code reference from our current refactored implementation leading to the same behavior in `SegmentInternal`:
https://github.com/VoletiRam/valkey-search/blob/multi-language-support/src/indexes/text/snowball_language.cc#L61-L107.
3. We have existing integration tests that assert on handling backslash as an escape character during ingestion: see https://github.com/valkey-io/valkey-search/blob/main/integration/test_fulltext.py#L294 where we assert that `\,` keeps the comma in the token during ingestion. 
4. The [RFC for the text field](https://github.com/valkey-io/valkey-search/blob/main/rfc/text-field.md) confirms we handle backslash as an escape character during ingestion: 

```
Note that individual punctuation characters can be escaped using the normal backslash prefix syntax, causing the next character to be treated as non-punctuation.
```

Let me know what you're seeing that indicates otherwise.

src/index_schema.cc
neerajr0: Addressed in https://github.com/valkey-io/valkey-search/pull/1263/commits/5838534261f988a166b754b21eb33671aafddc01 to return the correct language for non-English indexes with no text field. Also added a new test case.

src/indexes/text/fuzzy.h
neerajr0: Updated in https://github.com/valkey-io/valkey-search/pull/1263/commits/5d6fcf192f996c56e724e2d05ff50422a1dc0a98. In this commit I also fixed several other stale comments referencing 1.4.0 as the release version for multi-language support.

src/index_schema.cc
neerajr0: Addressed in https://github.com/valkey-io/valkey-search/pull/1263/commits/25dffaddd9effc11e262e43fca78f4d0ede9bfbc, so we now fail loudly if we attempt to create an index with an unrecognized language.

src/indexes/text/snowball_stem.cc
neerajr0: Addressed in https://github.com/valkey-io/valkey-search/pull/1263/commits/f39d99ba21cadfc8409db83fac16cae445285614 so we now invoke `BuildStemMap` instead of `GetStemRoot` after`SegmentInternal` completes.

src/indexes/text/snowball_language.cc
neerajr0: Thanks for raising. The cleanest way to handle this is to normalize first, then tokenize as you mentioned. I confirmed and for the current languages supported besides Arabic, tokenization + normalization are commutative, so it's safe to switch the order. 

Note that by doing so we now normalize the entire input text instead of normalizing token by token. This means delimiters now also get normalized. I confirmed this won't add overhead since ICU's `normalizeUTF8` has a [fast path](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/normalizer2impl.cpp#L754) for invariant punctuation characters; also the splitting logic has to iterate over those bytes anyways.

Changes addressed in https://github.com/valkey-io/valkey-search/pull/1263/commits/0b34ff94c8509da59e82e4e24d3aaeccbcebe430.

src/indexes/text/language.h
neerajr0: Removed dead code for `BuildStopWordsSet` in https://github.com/valkey-io/valkey-search/pull/1263/commits/6280b2fd3c7b7d8c7398749c5234cb861e12dcb4.

src/indexes/text/language.h
neerajr0: Added logic to treat unicode whitespace characters as punctuation in https://github.com/valkey-io/valkey-search/pull/1263/commits/cb9520e85a3c221cbf36ba380b8bfdcf2688554e. I suspected this could introduce a new compatibility divergence with RediSearch, since their tokenization [documentation](https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/escaping/) does not seem to mention Unicode whitespace. I confirmed this by by running the following against redis/redis-stack-server (via Docker) and Valkey:

```
FT.CREATE idx ON HASH PREFIX 1 doc: SCHEMA body TEXT NOSTEM

HSET doc:space body "hello world"                   # ASCII space (U+0020)
HSET doc:nbsp body "hello\xc2\xa0world"             # NO-BREAK SPACE (U+00A0)
HSET doc:enquad body "hello\xe2\x80\x80world"       # EN QUAD (U+2000)
HSET doc:ideographic body "hello\xe3\x80\x80world"  # IDEOGRAPHIC SPACE (U+3000)

FT.SEARCH idx "@body:hello" NOCONTENT
FT.SEARCH idx "@body:world" NOCONTENT
```

Redis only returns the first doc in the query results:

```
> FT.SEARCH idx "@body:hello" NOCONTENT
1) (integer) 1
2) "doc:space"

> FT.SEARCH idx "@body:world" NOCONTENT
1) (integer) 1
2) "doc:space"
```

Whereas Valkey returns all of the documents:

```
> FT.SEARCH idx "@body:hello" NOCONTENT
1) (integer) 4
2) "doc:space"
3) "doc:nbsp"
4) "doc:enquad"
5) "doc:ideographic"

> FT.SEARCH idx "@body:world" NOCONTENT
1) (integer) 4
2) "doc:space"
3) "doc:nbsp"
4) "doc:enquad"
5) "doc:ideographic"
```

I will update the RFC to document this.

boda26: Thanks! Great finding and we can document it as a minor inconsistency.

boda26 (approved):

neerajr0:
Takes dependency on https://github.com/valkey-io/valkey-search/pull/1317. We will use the workflow in this PR to benchmark the changes for multi-language support to ensure there are no regressions introduced and root cause any that are identified.

Frank-Gu-81:
Hi @neerajr0 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

Frank-Gu-81:
Benchmark results: https://github.com/Frank-Gu-81/valkey-search/pull/2#issuecomment-5617545542

Note: ignore the 2 failures for now. Pre-existing issue. Not caused by this PR.

integration/compatibility/generate_text.py
```diff
@@ -444,3 +446,211 @@ def test_text_search_escaped(self, key_type, dialect, schema_type):
...
+    character, the resulting query is distance 1 from W. For that query to also
+    be within distance 1 of a stem S, W must be within distance 2 of S (by
+    triangle inequality). Therefore we conservatively exclude any word W where
+    edit_distance(W, S) <= 2 for any stem S that differs from its original.
+
+    Returns a new vocab_by_field dict with only safe words per field.
+    If a field has no safe words, it is omitted from the result.
+    """
+    import snowballstemmer  # dev-only dep; used at generation time (regenerate.sh), not CI runtime
+    stemmer = snowballstemmer.stemmer(language)
```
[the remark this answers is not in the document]
allenss-amazon: lets add it to requirements.txt

src/indexes/text/text_index.cc
```diff
@@ -156,30 +152,33 @@ std::optional<std::reference_wrapper<const Rax>> TextIndex::GetSuffix() const {
...
-    stem_mappings_ptr = &in_progress_stem_mappings_[key];
+    // Lock briefly to obtain a stable pointer — node_hash_map guarantees
+    // pointer stability, so we can write to it after releasing the mutex.
+    InProgressStemMap *stem_mappings_ptr;
+    {
+      std::lock_guard<std::mutex> stem_guard(in_progress_stem_mappings_mutex_);
+      stem_mappings_ptr = &in_progress_stem_mappings_[key];
+    }
+    tokens = language_->TokenizeWithStemMap(data, min_stem_size_,
+                                            *stem_mappings_ptr);
```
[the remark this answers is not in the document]
VoletiRam: The code at `text_index.cc:169-177` is correct:

1. The mutex protects `in_progress_stem_mappings_` (the __map itself__) during `operator[]` to ensure safe node creation/lookup.
2. `absl::node_hash_map` guarantees __pointer stability__ — once a node exists, its address never changes even as other threads insert new keys.
3. The __value__ (`InProgressStemMap`) at that pointer is per-key. The mutation pipeline routes all fields for the same key to the same worker thread (via key-based sharding), so two threads won't write to the *same* `InProgressStemMap` concurrently. Different keys get different `InProgressStemMap` instances, so no shared state.
4. The comment in the code literally says: "Lock briefly to obtain a stable pointer — node_hash_map guarantees pointer stability, so we can write to it after releasing the mutex.

PR #1171, opened 2026-06-23
Fix indexing for zero-length keys

bandalgomsu:
Fixes zero-length key ("") indexing in Valkey Search. This removes empty-key rejection in the shared indexing path and vector indexes (HNSW/FLAT)

Closes : #1170

PR #1264, opened 2026-07-22
Fix numeric field precision loss in content materialization

### LINKED ISSUE #1262: [BUG] Numeric field values lose precision in content path (MaybeAddIndexedContent uses absl::StrCat(double) → 6 significant digits)
**Describe the bug**

Numeric field values lose precision when materialized in the search content path. `MaybeAddIndexedContent()` (`src/query/search.cc`) formats a numeric field's value with `absl::StrCat(*numeric)`, where `*numeric` is a `double`. `absl::StrCat`/`AlphaNum` formats doubles via abseil's `SixDigitsToBuffer` — only **6 significant digits**. Any integer-valued numeric field larger than ~6 digits is rounded on output.

For example, `20260201` is materialized as the string `"2.02602e+07"` (value `20260200`). The stored `double` is exact (well within the 53-bit mantissa); the loss happens purely in the double→string *formatting*, not in storage or comparison.

This surfaces most visibly in `FT.AGGREGATE ... GROUPBY` on a numeric field: because the grouping key is built from this truncated string, distinct values such as `20260201`, `20260202`, `20260203` all collapse into a single `20260200` bucket.

Note: PR #1043 rerouted **non-vector** queries so they no longer pass through `MaybeAddIndexedContent`, which incidentally fixed the non-vector `FT.SEARCH`/`FT.AGGREGATE` case. However, the defective line still exists and is still reachable — a **vector KNN** query that returns/LOADs a numeric field continues to truncate on current `main`.

Related: the codebase converts doubles to strings with inconsistent precision — `%.11g` (`src/expr/value.cc`, `src/commands/ft_aggregate.cc`), `%.12g` (`src/commands/ft_search.cc`), and bare `StrCat` / 6 digits (`src/query/search.cc`). None round-trips an IEEE double. These should be unified.

**To Reproduce**

Against a server with the search module (reproduces on `valkey-bundle:9.1.0`; the vector-path case still reproduces on `9.1.1` / current `main`):

Non-vector aggregate (fixed post-#1043, still broken on 1.2.0 / 9.1.0):
```
FT.CREATE idx ON HASH PREFIX 1 row: SCHEMA n NUMERIC
HSET row:1 n 20260201
HSET row:2 n 20260202
HSET row:3 n 20260203
FT.AGGREGATE idx "@n:[20260101 20260301]" LOAD 1 @n GROUPBY 1 @n REDUCE COUNT 0 AS c
```
Buggy output (search 1.2.0): a single group `n = "20260200"`, count 3 — instead of three distinct groups.

Vector KNN returning a numeric field (still broken on current `main`):
```
FT.CREATE vi ON HASH PREFIX 1 d: SCHEMA n NUMERIC emb VECTOR FLAT 6 TYPE FLOAT32 DIM 2 DISTANCE_METRIC L2
HSET d:1 n 20260201 emb "<8-byte little-endian float32 vector>"
FT.SEARCH vi "*=>[KNN 1 @emb $v AS dist]" PARAMS 2 v "<same vector>" RETURN 1 n DIALECT 2
```
Buggy output: `n = "2.02602e+07"` instead of `"20260201"`.

**Expected behavior**

Numeric field values returned via the content path should preserve their full value (round-trip), e.g. `20260201` should be returned/grouped as `20260201`, not `2.02602e+07` / `20260200`.

**Environment**

- Valkey Search Module Version: reproduced on 1.2.0; numeric content path (`absl::StrCat(*numeric)` in `MaybeAddIndexedContent`) still present on 1.2.1 and current `main`
- Valkey Core Version: 9.1.0 (also reproducible with a locally built valkey-server)

**Additional context**

Root cause line (current `main`): `src/query/search.cc` — `vmsdk::MakeUniqueValkeyString(absl::StrCat(*numeric))` in the `kNumeric` case of `MaybeAddIndexedContent`.

Suggested fix: format the numeric value with a precision that round-trips a `double` (e.g. `%.17g`) or route numeric content through the same formatting used elsewhere, and unify the double→string precision across the code paths listed above.

Aksha1812:
MaybeAddIndexedContent formatted numeric values with absl::StrCat(double), which uses 6 significant digits and truncates integers larger than ~1e6 (e.g. 20260201 -> "2.02602e+07"). This corrupts FT.AGGREGATE GROUPBY keys on numeric fields and numeric values returned by vector queries.

Route numeric content through expr::FormatDouble and switch FormatDouble to std::to_chars for shortest round-trip formatting. Also use it in the FT.AGGREGATE numeric reply path so both paths format consistently.

Fixes #1262


src/expr/value.cc
```diff
@@ -93,9 +94,9 @@ std::string FormatDouble(double d) {
       return "nan";
     }
   } else {
-    char storage[50];
-    size_t output_chars = snprintf(storage, sizeof(storage), "%.11g", d);
-    return {storage, output_chars};
+    char storage[32];
+    auto [ptr, ec] = std::to_chars(storage, storage + sizeof(storage), d);
+    return {storage, ptr};
```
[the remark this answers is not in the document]
allenss-amazon: How does to_chars control the # of significant digits in the output?

Aksha1812:
any more comments here @allenss-amazon ?

Aksha1812:
**Reviewers for this PR**

Aksha1812:
Ready for review

Frank-Gu-81 (approved):
LGTM.

Aksha1812:
@BCathcart . first pass done . can you do final review? only failing test is the recently added hnsw test . these changes are unrleated to that test

src/commands/ft_aggregate.cc
```diff
@@ -194,9 +194,8 @@ bool ReplyWithValue(ValkeyModuleCtx *ctx,
           if (!dble) {
             return false;
           }
-          auto double_size =
-              snprintf(double_storage, sizeof(double_storage), "%.11g", *dble);
-          value_view = std::string_view(double_storage, double_size);
+          double_storage = expr::FormatDouble(*dble);
+          value_view = double_storage;
```
BCathcart: Why not just use the double_storage variable directly and drop the string view?

testing/expr/value_test.cc
```diff
@@ -1354,4 +1354,15 @@ TEST_F(ValueTest, NestedArray_MixedTypesRecursive) {
   EXPECT_TRUE(inner2.GetArrayElement(1).IsDouble());
 }
 
+// Regression for #1262: large integers must round-trip without precision loss.
+TEST_F(ValueTest, FormatDoublePreservesLargeIntegers) {
```
BCathcart, line 1359: How about also testing a negative value and max size output?

BCathcart (approved):
Just a couple nits to address

Aksha1812:
done @BCathcart . can you rereview and merge if all good

src/commands/ft_aggregate.cc
BCathcart: Another nit: I guess we don't need this if/else and whole switch statement anymore. Maybe just add a function like this to wherever IndexerType is defined that we can call?

```
  inline void AssertValidIndexerType(indexes::IndexerType type) {
    CHECK(type == IndexerType::kTag ||
          type == IndexerType::kText ||
          type == IndexerType::kNone ||
          type == IndexerType::kNumeric ||
          type == IndexerType::kVector)
        << "Unexpected indexer type: " << static_cast<int>(type);
  }
```

Then this would maintain the same functionality:
```
AssertValidIndexerType(indexer_type);
value_view = *value.AsStringView();
```

Aksha1812:
Unrelated 
`FAILED ../../../integration/test_saverestore.py::TestZeroLengthKeySaveRestore::test_zero_length_hash_key_vector_restore_v2[HNSW-extra_args0] - util.waiters.WaitTimeout: Expected value to be: True, actual value is: False` failing . asked here https://github.com/valkey-io/valkey-search/pull/1171#issuecomment-5348576702

BCathcart (approved):

src/expr/value.cc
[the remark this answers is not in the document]
BCathcart: Looking online, worst case is 25 chars

src/commands/ft_aggregate.cc
Aksha1812: addressed

PR #1268, opened 2026-07-24
Add PR-triggered search benchmark comparison workflow

Carol-ggg:
### What
Adds `.github/workflows/search-benchmark-pr.yml` — a workflow that benchmarks a
pull request's search module against its base branch and posts a comparison
table as a PR comment.

For each run it builds the search module at **both** the PR merge commit and the
base branch, benchmarks each against the **same** valkey-server core commit, and
comments the delta. Holding core constant means the only variable is the search
module, so the comparison isolates the PR's performance impact.

### How it's triggered
Two entry points, normalized by a `resolve` job into one set of outputs:
1. **Manual** (`workflow_dispatch`) — type a PR number; optional inputs for core
   commit, number of runs, config file, cluster mode, and profiling toggles.
2. **Label** (`pull_request_target`) — add `run-search-benchmark` (or
   `run-search-benchmark-skip-profiling`) to a PR.

### Design notes
- **Same core commit to both benchmark runs** — the A/B holds valkey-server
  fixed and varies only the module .so.
- **Least-privilege permissions** — workflow-level is `contents: read`; the
  `resolve` job adds `pull-requests: read`; only the `benchmark` job gets
  `pull-requests: write` (to post the comment / remove the label).
- **All actions SHA-pinned** for supply-chain safety.
- **Concurrency** — label runs serialize per PR+label so re-labels queue rather
  than contaminate results; manual runs are keyed by run_id (independent).
- **Persistent dataset cache** on the runner via symlink, to avoid re-downloading
  datasets each run.
- Mirrors the pattern already used by `valkey-io/valkey`'s
  `benchmark-on-label.yml`.

### ⚠️ Runner setup required
This workflow runs on a **self-hosted** runner labeled `valkey-search-arm-pr`
and builds/runs **untrusted PR code**. Before enabling in production:

**Runner registration (recommended: organization-level)**
- Register **one** runner at the **org level** (`valkey-io`) inside a **runner
  group** whose repository access is scoped to just the benchmark repos
  (`valkey-search` + `valkey-perf-benchmark`).
- Give the runner the label `valkey-search-arm-pr`.
- Because the search and perf benchmarks may share one machine and both talk to
  the same target host / dataset cache and both run `pkill -f valkey`, they
  **must not run concurrently**. A single org-level runner process serializes
  jobs across both repos automatically (one process = one job at a time).
- ❗ Do **not** register two separate repo-level runners on the same box — that
  produces concurrent execution and contaminates benchmark results. Verify after
  setup: `systemctl list-units 'actions.runner.*'` should show exactly **one**.

**GitHub permissions / settings**
- `EC2_SEARCH_IP` secret configured in the repo (or an environment).
- Actions must be allowed to create PR comments (default `GITHUB_TOKEN` with the
  job's `pull-requests: write` scope).
- **Restrict who can apply the `run-search-benchmark` labels to maintainers** —
  this label is the security boundary for `pull_request_target` code execution.
  
  
  
 📌 Require https://github.com/valkey-io/valkey-perf-benchmark/pull/70 to be merged to output the correct comparison report.


.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,523 @@
...
+        uses: kishaningithub/setup-python-amazon-linux@63dc7bd642c6a564d0a93dd4b1cbe3e448ce3621
+        with:
+          python-version: "3.10"
+          cache: "pip"
+
+      - name: Checkout valkey_ram (for dataset-enabled valkey-benchmark)
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          repository: VoletiRam/valkey_ram
+          ref: npy_support
```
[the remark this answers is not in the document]
VoletiRam: We don't have vector support yet so gonna use the benchmark tool from my repo. Will switch to actual binary once we have support for vectors in valkey-benchmark.

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,523 @@
...
+        type: boolean
+        default: false
+      timeout_minutes:
+        description: "Job timeout in minutes (default 300)"
+        required: false
+        default: "300"
+        type: string
+  pull_request_target:
+    types: [labeled]
+
```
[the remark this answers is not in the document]
VoletiRam: Two things to add so this doesn't become a fake-fix:

1. `author_association` alone is not enough. It describes the PR opener, not the person who added the label. A triage-role account with label permissions on a hostile PR still trips this trigger. You need a check on `github.event.sender.login` (the labeler) as well.
2. Please also confirm the self-hosted runner `valkey-search-arm-pr` is ephemeral. `pull_request_target` on a persistent self-hosted host is the exact configuration GitHub warns against — one bad run persists as a backdoor on the host.

Suggested pattern (two-layer gate):

```yaml
jobs:
  resolve:
    if: |
      github.repository == 'valkey-io/valkey-search' && (
        github.event_name == 'workflow_dispatch' ||
        (
          github.event.action == 'labeled' &&
          (github.event.label.name == 'run-search-benchmark' ||
           github.event.label.name == 'run-search-benchmark-skip-profiling') &&
          contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'),
                   github.event.pull_request.author_association)
        )
      )
    steps:
      - name: Verify labeler is a trusted maintainer
        if: github.event_name == 'pull_request_target'
        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd  # v8
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
          script: |
            const labeler = context.payload.sender.login;
            const {owner, repo} = context.repo;
            const {data} = await github.rest.repos.getCollaboratorPermissionLevel({
              owner, repo, username: labeler,
            });
            if (!['admin', 'write'].includes(data.permission)) {
              core.setFailed(
                `Labeler '${labeler}' has permission '${data.permission}'; ` +
                `refusing to run benchmark on untrusted trigger.`
              );
            }
```

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,523 @@
...
+# For label-triggered runs, serialize per PR+label so re-labels queue rather
+# than contaminate results. On-demand (workflow_dispatch) runs have no
+# concurrency constraint — each runs independently.
+concurrency:
+  group: ${{ github.event_name == 'pull_request_target' && format('benchmark-{0}-{1}', github.event.pull_request.number, github.event.label.name) || format('on-demand-{0}', github.run_id) }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: "bash -Eeuo pipefail -x {0}"
```
VoletiRam, line 67: `-x` in the workflow default traces every command, including steps that pass `${{ secrets.EC2_SEARCH_IP }}` on the command line. GitHub's secret masking only masks the literal secret value; anything that transforms/joins/splits it (which bash tracing can do implicitly) can bypass masking. Recommend dropping `-x` from the default and adding `set -x` at the top of specific steps that don't touch secrets when needed.

```yaml
defaults:
  run:
    shell: "bash -Eeuo pipefail {0}"
```

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,523 @@
...
+    types: [labeled]
+
+# For label-triggered runs, serialize per PR+label so re-labels queue rather
+# than contaminate results. On-demand (workflow_dispatch) runs have no
+# concurrency constraint — each runs independently.
+concurrency:
+  group: ${{ github.event_name == 'pull_request_target' && format('benchmark-{0}-{1}', github.event.pull_request.number, github.event.label.name) || format('on-demand-{0}', github.run_id) }}
+  cancel-in-progress: true
+
+defaults:
```
VoletiRam, line 65: Two issues here:

1. `cancel-in-progress: true` — cancelling a mid-benchmark run wastes the compute and can leave orphaned `valkey-server` processes and half-populated `search-datasets-cache` on the self-hosted host, biasing whichever run picks up next.
2. Including `github.event.label.name` in the group key means adding a second label (e.g. someone toggles `run-search-benchmark-skip-profiling` after `run-search-benchmark`) races a new run against the in-progress one.

Suggested:

```yaml
concurrency:
  group: ${{ github.event_name == 'pull_request_target' && format('benchmark-{0}', github.event.pull_request.number) || format('on-demand-{0}', github.run_id) }}
  cancel-in-progress: false
```

Also, The header comment currently says "re-labels queue rather than contaminate results" but `cancel-in-progress: true` explicitly does not queue, it cancels. Worth aligning the comment with the behavior either way.

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,523 @@
...
+          ref: ${{ needs.resolve.outputs.merge_commit_sha }}
+          path: ${{ env.MODULE_DIR }}
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Checkout valkey-search (baseline)
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          repository: ${{ github.repository }}
+          ref: ${{ needs.resolve.outputs.base_ref }}
```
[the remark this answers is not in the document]
Carol-ggg: pass `base_sha` to downstream to fix the main used. [63058f5](https://github.com/valkey-io/valkey-search/pull/1268/commits/63058f570e9f5638be308862ac5c8c509885868c)

.github/workflows/search-benchmark-pr.yml
Carol-ggg: We use the same `-x` design for other workflows(and also for core workflow). A lot of the major steps have secrets, moving `set -x` to steps that don't have secrets might miss a lot of debugging prints

.github/workflows/search-benchmark-pr.yml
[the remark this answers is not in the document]
Carol-ggg: Discussed offline 

PR author check (author_association): We don't want to block benchmarking external contributors' PRs. Label-based design allows trusted maintainer reviews the code and decides it's safe to benchmark by applying the label. If we gate on the PR author's association, community PRs could never be benchmarked even when a maintainer explicitly triggers it.

Labeler permission check: This is redundant given our access controls. In GitHub, only users with triage role or above can apply labels to issues/PRs (docs). In this repo, we don't grant triage access to untrusted users — the labeler is already implicitly trusted by virtue of having permission to add labels at all.

How label permissions are controlled:

Repository Settings → Collaborators and teams → assign roles (Read, Triage, Write, Maintain, Admin)
Only Triage and above can apply/remove labels on issues and PRs
By keeping Triage+ access restricted to trusted maintainers, the label itself serves as the authorization gate
If we ever broaden triage access to less-trusted individuals, we can revisit adding an explicit labeler check at that point.

.github/workflows/search-benchmark-pr.yml
Carol-ggg: Discussed offline: 
For the concurrency group:
This concurrency design matches the existing `valkey-io/valkey benchmark-on-label.yml` pattern, which uses the same PR_number + label_name group key with cancel-in-progress: true.

The intent is:
- Same label re-added: cancels the in-progress run and starts fresh (useful when a PR is updated mid-benchmark and results would be stale anyway).
- Different labels on the same PR: these produce separate concurrency groups. In practice, parallel execution is prevented by the single self-hosted runner — only one job can physically run at a time on the valkey-search-arm-pr host, so the second job queues at the runner level.

The comments will be updated.

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,527 @@
...
+        type: string
+  pull_request_target:
+    types: [labeled]
+
+# For label-triggered runs, serialize per PR+label so re-labels queue rather
+# than contaminate results. On-demand (workflow_dispatch) runs have no
+# concurrency constraint — each runs independently.
+concurrency:
+  group: ${{ github.event_name == 'pull_request_target' && format('benchmark-{0}-{1}', github.event.pull_request.number, github.event.label.name) || format('on-demand-{0}', github.run_id) }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
Carol-ggg: same as above. the current design is intentional. the comment will be updated.

Carol-ggg:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (2)
> > .github/workflows/search-benchmark-pr.yml (2)> `514-527`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Move label removal to a dependent job with a job-level `always()` condition.**
> > > The `benchmark` job depends on `resolve`, so if `resolve` fails or is skipped, steps in `benchmark` are skipped and this label removal step never evaluates. A separate downstream job with `needs: resolve` and a job-level `if: always() && needs.resolve.outputs.label != ''` is needed so the trigger label is removed even when PR resolution fails.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In @.github/workflows/search-benchmark-pr.yml around lines 514 - 527, The
> > > “Remove benchmark label” step must run even when the benchmark job is skipped
> > > because resolve fails or is skipped. Move this step into a separate downstream
> > > job that depends on resolve, apply job-level `if: always() &&
> > > needs.resolve.outputs.label != ''`, and preserve the existing
> > > `actions/github-script` removal logic and token configuration.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 
> > > `197-203`: _🔒 Security & Privacy_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Disable persisted credentials on the remaining checkout steps.**
> > > `actions/checkout` persists the job token by default, and this job grants `pull-requests: write`. Add `persist-credentials: false` to the `VoletiRam/valkey_ram` checkout and the `valkey-perf-benchmark` checkout so PR-controlled actions can’t write to `pull_requests` after cloning these repositories.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In @.github/workflows/search-benchmark-pr.yml around lines 197 - 203, Add
> > > persist-credentials: false to the checkout steps for VoletiRam/valkey_ram and
> > > valkey-perf-benchmark in the workflow, while preserving their existing
> > > repository, ref, path, and fetch-depth settings.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In @.github/workflows/search-benchmark-pr.yml:
> - Around line 514-527: The “Remove benchmark label” step must run even when the
> benchmark job is skipped because resolve fails or is skipped. Move this step
> into a separate downstream job that depends on resolve, apply job-level `if:
> always() && needs.resolve.outputs.label != ''`, and preserve the existing
> `actions/github-script` removal logic and token configuration.
> - Around line 197-203: Add persist-credentials: false to the checkout steps for
> VoletiRam/valkey_ram and valkey-perf-benchmark in the workflow, while preserving
> their existing repository, ref, path, and fetch-depth settings.
> ```
> 
> ℹ️ Review info

1. we might not want to resolve the label if the workflow fails. 
2. added `persist-credentials: false` to the two other repos.

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+        type: boolean
+        default: true
+      skip_config_set:
+        description: "Skip config_sets (use default config set)"
+        type: boolean
+        default: false
+      timeout_minutes:
+        description: "Job timeout in minutes (default 1440)"
+        required: false
+        default: "1440"
```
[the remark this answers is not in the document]
Carol-ggg: The timeout is a safety net for stuck jobs, not an expected duration. Jobs run ~10 hours without profiling and ~20hour with profiling. 

If this becomes a real issue in the future, we can use GitHub App token before the reporting steps.

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+        type: boolean
+        default: true
+      skip_config_set:
+        description: "Skip config_sets (use default config set)"
+        type: boolean
+        default: false
+      timeout_minutes:
+        description: "Job timeout in minutes (default 1500)"
+        required: false
+        default: "1500"
```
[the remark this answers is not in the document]
Carol-ggg: see below comments.

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+      # Run parameters
+      RUNS: ${{ github.event.inputs.num_runs || '3' }}
+      CORE_COMMIT_INPUT: ${{ github.event.inputs.core_commit || '' }}
+      CLUSTER_FILTER: ${{ needs.resolve.outputs.cluster_filter }}
+      SKIP_PROFILING: ${{ needs.resolve.outputs.skip_profiling == 'true' && '1' || '' }}
+      SKIP_CONFIG_SET: ${{ github.event.inputs.skip_config_set == 'true' && '1' || '' }}
+    # Write access scoped to the job that posts the result comment / removes the label.
+    permissions:
+      contents: read
+      pull-requests: write
```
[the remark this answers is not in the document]
Carol-ggg: `pull-requests: write` is sufficient here. GitHub treats every PR as an issue — the `pull-requests` permission scope covers comments and label operations on pull requests even when using the Issues REST API endpoints (`issues.createComment`, `issues.removeLabel`). From GitHub's docs: "Every pull request is an issue, but not every issue is a pull request. For this reason, shared actions for both features, like managing assignees, labels, and milestones, are provided within the Issues endpoints." No separate `issues: write` is needed for PR-targeted operations.

VoletiRam (approved):

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+      - name: Set up Python
+        uses: kishaningithub/setup-python-amazon-linux@63dc7bd642c6a564d0a93dd4b1cbe3e448ce3621
+        with:
+          python-version: "3.10"
+          cache: "pip"
+
+      - name: Checkout valkey_ram (for dataset-enabled valkey-benchmark)
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          repository: VoletiRam/valkey_ram
```
KarthikSubbarao, line 201: Why is this needed? For Vector Query support? Is there a PR for this which you can link?

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+          - "fts-benchmarks-arm.json"
+          - "fts-benchmarks-shortened-arm.json"
+      cluster_mode:
+        description: "Cluster mode (empty=run config in file, true=cluster-mode only, false=non-cluster-mode only)"
+        type: choice
+        default: ""
+        options:
+          - ""
+          - "true"
+          - "false"
```
KarthikSubbarao, line 40: Are we using this today? ie, are we running with cluster mode enabled or disabled?

.github/workflows/search-benchmark-pr.yml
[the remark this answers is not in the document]
KarthikSubbarao: > Jobs run ~10 hours without profiling and ~20hour with profiling.


I don't mind the value used here. But what is the current estimate for this PR perf run to complete? Is it 4 hrs?

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+          echo "head_sha: ${{ steps.pr.outputs.head_sha }}"
+          echo "base_ref: ${{ steps.pr.outputs.base_ref }}"
+          echo "base_sha: ${{ steps.pr.outputs.base_sha }}"
+          echo "pr_number: ${{ steps.pr.outputs.pr_number }}"
+          echo "cluster_filter: ${{ steps.pr.outputs.cluster_filter }}"
+          echo "skip_profiling: ${{ steps.pr.outputs.skip_profiling }}"
+          echo "label: ${{ steps.pr.outputs.label }}"
+
+  benchmark:
+    needs: resolve
```
KarthikSubbarao, line 157: What does resolve mean? we need the branch to have resolved conflicts or comments before one can run the benchmark? Why?

It might still be useful to get some data.


Is the concern that the benchmark will be on stale module code , not in sync with the main branch?

.github/workflows/search-benchmark-pr.yml
KarthikSubbarao: Would be nice to remove the dependency on a forked pkg soon

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+            const baselineLink = `[\`${baselineShort}\`](https://github.com/${owner}/${repo}/commit/${baselineSha})`;
+            const cluster = '${{ needs.resolve.outputs.cluster_filter }}';
+            const clusterLabel = cluster === 'true' ? ' (Cluster Mode)' : cluster === 'false' ? ' (Standalone)' : '';
+            const skipProf = '${{ needs.resolve.outputs.skip_profiling }}' === 'true' ? ' | Profiling: skipped' : '';
+            const runUrl = `https://github.com/${owner}/${repo}/actions/runs/${{ github.run_id }}`;
+
+            let header = `**Search Benchmark${clusterLabel} ran on commit:** ${link}\n`;
+            header += `**Baseline (module):** ${baselineLink} (\`${base}\`)\n`;
+            header += `**Core (valkey-server):** ${coreLink}\n`;
+            header += `**Runs:** ${process.env.RUNS}${skipProf} | [Workflow Run](${runUrl})\n\n`;
```
KarthikSubbarao, line 492: Anything we want to add here? e.g. a comment saying if regression is more than 10%, a message saying this is a risk to merge?

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+            ${SKIP_PROFILING:+--skip-profiling}
+          echo "✓ Completed PR benchmarks"
+
+      - name: Kill valkey processes between runs
+        run: pkill -f valkey || echo "No valkey processes to kill"
+
+      - name: Run baseline benchmarks
+        working-directory: ${{ env.PERF_BENCHMARK_DIR }}
+        run: |
+          echo "Running search benchmarks for baseline..."
```
KarthikSubbarao, line 418: Observation: If we have the data saved for baseline in a DB (from latest run or daily run), we can save time here. We can run only if base commit has changed

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+    needs: resolve
+    runs-on: [self-hosted, valkey-search-arm-pr]
+    timeout-minutes: ${{ fromJSON(inputs.timeout_minutes || '1440') }}
+    env:
+      # Module configuration
+      MODULE_NAME: "search"
+      MODULE_DIR: "valkey-search"
+      MODULE_PATH: "${{ github.workspace }}/valkey-search"
+      MODULE_BASELINE_DIR: "valkey-search-baseline"
+      MODULE_BASELINE_PATH: "${{ github.workspace }}/valkey-search-baseline"
```
KarthikSubbarao, line 166: Is this forced to be the main branch or is this branch agnostic?

ie, what happens if I have a PR to the 1.2 branch? what happens? we should be running baseline on the 1.2 branch

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+      # Core configuration
+      CORE_GITHUB_REPO: "valkey-io/valkey"
+      CORE_BRANCH: "unstable"
```
KarthikSubbarao, line 169: Might be fine for now. Unless there is a breaking change in Valkey where we cannot load a 1.2 or 1.0 Search module onto it due to compat issues

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+        required: false
+        type: string
+        default: "3"
+      config_file:
+        description: "Benchmark config file (from valkey-perf-benchmark/configs/)"
+        type: choice
+        default: "fts-benchmarks-shortened-arm.json"
+        options:
+          - "fts-benchmarks-arm.json"
+          - "fts-benchmarks-shortened-arm.json"
```
KarthikSubbarao, line 32: How do the options ever get used? Is it for some one who is manually running workflows?

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+            # Create symlinks in /usr/local/bin so builds pick up GCC 14 by default
+            sudo ln -sf /usr/bin/gcc14-gcc /usr/local/bin/gcc
+            sudo ln -sf /usr/bin/gcc14-g++ /usr/local/bin/g++
+            sudo ln -sf /usr/bin/gcc14-gcc /usr/local/bin/cc
+            sudo ln -sf /usr/bin/gcc14-g++ /usr/local/bin/c++
+
+            echo "✓ GCC 14 installed and symlinked"
+            /usr/local/bin/gcc --version
+            /usr/local/bin/g++ --version
+          fi
```
KarthikSubbarao, line 303: Is this the conventional way to install dependencies ?

.github/workflows/search-benchmark-pr.yml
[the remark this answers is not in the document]
Carol-ggg: slightly over 4 hours for one job. but it runs for both baseline and merged version, so it will be ~8hr

.github/workflows/search-benchmark-pr.yml
Carol-ggg: default is "", which means use whatever setting the config.json has. Currently in the shortened test, we dont have cluster mode testings. But we have both non-cluster mode and cluster mode in the full test.

.github/workflows/search-benchmark-pr.yml
Carol-ggg: the install dependencies are what is previously used in the workflow. I added the GCC step for the new workflow. The GCC step is separately listed because we previously encountered some problem in installing GCC and have to install it manually using these code. 
The installation process is verified and the same as the one used in valkey-perf-benchmark continuous workflow.

.github/workflows/search-benchmark-pr.yml
Carol-ggg: yes. these options are for manual trigger

.github/workflows/search-benchmark-pr.yml
Carol-ggg: we need the npy support in ram's fork. We can remove them once we have the encode and decode support for vector query/

.github/workflows/search-benchmark-pr.yml
VoletiRam: For profiling, we don't need to run three times. We need one time and capture flamegraphs in that run.

.github/workflows/search-benchmark-pr.yml
VoletiRam: We are adding this specific step because search module requires gcc 13+ i believe. The machine we use for testing comes with gcc 12 by default. This is an one time setup.

.github/workflows/search-benchmark-pr.yml
Carol-ggg: This refers to the `resolve` job. It figures out the commits that the PR branch and base branch point to, and it also handles the merge of the PR branch into the base branch.

.github/workflows/search-benchmark-pr.yml
Carol-ggg: The comparison results are generated from the perf benchmark. We would need to add to the comparison workflow. We can add this as a follow up and shown it as a failed job once there is regression over 10%

.github/workflows/search-benchmark-pr.yml
Carol-ggg: sure. In the follow up, we can connect this to the existing database or a separate database that stores the baseline results.

.github/workflows/search-benchmark-pr.yml
Carol-ggg: The base uses whatever branch the PR targets.

.github/workflows/search-benchmark-pr.yml
Carol-ggg: fix profiling to run only one time is added in https://github.com/valkey-io/valkey-perf-benchmark/pull/79

Aksha1812:
**Reviewers for this PR**

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+          echo "base_ref: ${{ steps.pr.outputs.base_ref }}"
+          echo "base_sha: ${{ steps.pr.outputs.base_sha }}"
+          echo "pr_number: ${{ steps.pr.outputs.pr_number }}"
+          echo "cluster_filter: ${{ steps.pr.outputs.cluster_filter }}"
+          echo "skip_profiling: ${{ steps.pr.outputs.skip_profiling }}"
+          echo "label: ${{ steps.pr.outputs.label }}"
+
+  benchmark:
+    needs: resolve
+    runs-on: [self-hosted, valkey-search-arm-pr]
```
VoletiRam, line 158: We have to update the runner name if we decide to run it in same machine as core tests for PR profiling.

KarthikSubbarao (approved):

KarthikSubbarao:
Approving to help unblock the other PRs we have open to allow them to be perf tested. This is a useful functionality.

But as I commented in the PR before, it would be nice to clean up the build/installation steps. ie, can we have a single script for build and installation? Should we just be able to call ./build.sh where the same script includes any installation required? (edited) 

This Perf testing yml PR will not work if there is even a small preinstallation step changed.

There are other PRs where new installation steps are being added. So, we would need to keep updating this yml to include new installation steps.

Would be nice if we had a single script for any installation required. Then the changes to this yml can just use that same script and no changes will be required

KarthikSubbarao:
This PR yml is very tightly coupled to how we assume the build process is

This perf github yml should to be updated to use a single script for "pre-build installation activities".
We should maintain this script for any user as well who wants to build valkey-search on their machine

VoletiRam:
I am gonna own the pre-install script PR to install and setup necessary dependencies.

.github/workflows/search-benchmark-pr.yml
```diff
@@ -0,0 +1,530 @@
...
+            cp -r "$PERF_BENCHMARK_PATH/datasets/"* "$DATASET_CACHE/" || true
+          fi
+
+          rm -rf "$PERF_BENCHMARK_PATH/datasets"
+          ln -sf "$DATASET_CACHE" "$PERF_BENCHMARK_PATH/datasets"
+
+          echo "✓ Dataset directory linked to persistent cache: $DATASET_CACHE"
+          ls -lh "$PERF_BENCHMARK_PATH/" | grep datasets
+
+      - name: Install dependencies
```
yairgott, line 258: We should be running this benchmark workflow inside our `devcontainer` , similarly as done today by other workflows, for instance the [clang-tidy-format](https://github.com/valkey-io/valkey-search/blob/main/.github/workflows/clang_tidy_format.yml#L57) one.

Moving this execution into a container would provide a few major benefits:
1. **Simplifies the Workflow:** We could completely remove the ~45 lines of manual `dnf install`, GCC version checking, and symlinking (lines 258-303). The environment would be pre-packaged and less fragile to runner OS updates, just like our other containerized workflows.
2. **Guaranteed Cleanup:** It directly addresses the concerns about dirtying the self-hosted runner with orphaned processes. Instead of relying on `pkill -f valkey` to clean up, stopping the job simply kills the container, instantly tearing down all processes.
3. **Local Reproducibility:** It guarantees that developers investigating a performance regression locally are using the exact same toolchain, Python version, and GCC version as CI.

**One caveat to keep in mind:** Since we use `perf` for profiling, we would need to ensure the container is run with the necessary privileges (e.g., `--cap-add=SYS_ADMIN` and `seccomp=unconfined`) so that the kernel allows profiling inside the container.

neerajr0: Will take this as a future optimization so we can start using this for features in review.

.github/workflows/search-benchmark-pr.yml
neerajr0: Will raise a follow up PR for this.

neerajr0:
Remaining AIs:

- [x] 1. Ensure this workflow is configured for execution on label, see https://github.com/valkey-io/valkey/blob/unstable/.github/workflows/benchmark-on-label.yml in Valkey repo.
- [ ] 2. Use single script for build and installation, see: https://github.com/valkey-io/valkey-search/pull/1268#issuecomment-5358683282
- [ ] 3. Maintainer to add instance name and secret to `valkey-search` repo.
- [ ] 4. This workflow needs updating to use the runner and secret name from step 3. See: https://github.com/valkey-io/valkey-search/pull/1268#discussion_r3799089619
- [ ] 5. Later follow up: address optimization in https://github.com/valkey-io/valkey-search/pull/1268#discussion_r3824225136.

Frank-Gu-81:
Action item 2 and 4: https://github.com/valkey-io/valkey-search/pull/1317

.github/workflows/search-benchmark-pr.yml
Frank-Gu-81: Hello @yairgott, following up on your suggestion to containerize. I tried the existing .devcontainer image on an AL2023 arm64 runner (same OS/arch as our benchmark runner) and hit a blocker worth your input on.

Building in the container and running on the host doesn't work. The image is `ubuntu:noble` (glibc 2.39, gcc 13.3); AL2023 provides glibc up to 2.35. The resulting libsearch.so won't load on the host:

```
/lib64/libm.so.6: version `GLIBC_2.38' not found
/lib64/libc.so.6: version `GLIBC_2.36' not found
libmvec.so.1 => not found
```
`dlopen` fails outright, so `--loadmodule` would too. Not fixable with flags — `-static-libstdc++` would cover `GLIBCXX_3.4.32` but not the glibc versions or `libmvec`.

That leaves three paths I can think of, and I'd like your read:

1. Run everything in the container (build + benchmark). ABI-consistent by construction. The devcontainer already sets `--network host`, `seccomp=unconfined`, `SYS_PTRACE`, so overhead should be minimal — but it moves the measurement environment, so we'd want to validate perf fidelity before trusting results.

2. AL2023-based build image instead of the Ubuntu devcontainer. Matches the host ABI, keeps benchmarking on bare metal, but is a second image to maintain.

3. No container — keep host builds and make the install script multi-distro (apt + dnf). Which is my current PR #1317. 

Please let me know what you think. Thank you!

PR #1271, opened 2026-07-26
Extended Array compatibility fixes

allenss-amazon:
## Summary

PR #932 added `TOLIST` and with it a new ARRAY value, but the compatibility
suite only covered ARRAY as the *last* thing a pipeline produces. This adds
coverage for ARRAY as **input** to SORTBY, GROUPBY and APPLY, and fixes what
that coverage found: 132 of 362 answers differed from Redisearch when the tests
first ran, and none do now.

## Tests

Three datasets and a generator (`array-input-answers.pickle.gz`): every dyadic
operator in five operand shapes, unary `!`, every function in
`expr::function_table`, the unregistered array functions, SORTBY, a second
GROUPBY, every reducer, the empty array, and array-vs-array comparison.

Three harness fixes were needed before any of it could run: a command with both
GROUPBY and SORTBY was an unconditional `assert False`; rows keyed on an ARRAY
were compared against the wrong rows, because the engines return the collected
elements in different orders; and `parse_value` crashed on a RESP nil.

## Behavior fixes

| Symptom | Fix |
|---|---|
| Comparison against a scalar returned `kUNORDERED`, which `operator==` counts as equal, so `@array == 2` was true and `@array < "a"` false | ARRAY compares as the empty string, which is what Redisearch does — `@array == ""` is true there for an empty and a populated array alike |
| An array evaluated as false, inverting `&&`, `\|\|` and `!` | `AsBool` reads a populated array as truthy and an empty one as falsy, matching Redisearch on both |
| `abs`/`ceil`/`exp`/`floor`/`log`/`log2`/`sqrt` and `lower`/`upper` mapped over the elements | Redisearch accepts these and collapses the array — `nan` for the numeric ones, nil for the case-folding ones |
| A nil-valued APPLY or REDUCE output vanished from the reply | Redisearch names the field and replies nil. A `Nil` already carries a reason string clients never see, so it now records which nil it is: an unwritten record field reads as `kMissing` and stays out of the reply, every expression-produced nil is named with a nil value. Not array-specific — `APPLY lower(@n1)` over a numeric field dropped the field the same way |
| GROUPBY keyed on the whole array, merging groups that collected equal arrays | Redisearch treats an array key as a multi-value field; array-valued key components now expand to the cartesian product of their elements, bounded by the new `--max-group-key-expansion` (default 65536 keys per record) |
| `MIN`/`MAX` returned the array itself, `STDDEV` returned 0 | Redisearch reads an array as 0 for MIN/MAX and spreads it across the sample for STDDEV |

No compatibility gate guards any of this: ARRAY is new in #932, so there is no
released behavior to preserve.

## Deliberately not matched

`TOLIST`'s element order is deterministic but arbitrary — hash-table layout,
identical across redis-stack and Redis 8.8.1 and across restarts, but not
sorted, not insertion order, and it shifts when the ingest order changes or a
duplicate key is added. Two families of query depend on it and are left out of
the generator with a comment saying why: `SORTBY @array` with `LIMIT`/`MAX`, and
ordered comparison of two arrays. `integration/compatibility/array_input_compatibility.md`
records the measurements behind that call, along with the element-wise broadcast
that survives only where Redisearch rejects the query outright, and a few
divergences found on the way that are out of scope here.

## Testing

- Unit tests pass, including new coverage for array collapse and multi-value GROUPBY expansion
- Full integration suite: 265 passed, 8 skipped
- All three compatibility answer files pass, regenerated against `redis/redis-stack-server`

https://claude.ai/code/session_01PmQBtpNfmhS3GQqQiX3qqz


Aksha1812:
**Reviewers for this PR**

testing/ft_aggregate_exec_test.cc
```diff
@@ -271,6 +271,64 @@ TEST_F(AggregateExecTest, GroupTest) {
...
+        arr.emplace_back(e);
+      }
+      return expr::Value(std::move(arr));
+    };
+    rec->fields_[0] = to_value(n1);
+    rec->fields_[1] = to_value(n2);
+    return rec;
+  };
+
+  // {1,2} and {2,3} share the element 2, so the two records join five groups.
```
neerajr0: Shouldn't there be 3 groups? That's what the test is asserting below.

integration/compatibility/generate_array.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def test_apply_string_funcs(self, key_type):
+        self.setup_data(DATASET, key_type)
+        for fn in ["lower", "upper", "strlen", "exists"]:
+            self._apply(key_type, f"{fn}(@items)")
+            self._apply(key_type, f"{fn}(@sitems)")
+        self._apply(key_type, "substr(@sitems,0,3)")
+        self._apply(key_type, "substr(@items,0,1)")
+        # Array in a position argument rather than the subject.
+        self._apply(key_type, "substr(@sitems,@items,2)")
```
neerajr0: Nit: These repeated `_apply`s can be refactored.

src/commands/ft_aggregate_exec.cc
```diff
@@ -245,9 +300,16 @@ class Count : public GroupBy::ReducerInstance {
...
 };
 
+// MIN and MAX are numeric: Redisearch reads an array as the number 0 rather
+// than letting it become the reducer's result.
+static expr::Value NumericReducerArg(const expr::Value &value) {
+  return value.IsArray() ? expr::Value(0.0) : value;
+}
+
 class Min : public GroupBy::ReducerInstance {
   expr::Value min_;
```
neerajr0, line 401: `min_`, `max_` and `result_value_` use the default constructor which gives `Nil("missing")`. If a group has only nil inputs and no record updates the default object, `GetResult` returns this "missing". `ReplyWithValue` (`ft_aggregate.cc:164`) then calls `IsMissing()` → true and drops the reducer field from the reply entirely, instead of serializing it as a named nil value.

Fix: change `GetResult()` to return a non-missing nil when the sentinel was never updated, e.g.:

```
expr::Value GetResult() const override {
  return min_.IsMissing() ? expr::Value(expr::Value::Nil("no values")) : min_;
}
```

src/expr/value.cc
neerajr0, line 274: This is missing an `IsArray()` branch. In a debug build, if a record contains an array field with `GROUPBY` it will crash at this line: `DBG << "Record: " << *record`.

src/commands/ft_aggregate_exec.cc
```diff
@@ -181,6 +183,43 @@ absl::Status SortBy::Execute(RecordSet &records) const {
...
+    if (!value.IsArray()) {
+      alternatives.emplace_back(value);
+    } else if (value.IsEmptyArray()) {
+      // Nothing to group into: Redisearch keys these as nil.
+      alternatives.emplace_back(expr::Value::Nil("empty array group key"));
+    } else {
+      auto array = value.GetArray();
+      alternatives.assign(array->begin(), array->end());
+    }
+    if (keys.size() * alternatives.size() > max_expansion) {
```
neerajr0: Worth addressing overflow-safe comment below. This can be changed to:

```
if (alternatives.size() > 0 &&
    keys.size() > max_expansion / alternatives.size()) {
```

src/commands/ft_aggregate_exec.cc
```diff
@@ -245,9 +300,16 @@ class Count : public GroupBy::ReducerInstance {
...
+// than letting it become the reducer's result.
+static expr::Value NumericReducerArg(const expr::Value &value) {
+  return value.IsArray() ? expr::Value(0.0) : value;
+}
+
 class Min : public GroupBy::ReducerInstance {
   expr::Value min_;
-  void ProcessRecord(const ArgVector &values) override {
+  void ProcessRecord(const ArgVector &raw) override {
+    ArgVector values{NumericReducerArg(raw[0])};
```
neerajr0: This allocates a container on every `ProcessRecord` just to wrap a single value. Consider replacing with `const expr::Value value = NumericReducerArg(raw[0]);` and using `value` directly.

src/commands/ft_aggregate_exec.cc
```diff
@@ -273,7 +335,8 @@ struct ReducerInstanceVector : GroupBy::ReducerInstance {
 
 class Max : public GroupBy::ReducerInstance {
   expr::Value max_;
-  void ProcessRecord(const ArgVector &values) override {
+  void ProcessRecord(const ArgVector &raw) override {
+    ArgVector values{NumericReducerArg(raw[0])};
```
neerajr0: Same as above, can avoid the allocation using a local `const`.

integration/compatibility/generate_array.py
neerajr0, line 1: Some gaps in test coverage:

- No test for `FILTER` stage with array-valued fields (e.g., `filter @items > 2`, `filter exists(@items)`).
```
FT.AGGREGATE idx "@n1:[-inf inf]"
  GROUPBY 1 @t1 REDUCE TOLIST 1 @n1 AS items
  FILTER "@items > 2"
```
- No test for `FIRST_VALUE` reducer with array arguments.
```
FT.AGGREGATE idx "@n1:[-inf inf]"
  GROUPBY 1 @t1 REDUCE TOLIST 1 @n1 AS items
  GROUPBY 1 @t1 REDUCE FIRST_VALUE 1 @items AS first_arr
```
- No unit test for the `max_group_key_expansion` error path (`ResourceExhaustedError`).
```
// Unit test: two array keys whose product exceeds the limit
// Set max_group_key_expansion to 5, then:
auto record = make_record({1,2,3}, {10,20,30});  // 3*3 = 9 > 5
// Should return ResourceExhaustedError, currently untested
```
- `FormatDouble` boundary at `1e17` not tested — values near the fixed-vs-scientific switch.
```
// These straddle the fixed-vs-scientific switch:
EXPECT_EQ(FormatDouble(99999999999999999.0), "99999999999999999");  // just under 1e17?
EXPECT_EQ(FormatDouble(1e17), ???);  // which path does this take?
EXPECT_EQ(FormatDouble(-99999999999999.0), "-99999999999999");  // negative integral
```
- No unit test for `Stddev::Accumulate` recursive array spreading
```
// Unit test: Stddev reducer receives an array [1, 2, 3]
// Accumulate should spread it into three samples, not treat it as one value
// Also: what happens with nested arrays? [[1,2],[3]]? Non-numeric elements in array?
auto param = MakeStages("groupby 1 @n2 reduce stddev 1 @n1");
// Feed a record where n1 is an array value — no unit test does this currently
```
- No test for chained GROUPBYs where the second GROUPBY also uses TOLIST (TOLIST of an array-valued field exercises the flattening path — does Redisearch flatten the same way?)
```
FT.AGGREGATE idx "@n1:[-inf inf]"
  GROUPBY 1 @t1 REDUCE TOLIST 1 @n1 AS items
  GROUPBY 1 @items REDUCE TOLIST 1 @items AS nested_items
```
- No test for COUNT_DISTINCT verifying whether Redisearch treats an array as a single hashable value or flattens it like GROUPBY keys do
```
FT.AGGREGATE idx "@n1:[-inf inf]"
  GROUPBY 1 @t1 REDUCE TOLIST 1 @n1 AS items
  GROUPBY 1 @t1 REDUCE COUNT_DISTINCT 1 @items AS ncd
```

testing/ft_aggregate_exec_test.cc
allenss-amazon: Yes — the assertion was right and the comment was wrong. `{1,2}` and `{2,3}` yield three distinct groups (1, 2, 3); "five" was wrong under either reading, since there are 4 record→group memberships and 3 groups.

Fixed in `d9ddaae`; it now reads "the two records join three distinct groups: 1, 2 and 3."

integration/compatibility/generate_array.py
allenss-amazon: Done in `f117a09`. The fourteen calls are now a module-level `STRING_FUNC_EXPRS` table, matching how `MATH_FUNCS` and `ARRAY_FUNCS` are already declared, and the body is a two-line loop. Call order is preserved so the regenerated answers stay diffable.

src/commands/ft_aggregate_exec.cc
allenss-amazon: Good catch, and it was in scope for a reason worth recording: `87f2ddf` **in this PR** introduced the `IsMissing()` distinction. Before it every nil was dropped from the reply, so the reducers' default sentinels were consistent; the PR created the two-tier semantics and didn't update them.

Rather than assume the target behavior I added generator cases and measured against redis-stack-server. That changed the fix:

| reducer, group where every input was nil | RediSearch |
| --- | --- |
| `MIN` / `MAX` | **`0`** — numeric field and string field alike |
| `SUM` / `AVG` / `STDDEV` / `COUNT_DISTINCT` | `0` (already matched) |
| `TOLIST` | `[]` (already matched) |
| `FIRST_VALUE` | **named nil** |

So the suggested `Nil("no values")` is right for `FIRST_VALUE` but wrong for `MIN`/`MAX`, which need their numeric identity `0`. Both are fixed in `2eedf1e`.

The measurement also turned up a second defect on the same lines: `MIN`/`MAX` over a **populated** string group diverged too — we answered `Banana` by byte order where RediSearch answers `0`, because these reducers are strictly numeric. `NumericReducerArg` now keys off `AsDouble()` rather than the variant type, so hash fields (which arrive as strings) stay numeric and the comparison stops being lexicographic.

All of it is gated on `search.emulate-release` 1.3.0 as `compatibility-reduce_minmax_numeric`, with a row in COMPATIBILITY.md.

src/expr/value.cc
allenss-amazon: Confirmed and fixed in `d9ddaae`. `GroupKey::operator<<` streams each element, so once `ExpandGroupKeys` puts arrays in a key the `CHECK(false)` was reachable — only with `DBG` enabled, but that is precisely when someone is debugging this feature. Arrays now render recursively, so nested arrays work too, covered by `ValueTest.StreamArrayValue` (flat, empty, nested).

One note for the record: CodeRabbit filed the same finding separately and auto-resolved it with "✅ Addressed in commits 41a0158 to 8bce2e7". That was false — `git log -S "IsArray()" -- src/expr/value.cc` shows no commit added the branch until `d9ddaae`.

src/commands/ft_aggregate_exec.cc
allenss-amazon: Done in `d9ddaae`, in the division form you suggested. It also makes the `reserve(keys.size() * alternatives.size())` two lines below provably non-wrapping.

I dropped the `alternatives.size() > 0` half of the guard: `alternatives` is never empty, since every branch above pushes at least one element. Said so in a comment rather than coding around it.

Being straight about severity: this is practically unreachable. `keys.size()` is capped at `max_expansion` by the prior iteration and that config maxes at `UINT_MAX`, so wrapping `size_t` would need ~4.3 billion `expr::Value`s in a single document. Worth the one line anyway.

src/commands/ft_aggregate_exec.cc
allenss-amazon: Applied in `d9ddaae` — both `Min` and `Max` now use `const expr::Value value = NumericReducerArg(raw[0]);`.

The premise isn't quite right though: `ArgVector` is `absl::InlinedVector<expr::Value, 4>` (`ft_aggregate_parser.h:35`), so a one-element instance lives entirely inline and there is no heap allocation to avoid. The change is worth making for clarity, just not for performance.

src/commands/ft_aggregate_exec.cc
allenss-amazon: Applied in `d9ddaae` along with `Min` — see the note on the previous comment: `ArgVector` is an `absl::InlinedVector<expr::Value, 4>`, so the temporary was inline rather than a heap allocation. Still clearer as a local `const`.

integration/compatibility/generate_array.py
allenss-amazon: Worked through all seven. Five were real and are now covered; two were already covered before the review.

| # | gap | status |
| --- | --- | --- |
| 1 | `FILTER` with array-valued fields | `test_filter_array`, `test_filter_empty_array` (`f117a09`) |
| 2 | `FIRST_VALUE` with array arguments | `test_first_value_over_array` (`f117a09`) |
| 3 | `max_group_key_expansion` error path | `GroupByArrayKeyExpansionLimitTest` (`f117a09`) — over, under, and exactly at the limit |
| 4 | `FormatDouble` 1e17 boundary | `FormatDoubleIntegralUsesFixedNotation` (`f117a09`) |
| 5 | `Stddev::Accumulate` array spreading unit test | not added — see below |
| 6 | chained GROUPBY where the second also uses TOLIST | already covered (`groupby 1 @t1 reduce tolist 1 @items as flat`) |
| 7 | `COUNT_DISTINCT` over an array | already covered (`groupby 1 @t1 reduce count_distinct 1 @items as ncd`) |

On #4 — the suggested assertion would have failed. `99999999999999999.0` is not representable and rounds to exactly `1e17`, so it takes the scientific path; the largest double under the threshold is `99999999999999984`. The test pins that, plus `1e16` and `1e17` either side.

On #5 — deferred deliberately. The array-spreading behavior *is* verified end to end by `reduce stddev 1 @items` against real RediSearch, which is stronger evidence than a unit test. Your two follow-on questions (nested arrays, non-numeric elements inside an array) are genuinely uncovered and can only be reached from a unit test, so that is the shape it should take if picked up.

This was the most productive comment on the PR. Chasing it turned up that the pre-existing `test_empty_array` family never tested missing values at all: it runs `TOLIST` first, so the stage under test sees a *present* empty array. A new missing-field family across every stage then found real divergences in `!` (we answered `true` for an absent field), in every function proxy, and in GROUPBY key replies — all now fixed and gated on `search.emulate-release` 1.3.0.

It also exposed four defects in the compatibility harness itself, each of which was hiding information:

- rows could not sort on a key absent from some of them
- a nil in a numeric field compared unequal to itself, because the `n*` tolerance path calls `float(None)`
- a partial or failed generator run silently overwrote its answer file with a short one (a flaky docker start once cut text-search from 42612 answers to 11000)
- the generator's fixed container name and port let two runs on one machine destroy each other

Frank-Gu-81:

integration/compatibility/generate_array.py
neerajr0: Curious how addressing the test coverage gaps led to the discovery of the compatibility harness defects. Is there anything that needs a second look there?

neerajr0 (approved):

integration/compatibility/generate_array.py
allenss-amazon: Well, the last two were just a workflow thing. I was working on two different PRs at the same time. They are only a dev bug, not something that will ever be seen in production.

The numeric nil is genuine oversight on my part.

PR #1276, opened 2026-07-28
feat: overload protection - 2 mechanisms to prevent an ever growing query queue 

KarthikSubbarao:
1. IsCancelled check at top of Search() before mutex acquisition
   - Dead queries rejected instantly without wasting a reader mutex slot

2. Queue depth rejection before fan-out (search.max-query-queue-depth config)
   - Reject queries before creating N partition tasks when queue is too deep
   - Prevents the death spiral at the source
   - the Queue depth based rejection BEFORE fanout is set to 100K per node. But the check is done in both the local node path and the handling of the remote requests from fanouts


src/query/fanout.cc
```diff
@@ -315,6 +315,15 @@ absl::Status PerformSearchFanoutAsync(
     coordinator::ClientPool *coordinator_client_pool,
     std::unique_ptr<SearchParameters> parameters,
     vmsdk::ThreadPool *thread_pool) {
+  // Reject queries before fan-out if the reader thread pool queue is too deep.
```
allenss-amazon: When it comes to limiting the local fanouts. I think the limit needs to be adjusted by cluster size. Ultimately the number of queries outstanding could be as much as this local limit * number of nodes in the cluster (not shards, nodes). 

I think the limit needs to be adjusted based on the cluster size. The simplest scheme would be to simply divide the limit by the #of shards or # of nodes (both are cheaply available via the ClusterMap object).

I am concerned that perhaps this limit is too much for very large clusters. So a compromise would be to provide a floating-point parameter in the range 0..1 (or scaled to 0..100 -- same thing). When 0, you get the computation above, i.e., a fixed limit. When 1, you get the proposed fixed limit / # of nodes. For values in between, you get some blending. This means that the per-node pre-fanout queue size check is decreased sub-linearly.

Also, I have a concern about using the local thread pool size as the thing-to-limit. Isn't it the case that a coordinator that's issued a fan-out operation may schedule no local threads, meaning it shows up as 0 in the local thread pool. Even if it does schedule a local thread, that thread may complete long before the remainder of the operations do. I suspect that the correct thing is to keep a count of coordinator operations that are in-flight as the thing to limit.

src/query/search.cc
```diff
@@ -842,6 +842,13 @@ SerializationRange SearchResult::GetSerializationRange(
 }
 
 absl::Status Search(SearchParameters &parameters, SearchMode search_mode) {
+  // Reject already-cancelled queries before acquiring the time-slice mutex.
+  // Without this, expired queries that sat in the queue still acquire a reader
+  // slot and waste mutex time before discovering they're cancelled deep in the
+  // iteration loop.
+  if (parameters.cancellation_token->IsCancelled()) {
+    return absl::CancelledError("Query timed out while waiting in queue");
+  }
```
KarthikSubbarao, line 854: I am thinking about this....
If we are already draining the queue prior to task execution on the reader thread, then this may not be needed. There is a small benefit to this, but it will catch some cases.

IMO, we should have just one or the other: draining at the worker thread OR this simple check prior to timeslicemutex acquisition

I am going for the latter for now

src/valkey_search_options.cc
```diff
@@ -225,6 +225,25 @@ constexpr absl::string_view kEnableConsistentResults{
...
                                                  false);
 
+/// When enabled, local shard queries during fan-out get kMax priority in the
+/// thread pool, ensuring they complete even under overload. This makes
+/// partial-results load shedding work correctly.
+constexpr absl::string_view kLocalSearchMaxPriority{
+    "local-search-max-priority"};
+static auto local_search_max_priority =
+    config::BooleanBuilder(kLocalSearchMaxPriority, false)
+        .Dev()
```
KarthikSubbarao: Any thoughts on this being Dev?

IMO, if it is Dev and false, it will remain unused

src/valkey_search_options.cc
```diff
@@ -225,6 +225,25 @@ constexpr absl::string_view kEnableConsistentResults{
...
+        .Build();
+
+/// Maximum reader thread pool queue depth before rejecting new queries.
+/// When the queue exceeds this threshold, FT.SEARCH is rejected before fan-out
+/// to prevent cascading timeouts. 0 = unlimited (disabled).
+constexpr absl::string_view kMaxQueryQueueDepth{"max-query-queue-depth"};
+static auto max_query_queue_depth =
+    config::NumberBuilder(kMaxQueryQueueDepth, 0, 0, INT_MAX)
+        .Dev()
+        .Build();
```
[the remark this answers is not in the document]
KarthikSubbarao: Any thoughts on this being Dev?

IMO, if it is Dev and false, it will remain unused

src/query/fanout.cc
KarthikSubbarao: Thanks - that is a good idea. I am planning on making these configs the default and not dev since they can be released in the next version of valkey-search (1.3).

If they are dev configs, I fear it will just be unused and dead code (besides tests). 

In order to add this scaling-factor, we will need a new config. By default, should we have the scaling-factor be 0, meaning a queue-depth of 1000 means each node in the cluster allows 1000 as the queue size? Or should the scaling-factor be 100, so that we ensure it is skewed based on the cluster size by default?

src/valkey_search_options.cc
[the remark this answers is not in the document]
allenss-amazon: Making the query limit App is fine with me.

src/query/fanout.cc
allenss-amazon: We directly support floating point configs. So the scaling factor isn't required to make it fit into the configuration system. I don't care what the scaling factor is.

I do believe that the default should be close to 100%, i.e., default queue limit should be <limit> / # of nodes.

src/query/search.cc
allenss-amazon: What's the downside of having this test?

src/valkey_search_options.cc
allenss-amazon: Hmm, I'd be worried that this makes things worse, not better. Do we have evidence that this helps?

KarthikSubbarao: From tests with `local-search-max-priority=yes` (and when partial results are enabled), there were zero timeouts vs 668 with the config set to `no`. The fix was tested with delay on primaries only (simulating slow remote shards), queries sent to a replica (coordinator). Local shard completed instantly via kMax, returning partial results.

When it helps: Overload scenario where everything is timing out anyway. Getting partial results back is better than zero results.

When it hurts: Moderate load where remote partition requests WOULD complete in time, but get starved or rather delayed by local kMax priority.

This is why the config is set to false/no by default. It is opt in only

src/query/search.cc
KarthikSubbarao: I am keeping the check here and it occurs prior to the timeslice mutex reader lock being acquired

src/query/fanout.cc
```diff
@@ -315,6 +315,15 @@ absl::Status PerformSearchFanoutAsync(
...
+  // This prevents the death spiral where fan-out amplifies overload (each query
+  // creates N partition tasks that will never complete). Failing fast here means
+  // the customer gets a clear error and the system stays out of the hole.
+  auto max_depth = options::GetMaxQueryQueueDepth().GetValue();
+  if (max_depth > 0 && thread_pool->QueueSize() > static_cast<size_t>(max_depth)) {
+    return absl::ResourceExhaustedError(
+        "Search query queue depth exceeded");
+  }
```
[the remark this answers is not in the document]
KarthikSubbarao: @greptile-apps The queue depth check is an admission gate at the entry point — it decides whether to START a new query. Once a query is admitted, its partition work (including remote requests sent to other nodes) is committed work that must complete. If you reject remote partition requests mid-flight, the coordinator that already admitted the query gets a partial failure for a query it promised to execute. That's worse behavior than letting the work complete.

The overload protection model is: reject NEW queries at the door (commands.cc), don't yank the rug from queries already in progress. The "previously reported overload path" is already mitigated — the queue can't grow unbounded because new queries are rejected once it hits 10000. The remote partition requests in flight are bounded by the number of queries that were admitted BEFORE the limit was reached.

src/valkey_search_options.cc
```diff
@@ -225,6 +225,24 @@ constexpr absl::string_view kEnableConsistentResults{
 static config::Boolean prefer_consistent_results(kEnableConsistentResults,
                                                  false);
 
+/// Maximum reader thread pool queue depth before rejecting new queries.
+/// When the queue exceeds this threshold, FT.SEARCH is rejected before fan-out
+/// to prevent cascading timeouts. 0 = unlimited (disabled).
+constexpr absl::string_view kMaxQueryQueueDepth{"max-query-queue-depth"};
+constexpr int kDefaultMaxQueryQueueDepth{10000};
```
BCathcart: If a customer is running tons of queries on a tiny corpus on a huge machine, could 10,000 queries actually be processed pretty quickly? If we want to get more refined, we could track the average query RPS throughput over some short time and use that to estimate whether this query will time out or not. If we have high confidence it will time out, then we could reject it here.

I'm wondering though, why is there so much overhead in handling a timed out query? If it gets to the front of the queue and it's already timed out, it should be easy to throw away. Will rejecting already-cancelled queries before acquiring the time-slice mutex make a big difference here? If it times out while doing the search and partial results aren't enabled, then I can see there being a bunch of wasted work. So are we just trying to protect against the case where the queue depth sits at the point where each query isn't timed out when it begins processing but times out part way through?

BCathcart: Do we need to tag this PR as a breaking change?

src/valkey_search_options.cc
```diff
@@ -225,6 +225,24 @@ constexpr absl::string_view kEnableConsistentResults{
...
+static auto max_query_queue_depth =
+    config::NumberBuilder(kMaxQueryQueueDepth, kDefaultMaxQueryQueueDepth, 0,
+                          INT_MAX)
+        .Build();
+
+/// Controls how much the queue depth limit scales with cluster size (0.0-1.0).
+/// 0.0 = fixed limit (ignore cluster size). 1.0 = divide limit by shard count.
+/// Values in between blend sub-linearly.
```
BCathcart: Is this needed because the coordinator needs to do more work to gather results the more shards there are?

src/valkey_search_options.cc
BCathcart: As discussed offline, I think quickly throwing away timed out queries is the most important thing to prevent a situation where all queries are timing out and the queue is continuing to grow. Timed out queries should be drained significantly faster than new queries can be added. From a memory resource protection perspective, I definitely agree we should have a limit on the queue size, but I think erring on the side of too large is the way too go so we don't accidentally limit throughput.

src/valkey_search_options.cc
KarthikSubbarao: It was needed before. But I added another change later on to protect on the handling of the remote requests from other nodes

BCathcart (approved):

src/valkey_search_options.cc
KarthikSubbarao: The main limit here is to avoid an unbounded growth in the query queue since we did not have any limit up till now

The other fix with the performing the check before the timeslice mutex acquirig + heavy search will help with faster timeouts without slowing down other requests

Issue #1275, opened 2026-07-28
[BUG] search_rdb_indexes_restored_percent reports incorrect value when RDB load is done


KarthikSubbarao:
**Describe the bug**
```
# search_rdb
search_rdb_current_index_keys_restored_percent:100
search_rdb_indexes_restored_percent:66.666666666666671
search_rdb_load_failure_cnt:0
search_rdb_load_success_cnt:1
search_rdb_restore_backpressure_wait_cycles:3302702
search_rdb_restore_in_progress:0
search_rdb_save_failure_cnt:0
search_rdb_save_success_cnt:0
```

search_rdb_indexes_restored_percent reports incorrect value when RDB load is done

Nivesh-01:
This seems to be a bug that happens in CME because the total gets updated with the global metadata as well according to the initial analysis.

Nivesh-01:
PR for the fix - https://github.com/valkey-io/valkey-search/pull/1277

PR #1277, opened 2026-07-29
rdb restore percent metric bug fix

Nivesh-01:
In cluster/coordinator (CME) mode, `search_rdb_indexes_restored_percent` never reaches 100% after a successful RDB restore. For example, with 2 indexes it reports `66.67%` instead of `100%`.

The root cause: `PerformRDBLoad` counts all RDB sections (including the `GLOBAL_METADATA` section saved by `MetadataManager`) in the total, but only `SchemaManager::LoadIndex` increments the completed counter (`rdb_restore_completed_indexes`). The `MetadataManager::LoadMetadata` callback never increments it, so the final percentage is `completed/total = 2/3 = 66.67%`.

This does not affect standalone mode (no coordinator), where only `INDEX_SCHEMA` sections exist.

Fix
Added `Metrics::GetStats().rdb_restore_completed_indexes++` at the end of `MetadataManager::LoadMetadata()`, mirroring the existing pattern in `SchemaManager::LoadIndex()`. This ensures every successfully-loaded RDB section counts toward the completion percentage.

issue: https://github.com/valkey-io/valkey-search/issues/1275


integration/test_rdb_indexes_restored_percent.py
```diff
@@ -0,0 +1,174 @@
...
+            print(f"[Standalone] Progress samples: {observed_percents}")
+
+        # Verify no errors observed during monitoring
+        assert monitor_error[0] is None, monitor_error[0]
+
+        # Verify the metric progressed monotonically (non-decreasing)
+        if len(observed_percents) > 1:
+            for i in range(1, len(observed_percents)):
+                assert observed_percents[i] >= observed_percents[i-1], \
+                    f"Metric decreased: {observed_percents[i-1]} -> {observed_percents[i]}"
```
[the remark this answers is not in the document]
mohanrajendran:
```suggestion
        from valkey.exceptions import ValkeyError

        monitor_ready = threading.Event()
        done = threading.Event()
        observed_percents = []
        monitor_error = [None]
      
        def monitor():
            try:
                # Readiness handshake.
                monitor_client.info("search")
                monitor_ready.set()
      
                while not done.is_set():
                    info = monitor_client.info("search")
                    if info.get("search_rdb_restore_in_progress", 0):
                        pct = float(info["search_rdb_indexes_restored_percent"])
                        observed_percents.append(pct)
      
                        if pct > 100.0:
                            monitor_error[0] = f"Metric exceeded 100%: {pct}"
                            return
      
                    done.wait(0.001)
            except ValkeyError as exc:
                monitor_error[0] = exc
                monitor_ready.set()
      
        thread = threading.Thread(target=monitor, daemon=True)
        thread.start()
      
        assert monitor_ready.wait(timeout=5)
      
        try:
            self.client.execute_command("DEBUG", "RELOAD")
        finally:
            done.set()
            thread.join(timeout=5)
      
        assert not thread.is_alive()
        assert monitor_error[0] is None, monitor_error[0]

        # Verify the metric progressed monotonically (non-decreasing)
        if len(observed_percents) > 1:
            for i in range(1, len(observed_percents)):
                assert observed_percents[i] >= observed_percents[i-1], \
                    f"Metric decreased: {observed_percents[i-1]} -> {observed_percents[i]}"
```

mohanrajendran (approved):

yairgott (commented):
Once the coderabbitai comment is addressed, I'll approve this PR.

src/coordinator/metadata_manager.cc
```diff
@@ -683,6 +683,9 @@ absl::Status MetadataManager::LoadMetadata(
                                             /*trigger_callbacks=*/false,
                                             /*prefer_incoming=*/true));
   }
+
+  // Increment completed section counter for restore progress tracking
+  Metrics::GetStats().rdb_restore_completed_indexes++;
```
BCathcart, line 688: #1265 already added this in `SchemaManager::LoadIndex`

integration/test_rdb_indexes_restored_percent.py
```diff
@@ -0,0 +1,188 @@
+"""
+Verify search_rdb_indexes_restored_percent reports 100% after successful RDB
+restore in both standalone (CMD) and cluster/coordinator (CME) modes.
+
+This metric was previously broken in CME mode because the GLOBAL_METADATA RDB
+section was counted in the total but never incremented the completed counter,
+causing it to report 66.67% with 2 indexes (2/3 sections completed).
```
BCathcart, line 7: #1265 incremented `rdb_restore_completed_indexes` in the diskless path which previously bypassed it. This sounds like a different framing. Does it also address the problem you see here?

BCathcart:
>The root cause: PerformRDBLoad counts all RDB sections (including the GLOBAL_METADATA section saved by MetadataManager) in the total, but only SchemaManager::LoadIndex increments the completed counter (rdb_restore_completed_indexes). The MetadataManager::LoadMetadata callback never increments it, so the final percentage is completed/total = 2/3 = 66.67%.

It sounds like you may be mispurposing `rdb_restore_completed_indexes` to be the number of RDB sections and not the number of indexes (or is it just poorly named?). Maybe you need to re-evaluate the inputs you are using to the percentage calculation.

integration/test_rdb_indexes_restored_percent.py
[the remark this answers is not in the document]
Nivesh-01: Done during the last commit

Aksha1812:
**Reviewers for this PR**

Frank-Gu-81:
Hi @Nivesh-01 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @mnunberg1 — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

src/coordinator/metadata_manager.cc
```diff
@@ -683,6 +683,10 @@ absl::Status MetadataManager::LoadMetadata(
                                             /*trigger_callbacks=*/false,
                                             /*prefer_incoming=*/true));
   }
+
+  /* Deliberately does not touch rdb_restore_completed_indexes. That counter
+   * tracks index schemas only; the total it is compared against comes from
+   * SnapshotInfo::num_indexes, which excludes this section. */
```
BCathcart, line 689: Let's remove this. This is not the place for a comment about the semantics of rdb_restore_completed_indexes.

src/rdb_section.proto
```diff
@@ -25,6 +27,18 @@ enum RDBSectionType {
   RDB_SECTION_UNSET = 0;
   RDB_SECTION_INDEX_SCHEMA = 1;
   RDB_SECTION_GLOBAL_METADATA = 2;
+  RDB_SECTION_SNAPSHOT_INFO = 3;
+}
+
+// Payload-wide facts captured at RDB save time. Unlike the other RDBSection
+// types, this one is owned by the RDB framework itself rather than by a
+// component, and is always written as the first section. Intentionally
+// minimal; the containing oneof makes adding fields backward compatible.
```
BCathcart, line 36: Can you explain the connection between oneof and backwards compatibility?

integration/test_rdb_indexes_restored_percent.py
```diff
@@ -0,0 +1,530 @@
...
+            "Expected per-index key restore to report 100% after restore"
+        assert info["search_number_of_indexes"] == 2, \
+            f"Expected 2 indexes after restore, got {info['search_number_of_indexes']}"
+
+    def test_snapshot_info_tracks_actual_index_count(self):
+        """
+        The recorded count must follow the real number of index schemas rather
+        than a fixed value or the section count. Saving at two different index
+        counts and re-reading the RDB each time pins that down.
+        """
```
BCathcart, line 340: Is a separate test for this necessary?

integration/test_number_of_indexes_on_restore.py
```diff
@@ -59,3 +59,10 @@ def monitor():
         waiters.wait_for_true(lambda: index_1.backfill_complete(self.client))
         waiters.wait_for_true(lambda: index_2.backfill_complete(self.client))
         assert self.client.info("search")["search_number_of_indexes"] == 2
+
+        # Verify rdb_indexes_restored_percent is 100% after restore completes
+        info = self.client.info("search")
+        assert info["search_rdb_restore_in_progress"] == 0
+        percent = float(info["search_rdb_indexes_restored_percent"])
+        assert percent == 100.0, \
+            f"Expected search_rdb_indexes_restored_percent=100.0 after restore, got {percent}"
```
BCathcart, line 68: Just making a note that the changed touched on [here](https://github.com/valkey-io/valkey-search/pull/1312/changes#r3981205790) in #1312 would break this unless unless we use a pause point to prevent the RDB load from full completing

integration/test_rdb_indexes_restored_percent.py
```diff
@@ -0,0 +1,530 @@
...
+#
+# RDBSection.type is field 1 (varint), and RDB_SECTION_SNAPSHOT_INFO is enum
+# value 3. RDBSection.supplemental_count is field 2 and is zero for this
+# section, so proto3 omits it. RDBSection.snapshot_info_contents is field 5 and
+# is a nested message, giving wire type 2. Protobuf serializes in ascending
+# field-number order.
+_RDB_SECTION_TYPE_TAG = 0x08          # field 1, varint
+_RDB_SECTION_SNAPSHOT_INFO = 0x03     # RDB_SECTION_SNAPSHOT_INFO
+_SNAPSHOT_INFO_CONTENTS_TAG = 0x2A    # field 5, wire type 2 -> (5 << 3) | 2
+_NUM_INDEXES_TAG = 0x08               # SnapshotInfo.num_indexes, field 1 varint
```
BCathcart, line 41: Can you summarize your testing philosophy for me? This looks like it's getting pretty low-level and complex.

The testing I think is needed off the top of my head is:
1. v2 with SnapshotInfo save/load
2. v2 without SnapshotInfo save, v2 with SnapshotInfo load - i.e. backwards compatibility
3. v2 with SnapshotInfo save, v2 without SnapshotInfo load -  i.e. forwards compatibility across patch versions. I think we should fix this in a 1.2 patch but if we only fixed it in 1.3 we wouldn't even need this test.

Now for the third test it gets a bit tricky unless we commit an older libsearch.so binary for use in our testing. I would think this is a fine approach if it wasn't for the fact that libsearch.so is so large. Another option would be to download an old binary at test time.

BCathcart: Actually I see we've already established adding a old search binary under `/integration/module/` so I think we can just continue that pattern for now

src/rdb_section.proto
Nivesh-01: You're right that the comment was wrong, though the compatibility goal behind it is real — I want a 1.2 reader to still load a 1.3-written RDB.

The oneof isn't what provides that. What does is: proto3 open enums preserve the unknown type = 3, unknown field 5 is skipped, the loader's else branch drains unrecognized sections, and — the actual lever — SNAPSHOT_INFO deliberately doesn't feed min_version, so an older reader doesn't reject the RDB with require minimum version. It just sees one extra section, which inflates its rdb_restore_total_indexes by one.

The oneof is there because a section carries exactly one payload, matching index_schema_contents and global_metadata_contents; and the presence I rely on (has_snapshot_info_contents() staying true when num_indexes == 0, since proto3 drops the zero on the wire) comes from it being a message field, not from the oneof. Rewording to say all that, and moving the compatibility note next to the min_version decision where it belongs.

src/coordinator/metadata_manager.cc
Nivesh-01: agreed removing it.

integration/test_rdb_indexes_restored_percent.py
Nivesh-01: Fair, it doesn't need to stand alone. The only thing it adds over the round-trip test is that the recorded number moves with the real index count instead of being a constant, so I'll fold a second save at a different count into the round-trip test and drop this one.

integration/test_rdb_indexes_restored_percent.py
Nivesh-01: Agreed on both. Dropping the byte-level parsing — it asserts on wire encoding that protobuf is meant to abstract, and it breaks on any field renumbering. I was reaching for "prove the section is really in the file with the real count, independent of our own load path," but reading it back through the load path plus the metric gives that more robustly.

Building your matrix:

v2 save → v2 load. Round trip, with a second save at a different index count folded in, which lets me delete the standalone test you flagged above.
RDB without SnapshotInfo → v2 load. No old binary needed here — I'll commit a fixture RDB as base64, same pattern as test_skip_index_load.py. More deterministic than driving an old binary.
v2 save → old reader. Agreed on continuing the /integration/module/ pattern with a zipped binary. Worth recording what this actually verifies: the old loader takes the unknown-section branch, logs Ignoring unknown RDB section with type, drains the supplemental content, and doesn't reject the RDB because SNAPSHOT_INFO deliberately doesn't contribute to min_version. Its only degradation is total being inflated by one. So the assertion is "still loads clean."
I remeber we had discussed about this as well that we have blocker from loading the RDB when there is differnt minor or major version, but allow if there is a difference in R-patch. Since we are adding this to 1.3, I can remove this test since it will not really yield anything of significance, let me know your thoughts on this.

One blocker to raise on case 3: the 1.0.0 fixture we already ship can't be loaded on an older dev host. A direct dlopen gives libssl.so.3: cannot open shared object file, plus missing GLIBC_2.29–2.38 and 
3.4.32 — it's built against a modern userspace. So test_rdb_load_on_module_v1_0 fails locally for anyone on AL2, burning 90s in wait_for_logfile first, and a second binary test doubles that. The file already skips for unsupported arch and ASAN, so I'd add a loadability guard in the same spirit — probe the .so, skip with a clear reason if it won't load. Separate small PR since it's independent of this change. I mentioned this to you before as well, just did not tell the name of the test that time since I figured this always fails on local.

integration/test_number_of_indexes_on_restore.py
Nivesh-01: What I'm proposing: pull #1312's in_progress gate and the unsigned-wrap guard into this PR, and drop the counter zeroing from #1312. Two separate reasons — one a straight correctness fix, one a semantics decision I'd like your call on.

The correctness part. My change makes rdb_restore_total_indexes correct, but it doesn't stop a partially failed load from leaving completed < total behind permanently (this is the bug that I was talking about with you before) — the cleanup on this branch only clears the flag, and nothing reads the flag. So number_of_indexes stays inflated by the shortfall until the next load. #1312's in_progress gate is what actually fixes #1289, and it covers a case SnapshotInfo alone can't. Same for the unsigned-wrap guard: std::max(a - b, uint64_t{0}) wraps before max sees it, so it's a no-op today. Both belong here regardless of anything else.

The semantics part. Once the residual is gated, zeroing the counters buys nothing for #1289 — so the question becomes what these counters should mean at rest, decided on its own merits rather than as a side effect.

I'd argue for retaining all four (total, completed, and both current_index_keys_*, since the zeroing wipes those too). Retention makes the progress fields report last-restore state, consistent with rdb_last_restore_aux_load_duration_ms three fields above them, which deliberately does exactly that. More importantly it preserves signal: with zeroing, rdb_indexes_restored_percent reads 100.0 both for a clean restore and for one that died at 1 of 3, because the function early-outs on total == 0 (valkey_search.cc 336 - 345). An operator inspecting a replica that failed to sync sees 100% either way. With retention they see 33%, which is the truth.

That information loss is also why the assertion in this file goes quiet — it stops being able to tell success from failure for the same reason an operator can't. I'd treat that as evidence about the metric rather than as the thing to fix.

On the pause point — I don't think one is available. vmsdk::debug::PausePoint opens with CHECK(!IsMainThread()) (debug.cc 47), and PerformRDBLoad is main-thread-bound: LoadIndex writes staged_db_to_index_schemas_ (schema_manager.cc 718), a MainThreadAccessGuard (schema_manager.h 170) whose accessors call VerifyMainThread() (utils.h 57, class at :72). It would abort the server rather than pause it — which is why block_mutation_queue and mutation_processing work, both being on writer-pool threads.

Polling doesn't substitute. INFO is servable mid-load (LoadIndexExtension yields per key at 
index_schema.cc 1634).

If you'd rather keep the zeroing, that's a defensible call — it just means accepting that the restore fields are live-only and carry no post-mortem signal. In that case there's nothing meaningful left to assert at rest, since neither counter is exposed in INFO directly, so I'd move the completed → total convergence check into rdb_serialization_test.cc, which already pins the total across six RDB shapes but never asserts convergence. Better than adding INFO fields to serve a test.

Which way would you prefer?

Nivesh-01: As discussed if we make this change it will be a new CR so resolving this as of now.

PR #1279, opened 2026-07-31
RFC for multi-language support in FTS

neerajr0:
Request for Comments for Multi-language support feature to introduce 12 new languages for `TEXT` fields. See https://github.com/valkey-io/valkey-search for full feature PR.


Aksha1812:
**Reviewers for this PR**

- **First Pass Reviewer: @Aksha1812** — Please do your best to do a detailed review on the PR and get a response on your feedback. Once the first pass is done, notify the maintainer assigned to this PR to follow up on the final review and getting the PR merged. You can reach out to the people owning the relevant code paths for more help on the review.

neerajr0:
This PR needs updating after implementation changes in https://github.com/valkey-io/valkey-search/pull/1263. Converting back to draft until RFC is up to date.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+
+The existing English-only Full-Text Search (FTS) feature is extended to support
+11 additional languages: French, German, Spanish, Italian, Portuguese,
+Russian, Swedish, Turkish, Dutch, Indonesian, and Arabic. The current
+English Lexer class is replaced with a polymorphic `Language` interface —
+each language is a first-class, self-contained object implementing
+segmentation, normalization, stemming, and stop-word filtering. The
+implementation (`SnowballLanguage`) is wired to Snowball's per-language
+stemmers, punctuation sourced from the Unicode Common Locale Data Repository,
+Apache Lucene-derived stop words, and ICU-backed Unicode normalization and case
```
Aksha1812, line 18: Since ICU is a new hard dependency , have considered how it may affect build , and if it may increase the base memory footprint cost of the search binary . Also if we need all components from the ICU bundle , if not have we made efforts to trim some unused components ? If we considered any alternatives . ? also on a separate note if you have before vs after '.so' binary size coomparison / rss memory delta . not sure if all this was already considered , but don't see it apparrently in the rfc .

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+
+HSET doc:ar:1 body "المكتبة"
+    # basic Arabic letters:
+    # U+0627 U+0644 U+0645 U+0643 U+062A U+0628 U+0629
+
+HSET doc:ar:2 body "ﺍﻟﻤﻜﺘﺒﺔ"
+    # same word using Arabic Presentation Forms-B:
+    # U+FE8D U+FEDF U+FEE1 U+FEDA U+FE97 U+FE91 U+FE93
+
+# NFKC normalizes the presentation-form codepoints to their basic-letter
```
Aksha1812, line 688: AI pointed this out . i have less context but looks like it needs to be addressed

> NFKC handles the presentation-form case shown here, but the Arabic story stops short of what production Arabic text actually needs. NFKC does not strip harakat (short-vowel diacritics, which are combining marks) and does not unify the alef/hamza/teh-marbuta variants (أ/إ/آ→ا, ة→ه/ت) that dominate real corpora  so two documents differing only by diacritics won't match. Is Arabic-specific normalization (diacritic removal + letter folding) in scope for this RFC? If not, let's temper the Arabic recall claims so we don't over-promise.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+### Ingestion pipeline
+
+```
+Input text
+    │
+    ▼
+┌────────────────────────────┐
+│ SnowballLanguage::Tokenize │  UTF-8 validation, then:
+│  (SegmentInternal loop)    │   • NormalizeInPlace on the full input text
+│                            │     (NFC/NFKC + casefold) — ensures that
```
Aksha1812: Is there a specific order of execution to  NFC/NFKC and casefold . would that affect the final outcome? does that need to be made clear here ?

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+Each language's punctuation set and stop-word list live in that language's
+own header (`languages/french.h` defines French punctuation and French
+stop words). There is no shared composition with common constants — each
+language defines its complete, flat data set independently. Duplication
+across languages is intentional: each language is self-contained and
+independently modifiable, which is also what future non-Snowball languages
+(CJK) require.
+
+Stop-word sets are not built by a shared helper: `SnowballLanguage`
+normalizes each stop word through its own `NormalizeCaseFoldFilter` at
```
Aksha1812, line 298: Do we runs user-supplied STOPWORDS (and PUNCTUATION code points) through the same filter ? just making sure if there is no mismatch . It is not clear from the document.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+fail closed: nodes running a module version < 1.3 will reject index metadata
+carrying a non-English `LANGUAGE`. English and unspecified language values
+remain accepted on all versions for backward compatibility. Additionally,
+the registry rejects unrecognized language enum values at index creation
+time rather than silently falling back to English, ensuring that older
+module versions cannot incorrectly index data they cannot tokenize.
+
+## Compatibility divergences
+
+We intentionally diverge from RediSearch with the behaviors below.
```
Aksha1812, line 338: Do we have a maintainer sign off for this decision to diverge ?

Aksha1812 (commented):
I don't see a testing strategy which covers testing functioanlity for all 12 languages . Not sure if we usually include that in the RFC , but i think it should be in this case

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+  version gating, and stemming to the base.
+- **`SnowballStemFilter`** adapts the external `libstemmer` C API to the
+  `Stemmer` interface, using a thread-local `sb_stemmer` cache keyed by
+  language enum to avoid mutex contention on ingestion threads. Exposes
+  `GetStemRoot()` for single-word stemming (delete path, query expansion)
+  and `BuildStemMap()` for ingestion-time batch stem→surface-form mapping.
+- **`PunctuationSet`** (ASCII `std::bitset<128>` + non-ASCII
+  `flat_hash_set<uint32_t>`) is built by `BuildPunctuationSet()` in
+  `language.h`. It seeds from the per-language punctuation string and also
+  adds Unicode whitespace codepoints (U+00A0, U+2000–200A, U+202F,
```
Aksha1812: below https://github.com/valkey-io/valkey-search/pull/1279/files#diff-216150a9b96168c933d6dd2afd6f49e68803629d7bdf2243d81a8018f92236d6R432-R435  , the section "Unicode whitespace treated as word boundaries"  says the set is seeded from ICU \p{White_Space}. we can be conssitent if we are talking asbout the same thing here instead of listing specific unicodes

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+  U+3000, etc.) as word boundaries, ensuring that non-breaking spaces and
+  typographic whitespace in French, Arabic, and other scripts are correctly
+  treated as delimiters. The ASCII bitset preserves the `main`-branch O(1)
+  punctuation lookup for the common (English/ASCII) case; codepoint
+  decoding is triggered only by a non-ASCII lead byte.
+
+Language instances are immutable, stateless, and shared — heavyweight
+members (`PunctuationSet`, stop-word hash set, normalizer) exist once per
+language, not once per index. Two indexes using the same language share
+the same `Language*`; there is no per-index mutable language state, so
```
Aksha1812: " per-index mutable language state," is this also true for customizesd languages ?

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+    ├── Regular term ──────────────────────┐
+    │   language.NormalizeInPlace()         │
+    │   language.IsStopWord() → drop        │
+    │   Stemmer.GetStemRoot()               │→ TermPredicate
+    │                                       │
+    ├── Wildcard / Fuzzy ───────────────────┤
+    │   language.NormalizeInPlace() only     │→ Prefix/Suffix/FuzzyPredicate
+    │                                       │
+    └── Exact phrase (quoted) ──────────────┘
+        language.NormalizeInPlace() only     → TermPredicate(exact=true)
```
Aksha1812: nit : would update this diagram, loooks little confusing. something like this woulf be better .

```

                                Query string
                                     │
                                     ▼
          ┌──────────────────────────────────────────────────┐
          │ FilterParser::ParseTextTokens                       │
          │   byte-level scan → tokens                           │
          │   PunctuationSet bitset · IsNonAsciiDelimiter · \esc │
          └───────────────────────┬────────────────────────────┘
                                  │  per token
                                  ▼
                      ┌───────────────────────┐
                      │  NormalizeInPlace()     │  ◄── every token
                      └───────────┬────────────┘
                                  ▼
                            ◇  token shape?  ◇
                                  │
          ┌───────────────┬───────┴────────┬────────────────────┐
          │ word          │ word*  *word   │ %word%             │ "w1 w2 …"
          │ (regular)     │ *word* (wild)  │ (fuzzy)            │ (phrase)
          ▼               ▼                ▼                    ▼
    ┌───────────┐   (normalize only) (normalize only)   split into words
    │IsStopWord?│───drop→✗                                each → TermPredicate(exact)
    └─────┬─────┘         │                │                    │
       keep▼              │                │                    ▼
    ┌────────────┐        │                │        ┌───────────────────────────┐
    │Stemmer.    │        │                │        │ ComposedPredicate          │
    │GetStemRoot │        │                │        │  (AND, slop=0, inorder)    │
    └─────┬──────┘        │                │        └────────────┬──────────────┘
          ▼               ▼                ▼                     │
    TermPredicate   Prefix/Suffix/    FuzzyPredicate             │
    (exact=false)   InfixPredicate    (distance)                 │
          │               │                │                     │
          └───────────────┴────────────────┴─────────────────────┘
                                  │   all are TextPredicate
                                  ▼
          ┌──────────────────────────────────────────────────┐
          │ TextPredicate::BuildTextIterator()  → walk Rax      │
          │   • Term                      single-term lookup    │
          │   • Prefix/Suffix/Infix/Fuzzy expand terms, union   │
          │   • ComposedAND (phrase)      per-term iters + slop  │
          └───────────────────────┬────────────────────────────┘
                                  ▼
                              TextIterator
                              ```

File: rfc/multi-language-support.md [hunk 9]
```diff
@@ -0,0 +1,636 @@
...
+```
+
+`Fuzzy` DP is codepoint-aware so `MINSTEMSIZE` and edit distance work
```
Aksha1812: is `MINSTEMSIZE` even part of  fuzzy codepath . IIRC we don't care for stemming in for fuzzy search . in that case better to not mention them in the same line. let me know if i am wrong .

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+casefold locale.
+
+### Version gating and cluster safety
+
+Multi-language support ships behind a `MetadataVersion` bump to `1.3`.
+Non-English languages are gated by a version check so mixed-version clusters
+fail closed: nodes running a module version < 1.3 will reject index metadata
+carrying a non-English `LANGUAGE`. English and unspecified language values
+remain accepted on all versions for backward compatibility. Additionally,
+the registry rejects unrecognized language enum values at index creation
```
Aksha1812, line 310: what happens if specify actual non-English language names for older valkey-search versions . do they get rejected the same way as unrecognized language do now ?

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+section below describes how this restriction can be relaxed in a future
+release.
+
+### 1. Snowball 3.0.1 vs Snowball 2.1.0 (Dutch)
+
+Valkey Search uses Snowball 3.0.1, whereas RediSearch uses Snowball [2.1.0](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisearch/redisearch-2.0-release-notes/). Snowball 3.0.1 uses the
+Kraaij-Pohlmann algorithm for Dutch stemming, which strips grammatical
+prefixes; Snowball 2.1.0's Dutch Porter stemmer does not. The result is
+that Valkey Search reduces some Dutch words to the same stem where
+RediSearch reduces them to different stems.
```
Aksha1812: is that the only difference between 2 versions . or the only major difference to call out. Let's acknowledge some smaller difference in other languages we are introducing in 1.3 if any

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+An index in this release holds text in a single language, and there is no
+query-time `LANGUAGE` argument because a query-time language would only
+be meaningful if a single index could contain fields in multiple
+languages. Two forward-compatible extensions are envisioned, as discussed in
+the original [FTS RFC](https://github.com/valkey-io/valkey-rfc/pull/24/changes#diff-48ff4af40094c1df661c73561650315e01d872a043921be3a1a45eb2ffa39638R431)
+and related [comments](https://github.com/valkey-io/valkey-rfc/pull/24/changes#r2210835085):
+
+1. **`LANGUAGE_FIELD` in `FT.CREATE`.** A future release can add a
+   `LANGUAGE_FIELD` clause to `FT.CREATE` naming a document field whose
+   value marks each document's language. Ingestion would then route each
```
Aksha1812: nit : is `LANGUAGE_FIELD` per attribute or per document.? little unclear here. Not a blocker since we are talking about future possible extensions.

Aksha1812:
There are many unaddressed AI comments pending as well from before can we address them as well

rfc/multi-language-support.md
```diff
@@ -0,0 +1,539 @@
...
+Snowball stemming. This is replaced with a `LanguageProcessor` — a
+pipeline built from three pluggable primitive types.
+
+- **`Segmenter`** — splits a text span into tokens (1→N). The default
+  implementation `PunctuationSegmenter` is used by all 12 Snowball
+  languages and parameterized per-language via a `PunctuationSet` sourced
+  from the Unicode Common Locale Data Repository (CLDR) v46.
+- **`TokenFilter`** — transforms or drops a token (1→1 or 1→0). Default
+  implementations `NormalizeCaseFoldFilter` (Unicode NFC + case fold,
+  locale-aware for Turkish dotless-I) and `StopWordFilter` (per-language
```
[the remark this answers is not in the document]
neerajr0: The current feature implementation uses NFKC for Arabic. Normalization is language specific - clarified this in https://github.com/valkey-io/valkey-search/pull/1279/commits/8368d365eaec407bdb4af8b6781a8c04ec9d4152.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,539 @@
...
+
+Text handling was previously a single monolithic `Lexer` class with
+hard-coded English tokenization, ASCII-only punctuation, and English
+Snowball stemming. This is replaced with a `LanguageProcessor` — a
+pipeline built from three pluggable primitive types.
+
+- **`Segmenter`** — splits a text span into tokens (1→N). The default
+  implementation `PunctuationSegmenter` is used by all 12 Snowball
+  languages and parameterized per-language via a `PunctuationSet` sourced
+  from the Unicode Common Locale Data Repository (CLDR) v46.
```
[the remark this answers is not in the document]
neerajr0: The current implementation is backwards compatible with the existing English ASCII-only punctuation set. See https://github.com/valkey-io/valkey-search/blob/main/src/commands/ft_create_parser.h#L27-L29 in upstream and https://github.com/valkey-io/valkey-search/pull/1263/changes#diff-05c948e1f6e85d36e5a5dc2fbf675ced2ff4c4f1ac3f1d750d14ee1825d1df35R23-R24 in multi-language PR. Added detail in https://github.com/valkey-io/valkey-search/pull/1279/commits/79a3a38a9e158e5535d5c04b3581eb1249cf170b to clarify this.

Frank-Gu-81:
Hi @neerajr0 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @Aksha1812 — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,539 @@
...
+    ├── Regular term ──────────────────────┐
+    │   NormalizeCaseFoldFilter            │
+    │   StopWordFilter (if stop → drop)    │
+    │   Stemmer.GetStemRoot()              │→ TermPredicate
+    │                                      │
+    ├── Wildcard / Fuzzy ──────────────────┤
+    │   NormalizeCaseFoldFilter only       │→ Prefix/Suffix/FuzzyPredicate
+    │                                      │
+    └── Exact phrase (quoted) ─────────────┘
+        NormalizeCaseFoldFilter only        → TermPredicate(exact=true)
```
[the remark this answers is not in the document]
neerajr0: This is an existing behavior and consistent with Redisearch - confirmed empirically by indexing `the quick brown fox` and confirming an exact query for "the quick brown fox" returns 0 results. Updated the RFC in https://github.com/valkey-io/valkey-search/pull/1279/commits/e3f4ab7da6ae46b4f7351eca43def9e03ffaa00f.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+Input text
+    │
+    ▼
+┌────────────────────────────┐
+│ SnowballLanguage::Tokenize │  UTF-8 validation, then:
+│  (SegmentInternal loop)    │   • NormalizeInPlace on the full input text
+│                            │     (NFC/NFKC + casefold) — ensures that
+│                            │     compatibility mappings (e.g. NFKC:
+│                            │     U+FF0C fullwidth comma → U+002C) are
+│                            │     visible to the delimiter scanner
```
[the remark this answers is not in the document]
neerajr0: This is documented as an intended divergence in the RFC - `Unicode whitespace treated as word boundaries`. Re-indexing on RDB load will ensure forwards compatibility - an English index loaded from a 1.2 RDB on a 1.3 node will be re-tokenized (`IndexSchema::LoadFromRDB` eventually invokes `BackfillJob`) with the 1.3 English pipeline that uses Unicode whitespace as delimiters, so ingestion and query will stay consistent.

In practice the occurrence of Unicode whitespace in English text is rare - if needed users can override `PUNCTUATION` to opt out of this new behavior.

rfc/multi-language-support.md
neerajr0: This was considered during implementation. ICU's NFKC normalizer does not handle diacritic stripping or other Arabic-specific considerations like alef/hamza unification and tatweel shortening. Doing so would require a custom normalizer that extends the ICU implementation and would introduce a new divergence from Redisearch behavior (which also does not normalize these presentation forms). This can be a future improvement but for consistency in using ICU's provided normalizers will be deferred for the initial release.

rfc/multi-language-support.md
neerajr0: Normalization is applied prior to case folding. Ordering of normalization and case folding is commutative with our current implementation – however, we apply normalization first as a best practice because generally we want the word to be in a standard form before lowercasing it. What matters more is order between normalization and tokenization. These are not commutative for NFKC normalization (see https://github.com/valkey-io/valkey-search/pull/1263#discussion_r3855957446). Clarified this in https://github.com/valkey-io/valkey-search/pull/1279/commits/3354f0c73d60986383ae772039e89fe64b9949e6.

rfc/multi-language-support.md
neerajr0: Yes, these were reviewed with @allenss-amazon @BCathcart @KarthikSubbarao - intentional divergences are all intended to be search quality improvements over RediSearch.

rfc/multi-language-support.md
neerajr0: Yes — stop words are normalized before being filtered out, see `SnowballLanguage::Tokenize` section of Ingestion pipeline. Punctuation is normalized as well before tokenization because in the case of Arabic with NFKC normalization, some characters can be converted into delimiters during normalization, which are then used to separate text.

rfc/multi-language-support.md
neerajr0: Thanks for pointing this out. This language is not entirely accurate. There is no per-index mutable language state as mentioned. However at index creation time, `CreateLanguage()` wraps the registry singleton in a `CustomizedLanguage` that carries the effective punctuation and stop words (either user-supplied overrides from FT.CREATE or the language's defaults). Each index therefore owns its own immutable
`CustomizedLanguage` instance – heavyweight members (`PunctuationSet`,
stop-word hash set, normalizer, `SnowballStemFilter`) are built once at
construction and never modified. The overhead per instance is small (1-2 KB per language depending on stop-word list size). Updated this language in https://github.com/valkey-io/valkey-search/pull/1279/commits/1f0cc105b3ae7585b839bcdeca25ad48315e7e2b.

rfc/multi-language-support.md
neerajr0: Correct, any non-English language supported in version 1.3+ but not below will be rejected at parameter parsing if the version is < 1.3. See https://github.com/valkey-io/valkey-search/pull/1263/changes#diff-3f2ab72c507fe209b12ec1422ed0e732e7c8930fc9033e646f8545faec3c283aR306-R313.

rfc/multi-language-support.md
neerajr0: Double checked and there are a few optimizations for other languages introduced between 2.1.0 and 3.0.0. Added these in https://github.com/valkey-io/valkey-search/pull/1279/commits/fb052e4d3b42071576d7f23e756de3d0e6398aa0.

rfc/multi-language-support.md
neerajr0: We'd introduce this per document, eventually if we want further granularity we could it extend it per text field. Added in https://github.com/valkey-io/valkey-search/pull/1279/commits/8d8aed3794492a4059557169242814e9c76197d8.

rfc/multi-language-support.md
neerajr0: The build already disables unused ICU modules (`--disable-extras --disable-icuio --disable-layout --disable-tests --disable-samples`). Of the three static libraries we link, our code touches 3 APIs:

| ICU API | Used for |
|---------|----------|
| `Normalizer2::getNFCInstance` / `getNFKCInstance` | Unicode normalization (all languages) |
| `CaseMap::utf8Fold` | Generic case folding (all languages except Turkish) |
| `CaseMap::utf8ToLower` | Locale-aware lowercasing (Turkish dotless-i) |

Everything else in the linked libraries is unused today. 

Binary size comparison (Release build, same machine):

| Build | Unstripped | Stripped |
|-------|-----------|---------|
| `main` (no ICU) | 139 MB | 20 MB |
| `multi-language-support` (ICU) | 172 MB | 51 MB |
| **Delta** | **+33 MB** | **+31 MB** |

The +31 MB is almost entirely `libicudata.a` (31 MB) — the full Unicode CLDR data archive baked in as a static object. Most pages are demand-paged and never faulted in.

**Trimming path:** ICU supports a [data filter](https://unicode-org.github.io/icu/userguide/icu_data/buildtool.html) at build time to strip the archive to only normalization + case-folding tables. This should reduce `libicudata.a` from 31 MB to ~1–2 MB. We'll plan for this as a follow up optimization – the increased storage overhead should not affect runtime performance.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,539 @@
...
+pipeline built from three pluggable primitive types.
+
+- **`Segmenter`** — splits a text span into tokens (1→N). The default
+  implementation `PunctuationSegmenter` is used by all 12 Snowball
+  languages and parameterized per-language via a `PunctuationSet` sourced
+  from the Unicode Common Locale Data Repository (CLDR) v46.
+- **`TokenFilter`** — transforms or drops a token (1→1 or 1→0). Default
+  implementations `NormalizeCaseFoldFilter` (Unicode NFC + case fold,
+  locale-aware for Turkish dotless-I) and `StopWordFilter` (per-language
+  default lists sourced from Apache Lucene) are shared across all 12
```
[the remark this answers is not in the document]
neerajr0: Addressed in https://github.com/valkey-io/valkey-search/pull/1279/commits/8368d365eaec407bdb4af8b6781a8c04ec9d4152.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,539 @@
...
+│   Word extraction → QueryTokenizer                 │
+│     PunctuationQueryTokenizer (Snowball languages) │
+│     [future] CJK-specific tokenizers               │
+│   Escape handling → \<char>                        │
+└────────────────────────────────────────────────────┘
+    │
+    ├── Regular term ──────────────────────┐
+    │   NormalizeCaseFoldFilter            │
+    │   StopWordFilter (if stop → drop)    │
+    │   Stemmer.GetStemRoot()              │→ TermPredicate
```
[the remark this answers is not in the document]
neerajr0: Addressed in https://github.com/valkey-io/valkey-search/pull/1279/commits/9c8edbdee78f316dfba3caf3e8fb8e5766846189.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,539 @@
...
+on word-level positions in an already-tokenized stream and are unaffected
+by multi-byte handling.
+
+### Version gating and cluster safety
+
+Multi-language support ships behind a `MetadataVersion` bump to `1.3`.
+Non-English languages are gated by a version check so mixed-version clusters 
+fail closed: nodes running a module version < 1.3 will reject index metadata carrying a
+non-English `LANGUAGE`. English and unspecified language values remain
+accepted on all versions for backward compatibility.
```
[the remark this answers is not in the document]
neerajr0: Addressed in https://github.com/valkey-io/valkey-search/pull/1279/commits/9568bee951ab3ee983b653644cc5e2ec5a89e3fb.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,539 @@
...
+```
+
+Apache Lucene's Dutch analyzer uses Kraaij-Pohlmann-style prefix
+stripping, matching Valkey Search's behavior.
+
+### 2. ICU `utf8Fold` vs simple Unicode lowercasing
+
+`NormalizeCaseFoldFilter` uses ICU's `CaseMap::utf8Fold` during ingestion
+and querying. `utf8Fold` decomposes characters like `ß` to `ss` and
+ligatures like `ﬁ` to `fi` so searches match across these equivalent
```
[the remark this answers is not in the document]
neerajr0: https://github.com/valkey-io/valkey-search/pull/1279/commits/8368d365eaec407bdb4af8b6781a8c04ec9d4152 addresses this to specify that Turkish uses `CaseMap::utf8ToLower("tr")`.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,539 @@
...
+FT.SEARCH tr_custom "için" DIALECT 2
+  → 1 result: doc:tr_c:1
+```
+
+### 6. `NOSTEM` for exact-form indexes
+
+Without `NOSTEM`, Snowball stems both indexed content and the query, so
+a search for one Italian conjugation matches documents containing any
+other conjugation that stems to the same root. With `NOSTEM`, only the
+exact indexed surface form matches:
```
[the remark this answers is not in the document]
neerajr0: Updated in https://github.com/valkey-io/valkey-search/pull/1279/commits/c949d4cb1a6673aee5d96e5ca25dc7d0da61a320.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+NO-BREAK SPACE, U+3000 IDEOGRAPHIC SPACE, etc.) as token delimiters via
+ICU's `\\p{White_Space}` property. RediSearch only splits on ASCII
+whitespace (space, tab, newline) despite documentation stating "all
+whitespace separates tokens."
+
+This matters most for French typography (NBSP before `:` `;` `!` `?` and
+inside `« »`), Arabic text with non-breaking spaces, and CJK text with
+ideographic spaces.
+
+```
```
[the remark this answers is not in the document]
neerajr0: Added in https://github.com/valkey-io/valkey-search/pull/1279/commits/cfc5746b1167ee1d57114914deb536a181949468.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,636 @@
...
+HSET doc:tr:1 body "İstanbul çok güzel."
+HSET doc:tr:2 body "istanbul haritası"
+
+# NormalizeCaseFoldFilter with locale="tr" folds İ→i (not İ→i̇).
+# Both documents are found regardless of the case in the query.
+FT.SEARCH tr_idx "istanbul" DIALECT 2
+  → 2 results: doc:tr:1, doc:tr:2
+
+FT.SEARCH tr_idx "İSTANBUL" DIALECT 2
+  → 2 results: doc:tr:1, doc:tr:2
```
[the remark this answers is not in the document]
neerajr0: Addressed by https://github.com/valkey-io/valkey-search/pull/1279/commits/8368d365eaec407bdb4af8b6781a8c04ec9d4152.

rfc/multi-language-support.md
neerajr0: Updated in https://github.com/valkey-io/valkey-search/pull/1279/commits/6fd3efed38f7f753b653f6c52c6c698bb97b31aa.

rfc/multi-language-support.md
neerajr0: Updated in https://github.com/valkey-io/valkey-search/pull/1279/commits/dfab39cf46bcc2b5b031cf5368d681748a3d9c02.

rfc/multi-language-support.md
neerajr0: Thanks for catching, updated wording in https://github.com/valkey-io/valkey-search/pull/1279/commits/9d0de091b503f0051a4133854c37403b55193dce.

neerajr0:
> I don't see a testing strategy which covers testing functioanlity for all 12 languages . Not sure if we usually include that in the RFC , but i think it should be in this case

I chose to keep the RFC as a summary of the design decisions and behavior (including Redisearch compatibility) of the multi-language support feature based off previous RFCs like https://github.com/valkey-io/valkey-rfc/pull/24 which do the same. The PR description for the feature has extensive detail on the testing strategy: https://github.com/valkey-io/valkey-search/pull/1263. The attached document `snowball_language_benchmarking.md` also includes benchmarking results for precision, recall and hit rate against annotated multi-language corpora. I reran this using https://github.com/neerajr0/valkey-search-multilang-bench to confirm we're still seeing the same results after addressing comments. The doc also contains results on system performance (throughput, latency, QPS) from running on a dev desktop, but we will get definitive regression results from the new benchmark comparison GH workflow introduced into `valkey-search`: https://github.com/valkey-io/valkey-search/pull/1268. @Frank-Gu-81 and I are getting that set up currently.

rfc/multi-language-support.md
```diff
@@ -0,0 +1,771 @@
...
+# "corriamo" ≠ "correre" → no match.
+FT.SEARCH it_nostem "corriamo" DIALECT 2
+  → 0 results
+```
+
+### 7. Separating languages by key prefix
+
+Since an index holds a single language, deployments with mixed-language
+corpora create one index per language and route documents by key prefix
+(or, in a future release, by `FILTER` on a language field):
```
allenss-amazon: FILTER will be available in 1.3 release. Let's add another chapter showing that case.

allenss-amazon (changes requested):

rfc/multi-language-support.md
neerajr0: Added in https://github.com/valkey-io/valkey-search/pull/1279/commits/d9e509fb6497ad8ab9438907d4e969392db16320. Note that #984 hasn't merged yet so I manually reviewed the usage example which was derived from both PRs' (multi-lang and FILTER) source code.

allenss-amazon (approved):

Aksha1812 (approved):

rfc/multi-language-support.md
```diff
@@ -0,0 +1,833 @@
...
+partition documents into separate key namespaces. With `FILTER`,
+documents can share a single prefix and carry a field that indicates
+their language — each index self-partitions at ingestion time without
+requiring distinct key prefixes.
+
+```text
+# All documents share the "product:" prefix and carry a "lang" field.
+# Create one index per language, each restricted by FILTER.
+FT.CREATE fr_products ON HASH PREFIX 1 product:
+    FILTER '@lang == "french"'
```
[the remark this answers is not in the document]
allenss-amazon: FILTER is now merged.

Issue #1282, opened 2026-08-04
[BUG] Duplicate external labels in HNSW cause load validation to fail


BCathcart:
**Describe the bug**
The validation code asserts that there are only unique external labels in the `data_level0_memory_` slots. This is not always the case. I don't see any inherent problems with this state. The one place it can go wrong today is across a save/load. The load code rebuilds the `label_lookup_` from the labels in `data_level0_memory_`.  There can be one live slot and any number of tombstoned slots all carrying the same label. The `label_lookup_` reconstruction code needs to check that it points to the live slot for that label. Otherwise the HNSW state won't properly reflect the keyspace. Future updates or deletes to the record will error as it will see the label already belonging to a tombstoned node.

**To Reproduce**
The HNSW modify record code first calls `markDelete()` on the record to tombstone the existing slot with the label and then calls `addPoint()` to add the new vector with the exact same label. Theoretically, `std::unordered_set<tableint> deleted_elements` makes no promises about which slot `deleted_elements.begin()` points to. In practice though, it's always the last added slot. In the normal case, the same slot will be updated and there will be no duplicate labels. The problem occurs when another document is deleted between `markDelete()` and `addPoint()`. It will put another tombstoned slot at the front of `deleted_elements` and the `addPoint()` for the modified record will use it, adding the label to a new slot and still keeping the same label in the old slot it just tombstoned.

**Expected behavior**
Let's not have the load-side validation enforce unique slots. It just needs to properly reconstruct the `label_lookup_`.

**Environment**
Search 1.0, 1.1, and 1.2 are all susceptible.

**Additional context**
Additional problems that make mitigation hard:
 - Inability to modify Dev configs outside debug-mode.
 - All load validation being is coupled under a single enabled/disabled config, regardless if it's protecting against OOB accesses with security implications or just surfaces a small correctness bug like this one.

BCathcart:
> I don't see any inherent problems with this state.

My test actually does expose a use-after-free problem.

Vector bytes live in a refcounted `InternedString`. The only *owning* reference for an indexed vector is the entry in `tracked_vectors_`. The HNSW slot record stores a raw, *non-owning* `char*` to those bytes.

The free happens on modify, at `ModifyRecord` → `TrackVector(id, new_vec)`:

```
tracked_vectors_[label] = new_vec;   // overwrites → old InternedStringPtr destructs → refcount-- → frees old_vec
```

**Why only in the race case:**

[...]

PR #1283, opened 2026-08-04
Fix duplicate label handling

BCathcart:
Solves #1282.

I'll follow this up with an ingestion-side update to force same slot reuse for updates to make things more more intuitive and prevent future RDBs from also having duplicate labels.


integration/test_hnsw_allow_replace_deleted.py
```diff
@@ -184,3 +185,111 @@ def test_index_add_after_rdb_load_with_deleted_elements(self):
...
+        # The survivor should be doc:0 carrying its updated vector.
+        search_result = client.execute_command(
+            "FT.SEARCH", "idx",
+            "*=>[KNN 2 @vector $q]",
+            "PARAMS", "2", "q", vec0_new,
+        )
+        assert search_result[0] == 1, \
+            f"Expected 1 search result, got {search_result[0]}"
+        assert search_result[1] == b"doc:0", \
+            f"Expected doc:0 to survive, got {search_result[1]}"
```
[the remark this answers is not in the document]
BCathcart: My testing showed that a new deleted element always went to the front of the unsorted set iteration

integration/test_hnsw_allow_replace_deleted.py
```diff
@@ -184,3 +185,125 @@ def test_index_add_after_rdb_load_with_deleted_elements(self):
...
+        # deleted_elements.
+        client2 = self.server.get_new_client()
+        client2.delete("doc:0")
+        waiters.wait_for_equal(
+            lambda: hnsw_index.info(client2).num_docs, 1
+        )
+
+        # Resume the modify. Its addPoint pops a slot off deleted_elements to
+        # reuse. In practice we expect that to be doc:0's slot, so doc:1's live
+        # record mapped in label_lookup_ lands on slot 0 and its tombstone on slot 1.
```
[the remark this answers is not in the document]
BCathcart: In practice it always does with my testing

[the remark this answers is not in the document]
BCathcart: Not always. I'm now sure how much unsorted sets differing in this way across standard library versions is really a concern though. I'll add a counter to make this test fail if it doesn't hold true

Aksha1812 (commented):
AI said . looks correct . is this failing the ASAN test?


**Location:** `third_party/hnswlib/hnswalg.h:999-1001` (call site) + `src/indexes/vector_hnsw.cc:110-114` (`tracked_vectors_`)

**Mechanism.** `LoadIndex` interns a vector **per slot** but keys the owning reference **per label**:

```cpp
// hnswalg.h, loop over slots
*(char **)((*data_level0_memory_)[i] + offsetData_) =
    vector_tracker->TrackVector(id, chunk->data() + size_links_level0_, vector_size_);
```

```cpp
// vector_base.cc — the returned raw pointer is the ONLY thing the slot keeps
char *VectorBase::TrackVector(uint64_t internal_id, char *vector, size_t len) {
  auto interned_vector = StringInternStore::Intern(absl::string_view(vector, len),
                                                   vector_allocator_.get());
  TrackVector(internal_id, interned_vector);        // -> tracked_vectors_[label] = ptr
  return (char *)interned_vector->Str().data();     // local ptr dies here
}
```

```cpp
// vector_hnsw.cc — keyed by LABEL, so a second slot with the same label CLOBBERS
tracked_vectors_[internal_id] = vector;   // operator= decrements the old -> may free
```

With two slots sharing a label and holding *different* bytes, the second iteration drops the last reference to the first slot's vector. `Allocator::Free` runs. The first slot's `char*` at offset 136 dangles.

**Why this PR matters for it.** Pre-PR this RDB was rejected, so the code path was unreachable in production. Post-PR the load succeeds and leaves the dangling pointer in place.

**Why it's not benign.** Tombstoned slots are dereferenced on every traversal that visits them (`hnswalg.h:469-497`: phase 2 resolves the pointer, phase 3 calls `fstdistfunc_`, and `isMarkedDeleted` only filters the *result set* afterwards). Worse, in the PR's own scenario (§4, Phase 6) the dangling slot is the **live** one — slot ordering puts the live slot first, so it's `vec1_new` that gets freed. So `FT.SEARCH` computes distances on freed memory *and* `RETURN vector` copies freed bytes out to the client — a potential info leak once the allocator reuses the block.

**Proof / reproduction.** The PR's own new test. Run `integration/test_hnsw_allow_replace_deleted.py::TestHNSWDuplicateLabelRace` under ASAN and it should report `heap-use-after-free` in `fstdistfunc_` or in `GetValue`'s `result.assign`. The two duplicate slots necessarily hold different vectors — a same-vector modify is short-circuited to `kMissing` by `UpdateMetadata`/`IsVectorMatch` and never reaches `ModifyRecordImpl` — so the free is the *normal* case, not an edge case.

**Note it also occurs purely in memory**, without save/load: §4 Phase 4 shows `TrackVector(1, vec1_new)` freeing `vec1` while slot 1 still points at it.

**Root cause.** The comment on `UnTrackVector` states the broken assumption outright:

```cpp
// On key modification, the same internal_id is reused, so
// TrackVector overwrites the map entry in-place ...
```

That is exactly what no longer holds once a label spans multiple slots.

**Suggested fixes** (pick one; the first is smallest):

1. **During load, share one interned vector per label.** When a label is already present in `label_lookup_`, point the new slot at the *same* pointer already stored in `tracked_vectors_[label]` instead of interning fresh. Duplicate slots then share one allocation; nothing is freed; memory usage drops. The tombstone's routing distances shift slightly (it now uses the live vector), which affects recall marginally on a slot that is dead anyway. Cheapest and arguably the semantically nicest option.
2. **Hold a per-slot keepalive.** Add a `std::vector<InternedStringPtr>` sized `cur_element_count_` populated during load, so every slot owns its vector independently of the label map.
3. **Re-key `tracked_vectors_` by `tableint` slot** rather than label. Correct long-term, but `IsVectorMatch` is called from `UpdateMetadata` with a label, so it would need a label→slot resolution first; more invasive.

third_party/hnswlib/hnswalg.h
```diff
@@ -1275,8 +1293,9 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
     std::unique_lock<std::mutex> lock_deleted_elements(deleted_elements_lock);
     bool is_vacant_place = !deleted_elements.empty();
     if (is_vacant_place) {
-      internal_id_replaced = *deleted_elements.begin();
-      deleted_elements.erase(internal_id_replaced);
+      auto it = deleted_elements.begin();
+      internal_id_replaced = *it;
+      deleted_elements.erase(it);
```
Aksha1812, line 1319: **Location:** `third_party/hnswlib/hnswalg.h:1304-1309` — ten lines below a hunk this PR *does* touch, but left unchanged.

```cpp
labeltype label_replaced = getExternalLabel(internal_id_replaced);
setExternalLabel(internal_id_replaced, label);

std::unique_lock<std::mutex> lock_table(label_lookup_lock);
label_lookup_.erase(label_replaced);            // ◀── UNCONDITIONAL
label_lookup_[label] = internal_id_replaced;
lock_table.unlock();
```

`erase(label_replaced)` assumes *"the label on the slot I'm recycling is mapped to this slot."* With duplicates that is false: the recycled tombstone carries label L, but `label_lookup_[L]` points at a **different, live** slot. The erase destroys the live record's only mapping.

**Reproduction** (no save/load needed — see §4 Phase 8 for the full trace):

1. Trigger the race → slot 0 live/label 1, slot 1 tombstone/label 1, `label_lookup_ = {1→0}`, `deleted_elements = {1}`.
2. `HSET doc:2 ...` → `addPoint(label=2, replace_deleted=true)` pops slot 1 → `label_replaced = 1` → `label_lookup_.erase(1)`.
3. `label_lookup_ = {2→1}`. `doc:1` is live at slot 0, unmapped.
4. `FT.SEARCH` still returns `doc:1` (found by traversal, resolved via `key_by_internal_id_`).
5. `DEL doc:1` → `markDelete(1)` → `find(1) == end()` → throws "Label not found" → `hnsw_remove_exceptions_cnt++`. **The doc cannot be deleted, ever.** Searches keep returning a key the user deleted.

**This directly refutes the premise of the PR and the issue.** The issue says *"There can be one live slot and any number of tombstoned slots all carrying the same label. This is not fundamentally incorrect"*, and the code comment repeats it. But the duplicate tombstone is in `deleted_elements` and is therefore *live ammunition*: the next insert corrupts the index. The PR's load fix produces an index that is **one insert away** from unremovable-ghost corruption. Making the RDB loadable without making the state safe is half a fix.

**Suggested fix — small and complete:**

```cpp
std::unique_lock<std::mutex> lock_table(label_lookup_lock);
// Only drop the old label's mapping if it actually points at the slot we are
// recycling. With duplicate labels the mapping may belong to a different,
// live slot, which must not be unmapped.
auto it = label_lookup_.find(label_replaced);
if (it != label_lookup_.end() && it->second == internal_id_replaced) {
  label_lookup_.erase(it);
}
label_lookup_[label] = internal_id_replaced;
```

This makes duplicate labels genuinely benign — which is what the PR's comment claims they already are. I'd argue **this belongs in this PR**, not a follow-up, because this PR is what makes duplicate-label indexes loadable and therefore common.

The author does say *"I'll follow this up with an ingestion-side update to force same slot reuse for updates"*, which would prevent *new* duplicates. But it doesn't help RDBs already written by 1.0/1.1/1.2 — which is the entire population this PR exists to load. Those will load, then hit this bug on the next insert.

Aksha1812:
workflows still failing 
```
            f"New metrics were added. Please update the expected_metrics baseline in this test."
        )
E       AssertionError: Unexpected new APP metrics: ['hnsw_duplicate_label_on_load_count']
E         New metrics were added. Please update the expected_metrics baseline in this test.
E       assert not {'hnsw_duplicate_label_on_load_count'}

../../../integration/test_info.py:182: AssertionError
```

third_party/hnswlib/hnswalg.h
BCathcart: Skimming this, I think this is just calling out the existing issue that was surfaced by the test I added with ASAN. I'm already working on this. I'm resolving this but feel free to validate with your AI it's fixed after I push the next revision.

BCathcart:
@Aksha1812 Can you check if the changes I've been making are addressing your AI's concerns and provide a quick summary of any remaining issues you've found?

Aksha1812:
> @Aksha1812 Can you check if the changes I've been making are addressing your AI's concerns and provide a quick summary of any remaining issues you've found?

yes looks good. doing one final check

Aksha1812:
can you add some more unit tests . e.g
```
TEST_F(VectorIndexTest, AcceptDuplicateLabelTombstoneThenLive)   
TEST_F(VectorIndexTest, AcceptDuplicateLabelLiveThenTombstone)   
TEST_F(VectorIndexTest, AcceptDuplicateLabelAllTombstoned)       
```

Aksha1812:
AI pointed out some minor nits and edge cases . I don't think these need to be addressed . I think we have everything for this PR . but just sharing if you want to take a quick look
[pr-1283.md](https://github.com/user-attachments/files/30752965/pr-1283.md)

Aksha1812 (approved):

third_party/hnswlib/hnswalg.h
```diff
@@ -1035,9 +1037,24 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
...
+      // A label can appear on multiple slots in RDBs written by older versions.
+      // There can be one live slot and any number of tombstoned slots all
+      // carrying the same label. We just need to rebuild the label lookup to
+      // point the label at the live slot if there is one.
+      auto dup_it = label_lookup_.find(ext_label);
+      if (dup_it == label_lookup_.end()) {
+        label_lookup_[ext_label] = i;
+      } else {
+        valkey_search::Metrics::GetStats().hnsw_duplicate_label_on_load_cnt +=
+            1;
```
KarthikSubbarao, line 1050: Question for now or future: Do we want to handle these duplicate cases / tombstones by cleaning them up during load and not leaving the tombstones to exist?

KarthikSubbarao (approved):

third_party/hnswlib/hnswalg.h
BCathcart: That gets into the tombstone reclamation discussion. If we had a routine to reclaim a tombstone and repair the incoming connections, we could use that at load time. The tombstones with duplicate labels will potentially have garbage vector data, so they may be detrimental to the structure of the graph already and it would be good to remove them, but repairing the structure is complicated. Definitely something worth considering as a follow-up as part of the space reclamation discussion.

yairgott:
A point is marked for deletion only if the attribute, or the entire hash record, is deleted. As part of the the vector RemoveRecord handling, a call to `UnTrackKey` is issued to remove the mapping between the internal(label) and the engine key and vise-versa.  Therefore, there is no usage for the label after the remove record is completed. 

As a longer term solution, it would be cleaner to delete the label from label_lookup_ and clear the point label value as part the `markDelete` function. As pointed, there is no logical usage of the label value of marked as deleted points.

BCathcart:
Label lookup table is cleaned up in #1291 along with other changes to continue getting allow-replace-deleted enabled mode into a better place

Issue #1288, opened 2026-08-06
[BUG] Leak old vectors when tombstoned slot is overwritten


BCathcart:
Vectors are owned by `tracked_vectors_`. Tombstoned slots still need to have their vectors tracked or else they’re freed prematurely and the raw vector pointer in the HNSW slots are dangling. This was the issue when an update overwrote the tracked_vectors_ entry but didn’t overwrite the same slot at the HNSW level addressed in #1283.

I’m now seeing we also have a clear memory leak with allow-replace-deleted enabled. `UnTrackVector()` is a no-op no matter what. When a tombstoned slot is overwritten, we need to pass the label up and clean it up from tracked_vectors_. Alternatively we could pass ownership of the vector to the HNSW layer for tombstoned slots and it could be deleted at the library level.

PR #1291, opened 2026-08-10
Fixes for HNSW allow-replace-deleted mode

BCathcart:
Finishes solving #1282 and solves #1288.

There are three things addressed here:
##### 1. Fully recovering to a healthy state when loading an RDB that contains an HNSW index with duplicate labels.

#1283 prevents new instances of duplicate labels being created. It also prevents the validation logic from rejecting existing RDBs with duplicate labels and restores the HNSW index to the state it was in at the time of save, but that isn't a fully healthy state. There will be dangling vector pointers in the index (see https://github.com/valkey-io/valkey-search/issues/1282#issuecomment-5186675853). This PR actually repairs the HNSW state so that there are no dangling pointers by enforcing label uniqueness. This gives a clear one-to-one mapping of labels in `tracked_vectors_` and labels in `data_level0_memory_`.

##### 2. Cleaning up an overwritten slot's vector from `tracked_vectors_`.

The whole point of enabling `allow-replace-deleted` is to reduce memory usage with the potential risk of reducing the quality of the graph structure (incoming edges lose effectiveness when the vector is arbitrarily replaced). The problem was, when a slot was overwritten, we weren't actually freeing the most significant part of the memory associated with it: the old vector (#1288). Now the HNSW library layer returns the label of an overwritten slot so that `tracked_vectors_` can be cleaned up.

##### 3. Removes tombstoned labels from `label_lookup_`.

Labels are used by the search module to operate on live vectors. Once a vector is deleted from the keyspace, and the module has told the HNSW library to tombstone the slot with that label, there is no more use for mapping from label -> slot. `label_lookup_` now only contains live labels as a simplification and space optimization.


integration/test_hnsw_allow_replace_deleted.py
```diff
@@ -290,7 +296,60 @@ def test_load_rdb_with_duplicate_label(self):
...
+            "FT.SEARCH", "idx",
+            "*=>[KNN 1 @vector $q]",
+            "PARAMS", "2", "q", self.LIVE_VEC,
+            "RETURN", "1", "vector",
+        )
+        returned_fields = search_result[2]
+        vector_value = returned_fields[returned_fields.index(b"vector") + 1]
+        assert vector_value == self.LIVE_VEC, \
+            f"Expected survivor to carry its vector, got {vector_value!r}"
+
```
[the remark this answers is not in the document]
BCathcart: To your first claim, tests are run sequentially.

To your second claim, the test will fail either way but I can make the change.

third_party/hnswlib/hnswalg.h
```diff
@@ -1180,6 +1214,7 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
       throw std::runtime_error("Label not found");
     }
     tableint internalId = search->second;
+    label_lookup_.erase(search);
```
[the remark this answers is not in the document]
BCathcart: Labels are never re-used across keys. They're only re-used if an existing key is updated with a new vector. I'll update the code to make this clearer

src/indexes/vector_hnsw.cc
```diff
@@ -180,8 +180,12 @@ absl::Status VectorHNSW<T>::AddRecordImpl(uint64_t internal_id,
...
       absl::ReaderMutexLock lock(&resize_mutex_);
 
-      algo_->addPoint((T *)record.data(), internal_id,
-                      algo_->allow_replace_deleted_);
+      auto evicted = algo_->addPoint((T *)record.data(), internal_id,
+                                     algo_->allow_replace_deleted_);
+      if (evicted.has_value()) {
+        absl::MutexLock lock(&tracked_vectors_mutex_);
+        tracked_vectors_.erase(*evicted);
+      }
```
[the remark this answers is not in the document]
BCathcart: A label won't be reused by AddRecordImpl. Only ModifyRecord re-uses labels.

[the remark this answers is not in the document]
BCathcart: Why is your confidence score still 4/5?

src/indexes/vector_hnsw.cc
```diff
@@ -180,8 +180,12 @@ absl::Status VectorHNSW<T>::AddRecordImpl(uint64_t internal_id,
     try {
       absl::ReaderMutexLock lock(&resize_mutex_);
 
-      algo_->addPoint((T *)record.data(), internal_id,
-                      algo_->allow_replace_deleted_);
+      auto evicted = algo_->addPoint((T *)record.data(), internal_id,
```
yairgott, line 183: Once the memory sharing PR lands, this change will no longer be necessary. Merging it into main now will just create extra overhead and potential merge conflicts for that PR. That said, this fix is highly valuable for backporting to older versions.

third_party/hnswlib/hnswalg.h
```diff
@@ -987,6 +989,11 @@ class HierarchicalNSW : public AlgorithmInterface<dist_t> {
     data_level0_memory_ = std::make_unique<ChunkedArray>(
         size_data_per_element_, k_elements_per_chunk, max_elements);
 
+    // Stage each slot's vector keyed by slot (not label) so duplicate labels
```
yairgott: Can you elaborate why this is needed given that on mark as deleted triggers now removal from label_lookup_? I believe that if the point stored label is ignored when the point is marked as deleted, it should be sufficient.

BCathcart: Oh that's actually an outdated comment and doesn't make sense there. I've updated the comments. The tombstoned list is just temporary during load to detect and recover from duplicate label situations.

src/indexes/vector_hnsw.cc
BCathcart: What the status of the shared memory PR? From what Allen mentioned it was close to being merged. Is it just the tracked_vectors_ handling update that will be a merge conflict? I can easily revert this part in main. If it's more than just this I could revert all the source changes in main and leave the tests skipped for now and we'll need to coordinate on making sure all gaps are covered after the shared memory PR lands. WDYT?

third_party/hnswlib/hnswalg.h
yairgott: As label duplication may only happen with points that are already marked as deleted. Wouldn't it be simpler to just ignore the stored label for those deleted points (treating it as empty while checking for duplication)?

src/indexes/vector_hnsw.cc
yairgott: The memory sharing PR stack is in fairly good state. Allen has already started to reviewed them but haven't stamp them yet. I expect it to land shortly after Allen is back, 8/17.

https://github.com/valkey-io/valkey-search/pull/1168
https://github.com/valkey-io/valkey-search/pull/1232
https://github.com/valkey-io/valkey-search/pull/1233

third_party/hnswlib/hnswalg.h
BCathcart: Tombstoned slots still need a unique label today since the vectors for all slots are maintained in `tracked_vectors_` keyed by the label. It sounds like after the shared memory PR though, a label may serve no purpose once the slot is marked deleted?

Aksha1812:
**Reviewers for this PR**

BCathcart:
Confirmed with @yairgott that his shared memory PRs will address these gaps in main. For one, there will be no more `tracked_vectors_` table causing issues. He's going to pull the tests from this PR into his shared memory PR. I've raised a new PR with the same diff directly to 1.2

BCathcart:
New PR: #1321

PR #1292, opened 2026-08-10
Avoid main-thread deadlock when draining the multi/exec queue during a fork

### LINKED ISSUE #1260: [BUG] Permanent main-thread deadlock: FT.SEARCH during a fork window blocks in ProcessMultiQueue() on the suspended writer pool
**Describe the bug**

The main thread deadlocks permanently when an `FT.SEARCH`/`FT.AGGREGATE` arrives while a fork child (BGSAVE, AOF-rewrite preamble, or replica full sync) is alive and there are pending multi-queue mutations. The server never recovers: the event loop is dead, `SHUTDOWN`/SIGTERM are accepted by the signal handler but never processed, and only SIGKILL ends the process. We hit this three times in one day on a production-shaped environment (standalone + Sentinel, no cluster mode), always on the primary — twice on a routine `BGSAVE`, once on the diskless full-sync fork feeding a rejoining replica.

**Mechanism (verified against the 1.2.1 tag)**

1. Writes to indexed keys executed inside `MULTI/EXEC` or Lua (note: redis-py pipelines default to `MULTI/EXEC`) enqueue their mutations into `multi_mutations_keys_` via `EnqueueMultiMutation` (`src/index_schema.cc:871-873`). The queue is drained lazily, on the query path.
2. `AtForkPrepare` (`src/valkey_search.cc:1021-1045`) suspends the writer/reader/utility pools. `AfterForkParent` resumes reader/utility/gRPC but intentionally **not** the writer pool — it is resumed only by the fork-child-died notification (`src/valkey_search.cc:1222`), which is delivered through the module event fired from `checkChildrenDone`, i.e. **by the main thread**.
3. If an `FT.SEARCH` arrives while the child is alive, `QueryCommand::Execute` calls `ProcessMultiQueue()` (`src/commands/commands.cc:152`) **on the main thread**, which schedules the queued mutations onto the (suspended) writer pool and blocks on `absl::BlockingCounter::Wait()` (`src/index_schema.cc:725`).
4. Every path that could resume the writer pool runs on the main thread, which is now blocked → circular wait, permanent hang.

The code in this region is unchanged on `main` as of 2026-07-21. This is distinct from #761 (CME cross-shard RPC deadlock; we run no coordinator) and from the Suspend/Resume hang addressed by PR #1139 (stale thread count after #1062); the writer-pool-suspension-until-child-died design dates back to #277.

**Log signature**

Successful save (parent, pid 1):

```
1:M ... * Background saving terminated with success
1:M ... # <search> [warning] ... /src/valkey_search.cc:1222: Fork child died notification received. Resuming writer worker thread pool returned message:  Suspend duration: 366.217953ms
```

Hung save — the child completes normally, then the parent goes permanently silent (no "Background saving terminated", no child-died line):

```
1:M 21 Jul 2026 06:47:41.101 * 100 changes in 300 seconds. Saving...
1:M 21 Jul 2026 06:47:41.103 # <module> [warning] ... /src/valkey_search.cc:1030: At prepare fork callback, suspend writer worker thread pool returned message:
1:M 21 Jul 2026 06:47:41.200 # <module> [warning] ... /src/valkey_search.cc:1055: After fork parent callback, resume reader worker thread pool returned message:
1:M 21 Jul 2026 06:47:41.201 * Background saving started by pid 81692
81692:C 21 Jul 2026 06:47:42.135 * DB saved on disk
81692:C 21 Jul 2026 06:47:42.141 * Fork CoW for RDB: current 5 MB, peak 5 MB, average 4 MB
1:signal-handler (1784616537) Received SIGTERM scheduling shutdown...     <- never executed; SIGKILL after grace period
```

External probes date the lockup precisely: the parent still answered `PING` ~0.4-1.4 s *after* fork return (i.e. through the child's lifetime) and went silent around child-exit processing time.

**To Reproduce**

Race repro (how we hit it): standalone server, module loaded with default thread pools; an index over a small JSON/HASH corpus; loop A writes indexed keys via `MULTI/EXEC` (~0.5 op/s is enough), loop B runs `FT.SEARCH` (~0.2 op/s), loop C triggers forks (`BGSAVE` or RDB save points every few minutes). With these rates we observed 3 permanent hangs across ~90 forks in one day (and 0 in ~290 forks the day before) — a low-probability race per fork, but certain over time. A deterministic reproduction should be possible by holding the fork child alive (large dataset or a debug pausepoint) and issuing one `MULTI` write to an indexed key followed by one `FT.SEARCH`.

**Expected behavior**

`FT.SEARCH` during a fork window must not be able to block the main thread on work that only the main thread can unblock — e.g. `ProcessMultiQueue()` should not block on the writer pool while it is fork-suspended (defer the drain, or resume the writer pool from a non-main-thread path).

**Workaround (verified)**

Loading the module with `--reader-threads 0 --writer-threads 0` makes ingestion synchronous (`SyncProcessMutation`, `src/index_schema.cc:865-870`), so `EnqueueMultiMutation`/`ProcessMultiQueue` and the suspend/resume machinery never engage; the deadlock becomes structurally impossible. Note the asymmetry: both configs are registered with min=1, so `CONFIG SET`/config-file reject 0, but the module-argv path accepts 0/0 (covered by the `zero_rw_threads` unit test). Running this in our environment since 2026-07-21.

**Environment**

- `valkey/valkey-bundle:9.1.1-trixie@sha256:477e4579ddcdf15c9eedaae4c7c385774961da61227cd14697aba5ecbedc3aa2` (valkey core 9.1.0, search module ver 66049 = 1.2.1, json 10002 also loaded)
- Linux amd64, Kubernetes, standalone primary + replica + Sentinel (no cluster mode, no coordinator)
- `appendonly yes` (everysec), RDB save points active at the time, `repl-diskless-sync yes`
- 5 non-vector indexes (`FT.CREATE ... ON JSON`, tag/text fields), tiny corpus (~20 docs) — corpus size is clearly not a factor

We can provide full sanitized logs of all three incidents and additional timing data on request.

txchmechanicus:
## Summary

Fix a permanent main-thread deadlock when `FT.SEARCH` drains the multi/exec mutation queue
while a fork is in flight. Adds a deterministic regression test.

Fixes #1260

## Root cause

`IndexSchema::ProcessMultiQueue` unconditionally schedules pending multi/exec mutations onto
`mutations_thread_pool_` and waits for them on the calling thread. During a fork (BGSAVE,
AOF rewrite or replica full sync) that pool is suspended by `AtForkPrepare` and stays
suspended until the child is reaped, while every path that resumes it —
`OnForkChildCallback` and the `max-worker-suspension-secs` check in `OnServerCronCallback` —
runs on the main thread, i.e. the thread parked in `BlockingCounter::Wait()`. The server
stops answering, never processes SIGTERM, and has to be SIGKILLed.

Reproduced on 1.2.1 with a single `MULTI`/`HSET`/`EXEC` followed by one `FT.SEARCH` inside
the fork window; details in #1260.

## Fix

When the pool is suspended, drain the queue inline on the calling thread instead of handing
work to parked workers. The workers are stopped, so there is no concurrent writer to race
with, and the normal path is untouched.

`ProcessMultiQueueInline` deliberately does not hold `time_sliced_mutex_`:
`ProcessSingleMutationAsync` acquires it per mutation and the mutex is not recursive.
`stats_.mutation_queue_size_` is incremented before the call to stay balanced with the
decrement inside, mirroring `ScheduleMutation`.

This leaves the `max-worker-suspension-secs` valve still being checked from
`OnServerCronCallback`, i.e. on the main thread, so it cannot fire while that thread is
blocked — I can address that in a follow-up if you'd like it out of this PR.

## Testing

`integration/test_multi_queue_fork_deadlock.py` sends `BGSAVE`, one `MULTI`/`HSET`/`EXEC`
into an indexed prefix and `FT.SEARCH` as a single pipeline, so the server handles them in
one event loop iteration and the fork child cannot be reaped in between — deterministic
rather than timing dependent. It hangs and fails without the fix, passes with it.

Unit tests 22/22 binaries, integration suite 269 passed / 8 skipped, `clang-format` clean,
no new warnings including `-Wthread-safety`. Also verified A/B on `valkey-bundle:9.1.1` with
the same pod and dataset, only the module path differing: the stock module hangs on the
first attempt with every thread in `futex_wait_queue_me`, the patched one survives 5/5 runs.

### DIFF
--- a/integration/test_multi_queue_fork_deadlock.py
+++ b/integration/test_multi_queue_fork_deadlock.py
@@ -1,0 +1,54 @@
+"""Regression test for the fork-window deadlock reported in issue #1260.
+
+A write to an indexed key inside MULTI/EXEC is deferred to the multi/exec queue
+and only drained on the next FT.SEARCH. If a fork is in flight at that point the
+mutations thread pool is suspended, and draining it from the main thread never
+completes: the only paths that resume the pool run on that same thread.
+"""
+
+from valkey.client import Valkey
+from valkey_search_test_case import ValkeySearchTestCaseBase
+from valkeytestframework.conftest import resource_port_tracker
+from utils import run_in_thread
+from indexes import *
+
+# A single pending mutation must stay below the auto-drain threshold, which is
+# search.writer-threads.
+WRITER_THREADS = 2
+FILLER_KEYS = 1000
+SEARCH_TIMEOUT = 10
+
+index = Index("idx", [Tag("t")], prefixes=["p:"])
+
+
+class TestMultiQueueForkDeadlock(ValkeySearchTestCaseBase):
+    def test_search_during_fork_with_pending_multi_mutation(self):
+        """FT.SEARCH must not hang while a fork is in flight."""
+        client: Valkey = self.server.get_new_client()
+        client.execute_command("CONFIG SET", "search.writer-threads", WRITER_THREADS)
+        index.create(client)
+
+        # Give the RDB child enough work to outlive the pipeline below
+        for i in range(FILLER_KEYS):
+            client.hset(f"filler:{i}", mapping={"t": "x"})
+
+        # One pipeline, so the server handles all of it within a single event
+        # loop iteration and the fork child cannot be reaped in between
+        def bgsave_then_search():
+            pipe = self.server.get_new_client().pipeline(transaction=False)
+            pipe.execute_command("BGSAVE")
+            pipe.execute_command("MULTI")
+            pipe.execute_command("HSET", "p:1", "t", "run")
+            pipe.execute_command("EXEC")
+            pipe.execute_command("FT.SEARCH", index.name, "@t:{run}", "LIMIT", "0", "1")
+            return pipe.execute(raise_on_error=False)
+
+        thread, result, error = run_in_thread(bgsave_then_search)
+        thread.join(SEARCH_TIMEOUT)
+        assert not thread.is_alive(), "FT.SEARCH deadlocked during the fork window"
+        assert error[0] is None, error[0]
+
+        search_reply = result[0][-1]
+        assert not isinstance(search_reply, Exception), search_reply
+        assert search_reply[0] == 1
+        assert client.ping()
--- a/src/index_schema.cc
+++ b/src/index_schema.cc
@@ -820,6 +820,15 @@ void IndexSchema::ProcessMultiQueue() {
   Metrics::GetStats().ingest_last_batch_size = multi_mutations_keys.size();
   Metrics::GetStats().ingest_total_batches++;
 
+  // While a fork is in flight the mutations thread pool is suspended, and every
+  // path that resumes it runs on this thread, so waiting below would never
+  // return. Drain inline instead: the workers are parked, hence no concurrent
+  // writer to race with.
+  if (ABSL_PREDICT_FALSE(mutations_thread_pool_->IsSuspended())) {
+    ProcessMultiQueueInline();
+    return;
+  }
+
   absl::BlockingCounter blocking_counter(multi_mutations_keys.size());
   vmsdk::WriterMutexLock lock(&time_sliced_mutex_, false, true);
   while (!multi_mutations_keys.empty()) {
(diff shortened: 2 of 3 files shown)


mohanrajendran (approved):

Aksha1812:
**Reviewers for this PR**

Frank-Gu-81:
Hi @txchmechanicus 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

KarthikSubbarao (approved):

PR #1295, opened 2026-08-11
perf(query): cancel pending shards when fan-out fails

### LINKED ISSUE #1284: [OPTIMIZATION] Cancel pending shards when fan-out fails in ALLSHARD search
**What operations aren't optimal?**

  When an `ALLSHARDS` query receives its first shard error, the coordinator cancels only its own token. Active remote searches continue running because the coordinator never calls `grpc::ClientContext::TryCancel()`.

  Remote leaves already receive a `CallbackServerContext` and poll it through their existing cancellation checks. However:

  - `Client::SearchIndexPartition()` does not expose a cancellation handle.
  - Each outbound `ClientContext` is private to request state that is released from the completion callback.
  - Remote searches continue until their normal completion or the RPC deadline, currently 25 seconds by default.
  - gRPC retries may continue while the fan-out is already unrecoverable.
  - Late successful callbacks can still perform conversion, interning, and aggregation after the result is known to be unusable.

  The original first_node_error must remain the final client error. A cancellation caused by another shard failure must not replace it with `CANCELLED` or `DEADLINE_EXCEEDED`.

  **How could they be improved?**

  When the fan-out becomes unrecoverable:

  - Transition the tracker to a thread-safe terminal state exactly once.
  - Register each outbound search’s cancellation handle before dispatch.
  - Safely call `TryCancel()` on active remote RPCs, including requests queued for asynchronous dispatch.
  - Ensure terminal-state publication and handle registration are synchronized, so a request cannot start after cancellation without immediately receiving it.
  - Keep request state and `ClientContext` alive until the RPC callback completes.
  - Perform RPC cancellation outside the tracker mutex.
  - Skip late callback aggregation while retaining callback cleanup.
  - Preserve the first shard error and distinguish parent cancellation from deadline cancellation.
  - Reuse the existing remote cancellation polling; no new leaf polling checks are required.
  - Add deterministic remote tests, including cancellation versus dispatch/completion races, retry behavior, late callbacks, and verification that leaves observe gRPC cancellation.

  Consistency-failure and `SOMESHARDS` semantics should be handled separately.

mohanrajendran:
## Summary
  When an ALLSHARDS fan-out becomes unrecoverable (a shard error or local
  failure with partial results disabled, or a consistency failure), the
  coordinator now cancels the remaining in-flight remote RPCs via
  `ClientContext::TryCancel()` instead of letting them run to the RPC
  deadline (25s default).
  
  - Unified cancellation on `std::stop_source`; `cancel::Token` exposes a
    `std::stop_token`.
  - `Client::SearchIndexPartition` takes a stop token; each outbound
    `ClientContext` registers a `stop_callback` that calls `TryCancel()`,
    including async-queued dispatch. Request state/context stay alive
    until the RPC callback completes.
  - One shared token across local + remote paths; `CancelFanout()` is the
    single choke point, invoked outside the tracker mutex.
  - First shard error is preserved — an induced `CANCELLED` never
    overwrites `first_node_error`.
  - Cancelled RPCs counted separately from failures; added a
    `fanout-cancels` dev counter.
  
  ## Testing
  - Unit: token contract, queued async dispatch observing stop,
    first-error-wins, partial-results both directions, in-process gRPC
    TryCancel propagation.
  - Integration: end-to-end verification that leaves observe gRPC
    cancellation (`gRPCCancels`), plus consistency suites. Full suite green
    except pre-existing environment-only failures (compatibility ROOT_DIR,
    v1.0 module artifact).
  
  Closes #1284


Aksha1812:
**Reviewers for this PR**

Frank-Gu-81:
Hi @mohanrajendran 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @zackcam — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

PR #1304, opened 2026-08-18
Skip metadata reconciliation after replica demotion

mohanrajendran:
## Summary

Recheck the node role inside the queued `RunByMain` callback before reconciling asynchronously fetched metadata. If the node was demoted to a replica while the fetch was in flight, discard the stale reconciliation and avoid issuing `FT.INTERNAL_UPDATE` from a read-only node.

The regression test deterministically queues the callback, demotes the node, and verifies that neither metadata callbacks nor `FT.INTERNAL_UPDATE` run.

## Tests

- `./.build-debug/tests/coordinator_test --gtest_filter=*SkipsQueuedReconciliationAfterDemotion --gtest_brief=1`
- `./.build-debug/tests/coordinator_test --gtest_brief=1` (44 tests passed)
- `clang-format --dry-run --Werror src/coordinator/metadata_manager.cc testing/coordinator/metadata_manager_test.cc`


Aksha1812:
lets add this to 1.2 project board  as well and mark it to be backported @allenss-amazon since we saw this crash in 1.2 first

Nivesh-01 (approved):

VoletiRam (approved):

src/coordinator/metadata_manager.cc
```diff
@@ -465,6 +465,16 @@ void MetadataManager::HandleBroadcastedMetadata(
                           schema = std::unique_ptr<GlobalMetadata>(
                               response.release_metadata()),
                           address = std::move(address)] {
+          // The node may have been demoted after accepting the broadcast but
+          // before the asynchronous metadata fetch completed. Recheck the
+          // role on the main thread immediately before applying the update.
+          if (ValkeyModule_GetContextFlags(ctx) &
+              VALKEYMODULE_CTX_FLAGS_SLAVE) {
+            VMSDK_LOG_EVERY_N_SEC(DEBUG, ctx, 1)
```
allenss-amazon: I think this should get logged by default, a higher level (NOTICE?) should get used.

allenss-amazon: Also, let's move the ValkeyModule..... stuff into a helper function. Maybe in utils.cc in vmsdk.

mohanrajendran: Done in fb9c78e. The discard now logs at NOTICE instead of DEBUG (still rate-limited via VMSDK_LOG_EVERY_N_SEC).

mohanrajendran: Done in fb9c78e. Added vmsdk::IsReplica(ctx) in vmsdk/src/utils.{h,cc} alongside the existing IsRealUserClient/MultiOrLua context-flag helpers, and switched the three replica checks in metadata_manager.cc to use it.

PR #1312, opened 2026-08-21
Fix stale RDB restore counters inflating number_of_indexes metric

Shikha-code36:
## Summary
- \`rdb_restore_total_indexes\` counts RDB *sections* (includes the coordinator's non-index \`GLOBAL_METADATA\` section) while \`rdb_restore_completed_indexes\` only counts index schemas, so the two never converge when coordinator metadata is present, leaving a permanent residual.
- That residual was added into \`number_of_indexes\` unconditionally, even at rest.

## Fix
- Gate the residual term in \`number_of_indexes\` on \`rdb_restore_in_progress\`.
- Reset the restore counters on every exit path of \`PerformRDBLoad\` (success or failure) via \`absl::Cleanup\`.

## Test plan
- [x] \`bash ci/build_ubuntu.sh --run-tests\` — build succeeds, full unit test suite passes (incl. \`core_test\` / \`rdb_serialization_test.cc\`)"


mohanrajendran (approved):

Shikha-code36:
@mohanrajendran, I have rerun it let me know anything else require on my end.
Thanks :)

Shikha-code36:
Hi @allenss-amazon, could you please take a look at this PR when you get a chance? It fixes stale RDB restore counters inflating the number_of_indexes metric and ensures the restore counters are reset on all exit paths.

Tests are passing. Would appreciate a review and approval if everything looks good. Thanks!

Frank-Gu-81 (approved):

Frank-Gu-81:
Hi @Shikha-code36 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @Frank-Gu-81 — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

Aksha1812 (approved):
Fix looks right. One thing before it lands: I don't think this is actually covered by a test yet. test_number_of_indexes_on_restore.py asserts the at-rest count, but it runs standalone with no coordinator, so section_count == index_count and the residual is always 0 there, it passes even without this fix. The leak only shows up when there's a non-index section in the RDB, i.e. the GLOBAL_METADATA section in coordinator/cluster mode, which is what makes total > completed.
 
Could we add a small coordinator-mode case (e.g. reusing ValkeySearchClusterTestCaseDebugMode) that creates indexes, does SAVE + DEBUG RELOAD, and asserts search_number_of_indexes equals the true count at rest? That's the path that actually regressed, and it'd lock in both the gate and the counter reset.

  Want it shorter/blunter, or should I fold in the during-load over-count point too?

Shikha-code36:
> Fix looks right. One thing before it lands: I don't think this is actually covered by a test yet. test_number_of_indexes_on_restore.py asserts the at-rest count, but it runs standalone with no coordinator, so section_count == index_count and the residual is always 0 there, it passes even without this fix. The leak only shows up when there's a non-index section in the RDB, i.e. the GLOBAL_METADATA section in coordinator/cluster mode, which is what makes total > completed.
> 
> Could we add a small coordinator-mode case (e.g. reusing ValkeySearchClusterTestCaseDebugMode) that creates indexes, does SAVE + DEBUG RELOAD, and asserts search_number_of_indexes equals the true count at rest? That's the path that actually regressed, and it'd lock in both the gate and the counter reset.
> 
> Want it shorter/blunter, or should I fold in the during-load over-count point too?

Thanks for the review! I added `TestNumberOfIndexesOnRestoreCluster::test_number_of_indexes_at_rest_coordinator_mode` in `integration/test_number_of_indexes_on_restore.py`, following your suggestion, creates indexes, SAVE + DEBUG RELOAD in coordinator mode, asserts `search_number_of_indexes` equals the true count at rest. Verified it exercises the actual regression path: with indexes present, `MetadataManager` writes a GLOBAL_METADATA section alongside the index sections, which is exactly the mismatch (`rdb_restore_total_indexes` counting sections vs. `rdb_restore_completed_indexes` counting only index schemas) that caused the original bug. Ran it locally, passes against the fix. Kept it scoped to the at-rest assertion for now; happy to add the during-load over-count check too if you'd like it folded in.

src/rdb_serialization.cc
```diff
@@ -173,6 +174,15 @@ absl::Status PerformRDBLoad(ValkeyModuleCtx *ctx, SafeRDB *rdb, int encver) {
   Metrics::GetStats().rdb_restore_completed_indexes = 0;
   Metrics::GetStats().rdb_restore_current_index_keys_total = 0;
   Metrics::GetStats().rdb_restore_current_index_keys_loaded = 0;
+  // Clear progress tracking on every exit path (success or failure) so a
+  // stale residual can't leak into the at-rest number_of_indexes metric.
```
BCathcart, line 178: Thanks for fixing `number_of_indexes`! I'm considering whether we should make the change to clear them at the end of RDB load though (besides `rdb_restore_in_progress` of course). Right now they provide retrospective insight into the last RDB load. If the only motivation for clearing them is to fix `number_of_indexes`, then I'd suggest leaving them as is since the `rdb_restore_in_progress` gate should be enough. If it's more philosophical about the metric behaviour then I definitely don't mind having this discussion.

BCathcart: I'm going to merge this now without zeroing the metrics and we can always follow-up about it. Changing the behaviour of them is probably better gated to the next major version.

src/rdb_serialization.cc
```diff
@@ -173,6 +174,15 @@ absl::Status PerformRDBLoad(ValkeyModuleCtx *ctx, SafeRDB *rdb, int encver) {
...
+  absl::Cleanup clear_restore_progress = [] {
+    Metrics::GetStats().rdb_restore_in_progress = false;
+    Metrics::GetStats().rdb_restore_total_indexes = 0;
+    Metrics::GetStats().rdb_restore_completed_indexes = 0;
+    Metrics::GetStats().rdb_restore_current_index_keys_total = 0;
+    Metrics::GetStats().rdb_restore_current_index_keys_loaded = 0;
+  };
```
BCathcart, line 181:
```suggestion
  absl::Cleanup clear_restore_progress = [] {
    Metrics::GetStats().rdb_restore_in_progress = false;
  };
```

BCathcart (approved):

PR #1316, opened 2026-08-24
introduce VectorRegistry for memory sharing

yairgott:
This commit enables memory sharing between the module and the engine to reduce overall memory footprint.

Changes:

Implement VectorRegistry to maintain a single copy of the vector payload within the module and between the module and the engine.
Modify the load index flow to ignore the serialized hnswlib RDB during load keys, avoiding OOMs by discarding the RDB. Instead, once the mapping to the keys data is loaded, by VectorBase::LoadTrackedKeys, the vectors are loaded from the engine, shared with the engine and stored in hnswlib.


integration/test_vector_registry.py
```diff
@@ -0,0 +1,511 @@
...
+            vec_data = [float(i + j) for j in range(dim)]
+            vec_bytes = float_to_bytes(vec_data)
+            expected_vectors[key] = vec_bytes
+            client.hset(key, mapping={"vec": vec_bytes})
+
+        # Wait for mutations / backfill to finish indexing
+        waiters.wait_for_equal(
+            lambda: vector_index.info(client).num_docs,
+            num_vectors,
+        )
```
allenss-amazon: unneeded, correct?

integration/test_vector_registry.py
```diff
@@ -0,0 +1,511 @@
...
+        # 3. Using hget command, ensure received replies match expected vectors
+        for key, expected_bytes in expected_vectors.items():
+            got_bytes = client.hget(key, "vec")
+            assert got_bytes == expected_bytes, f"HGET returned unexpected value for key {key}"
+
+        # 4. Drop the index and ensure vector registry indicates it is empty
+        vector_index.drop(client)
+
+        waiters.wait_for_equal(
+            lambda: int(_get_vmsdk_info(client).get("vector_registry_entry_cnt", -1)),
```
allenss-amazon: How can this ever be true?

integration/test_vector_registry.py
```diff
@@ -0,0 +1,511 @@
...
+        waiters.wait_for_equal(lambda: vector_index.info(client).num_docs, 2)
+
+        # 3. Delete doc:1
+        client.delete("doc:1")
+
+        waiters.wait_for_equal(lambda: vector_index.info(client).num_docs, 1)
+
+        # 4. HGET doc:2 vec and verify it matches vec_bytes
+        val = client.hget("doc:2", "vec")
+        assert val == vec_bytes, f"Expected {vec_bytes}, got {val}"
```
allenss-amazon: This isn't verifying anything in the shared memory world. Not sure it even needs to be present. Either enhance it to do actual shared memory testing or remove it.

integration/test_vector_registry.py
```diff
@@ -0,0 +1,511 @@
...
+        server, client = self.create_server(
+            testdir=testdir,
+            server_path=server_path,
+            args=args,
+            port=port,
+            conf_file=conf_file,
+        )
+        self.wait_for_logfile(logfile, "Ready to accept connections")
+        client.ping()
+        return server, client
```
allenss-amazon: This seems like a lot of unnecessary duplicated code. Don't the base classes provide the proper mechanism for this?

integration/test_vector_registry.py
```diff
@@ -0,0 +1,511 @@
...
+import struct
+import pytest
+from valkey import Valkey
+from valkey_search_test_case import ValkeySearchTestCaseDebugMode, LOGS_DIR
+from valkeytestframework.conftest import resource_port_tracker
+from indexes import Index, Vector, KeyDataType, float_to_bytes
+from util import waiters
+
+
+def _get_vmsdk_info(client: Valkey) -> dict[str, str]:
```
allenss-amazon, line 11: style here. virtually all of the code in the test uses the template .... get('field', 0)  This certainly protects the pytest from failing. But makes it harder to read and seems to imply that these fields are optional -- which in general I think is not true. I'd much rather see them replaced with ['field'] which makes it easier to read and properly tests that the expected field is present.

integration/test_vector_registry.py
```diff
@@ -0,0 +1,511 @@
...
+        # 2. Ingest a vector and verify increments
+        key1 = "doc:1"
+        vec_data1 = [1.0] * dim
+        vec_bytes1 = float_to_bytes(vec_data1)
+        client.hset(key1, mapping={"vec": vec_bytes1})
+
+        waiters.wait_for_equal(
+            lambda: vector_index.info(client).num_docs,
+            1,
+        )
```
allenss-amazon: again, unneeded.

allenss-amazon (commented):
The new lifecycle test file does a good job of creating stimulus. Where it's weak is that the pass/fail conditions are typically 1) no crash and 2) ended up the with the right number of vectors in the registry. I would like to see the pass/fail criteria more difficult. An example of improved criteria would be whether there have there been any leaks of vectors, i.e., does the allocators usage count == expected ?

I'd also like to come up with some externally visible metric that would provide insight into the expected state of the HNSW index (w.r.t. sharing). Perhaps the deleted/unused entries count?

PR #1320, opened 2026-08-25
fix: guard HNSW prefetches against empty neighbor lists

bluayer:
## Summary

- Guard first-neighbor prefetches when the link list is empty.
- Guard second-neighbor prefetches when fewer than two neighbors exist.
- Skip upper-layer greedy traversal when the current node has no neighbors.

## Problem

HNSW reserves storage for neighbor slots independently of the active neighbor count. Values outside `getListCount()` are therefore not valid neighbors and may contain stale or uninitialized internal IDs.

Some prefetch paths evaluated `datal[0]` and `datal[1]` without first checking the neighbor count. Those IDs were then used to calculate addresses in `visited_array` and `getDataByInternalId()`.

## Behavior

This does not change graph topology or search behavior.

When an upper-layer link list is empty, the current entry point remains unchanged and traversal continues at the next lower layer.

## Tests

- Added `HnswHandlesEmptyNeighborLists`.
- `indexes_test`: 202 tests passed.
- `TestHNSWAllowReplaceDeleted`: 2 integration tests passed.


yairgott (approved):

Frank-Gu-81:
Hi @bluayer 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

bluayer:
Hi @Frank-Gu-81, Thank you for the reminder.
Hi @yairgott , Please let me know if there're any other steps left for merging this PR.

PR #1325, opened 2026-08-27
Fix VectorRegistry lifecycle management, multi-DB isolation, and server event handling

yairgott:
#### Summary of Changes
This PR addresses several lifecycle, synchronization, and memory-tracking issues in `VectorRegistry`, ensuring robust multi-index vector sharing, proper database-level (`dbnum`) isolation, and clean handling of Valkey server events (`FLUSHDB`, `FLUSHALL`, `SWAPDB`).
#### Key Improvements
1. **VectorRegistry Lifecycle & Multi-Schema Tracking**
   - Refactored `VectorRegistry` to track vector records per `(key, attribute, db_num)` and manage shared consumers across multiple index schemas.
   - Replaced redundant wrappers with direct deduplication and record tracking via `DedupOrConstruct`.
   - Corrected entry retention semantics: dropping an index schema (`FT.DROPINDEX`) does not purge database keys or registry entries that still exist in the keyspace.
2. **Server Event Handling & Multi-DB Isolation (`FLUSHDB` / `SWAPDB`)**
   - Implemented `VectorRegistry::OnFlushDB(const ValkeyModuleFlushInfo *flush_info)` to handle per-database `FLUSHDB` and global `FLUSHALL` (`dbnum == -1`), ensuring flush operations on one database do not delete entries in other databases.
   - Implemented `VectorRegistry::OnSwapDB(const ValkeyModuleSwapDbInfo *swap_info)` to properly re-map `dbnum` associations for all tracked entries on database swap.
   - Wired `OnFlushDBCallback` and `OnSwapDBCallback` in `src/server_events.cc`.
3. **IndexSchema & Attribute Pipeline Cleanup**
   - Streamlined attribute data handling (`AttributeDataType`, `AttributeData`, `VectorBase`) for consistent indexing and record mutation across HASH and JSON data types.
   - Ensured string normalization and raw vector tracking avoid unnecessary memory allocations.
4. **Testing & Verification**
   - **C++ Unit Tests** (`testing/vector_registry_test.cc`):
     - Added `FlushDBPreservesOtherDBEntries` to verify `dbnum` isolation during `FLUSHDB`.
     - Added `FlushDBAllClearsAllEntries` to verify complete clearance during `FLUSHALL`.
     - Added `SwapDBExchangesDBEntries` to verify correct entry re-mapping on `SWAPDB`.
   - **Integration Tests** (`integration/test_vector_registry_lifecycle.py` & `integration/test_vector_registry.py`):
     - Added `test_flushdb_preserves_other_database_entries` for both `HNSW` and `FLAT` algorithms.
     - Fixed lifecycle tests to remove invalid post-`FT.DROPINDEX` assertions expecting `entry_cnt == 0` when keys remain in the database.
     - Verified full test suite passes (100% passing across all 13 unit test suites and 101 integration tests).


src/vector_registry.cc
```diff
@@ -186,25 +195,25 @@ bool VectorRegistry::ShareWithValkey(
...
-              .db_num = db_num,
-              .key = key,
-              .attribute_identifier = attribute_identifier,
-          };
-          LockFreeUntrackIfUnused(search_key);
-        }
-      });
+    int db_num, const InternedStringPtr &interned_attribute_identifier,
+    uint64_t index_id) {
+  vmsdk::VerifyMainThread();
```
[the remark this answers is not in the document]
yairgott: VectorHNSW/VectorFlat are using DestructByMainThread

integration/test_vector_registry_lifecycle.py
```diff
@@ -737,7 +735,410 @@ def test_restore_replace_and_copy_replace_untracks_vector(self):
...
+        idx_b = self.build_index(name="idx_b", algo=algo)
+        idx_a.create(self.client, wait_for_backfill=True)
+        idx_b.create(self.client, wait_for_backfill=True)
+
+        count = 20
+        for i in range(count):
+            self.write_key(idx_a, str(i), [float(i + j) for j in range(DIM)])
+
+        self.wait_for_docs(idx_a, count)
+        self.wait_for_docs(idx_b, count)
```
allenss-amazon, line 1244: This should irrelvant. But what might be a concern is backfill from the index creation. Is there any possibility that backfill would make this test flakey? Or conversely do we need some backfill oriented tests of functionality in this area?

integration/test_vector_registry_lifecycle.py
```diff
@@ -737,7 +735,410 @@ def test_restore_replace_and_copy_replace_untracks_vector(self):
...
+
+        t.join()
+        assert not errors, f"Ingestion worker encountered errors: {errors}"
+
+        # Wait for index B to finish indexing all 50 keys
+        self.wait_for_docs(idx_b, 50)
+        assert self.registry_stat("entry_cnt") == 50
+
+        # Drop index B
+        idx_b.drop(self.client)
```
allenss-amazon: This test accomplishes none of what it purports to do. The creations are waiting for backfill to complete and the index drop happens when nothing is outstanding.

I would recommend that this test use pausepoints to suspend a backfill and to freeze the mutation queue in order to perform a drop while an index is actively ingesting data.

integration/test_vector_registry_lifecycle.py
```diff
@@ -737,7 +735,410 @@ def test_restore_replace_and_copy_replace_untracks_vector(self):
...
+                algo=rng.choice(algos)
+            )
+            idx.create(self.client, wait_for_backfill=True)
+            active_indexes[name] = idx
+            index_counter += 1
+
+        stop_event = threading.Event()
+        errors = []
+
+        def worker_mutations():
```
allenss-amazon, line 1489: This is a great test. I would like to see it extended to cover the unusual scenario where a key exists which would normally be present in two indexes. However, due to the sequencing of backfills, it's actually only present in one of the two indexes (this can easily be created by sequencing of backfills and index creations). This unusual scenario seems relatively untested.

src/indexes/vector_base.h
```diff
@@ -333,6 +335,8 @@ class VectorBase : public IndexBase {
   std::shared_ptr<const VectorRecord> GetOrConstructVectorRecord(
       const InternedStringPtr &key, absl::string_view record) const;
   UniqueFixedSizeAllocatorPtr vector_allocator_{nullptr, nullptr};
+  static inline std::atomic<uint64_t> next_index_id_{1};
+  uint64_t index_id_{next_index_id_.fetch_add(1, std::memory_order_relaxed)};
```
allenss-amazon: I see you're introducing a new way to identify indexes. It makes sense that for objects like vectors that inherently have a lifetime beyond a single index, that they would need a way to uniquely identify indexes other than by name. 

It seems to me that conceptually, you'd put a ptr to the IndexSchema, but a shared_ptr would create a circular reference, so why not use a weak_ptr ? Shouldn't it always be the case that a weak_ptr would be valid, as the IndexSchema shouldn't actually be destructed before the vector_base was destructor.

In thinking about this, it seems to me that there's no reason that IndexBase shouldn't have a weak_ptr to the containing IndexSchema for anybody's use -- it's sometimes useful for an index to be able to look at the containing IndexSchema....

src/index_schema.cc
```diff
@@ -595,23 +585,18 @@ void TrackRecord(const Key &key, const Attribute &attribute,
   }
   auto *vector_base =
```
allenss-amazon: Seems like CHECK(vector_base) is the correct code here.

src/index_schema.cc
```diff
@@ -595,23 +585,18 @@ void TrackRecord(const Key &key, const Attribute &attribute,
...
+  VectorRegistry::Instance().Track(key, record, attribute_data_type, db_num,
+                                   vector_base);
 }
 
 void IndexSchema::ProcessKeyspaceNotification(ValkeyModuleCtx *ctx,
                                               ValkeyModuleString *key,
                                               bool from_backfill) {
   auto key_cstr = vmsdk::ToStringView(key);
-  if (key_cstr.empty()) {
+  if (key_cstr.empty() || is_destructing_) {
```
allenss-amazon: Why is this needed?

src/indexes/vector_base.h
allenss-amazon: Obsolete — closing.

The `next_index_id_`/`index_id_` pair this comment is about is gone from the current head. The design moved away from per-index consumer tracking entirely: `VectorRegistry` now keys a single `VectorRecordWithSize` by `(db_num, key, attribute_identifier)` and holds no notion of which index schemas are consuming it. Retention is driven purely by keyspace presence (`DedupOrConstruct` with a null vector erases; `OnFlushDB`/`OnSwapDB` remap or drop whole databases), so there is no longer anything that needs to identify an index.

The broader suggestion — `IndexBase` holding a `weak_ptr` to its containing `IndexSchema` — is still a reasonable standalone idea, but it no longer has a dependency in this PR. Worth a separate issue if you want it.

src/index_schema.cc
allenss-amazon: Addressed (structurally, not literally) — closing.

The `TrackRecord` free function this was on no longer exists. It was replaced by `AppendMutatedAttribute` in `src/index_schema.cc`, where the `dynamic_cast` is now inside an `if (indexes::IsVectorIndex(index))` guard:

```cpp
if (indexes::IsVectorIndex(index)) {
  auto *vector_base = dynamic_cast<indexes::VectorBase *>(index.get());
  ...
}
```

So the unguarded dereference of a possibly-null `vector_base` that prompted the comment is no longer reachable — a null result would require a `kVector`/`kHNSW`/`kFlat` index that is not a `VectorBase`, i.e. a class-invariant violation rather than a runtime condition. A `CHECK(vector_base)` would still document that invariant for free if you want it as a follow-up nit, but the defect itself is gone.

src/vector_registry.cc
```diff
@@ -45,108 +43,140 @@ void VectorRegistry::Init(ValkeyModuleCtx *ctx) {
       << "Valkey version should be 9.0.1 and above";
```
allenss-amazon: Wrong version number and this is checked on module load.

src/server_events.cc
```diff
@@ -25,7 +26,11 @@ void OnForkChildCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
 
 void OnFlushDBCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent eid,
                        uint64_t subevent, void *data) {
-  SchemaManager::Instance().OnFlushDBCallback(ctx, eid, subevent, data);
+  if (subevent & VALKEYMODULE_SUBEVENT_FLUSHDB_END) {
```
allenss-amazon, line 29: This changes the qualification for call ing SchemaManager::Instance().OnFlushDBCallback. Which happens to not matter because it does nothing without this subevent flag anyways. But why change it? Altternatively -- rename the OnFlushDbCallback to be OnFlushEndDbCallback.

allenss-amazon (commented):
1. Can we split the changes to the benchmark flow into a separate PR?
2. Need to update the FT.INFO command documentation to reflect the new fields.
3. The default batch size for the ProcessPendingUnshares is a problem. Because the cost for this operation is variable on the vector size. Claude claims that the current value for 128-element FP32 vectors causes 167mSec during the timer call. This is too large.  I'd recommend switching to a time-based processing of the unshares as this makes the batch size independent of vector size.
4. In the testing. ExpectTracked is calling DedupOrConstruct. Which would guarantee that the output matches -- rendering the subsequent data compare test questionable. I'd recommend either a boolean parameter to disable the construct path OR a separate function that only looks up in the registry or fails.
5. I'd like to see a counter of unshares and a global count of pending_unshares, this would enable testing to ensure these are actually getting processed in an integration test.
6. OnUnload should still CHECK if there is a non-empty registry.
7. Flushall testing needs to be enhanced to include multiple databases defined. 
8. I'd like to see # of ActiveAllocations  exposed in INFO. Where the integration tests assert that the registry is empty, I'd like to see an additional check that the allocations are zero.
9. I'd like to see # of chunks exposed in INFO. Operationally, this will give us insight into vector fragmentation.

Frank-Gu-81:
Hi @yairgott 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @chinguyen21 — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

yairgott:
> 1. Can we split the changes to the benchmark flow into a separate PR?

Done. 

> 2. Need to update the FT.INFO command documentation to reflect the new fields.

Done.

> 3. The default batch size for the ProcessPendingUnshares is a problem. Because the cost for this operation is variable on the vector size. Claude claims that the current value for 128-element FP32 vectors causes 167mSec during the timer call. This is too large.  I'd recommend switching to a time-based processing of the unshares as this makes the batch size independent of vector size.

Great call. I've completely retired `options::GetVectorUnshareBatchSize()` in favor of `options::GetVectorUnshareTimeLimitMs()` (defaulting to 10ms). The main thread now processes pending unshares continuously until the time budget is exhausted. 

> 4. In the testing. ExpectTracked is calling DedupOrConstruct. Which would guarantee that the output matches -- rendering the subsequent data compare test questionable. I'd recommend either a boolean parameter to disable the construct path OR a separate function that only looks up in the registry or fails.

Fixed. I removed the `DedupOrConstruct` call from `ExpectTracked`. It now strictly utilizes `VectorRegistry::Instance().GetTrackedRecordForTest(...)` which purely looks up the record in the registry and fails the test if it's absent, guaranteeing we never accidentally construct state during verification.


> 5. I'd like to see a counter of unshares and a global count of pending_unshares, this would enable testing to ensure these are actually getting processed in an integration test.

Done. We added `completed_unshares` to the registry tracking and exposed all of this through `FT._DEBUG VECTOR_SHARING_STATS`. The newly added lifecycle integration tests explicitly poll these numbers to mathematically verify that the background unshare operations are fully executing and draining the pending queue successfully.

> 6. OnUnload should still CHECK if there is a non-empty registry.

Done. I have placed `CHECK_EQ(VectorRegistry::Instance().GetStats().entry_cnt, 0);` alongside a check for 0 pending unshares directly inside `ValkeySearch::OnUnload()` to strictly enforce this invariant.

> 7. Flushall testing needs to be enhanced to include multiple databases defined.

Added! `test_multi_index_flushall_cleanup` is now in the python test suite. It spans vector indexes across both `db0` and `db1`, triggers `FLUSHALL`, and cleanly asserts that the registry's `entry_cnt` completely zeroes out alongside the databases.

> 8. I'd like to see # of ActiveAllocations  exposed in INFO. Where the integration tests assert that the registry is empty, I'd like to see an additional check that the allocations are zero.

Done.

> 9. I'd like to see # of chunks exposed in INFO. Operationally, this will give us insight into vector fragmentation.

Done.

integration/test_vector_registry_lifecycle.py
yairgott: In this test (`test_multi_index_sharing_and_drop_one_by_one`), both indexes are created before data is ingested (`wait_for_backfill=True`), so backfill encounters zero existing keys and completes deterministically without race conditions.

To thoroughly test backfill dynamics and potential race conditions across multiple indexes, I added `test_multi_index_partial_backfill_state`. That test explicitly pre-populates keys and freezes the mutation/backfill queue using pausepoints for one index while the other completes, verifying that dropping or sharing works reliably under partial backfill states.

integration/test_vector_registry_lifecycle.py
yairgott: I refactored `test_multi_index_concurrent_ingest_and_drop` to use pausepoints (`block_mutation_queue` and `mutation_processing`). 
The test now explicitly freezes worker mutation processing at the pausepoint while writes accumulate, drops index A while mutation tasks are frozen/in-flight, and then unfreezes processing. It verifies that index B (which survived the drop) safely finishes indexing the remaining documents and the registry cleans up cleanly without leaks.

integration/test_vector_registry_lifecycle.py
yairgott: Done! I added `test_multi_index_partial_backfill_state` to cover this exact scenario:
1. Documents are written before any index exists.
2. Index A is created and allowed to fully backfill.
3. Index B is created, but its backfill is paused mid-flight via the `block_mutation_queue` pausepoint.
4. Index A is dropped while Index B is still partially backfilled.
5. The pausepoint is released, index B finishes backfilling, and we verify that the registry retains the records until Index B is also dropped.

src/index_schema.cc
yairgott: When an `IndexSchema` is being torn down (e.g. during `FT.DROPINDEX` or shutdown), `is_destructing_` is set to `true`. 
Checking `is_destructing_` ensures that late-arriving keyspace notification events or queued notifications are discarded immediately rather than attempting to mutate, allocate memory, or track records on a schema that is actively in the process of destruction.

src/server_events.cc
yairgott: Renamed the OnFlushDbCallback to be OnFlushEndDbCallback.

src/vector_registry.cc
yairgott: Updated! Corrected the error message to specify that `ValkeyModule_HashSetStringRef` and `ValkeyModule_HashHasStringRef` require Valkey 9.1.0 and above.

docs/commands/ft.info.md
```diff
@@ -75,6 +75,8 @@ An array of key value pairs.
   - `dimensions` (integer) Dimension count
   - `distance_metric` (string) Possible values are `L2`, `IP` or `COSINE`
   - `size` (integer) Number of valid vectors for this attribute
+  - `ActiveAllocations` (integer) Number of active memory allocations for vector data
+  - `ChunkCount` (integer) Number of memory chunks currently allocated by the vector memory allocator
```
allenss-amazon: This doesn't make sense to me. Aren't these values global? shouldn't they be part of INFO, not ft.info?

integration/test_vector_registry.py
```diff
@@ -22,6 +22,19 @@ def _get_vmsdk_info(client: Valkey) -> dict[str, str]:
...
+    try:
+        info = client.execute_command("FT.INFO", index_name)
+        attrs = info[info.index(b'attributes') + 1]
+        for attr in attrs:
+            if attr[attr.index(b'identifier') + 1] == attr_name.encode():
+                if b'ActiveAllocations' in attr:
+                    return int(attr[attr.index(b'ActiveAllocations') + 1])
+        return 0
+    except Exception:
+        return 0
```
allenss-amazon: If this is global, then all you care about is waiting for any pending unshares to complete, right?

src/schema_manager.cc
```diff
@@ -874,9 +874,9 @@ absl::Status SchemaManager::LoadIndex(
   return absl::OkStatus();
 }
 
-void SchemaManager::OnFlushDBCallback(ValkeyModuleCtx *ctx,
-                                      ValkeyModuleEvent eid, uint64_t subevent,
-                                      void *data) {
+void SchemaManager::OnFlushEndDBCallback(ValkeyModuleCtx *ctx,
+                                         ValkeyModuleEvent eid,
+                                         uint64_t subevent, void *data) {
   if (subevent & VALKEYMODULE_SUBEVENT_FLUSHDB_END) {
```
allenss-amazon: Isn't the IF statement redundant now?

allenss-amazon (changes requested):
Need to fix the INFO vs FT.INFO. Also, I'm happy with making the new fields Dev fields, so no doc is required.

allenss-amazon (approved):
assuming it builds. let's merge it.

PR #1326, opened 2026-08-27
Fix permanent main-thread deadlock in ProcessMultiQueue during fork

### LINKED ISSUE #1260: [BUG] Permanent main-thread deadlock: FT.SEARCH during a fork window blocks in ProcessMultiQueue() on the suspended writer pool

Shikha-code36:
`ProcessMultiQueue()` schedules pending multi/exec mutations onto the
writer thread pool and blocks the calling thread on an
`absl::BlockingCounter::Wait()`. `FT.SEARCH`/`FT.AGGREGATE` calls this
synchronously from the main thread (`QueryCommand::Execute`).

The writer pool is suspended for the entire lifetime of a fork child
(BGSAVE, AOF-rewrite preamble, replica full sync) via
`AtForkPrepare`/`AfterForkParent`, and is only resumed later by the
fork-child-died module event — which is itself delivered on the main
thread. If a `MULTI/EXEC` write queues a mutation and an `FT.SEARCH`
arrives while a fork child is still alive, the main thread blocks
waiting on a pool that only the main thread can resume: a permanent
deadlock. Only `SIGKILL` recovers.

Drain the multi-queue synchronously on the calling thread when the
writer pool is fork-suspended, instead of scheduling onto it and
blocking. This avoids the circular wait while preserving the pool's
suspension for its original purpose (limiting dirty-page growth from
concurrent writes during the fork child's COW lifetime).

Fixes #1260


src/index_schema.cc
```diff
@@ -846,6 +846,25 @@ void IndexSchema::ProcessMultiQueue() {
...
+    // would deadlock the main thread waiting on itself, so drain the queue
+    // synchronously on the calling thread instead.
+    while (!multi_mutations_keys.empty()) {
+      auto key = multi_mutations_keys.front();
+      multi_mutations_keys.pop_front();
+      {
+        absl::MutexLock lock(&stats_.mutex_);
+        ++stats_.mutation_queue_size_;
+      }
+      ProcessSingleMutationAsync(detached_ctx_.get(), /*from_backfill=*/false,
```
yairgott, line 862: With this approach, we lose the guarantee that multi-exec mutations received prior to a search command are fully ingested before the search executes.

As an alternative, consider using `vmsdk::BlockedClien`t to block the client issuing the search command. We could maintain a list of these blocked clients and resume their search commands as part of `AfterForkParent`. This resumption would re-enter the search flow, first waiting for the queued multi-exec mutations to be fully ingested before proceeding forward. WDYT?

yairgott: Also, it's important to cover this fix with tests.

Shikha-code36: Thanks for the detailed look, Yair.

I traced through the call chain and I don't think the pre-search ingestion guarantee is actually lost here — let me lay out why, and please push back if I'm missing a scenario:

- `ProcessMultiQueue()` is only ever called synchronously on the main thread, right before `query::Search()` runs (see `commands.cc:169`). That's true for both branches of this fix the normal thread-pool path and the new suspended-pool fallback.
- `ProcessSingleMutationAsync` (despite the name) does fully synchronous work, it calls `SyncProcessMutation` directly under `time_sliced_mutex_`, no further async dispatch. The fallback calls it inline in the loop, so every queued key is fully ingested before `ProcessMultiQueue()` returns.
- `ThreadPool::SuspendWorkers()` only lets a worker pause at a task boundary (`AwaitSuspensionCleared` is checked between tasks), so no worker can be suspended mid-task holding `time_sliced_mutex_`, no lock-ordering conflict with the fallback's direct call.

So functionally, the fallback branch trades "drain via the writer pool + block on a counter" for "drain inline on the main thread", same ingestion-before-search guarantee, different (worse) throughput for that one search call while the pool is suspended.

To check this against real behavior rather than just reading the code, I added `MultiExecTest.FtSearchMultiWriterPoolSuspended`, the writer-pool-suspended version of the existing `FtSearchMulti` test. It suspends the writer pool, queues multi-exec mutations, issues `FT.SEARCH`, and asserts all queued keys were ingested before the search ran. It passes.

That said, I do think your `BlockedClient` idea is a reasonable alternative for a different reason: it avoids stalling the main thread while it drains a potentially large queue during writer-pool suspension (up to 60s), at the cost of a bigger change (tracking blocked clients, resuming them at the right point). One note on the mechanics: the writer pool isn't actually resumed in `AfterForkParent`, that only resumes the reader/utility pools (`valkey_search.cc:1064-1084`); the writer pool comes back either when the fork child dies (`OnForkChildCallback`) or on the cron-driven suspension timeout, so resumption would need to hook one of those instead. Happy to discuss scoping that as a follow-up if you'd rather optimize for the main-thread-stall case, but wanted to flag that I don't see a correctness gap in the current fix.

Frank-Gu-81:
Hi @Shikha-code36 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @zackcam — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

zackcam (approved):
Looks good to me, the one comment if have is about https://github.com/valkey-io/valkey-search/pull/1326/changes#r3912356496 where you mention the idea around the stall that could occur. Said would be fixed as a follow up I think thats fine or doing in here. Up to maintainer on approach they want as I think the stall is a real issue

KarthikSubbarao:
closing  this as it is a duplicate of https://github.com/valkey-io/valkey-search/pull/1292 which was opened up with the fix first

PR #1331, opened 2026-08-28
Fix ThreadPool suspend/resume race with concurrent Resize

### LINKED ISSUE #1330: [BUG] ThreadPool suspend/resume crashes when the pool is resized during a fork (BGSAVE / full sync)
**Describe the bug**
 Resizing the worker thread pools via CONFIG SET search.reader-threads / search.writer-threads while a fork (BGSAVE / full sync) has them suspended corrupts the suspend/resume handshake, since it
  waits on a stale snapshot of the worker count. A worker created by the resize acknowledges a counter the suspension has already destroyed, crashing the server with a SIGSEGV

```
 === VALKEY BUG REPORT START: Cut & paste starting from here ===
  valkey 9.1.1 crashed by signal: 11, si_code: 1
  Accessing address: 0x8
  
  ------ STACK TRACE ------
  EIP:
  libsearch.so(_ZN4absl12lts_2024072215BlockingCounter14DecrementCountEv+0xd)
  
  255 write-worker-24 *      <-- crashing thread
  /lib/x86_64-linux-gnu/libc.so.6(+0x45330)
  libsearch.so(absl::lts_20240722::BlockingCounter::DecrementCount()+0xd)
  libsearch.so(vmsdk::ThreadPool::WorkerThread(std::shared_ptr<vmsdk::ThreadPool::Thread>)+0x75)
  libsearch.so(+0x629745)
  /lib/x86_64-linux-gnu/libc.so.6(+0x9cb84)
  /lib/x86_64-linux-gnu/libc.so.6(+0x129d6c)

```

**To Reproduce**
  
  1. Create an index, load a few thousand docs.
  2. Loop BGSAVE continuously (each fork suspends/resumes the pools).
  3. Concurrently flip CONFIG SET search.writer-threads / search.reader-threads between 1 and 8

baswanth09:
Fixes #1330 
Count workers instead of a stale BlockingCounter snapshot to keep the handshake balanced when the pool is resized during a fork suspension.


vmsdk/src/thread_pool.cc
```diff
@@ -179,77 +178,88 @@ void ThreadPool::JoinTerminatedWorkers() {
...
     {
       absl::MutexLock lock(&queue_mutex_);
-      AwaitSuspensionCleared();
+      AwaitSuspensionCleared(*thread);
+      if (suspend_workers_ && thread->IsShutdown()) {
+        // A worker retired during suspension exits immediately; queued tasks
+        // are left to the remaining workers.
+        --active_workers_;
+        thread->MarkJoinable();
+        return;
```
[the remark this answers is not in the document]
baswanth09: fixed in https://github.com/valkey-io/valkey-search/pull/1331/commits/cae2a0e32b7dad3b396c6c831305d3b661f4b43a

vmsdk/src/thread_pool.cc
```diff
@@ -179,77 +178,89 @@ void ThreadPool::JoinTerminatedWorkers() {
...
+    absl::MutexLock lock(&queue_mutex_);
+    ++active_workers_;
+  }
   while (true) {
     absl::AnyInvocable<void()> task;
     {
       absl::MutexLock lock(&queue_mutex_);
-      AwaitSuspensionCleared();
+      const bool was_suspended = AwaitSuspensionCleared(*thread);
+      if ((was_suspended || suspend_workers_) && thread->IsShutdown()) {
```
mohanrajendran: Overall, if there was no suspension (`was_suspended == false` and `suspend_workers_ == false`), this shutdown check will not happen. Subsequently, if we have tasks in queue too (`ThreadPool::QueueReady == true`), we will never enter a thread shutdown. We should add a test to catch this case. Looks like this was a bug earlier too, we only process the shutdown signal when the queue is drained.

```suggestion
      AwaitSuspensionCleared(*thread);
      if (thread->IsShutdown()) {
```

baswanth09: Makes sense. I will update it

baswanth09: Fixed in https://github.com/valkey-io/valkey-search/pull/1331/commits/e1491eb76957e32551f31addb8ce0342fa999897

vmsdk/src/thread_pool.cc
```diff
@@ -179,77 +178,88 @@ void ThreadPool::JoinTerminatedWorkers() {
...
+  {
+    absl::MutexLock lock(&queue_mutex_);
+    ++active_workers_;
+  }
   while (true) {
     absl::AnyInvocable<void()> task;
     {
       absl::MutexLock lock(&queue_mutex_);
-      AwaitSuspensionCleared();
+      AwaitSuspensionCleared(*thread);
```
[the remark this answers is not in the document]
baswanth09: Not an issue — `QueueReady()` also returns true when `suspend_workers_` is set, so an idle worker wakes, takes the `if (suspend_workers_) continue;` path, and gets counted in `AwaitSuspensionCleared`. Tests
  like `StartSuspendAndJoin` suspend fully idle pools and cover this.

mohanrajendran (approved):

allenss-amazon (approved):

PR #1333, opened 2026-08-29
Fix CMake build-type, target-arch, compiler gate, proto regen, and build.sh issues

mohanrajendran:
## Summary

This PR corrects a few build-system issues in the CMake modules and `build.sh`. The current `build.sh` flow does not hit them, but they affect `RelWithDebInfo`/`MinSizeRel` builds, cross-compilation, some compiler version strings, and regeneration of stale protobuf files.

## Changes

**`cmake/Modules/valkey_search.cmake`**
- Build-type classification used `if(CMAKE_BUILD_TYPE MATCHES Release)`, a regex substring test. `RelWithDebInfo` and `MinSizeRel` do not contain the literal `Release`, so they fell into the debug branch and got `-O0`, `-fno-lto`, and `NDEBUG` — an unoptimized binary with assertions stripped. Now uses exact `STREQUAL` checks for the three optimized build types.
- Stopped defining `NDEBUG` in the debug branch, so `assert()` stays active in debug builds.
- Target ISA flags now key off `CMAKE_SYSTEM_PROCESSOR` (target) instead of `CMAKE_HOST_SYSTEM_PROCESSOR` (build host), so cross-compilation emits the right code. Non-x86 behavior is unchanged.
- Compiler-version gate uses `CMAKE_CXX_COMPILER_VERSION VERSION_LESS` instead of a 3-component version regex that left the major version empty (and silently bypassed the check) for 1- or 2-component version strings.
- `message(WARN)` → `message(WARNING)` (`WARN` is not a valid mode; it was printed as literal text).

**`cmake/Modules/protobuf_generate.cmake`**
- Proto regeneration used `if(${GEN_CC_REQUIRED} OR ${GEN_FILE_H})`, which expands to `if(OFF OR /path/foo.pb.h)`; a bare path string is falsy, so header staleness (`GEN_H_REQUIRED`) was never consulted and a stale/missing `.pb.h` would not be regenerated. Now compares the `GEN_H_REQUIRED` boolean.

**`build.sh`**
- Added a `TEST_PATTERN=""` default so a later reference is safe under `set -u`.
- Removed a dead `BUILD_TOOL` assignment (always overwritten before use; the dead copy also used `ninja`, wrong on RedHat where it is `ninja-build`).
- `--clean` now runs the clean target on an existing build dir and exits, instead of also doing ICU setup and a conditional configure.
- `--jobs` is now honored by the ICU build (was hardcoded to all cores).
- Removed the CMake-file mtime scan from `is_configure_required`; Ninja and Make already re-run CMake when a `CMakeLists.txt`/`.cmake` file changes. `check_and_clean_on_branch_change` now sets `RUN_CMAKE=yes` when it deletes the configure-time `.pb.*` outputs, so a branch switch still forces a reconfigure.

**`scripts/common.rc`**
- The three TSAN valkey-server `sed` patches now fail loudly if the text they expect is missing (upstream drift), instead of silently doing nothing (`sed` returns success even when it matches nothing). The helper is idempotent for the cached clone.

## Testing
Individual changes were checked with CMake script-mode parse checks, `bash -n`, and small standalone reproductions of the changed logic. A full local build did not complete due to unrelated dependency issues in the local environment. Relying on CI for the full build and test run. Opening as a **draft** for that reason.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,6 +17,7 @@ ARGV=$@
 EXIT_CODE=0
 INTEG_RETRIES=1
 JOBS=""
+TEST_PATTERN=""
 CMAKE_GENERATOR=${CMAKE_GENERATOR:-"Ninja"}
 
 echo "Root directory: ${ROOT_DIR}"
@@ -182,12 +183,6 @@ export SAN_BUILD
 export ROOT_DIR
 . "${ROOT_DIR}/scripts/common.rc"
 
-if [[ "${CMAKE_GENERATOR}" == "Ninja" ]]; then
-  BUILD_TOOL="ninja"
-else
-  BUILD_TOOL="make -j$(num_proc)"
-fi
-
 function build_icu_if_needed() {
     printf "${BOLD_PINK}Checking ICU dependencies...${RESET}\n"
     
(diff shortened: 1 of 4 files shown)


mohanrajendran:
Will track here: https://github.com/valkey-io/valkey-search/pull/1225

Issue #1340, opened 2026-08-30
[BUG] FT.AGGREGATE: fields referenced only by a pipeline stage are never loaded, silently corrupting GROUPBY results


allenss-amazon:
**Describe the bug**

`FT.AGGREGATE` only fetches the fields named in the `LOAD` clause. A field referenced *solely* by a later pipeline stage (`GROUPBY`, `REDUCE`, `FILTER`, `SORTBY`, `APPLY`) is never fetched, so that stage sees a Nil value.

The most serious symptom is `GROUPBY`: grouping on a field that was not explicitly `LOAD`ed collapses **every record into a single group** and drops the grouping key from the reply. There is no error — the client receives a plausible-looking but wrong aggregation.

RediSearch auto-loads a field that a `GROUPBY` or `SORTBY` references. Affects `HASH` and `JSON` indexes equally.

**To Reproduce**

```bash
valkey-cli FT.CREATE h_idx ON HASH PREFIX 1 h: SCHEMA cat TAG price NUMERIC
valkey-cli HSET h:1 cat a price 10
valkey-cli HSET h:2 cat b price 20
valkey-cli HSET h:3 cat a price 30

# GROUPBY on a field that was not LOADed
valkey-cli FT.AGGREGATE h_idx "@price:[-inf inf]" GROUPBY 1 @cat REDUCE COUNT 0 AS cnt
```

**Actual result** — one group, grouping key absent:
```
1) (integer) 1
2) 1) "cnt"
   2) "3"
```

**Expected behavior** — what RediSearch returns:
```
1) (integer) 2
2) 1) "cat"
   2) "b"
   3) "cnt"
   4) "1"
3) 1) "cat"
   2) "a"
   3) "cnt"
   4) "2"
```

Adding `LOAD 1 @cat` makes valkey-search agree with RediSearch, which isolates the cause.

**Other stages, same root cause**

| Query (`LOAD 1 @__key` only) | RediSearch | valkey-search |
| --- | --- | --- |
| `GROUPBY 1 @cat REDUCE COUNT 0 AS cnt` | 2 groups, `cat` emitted | 1 group, `cat` absent |
| `GROUPBY 1 @cat REDUCE SUM 1 @price AS total` | `cat b total 20`, `cat a total 40` | `total 0` (key *and* reducer input Nil) |
| `SORTBY 2 @price ASC` | auto-loads, emits `price`, sorted | `price` absent, unsorted |
[...]

PR #1341, opened 2026-08-30
Implicitly load fields referenced by a pipeline stage (#919)

### LINKED ISSUE #919: [BUG] FT.AGGREGATE GROUPBY Without LOAD Produces Missing Group Key Fields

allenss-amazon:
## Problem

`FT.AGGREGATE` only fetches the fields named in the `LOAD` clause. A field referenced *solely* by a later pipeline stage is never fetched, so the stage sees a Nil value:

| query (nothing loaded, or `LOAD 1 @__key`) | Redisearch | valkey-search |
| --- | --- | --- |
| `GROUPBY 1 @cat REDUCE COUNT 0 AS cnt` | one group per distinct `cat`, each emitting it | **one group total, `cat` absent** |
| `GROUPBY 1 @cat REDUCE SUM 1 @price AS total` | `cat b total 20`, `cat a total 40` | `total 0` — key *and* reducer input Nil |
| `SORTBY 2 @price ASC` | auto-loads, emits `price`, sorted | `price` absent, rows unsorted |
| `FILTER @price>15` | error: ``Property `price` not loaded nor in pipeline`` | returns 0 records |
| `APPLY @price*2 AS dbl` | same error | `dbl` silently absent |

The `GROUPBY` case is the most damaging: the reply is well-formed and carries no error, so a silently wrong aggregation is easy to ship. Affects HASH and JSON alike.

## Fix

Every `@name` in a `GROUPBY` key, `REDUCE` argument, `SORTBY` key, `APPLY` or `FILTER` expression already becomes a record column during parsing — `GROUPBY` keys through `AggregateParameters::MakeReference()` directly, the rest through `expr::Expression::Compile`. The column table is therefore already a complete list of the names the pipeline references, so `ManipulateReturnsClause` consults it directly instead of walking the stages and their expression trees.

Only names that resolve to a declared attribute are loaded. Anything else is produced by the pipeline itself — an `APPLY` or `REDUCE` output, a chained `GROUPBY` over a reducer alias — and has no stored value to fetch.

Vector fields are deliberately skipped. Loading one is rejected outright, so auto-loading would turn a query that merely produced a Nil into a hard error; that is a larger behavior change than this fix intends.

## One deliberate extension

Redisearch auto-loads for `GROUPBY` and `SORTBY`, but *rejects* an un-loaded reference in `APPLY` and `FILTER`. This change loads them instead. That is strictly more permissive, so no query that works against Redisearch changes meaning — but it is a divergence, and worth an explicit decision during review. Rejecting instead would require the ordered "not loaded nor in pipeline" validation described below.

## Tests

`GROUPBY` and `SORTBY` are covered in the compatibility suite against captured Redisearch answers — 13 new cases across both key types, 104 captured answers. Verified failing before the fix (`KeyError: 't1'`, with valkey-search returning `[1, [b'cnt', b'15']]` where Redisearch returns 15 groups) and passing after.

`APPLY` and `FILTER` cannot be compatibility-tested: Redisearch errors on those, and the harness skips any case where the reference engine itself failed (`if expected["exception"]: return True`), so such cases would be silently vacuous. They are asserted directly in `integration/test_aggregate_autoload.py`, together with two guard cases — derived names must not be auto-loaded, and unknown fields must still be rejected.

- Unit tests: 22/22 suites
- Integration: 345 passed, 8 skipped, 0 failures
- Compatibility replay: 3/3 (both pickles, CMD and CME)
- `clang-format` clean; `typos` unchanged from the `main` baseline

## Issues and PRs

- Closes #919
- Obsoletes #997, which fixes the same defect but only for `GROUPBY` **keys**. Reducer arguments live in `reducers_[].args_` rather than `groups_`, so `GROUPBY 1 @cat REDUCE SUM 1 @price AS total` still returns `total 0` under that PR — with the group keys now correct, which arguably makes the wrong aggregate *harder* to notice. `SORTBY`, `APPLY` and `FILTER` are untouched by it. Verified by building #997 and running all five cases.
- Resolves four of the five cases in #1340.

## Not addressed

A reference to a name that is neither a declared attribute nor produced by an earlier stage. Redisearch rejects it; valkey-search still yields Nil:

```
LOAD 1 @price FILTER @dbl>25 APPLY @price*2 AS dbl   -> Redisearch: Property `dbl` not loaded nor in pipeline
LOAD 1 @price APPLY @price*2 AS dbl FILTER @dbl>25   -> works in both
```

Note this is order-sensitive — the same two stages reordered — so it needs an ordered walk over the stages against an accumulating set of available names, rather than the flat membership test used here. Left for a follow-up; the remaining part of #1340.


boda26 (approved):
lgtm

Issue #1306, opened 2026-08-19
Agent Steering Docs


BCathcart:
#1021 added a simple philosophical `CLAUDE.md`. We should flesh it out with some more specifics and add an `AGENTS.md` file as well.

PR #1343, opened 2026-08-31
Fix aarch64 RDB compatibility integration tests

mohanrajendran:
## Summary
- add an architecture-matched Valkey Search 1.0.0 module fixture for aarch64
- select the legacy module fixture by host architecture
- fix normal dependency packaging to include required dependencies
- document both integration test harnesses and compatibility fixture requirements

## Validation
- focused Python CMD/CME RDB compatibility tests: 2 passed
- C++ vector_search_integration suite: 47 passed
- full integration command started both harnesses; C++ suite passed, and the long Python suite was manually stopped after 33 tests passed during end-to-end skill validation
- Python syntax, shell syntax, ZIP integrity, and ELF architecture checks passed


mohanrajendran:
Also addresses https://github.com/valkey-io/valkey-search/issues/1306 by creating a skills folder.

.skills/run-integration-tests/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+   Use `build.sh --run-integration-tests=<pattern>` to select Python tests;
+   this skips the C++ harness. Use
+   `testing/integration/run.sh --test <suite>` for C++ suite selection.
+
+## Python Cluster Tests
+
+For coordinator, shard fan-out, replica, failover, and recovery changes:
+
+- Test standalone and cluster/coordinator modes where the path is shared.
+- Cover single-shard and cross-shard keys, indexing, backfill, search, and
```
BCathcart: > Cover single-shard and cross-shard keys

What do you mean by "cross-shard keys"?

.skills/run-integration-tests/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+   testing/integration/run.sh --help
+   ```
+
+   Use `build.sh --run-integration-tests=<pattern>` to select Python tests;
+   this skips the C++ harness. Use
+   `testing/integration/run.sh --test <suite>` for C++ suite selection.
+
+## Python Cluster Tests
+
+For coordinator, shard fan-out, replica, failover, and recovery changes:
```
BCathcart: This section gets a bit abstract. It might be simpler for an agent to just run the whole test suites at the end of its changes rather than trying to figure out which tests are relevant? Or if a developer is asking it to run a certain test and it sees it's related to replication, might it start running a bunch of tests?

.skills/run-integration-tests/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+- Include partial shard failures, replicas, failover, and recovery where
+  applicable.
+- Confirm the harness created the expected nodes, slots, replicas, and
+  generated per-node configuration files.
+
+The harness starts nodes through `integration/valkeytestframework`, writes
+module arguments into configuration files, creates clusters with
+`valkey-cli`, and waits for topology convergence.
+
+## Compatibility Tests
```
BCathcart: We generally refer to "compatibility tests" in this repo as comparing compatibility with the Redis API

.skills/run-integration-tests/SKILL.md
```diff
@@ -0,0 +1,91 @@
...
+
+RDB or module-version tests need the exact historical server/module fixtures,
```
BCathcart: A lot of RDB tests don't require historical server/module fixtures.

I think the overall gist here is good of not testing an old module plus a current server and counting it as full historical-runtime coverage. Is this relevant for a skill about running integration tests though? This sounds like more advice on how to not write bad integration tests.

BCathcart (commented):
Overall LGTM. Left a few comments on the skill. Let's backport this to 1.2 since it has the broken test

.agents/skills/run-integration-tests/SKILL.md
```diff
@@ -0,0 +1,60 @@
...
+   testing/integration/run.sh --help
+   ```
+
+   Use `build.sh --run-integration-tests=<pattern>` for Python tests or
+   `testing/integration/run.sh --test <suite>` for C++ suite selection.
+
+## Python Cluster Tests
+
+For coordinator or replica changes, run the applicable test in standalone and
+cluster modes. For fan-out coverage, use data whose matching documents span at
```
BCathcart: There is no coordinator in standalone mode. I'd consider removing this section. It's the one thing that takes this skill from "how" to run tests towards "what" tests to run.

BCathcart (approved):
Approved with one more comment needing resolution

.agents/skills/run-integration-tests/SKILL.md
mohanrajendran: Pared it down even further, I still think some prescription on what tests to run should be useful. Eg. smaller suite is run in the middle of a development cycle, vs full suite right before PR.

PR #1287, opened 2026-08-05
Scoring: BM25STD for text, numeric and tag

boda26:
### Summary

  This branch adds text relevance scoring to non-vector `FT.SEARCH` queries, introducing a pluggable scorer framework with a standard Okapi **BM25STD** implementation. Documents matched by
  text, tag, and numeric predicates are now ranked by relevance instead of arriving unordered, and clients can retrieve per-document scores via a new `WITHSCORES` option. Scoring works in both
  standalone and cluster (fanout) modes, honors per-predicate `$weight` query modifiers, and stays scale-consistent even when documents mutate mid-query.

  ### Motivation

  Previously, non-vector queries returned matches with no relevance ordering (`distance` was reported as the "score"). This adds true full-text ranking comparable to Redis's BM25STD, so
  results are ordered by how well they match the query.

  ### Key changes

  **Scoring framework** (`src/indexes/scoring/`)
  - New stateless, thread-safe `Scorer` interface with `PrecomputeIDF` / `ScoreLeaf` / `ComposeDocumentScore` seams, plus a `Bm25StdScorer` (k1=1.2, b=0.75) and a `GetScorer(ScorerType)`
  registry.
  - IDF is precomputed once per term; leaf scoring is an inlinable hot path.

  **Query API** (`src/commands/`)
  - `FT.SEARCH` accepts `WITHSCORES` (score reported as a top-level reply element, Redis-compatible) and `SCORER` (currently `BM25STD`).
  - QMA `$weight:N` block parsing (`=> {$weight:N}`) for per-predicate weighting, with positive-float validation.

  **Scoring engine** (`src/query/search.cc`)
  - Two scoring paths: an **in-iterator fast path** for pure-text queries, and an **extra-step `ScoreTextQuery`** post-filter tree walk for combined queries (text + numeric/tag/negate) and recompute, so both
  leaf and group weights are honored.
  - `ResolveLeaves` hoists all document-independent work (posting lists, stem-variant expansion, dt, IDF) out of the per-document loop; `ScoreNode` recursively composes AND (all-match) / OR
  (any-match) subtrees.
  - Tag values scored as BM25 terms (F≡1); numeric ranges treated as filters (contribute 0).
  - Results sorted by score descending via `partial_sort` up to the requested limit, with a deterministic key tie-break.

  **Precomputation & storage**
  - Per-document length/norm and corpus `total_doc_len` tracked in `TextIndexSchema`; BM25's N counts all indexed docs via a total-docs provider.

[...]

PR #1345, opened 2026-09-01
Scoring: score all-text OR by fanning out fetchers instead of building a OR composite iterator

boda26:
# Score all-text OR by fanning out fetchers instead of building a composite iterator

## Motivation

Follows up the `TODO: optimize with eliminate BuildTextIterator for OR in followup PR` left in `EvaluateFilterAsPrimary` by #1287.

Before this change, a pure-text OR was planned as a **single** fetcher wrapping an `OrProximityIterator`: `BuildTextIterator` was called on the whole `kComposedOr` subtree, the iterator unioned the branches internally, and it emitted each matching doc once with the branch scores already summed. That works, but it means OR is the only operator that has to construct a composite iterator purely to compute a sum the drain loop is already positioned to compute, and it forces every leaf type to be reachable through `OrProximityIterator` before it can participate in an OR.

This PR removes that block. OR now recurses per child like every other multi-child case, appending one fetcher per branch to `entries_fetchers`.

## Performance
```
# Before the change (build OR composite iterator)
OR of two          7551 (73.73)         4895 (115.33)            -35.6%     +56.6%
OR of three        5404 (105.54)        3377 (165.12)            -37.5%     +56.5%

# After the change (populate or_weight_multiple in entry fetcher)
OR of two          7619 (73.73)         6549 (84.86)             -14.8%     +15.1%
OR of three        5376 (105.79)        4690 (119.81)            -12.7%     +13.7%
```
The refactor brings significant performance improvement, augmenting throughoutput regression from -35% to -13%, and latency regression from +56% to +14%. This closes a significant performance gap specifically on pure-text OR queries.

## Changes

**1. Group weights ride down to the leaf.** `OrProximityIterator` was the thing applying a group weight for OR; with it gone, the weight has to reach the leaf another way. `EvaluateFilterAsPrimary` and `TextPredicate::BuildTextIterator` take a new `or_weight_multiplier` (default `1.0f`), and the OR branch multiplies it by its own group weight before recursing:

```cpp
const float child_multiplier = or_weight_multiplier * composed_predicate->GetWeight();
```

`TermPredicate::BuildTextIterator` folds it into the leaf as `GetWeight() * or_weight_multiplier`. This is sound specifically because **OR is a sum**, so a group weight distributes onto the leaves exactly — unlike an AND with slop, where the group weight is not a per-leaf scalar. AND therefore keeps applying its weight at the node, unchanged.

**2. Dedup accumulates instead of dropping.** `absl::flat_hash_set<const char*> seen_keys` becomes `absl::flat_hash_map<const char*, size_t> seen_at`, mapping the interned key pointer to the doc's slot in `borrowed`. A repeat sighting adds into that slot rather than skipping.

**3. Compose once, after the drain.** The per-doc `ComposeDocumentScore` moved out of the loop into a trailing pass over `borrowed`.

## Why dedup-add rather than dedup-drop

Dedup used to be lossless, and it stopped being lossless for one reason: **the OR union moved out of the iterator and into the dedup map.**

In the old plan, the only paths that produced duplicate sightings were tag and (on the prefilter path) negate — and on both, in-iterator scoring was off, so a dropped sighting carried a score of `0.0f`. Nothing was lost. Pure-text OR, the one case where a duplicate *would* have carried real score, never produced duplicates at all, because `OrProximityIterator` had already merged the branches upstream. The `seen_keys` set was dead weight on that path.

Fanning out inverts that. A doc matching *k* branches is now emitted *k* times, once per fetcher, and each sighting carries **only that branch's contribution** — its own leaf score times its own accumulated multiplier. So for `hello | (rare | unique)`, `doc:6` arrives twice: once worth `rare`, once worth `unique`. Dropping the second sighting would report `doc:6` scored on `rare` alone and silently under-rank it. The doc's true score only exists as the sum across sightings, so dedup has to be the thing that sums it:

```cpp
auto [it, inserted] = seen_at.try_emplace(key->Str().data(), borrowed.size());
if (!inserted) {
  borrowed[it->second].score += raw;   // another branch matched this doc
  iterator->Next();
  continue;
}
```

The score is also read *before* the dedup check now, since under the old ordering a repeat sighting returned early and never reached the score read.

The map stores an index rather than a pointer or reference because `borrowed` keeps growing via `push_back` during the drain, and reallocation would invalidate either of those. Pointer keys remain valid identity because `InternedStringPtr` guarantees one allocation per distinct string — unchanged from `seen_keys`.

## Why compose moved out of the loop

`ComposeDocumentScore(sum_of_terms, document_score)` is a transform on the *final* sum, and with fan-out the final sum isn't known until the whole fetcher queue is drained — the branch that bumps a doc may be several fetchers later. Composing at push time would compose a partial sum and then pile raw addends on top of an already-composed value.

`Bm25StdScorer` happens to be a multiply, so it algebraically distributes over addition; but it is a virtual on the `Scorer` interface, nothing binds a future scorer to being linear, the inf short-circuits are not linear, and `a*d + b*d` is not bit-identical to `(a+b)*d` in float. Summing first and composing once reproduces the old `OrProximityIterator` ordering exactly. It also drops a `GetDocumentScore` lookup and a `BorrowedInternedStringPtr` construction per *sighting* down to one per unique doc.

## Blast radius

The accumulation only ever fires for pure-text OR. `score_in_drain` requires `!has_non_text_predicate`, which excludes numeric, tag, and negate; `needs_dedup` is `has_or || has_tag || has_negate`; intersecting the two collapses the disjunction to `has_or`. Concretely:

- **tag** reaches the dedup path but with `score_in_drain == false`, so `raw` stays `0.0f` and `+= raw` is a no-op — identical zeros, then scored by `ScoreTextQuery` as before.
- **negate** never reaches it: `IsUnsolvedQuery` returns true for any `kContainsNegate`, so control goes to the prefilter branch and the drain loop doesn't run.
- **pure-text AND** has `needs_dedup == false` and one fetcher, so no map is built and repeats are impossible.

`iterator_scoring_enabled` was renamed `score_in_drain` and gained `&& !is_match_all`, which folds a condition that was previously re-tested at the extra-step call site into the flag itself — match-all's universal-set scan carries no `TextIterator`, so in-iterator scoring was already inert for it.

## Testing

Two cases added to `test_scoring.py` Group 2 (`test_and_or`):

- `hello | (rare | unique)` — a nested OR is one flat union; `doc:6` matches both inner branches and must accumulate `rare + unique`, `doc:8` only `rare`.
- `((hello)=>{$weight:2} | ((rare)=>{$weight:3} | (unique)=>{$weight:2})=>{$weight:2})=>{$weight:2}` — multipliers accumulate down both OR levels (2, then 4), so hello lands at 4×, rare at 12×, unique at 8×, and `doc:6` sums its two inner branches at *different* multipliers.

Expectations are written against baselines fetched in-test rather than hardcoded floats, so they assert the multiplier algebra rather than a transcription of it. Full scoring suite passes (22/22, including `test_scoring_cluster.py`). Confirmed the new assertions are discriminating by mutating `12 *` → `11 *` and observing the expected failure.

## Known gap (pre-existing, not introduced here)

`PrefixPredicate`, `SuffixPredicate`, `FuzzyPredicate` accept `or_weight_multiplier` but ignore it — they construct `TermIterator` via the defaulted `leaf_weight = 1.0f, scorer = nullptr` path, so they are unscored entirely and already ignore their own `GetWeight()`. This PR doesn't regress them, but it does add a parameter they silently discard. Worth a follow-up to either wire them into the scored constructor or document the limitation.

### DIFF
--- a/src/query/search.cc
+++ b/src/query/search.cc
@@ -231,10 +231,12 @@ inline bool NeedsDeduplication(QueryOperations query_operations) {
 
 // Builds TextIterator for text predicates. Returns pair of iterator and
 // estimated size.
+// or_weight_multiplier lands on this node only; children are built with 1.0f.
 std::pair<std::unique_ptr<indexes::text::TextIterator>, size_t>
 BuildTextIterator(const Predicate *predicate, bool negate,
                   bool require_positions, bool is_vec_query,
-                  const indexes::scoring::Scorer *scorer) {
+                  const indexes::scoring::Scorer *scorer,
+                  float or_weight_multiplier = 1.0f) {
   if (predicate->GetType() == PredicateType::kComposedAnd ||
       predicate->GetType() == PredicateType::kComposedOr) {
     auto composed_predicate =
@@ -266,7 +268,7 @@ BuildTextIterator(const Predicate *predicate, bool negate,
       size_t total_size = min_size == SIZE_MAX ? 0 : min_size;
       return {std::make_unique<indexes::text::ProximityIterator>(
                   std::move(iterators), slop, inorder, skip_positional,
-                  composed_predicate->GetWeight()),
+                  composed_predicate->GetWeight() * or_weight_multiplier),
               total_size};
     } else {
       absl::InlinedVector<std::unique_ptr<indexes::text::TextIterator>,
@@ -290,7 +292,8 @@ BuildTextIterator(const Predicate *predicate, bool negate,
         return {nullptr, 0};
       }
       return {std::make_unique<indexes::text::OrProximityIterator>(
-                  std::move(iterators), composed_predicate->GetWeight()),
+                  std::move(iterators),
+                  composed_predicate->GetWeight() * or_weight_multiplier),
               total_size};
     }
   }
@@ -302,8 +305,8 @@ BuildTextIterator(const Predicate *predicate, bool negate,
     // Stamp the query-selected scorer so TermPredicate::BuildTextIterator can
     // build a scored TermIterator without a hardcoded scorer.
     text_predicate->SetScorer(scorer);
-    auto result = text_predicate->BuildTextIterator(text_index, field_mask,
-                                                    require_positions);
+    auto result = text_predicate->BuildTextIterator(
+        text_index, field_mask, require_positions, or_weight_multiplier);
     return {std::move(result), size};
   }
   if (predicate->GetType() == PredicateType::kNegate) {
@@ -317,7 +320,7 @@ BuildTextIterator(const Predicate *predicate, bool negate,
 size_t EvaluateFilterAsPrimary(
     const SearchParameters &parameters, const Predicate *predicate,
     std::queue<std::unique_ptr<indexes::EntriesFetcherBase>> &entries_fetchers,
-    bool negate) {
+    bool negate, float or_weight_multiplier) {
   const QueryOperations query_operations =
       parameters.filter_parse_results.query_operations;
   const IndexSchema *index_schema = parameters.index_schema.get();
@@ -341,9 +344,9 @@ size_t EvaluateFilterAsPrimary(
     auto predicate_type =
         EvaluateAsComposedPredicate(composed_predicate, negate);
     if (predicate_type == PredicateType::kComposedAnd) {
-      auto [text_iter, size] =
-          BuildTextIterator(composed_predicate, negate, false, is_vec_query,
-                            indexes::scoring::GetScorer(parameters.scorer));
+      auto [text_iter, size] = BuildTextIterator(
+          composed_predicate, negate, false, is_vec_query,
+          indexes::scoring::GetScorer(parameters.scorer), or_weight_multiplier);
       if (text_iter) {
         entries_fetchers.push(
             std::make_unique<indexes::text::TextIteratorFetcher>(
@@ -354,8 +357,8 @@ size_t EvaluateFilterAsPrimary(
       std::queue<std::unique_ptr<indexes::EntriesFetcherBase>> best_fetchers;
       for (const auto &child : composed_predicate->GetChildren()) {
         std::queue<std::unique_ptr<indexes::EntriesFetcherBase>> child_fetchers;
-        size_t child_size = EvaluateFilterAsPrimary(parameters, child.get(),
-                                                    child_fetchers, negate);
+        size_t child_size = EvaluateFilterAsPrimary(
+            parameters, child.get(), child_fetchers, negate, or_weight_multiplier);
         if (child_size < min_size) {
           min_size = child_size;
           best_fetchers = std::move(child_fetchers);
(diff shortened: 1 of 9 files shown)


zackcam (approved):
Looks good to me, should make sure when we merge this or prefix we get the conflicts sorted correctly. Specifically the or_weight_multiplier

BCathcart:
Can you explain to me the outcome of the conversation about this:

https://github.com/valkey-io/valkey-search/pull/1287#discussion_r3884123296
> This makes sense to me but let's ask @KarthikSubbarao why we didn't build a text iterator for composed OR before and instead fell back to the flat_hash_set de-dup method. Also since NeedsDeduplication() is true if query_operations & QueryOperations::kContainsOr, we'll unnecessarily add the keys to a flat_hash_set for composed OR with only text even though the text iterator already skips duplicates I think. We probably want to optimize this.

I want to understand things at a high level. In this PR you took out building the text iterator for all-text OR which would have done the de-duping. Instead we use a flat_hash_map at the end to de-dup (aka the "drain loop"). We still use the text iterator for all-text AND and skip the drain loop. Why did we make the decision to use the drain loop for OR and not for AND in the all-text case?

Essentially, why can't we simply make this change:
```
inline bool NeedsDeduplication(QueryOperations query_operations) {
  bool has_tag = query_operations & QueryOperations::kContainsTag;
  bool has_num = query_operations & QueryOperations::kContainsNumeric;
  return has_tag || has_num;
}
```

boda26:
I benchmarked again with skipping the outer dedup and only building the or iterators. The perf results improved ~10% but still far behind the results of the current PR.
```
  scenario           baseline rps(p50)    candidate rps(p50)      rps Δ%    p50 Δ%
  OR of two          7701 (73.47)         5416 (104.06)            -29.3%     +41.3%
  OR of three        5325 (105.15)        3836 (150.27)            -29.4%     +42.8%
  ```

**Comment by boda26** [comment]
I think the OR iterator is designed to touch and emit far more documents than the AND iterator. There is no single point that OR cost more than AND, but the total amount of operations. I think this PR is the right way to solve the perf problem
```
┌────────────────┬────────────────────────────┬──────────────────────┐
│                │  posting entries touched   │  documents emitted   │
├────────────────┼────────────────────────────┼──────────────────────┤
│ a b c (AND)    │ ≤ |A|+|B|+|C|, sometimes ≪ │ ≤ min(|A|,|B|,|C|)   │
├────────────────┼────────────────────────────┼──────────────────────┤
│ a | b | c (OR) │ = |A|+|B|+|C|, exactly     │ = |A∪B∪C| ≥ max(...) │
└────────────────┴────────────────────────────┴──────────────────────┘
```

File: src/query/search.cc [hunk 1]
```diff
@@ -1201,12 +1201,25 @@ absl::StatusOr<std::vector<indexes::BorrowedNeighbor>> DoSearchNonVector(
       while (!iterator->Done()) {
         const auto &key = **iterator;
         BACKGROUND_PAUSEPOINT("search_entries_fetcher");
+        // Read before dedup: a repeat sighting is another branch's match.
+        float raw = 0.0f;
+        if (score_in_drain) {
+          if (auto *text_iter = iterator->GetTextIterator()) {
+            raw = text_iter->GetScore() * text_iter->GetWeight();
```
BCathcart: nit: you could consider applying the weight inside GetScore() to avoid the callers having to remember to apply it

BCathcart (approved):

BCathcart:
Thanks for grabbing some perf numbers @boda26! Creating a heap with `FindMinimumKey` every step of the OrProximityIterator is probably where the overhead is coming from compared to the drain loop

Issue #1346, opened 2026-09-01
Reviewer Triage Board


Aksha1812:
# 🗂️ Reviewer Triage Board

![open PRs](https://img.shields.io/badge/open_PRs-61-blue) ![P1](https://img.shields.io/badge/P1-15-red) ![P2](https://img.shields.io/badge/P2-10-orange) ![P3](https://img.shields.io/badge/P3-14-yellow) ![needs first-pass](https://img.shields.io/badge/needs_first--pass-1-critical) ![ready for maintainer](https://img.shields.io/badge/ready_for_maintainer-9-blueviolet) ![approved](https://img.shields.io/badge/approved-4-brightgreen) ![stale >7d](https://img.shields.io/badge/stale_%3E7d-6-lightgrey)

_Auto-updated every ~15 min from **valkey-io/valkey-search** open PRs. Last run: **2026-09-11 09:34 UTC**._

```mermaid
pie showData title PRs by priority
    "P1" : 15
    "P2" : 10
    "P3" : 14
    "P4/none" : 22
```

## ✍️ How to update this board

**Repository collaborators and maintainers** can drive this board: add a **comment** with one or more commands; the bot applies it, then deletes your comment so this page stays fast.

**General commands** — act on a PR (the priority tables below):

```
/priority 1234 P2      set priority (P1–P4; `/priority 1234 clear` to remove)
/note 1234 some text   set a note (≤500 chars; empty clears it)
/stage 1234 maintainer override stage: review | changes | maintainer | approved | auto
/needs-review 1234     re-request maintainer review (clears maintainers’ Reviewed)
/update                refresh the board from the latest PR data now
```

**Personal commands** — act on you (your row in *Your queue*):

```
/claim 1234        add the PR to your queue / put your name on it
/unclaim 1234      remove your claim
/reviewed 1234     mark reviewed (only if you did NOT approve on GitHub)
/unreviewed 1234   un-mark it
```

> **Approving a PR on GitHub already marks it Reviewed ✅ for you — you do not need `/reviewed`.** And if you later withdraw or change that approval, it clears itself again. The board picks this up on the next ~15-min refresh, or immediately if anyone runs `/update`. `/reviewed` / `/unreviewed` are only for a review you did *without* a formal GitHub approval.

## ℹ️ Legend

_Expand a column to see what each symbol means. AI reviewers (CodeRabbit, Greptile, …) are excluded from every count._

[...]

PR #1350, opened 2026-09-01
Scoring: BM25STD scoring for prefix, suffix and fuzzy queries

boda26:
## Overview

  Previously, term-expansion queries were **not scored** — prefix (`cat*`), suffix
  (`*ing`), fuzzy (`%cat%`), and tag prefix (`@cat:{re*}`) all contributed `0` to a
  document's BM25STD score. This PR makes them produce real BM25STD scores on both
  the in-iterator path (pure-text queries) and the extra-step path (combined
  text + numeric/tag/negate queries), for **text** prefix/suffix/fuzzy and for
  **tag** prefix.

  ## Core contract

  An expansion is scored on **exactly one matched term/value's** BM25 (its own IDF
  + its own F), **never the sum** over all matched terms — matching Redis's
  single-representative behavior (`docs/redis_prefix_suffix_fuzzy_scoring.md`).
  *Which* term wins is an unspecified, corpus-dependent union artifact, so the two
  code paths may pick different representatives on a doc matching several terms;
  both honor the one-term invariant. An explicit tag union (`{red|blue}`) still
  **sums** its members — only expansions pick one.

  ## Changes by file

  ### Text expansion scoring — in-iterator path
  - **`src/indexes/text/term.cc` / `term.h`** — `TermIterator` gains a `per_term_dt`
    arg; in expansion mode it precomputes `per_term_idf_` (one IDF per matched
    term) and `GetScore()` contributes a single matched term's BM25 instead of the
    summed-F exact/stem behavior.
  - **`src/indexes/text.cc`** — `Prefix`/`Suffix`/`Fuzzy` `BuildTextIterator` collect
    each matched term's document frequency (`GetKeyCount()`) into `per_term_dt`.
  - **`src/indexes/text/fuzzy.h`** — `FuzzySearch::Search` now returns an `Expansion`
    struct (`key_iterators` + `per_term_dt` + `postings`).
  - **`src/query/predicate.cc`** — `FuzzyPredicate::Evaluate` adapted to the new
    `Expansion` API.

  ### Text expansion scoring — extra-step path (combined queries)
  - **`src/query/search.cc`** — `ResolvedLeaf` gains `expansion_terms` (posting list
    + precomputed IDF per matched term). `ResolveLeaves` dynamic_casts
    Prefix/Suffix/Fuzzy predicates and resolves their expansion terms; `ScoreNode`
    contributes the first expansion term whose posting contains the key. Fixes
    combined queries like `cat* @rank:[0 100]` that previously scored expansions 0.

  ### Tag prefix scoring
  - **`src/indexes/tag.h` / `tag.cc`** — new `Tag::GetPrefixMatchedValues(prefix)`:
    walks the rax subtree (mirroring `Tag::Search`'s prefix branch) and returns each
    matched value + its document count. The exact `GetTagValueDocCount` primitive
    stays exact — prefix values are no longer passed to it.
  - **`src/query/search.cc`** — `ResolvedLeaf` gains `tag_prefix_groups` (one group
    per `foo*` query value, holding each expansion value's IDF). `ResolveLeaves`
    (kTag) routes `*`-suffixed values to `GetPrefixMatchedValues`; exact values keep
    the existing summed path. `ScoreNode` (kTag) credits **one** matched value per
    prefix group, summed alongside exact values and other union members. F ≡ 1, as
    for exact tag values.

  ## Not implemented (no scoring)

  - **Tag suffix / infix** (`@cat:{*llo}`, `@cat:{*ll*}`) — tag suffix/infix
    *matching* itself is unimplemented (`tag.cc` TODO `b/357027854`; no suffix trie
    for tags), so there is nothing to score. When that matching lands, such patterns
    will fall through to the exact `GetTagValueDocCount` branch and score 0 until the
    expansion routing here is extended.
  - **Tag fuzzy** (`@cat:{%hello%}`) — not supported by Redis either (syntax error
    on a tag field), so there is nothing to implement.
  - **Text infix** (`*cat*`) — `InfixPredicate::BuildTextIterator`/`Evaluate` still
    `CHECK(false)`, so infix queries abort before scoring.

  ## Tests

  - **`testing/search_test.cc`** — in-iterator + extra-step tests for text
    prefix/suffix/fuzzy (single-match == exact term; multi-match == one term, `<`
    sum), plus 3 tag-prefix tests (`TagPrefixSingleMatchEqualsExactValue`,
    `TagPrefixMultiMatchScoresOneValueNotSum`, `TagPrefixInCombinedQueryScored`).
  - **`integration/test_scoring.py`** — `INDEX_PSF` (text) and `INDEX_TPX` (tag) with
    `WITHSUFFIXTRIE` on `idxMix`. Group 15 covers text prefix/suffix/fuzzy; Group 16
    covers tag prefix (single-match == exact value; multi-match == one value and `<`
    the union `{redis|redcap}`; combined with numeric still scored). Tests assert
    self-consistent relationships against the suite's own exact scores rather than
    pinning Redis numbers on multi-match docs (the representative pick can diverge).

  ## Notes

  - Document-frequency counts are clamped to `total_docs` to stay consistent with
    the transiently-out-of-sync counters (same pattern as `ResolveLeaves`).
  - A pure tag query sets `kContainsTag`, so tag prefix scoring flows through the
    extra-step `ScoreTextQuery` → `ScoreNode` path (verified), same as combined
    queries.


src/query/search.cc
```diff
@@ -851,14 +945,33 @@ std::optional<float> ScoreNode(const Predicate *predicate,
...
+      // contains this key -- the representative is unspecified per the oracle,
+      // so this may differ from the in-iterator heap-order pick on multi-match
+      // docs; both honor the one-term invariant. doc_len is co-located in the
+      // matched posting entry.
+      if (!leaf.expansion_terms.empty()) {
+        for (const auto &term : leaf.expansion_terms) {
+          if (auto entry = term.postings->LookupKey(key)) {
+            return score_ctx.scorer->ScoreLeaf(
+                {term.idf, entry->tf, entry->doc_len, score_ctx.avg_doc_len,
+                 predicate->GetWeight()});
```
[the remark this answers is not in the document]
boda26: Pushed a new commit to resolve.

zackcam (commented):
Looks mostly good to me, just a question on:
>Line 874 records every prefix spelling without using seen. On a case-insensitive index, @cat:{red*|RED*} scores each matching value twice. Normalize and deduplicate prefix values before appending them to leaf.tag_prefixes, as exact tag values already do.

I think the AI is correct

Frank-Gu-81:
Hi @boda26 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

src/indexes/text.cc
```diff
@@ -236,16 +236,25 @@ std::unique_ptr<indexes::text::TextIterator> PrefixPredicate::BuildTextIterator(
...
+    auto postings = word_iter.GetPostingsTarget();
+    per_term_dt.push_back(postings->GetKeyCount());
+    key_iterators.emplace_back(postings->GetKeyIterator());
     word_iter.Next();
     ++word_count;
   }
   return std::make_unique<indexes::text::TermIterator>(
-      std::move(key_iterators), field_mask, require_positions);
+      std::move(key_iterators), field_mask, require_positions,
+      /*stem_field_mask=*/0, /*has_original=*/false, GetWeight(),
```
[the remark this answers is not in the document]
boda26: pushed a new commit to fix this

zackcam (approved):

integration/test_scoring.py
```diff
@@ -574,3 +661,98 @@ def test_scores_follow_mutations(self):
...
+        _, category = search(client, IDX_EXPANSION, "category")
+        _, catalog = search(client, IDX_EXPANSION, "catalog")
+        got = prefix["exp:multi"]
+        one, two = category["exp:multi"], catalog["exp:multi"]
+        assert got < one + two - SCORE_ABS_TOL
+        assert (got == pytest.approx(one, abs=SCORE_ABS_TOL)
+                or got == pytest.approx(two, abs=SCORE_ABS_TOL)), (
+            f"prefix={got} category={one} catalog={two}")
+
+        # A text+numeric/tag query takes the extra-step path. Each pattern
```
BCathcart, line 698: Does adding a numeric term affect score at all? You may want to add a case to validate that the score is the same in both the extra-step path and iterator scoring path

src/query/search.cc
```diff
@@ -843,27 +974,53 @@ std::optional<float> ScoreNode(const Predicate *predicate,
...
+      // representative is unspecified per the oracle, so this may differ from
+      // the in-iterator heap-order pick on multi-match docs; both honor the
+      // one-term invariant. doc_len is co-located in the matched posting entry.
+      if (!leaf.expansion_terms.empty()) {
+        for (const auto &term : leaf.expansion_terms) {
+          if (auto entry = term.postings->LookupKey(key, leaf.field_mask)) {
+            return score_ctx.scorer->ScoreLeaf(
+                {term.idf, entry->tf, entry->doc_len, score_ctx.avg_doc_len,
+                 predicate->GetWeight()});
+          }
```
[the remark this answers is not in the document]
boda26: This is not true. field_mask is passed in LookupKey, which returns nullopt unless some position for that key carries a field in the mask.

integration/test_scoring.py
boda26: I think there is already a case in group 10, the query "hello" (in-iterator) and "hello | @rank:[5 5]" (extra-step) yields same scores.
```
_, hello = search(client, IDX_MAIN, "hello")

# ORing a numeric branch admits a text-less doc and changes no score.
keys, mixed = search(client, IDX_MAIN, "hello | @rank:[5 5]")
assert keys == ["doc:3", "doc:1", "doc:2", "doc:5"]
assert mixed == pytest.approx({**hello, "doc:5": 0.0}, abs=SCORE_ABS_TOL)
```

src/indexes/text/posting.cc
```diff
@@ -87,12 +87,38 @@ size_t Postings::GetTotalTermFrequency() const {
...
+                            uint64_t field_mask) {
+  PositionIterator iter(flat_map);
+  while (iter.IsValid()) {
+    if ((iter.GetFieldMask() & field_mask) != 0) {
+      return true;
+    }
+    iter.NextPosition();
+  }
+  return false;
+}
```
BCathcart, line 103: Position maps can get pretty big for common terms across a large corpus. A space vs time tradeoff we can make is to maintain an overall field mask per position map on the ingestion side to skip this iteration. Let's keep this in mind when we analyze the perf results on large corpuses.

src/indexes/text/term.cc
```diff
@@ -66,6 +78,17 @@ float TermIterator::GetScore() const {
     return 1.0f;
   }
 
+  // Expansion mode: contribute a SINGLE matched term's BM25 (own IDF + own F),
+  // never the sum. Which term is unspecified, so take the merge's front.
```
BCathcart, line 82: This is something we need to call out clearly in the docs

src/query/search.cc
```diff
@@ -915,6 +1075,19 @@ std::optional<float> ScoreNode(const Predicate *predicate,
                                             score_ctx.avg_doc_len,
                                             predicate->GetWeight()});
       }
+      // Each prefix contributes ONE matched value (the doc's first), never the
+      // sum; a union still sums. Clamp as ResolveLeaves does -- dt and
+      // total_docs come from independently-locked counters.
```
BCathcart, line 1080: Important to document this for tag as well

src/indexes/tag.cc
```diff
@@ -517,6 +518,32 @@ size_t Tag::GetTagValueDocCount(absl::string_view value) const {
...
+
+  for (const auto &part :
+       absl::StrSplit(it->second.raw_tag_string->Str(), separator_)) {
+    const absl::string_view tag = absl::StripAsciiWhitespace(part);
+    // Empty tags are never indexed (ParseRecordTags drops them), and a bare `*`
+    // query gives an empty prefix that would otherwise match one.
+    if (tag.empty()) continue;
+    if (case_sensitive_ ? absl::StartsWith(tag, prefix)
+                        : absl::StartsWithIgnoreCase(tag, prefix)) {
+      return GetTagValueDocCount(tag);
```
BCathcart, line 541: Here we're scoring a key in the extra step approach. We compare the key's tags, then traverse the tree to find the `dt` if it's a match. As we know, repeated Rax traversals aren't ideal. What do you think about pre-fetching `dt` like is done for text?

src/indexes/text.cc
```diff
@@ -265,16 +275,24 @@ std::unique_ptr<indexes::text::TextIterator> SuffixPredicate::BuildTextIterator(
...
+    key_iterators.emplace_back(postings->GetKeyIterator());
     word_iter.Next();
     ++word_count;
   }
   return std::make_unique<indexes::text::TermIterator>(
-      std::move(key_iterators), field_mask, require_positions);
+      std::move(key_iterators), field_mask, require_positions,
+      /*stem_field_mask=*/0, /*has_original=*/false,
+      GetWeight() * or_weight_multiplier,
+      /*num_doc_contain_term=*/0, GetTextIndexSchema().get(), GetScorer(),
```
BCathcart, line 294: nit: You could pass `IndexScoringStats` instead of the whole text_index_schema_. If you really wanted to make it nice IMO, you could create a `TermScoringInputs` struct and group all the scoring inputs together.

src/query/search.cc
```diff
@@ -664,26 +675,72 @@ struct ResolvedLeaf {
   // per-candidate ScoreNode walk needs only a cheap map lookup.
   const indexes::Tag *tag_index = nullptr;
   absl::InlinedVector<std::pair<std::string, float>, 4> tag_values;
+
+  // Tag prefix query values (`foo*`), as views into the TagPredicate's tag
+  // strings. Nothing is precomputed: the representative value is per-document,
+  // so Tag::GetPrefixMatchDocCount resolves its dt per candidate.
+  absl::InlinedVector<absl::string_view, 2> tag_prefixes;
```
BCathcart, line 682: A small space optimization that would also separate data clearly is to use a union to separate Tag, Term, and Expansion.

src/query/search.cc
```diff
@@ -698,17 +755,77 @@ void ResolveLeaves(const Predicate *predicate, uint32_t total_docs,
...
+        auto it = tree.GetWordIterator(pattern);
+        for (uint32_t n = 0; !it.Done() && n < max_words; ++n, it.Next()) {
+          AddExpansionTerm(it.GetPostingsTarget(), total_docs, scorer,
+                           expansion_leaf);
+        }
+      };
+      bool is_expansion = true;
+      if (auto *p = dynamic_cast<const PrefixPredicate *>(predicate)) {
+        expansion_leaf.field_mask =
+            ScoringFieldMask(p->GetFieldMask(), p->GetTextIndexSchema().get());
```
BCathcart, line 780: nit: Call `GetNumTextFields()` once and pass it to each `ScoringFieldMask()` call

src/query/search.cc
```diff
@@ -655,6 +656,16 @@ struct ResolvedLeaf {
   // per candidate document.
   float term_weight = 0.0f;
```
BCathcart, line 657: Let's call this `idf` like we do for ExpansionTerm. "weight" is overloaded

src/query/search.cc
```diff
@@ -723,9 +840,15 @@ void ResolveLeaves(const Predicate *predicate, uint32_t total_docs,
         }
       };
```
BCathcart, line 841: nit: you could probably unify this code with creating an "ExpansionTerm". They are fetching the same data

src/query/search.cc
```diff
@@ -843,27 +974,53 @@ std::optional<float> ScoreNode(const Predicate *predicate,
...
+      // the doc carries only elsewhere would supply the IDF and F. The
+      // representative is unspecified per the oracle, so this may differ from
+      // the in-iterator heap-order pick on multi-match docs; both honor the
+      // one-term invariant. doc_len is co-located in the matched posting entry.
+      if (!leaf.expansion_terms.empty()) {
+        for (const auto &term : leaf.expansion_terms) {
+          if (auto entry = term.postings->LookupKey(key, leaf.field_mask)) {
+            return score_ctx.scorer->ScoreLeaf(
+                {term.idf, entry->tf, entry->doc_len, score_ctx.avg_doc_len,
+                 predicate->GetWeight()});
```
BCathcart, line 999: We pre-fetch the postings and IDF for every term so the benefit really depends on whether some keys only have the later terms. Just something to keep in mind if we start really digging into optimizing

BCathcart (approved):
Looks good overall. I don't see any correctness issue. Left several small comments that can be addressed here or in a follow-up. I also think it's worth future discussion about the strategy of scoring a single arbitrary term during term expansion.

BCathcart:
Leaving the last minor review items as follow-ups.

>I also think it's worth future discussion about the strategy of scoring a single arbitrary term during term expansion

Intuitively for me, a prefix matching multiple terms should have every matching term contribute to the score like how a text AND or a term with stem roots is handled. I'd like us to look into how it's typically handled in other search engines besides Redisearch.

src/query/search.cc
```diff
@@ -843,27 +974,53 @@ std::optional<float> ScoreNode(const Predicate *predicate,
...
+      // Expansion leaf: contribute exactly ONE matched term's BM25 (its own IDF
+      // + own F), never the sum. Pick the first expansion term whose posting
+      // contains this key in a requested field -- without the field gate a term
+      // the doc carries only elsewhere would supply the IDF and F. The
+      // representative is unspecified per the oracle, so this may differ from
+      // the in-iterator heap-order pick on multi-match docs; both honor the
+      // one-term invariant. doc_len is co-located in the matched posting entry.
+      if (!leaf.expansion_terms.empty()) {
+        for (const auto &term : leaf.expansion_terms) {
+          if (auto entry = term.postings->LookupKey(key, leaf.field_mask)) {
```
BCathcart, line 996: We need to be a bit careful here. In a large corpus a prefix could have many matching terms, but each key might only contain a small subset of the matches. Doing up to 200 btree map lookups for a key is going to be more expensive than traversing the per-key rax tree subtree and simply finding the existence of a target.

src/indexes/text/posting.cc
boda26: Adding a comment to record this for future

src/indexes/text.cc
boda26: Fixed in stemming pr

src/query/search.cc
boda26: Splitted ResolvedLeaf to TermLeaf, ExpansionLeaf and TagLeaf

PR #1352, opened 2026-09-02
fix breaking changes introduced in #1008

Frank-Gu-81:
Used COMPATIBILITY_FIX macro to preserve both paths.


Frank-Gu-81:
This would be a search 1.3 launch blocker PR. This needs to be merged to address the breaking change introduced in #1008.

allenss-amazon (approved):

PR #1354, opened 2026-09-02
Scoring: BM25STD scoring for stemming queries

boda26:
## Overview

  Previously stemmed query terms folded all stem variants into a single BM25 leaf, and their combined document frequency (`dt`) was double-counted when a doc was indexed under several variants
  (marked `TODO: scoring for stemming` in `search.cc`). This PR replaces that with the industry-standard split: a stemmed query term expands to a **union of up to three independent BM25 leaves 
  that are SUMMED**, each carrying its own IDF and its own term frequency `F`:

  1. **Exact surface term** — the query word as typed.
  2. **Stem root literal** — e.g. `run` for query `running`, scored as its own leaf when present and distinct from the query word.
  3. **Stem inflection group** — all variants sharing the stem root; `F` sums per-doc frequencies, but `dt` is the **distinct** doc count (a doc holding several inflections counts once).

  Because the exact word is also a parent of its own stem, it contributes to both leaf 1 and leaf 3 — a deliberate **exact-match boost** so `running` (TF 1) outranks `runs runs runs` (TF 3).
  The split is applied identically on both scoring paths (in-iterator and extra-step).

  ## Changes by file

  ### Distinct-doc counting (shared helper)
  - **`src/indexes/text/posting.cc` / `posting.h`** — new `CountDistinctKeys(span<KeyIterator>)`: k-way merge over key-sorted iterators that counts a key present in multiple iterators once.
  Used by **both** scoring paths to compute the stem-inflection leaf's `dt` consistently.

  ### In-iterator path (pure-text queries)
  - **`src/indexes/text.cc`** — `TermPredicate::BuildTextIterator` now computes three separate `dt` values (exact / root / stem-inflection), the root via `TryAddWordKeyIterator` (extended with
  an `out_doc_count`), and the inflection group's distinct `dt` via `CountDistinctKeys`.
  - **`src/indexes/text/term.cc` / `term.h`** — `TermIterator` takes `stem_num_doc_contain_term`, `root_num_doc_contain_term`, and `has_root`; precomputes `idf_stem_` and `idf_root_` alongside
  `idf_`. `GetScore()` now partitions positioned iterators by index into exact / root / stem TFs and sums up to three `ScoreLeaf` calls (each only when its TF > 0) instead of a single summed-F
  leaf.

  ### Extra-step path (combined text + numeric/tag/negate queries)
  - **`src/query/search.cc`** — `ResolvedLeaf` is refactored from flat `postings`/`num_doc_contain_term`/`term_weight` into a vector of `TermGroup`s (1 for a plain term, up to 3 for a stemmed
  term), each with its own postings and precomputed IDF. `ResolveLeaves` builds the exact / root-literal / inflection groups (distinct `dt` via `CountDistinctKeys`); `ScoreNode` sums each
  group's `ScoreLeaf`, fetching `doc_len` once, and returns `nullopt` only when no group matches. Removes the old double-counting TODO.

  ### Tests
  - **`integration/test_scoring.py`** (+126) — new stemming index `INDEX_STEM` (7 `run`-family docs) with Redis-verified `STEM_RUNNING_SCORES`, plus `INDEX_STEM_MIX` (same docs + numeric `rank`
  + tag `cat`) to exercise the extra-step path. Group 15 covers pure-text stemming; Group 16 covers combined AND / OR / tag queries and asserts both paths produce identical per-leaf scores.
  - **`testing/text_test.cc`** (+75) — new `StemScoringTest` fixture driving the in-iterator path: pins the three-leaf oracle values (`d1 running` 1.450833, `d3 run` 0.980829, `d2 runs`
  0.470004) and covers the `has_original_ == false` path (stem-only match when the exact word is absent).
  - **`testing/search_test.cc`** (+65) — `BuildTextTagSchema` gains a `no_stem` flag; adds extra-step tests pinning the same oracle values, verifying `$weight` scales the whole expansion, and 
  that the recompute path (`SingleDocumentScorer`) matches the shard-side extra-step path on a stemmed query.

  ## Notes

  - The two paths may still be exercised independently, so tests pin the **same** oracle values on both to guard against divergence in the stem split.
  - Oracle scores are verified against the Redis 8.6 baseline / industry-standard reference; behavior is documented in `docs/redis_stemming_scoring.md`.


src/utils/string_interning.h
```diff
@@ -217,6 +217,17 @@ class BorrowedInternedStringPtr {
...
+  // Reinterpret this borrowed pointer as a non-owning InternedStringPtr
+  // reference without touching the ref count. Both types are single-pointer
+  // wrappers with identical layout (asserted below), so the reference aliases
+  // the same InternedString. Valid only for read-only use (map lookups,
+  // hashing) while the underlying InternedString is alive; the callee must not
+  // take ownership (copy/move) of the reference.
+  // Delete this if extra-step scoring not used
+  const InternedStringPtr &AsInternedRef() const {
+    return *reinterpret_cast<const InternedStringPtr *>(this);
+  }
```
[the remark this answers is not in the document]
boda26: This is accidentally brought back by the merge from main. I will delete that as we have transparent comparators now.

boda26:
There is a field mask bug on extra-step path in current code. The fix is in prefix scoring PR https://github.com/valkey-io/valkey-search/pull/1350, so merging that first might help.

boda26:
```
  scenario             baseline rps(p50)    candidate rps(p50)      rps Δ%    p50 Δ%
  fanout 1 unstemmed   45482 (11.63)        44588 (11.96)             -1.3%      +3.1%
  fanout 1 stemmed     37500 (14.66)        35034 (15.68)             -5.6%      +6.8%
  fanout 2             40846 (13.22)        38290 (14.34)             -6.7%      +8.2%
  fanout 4             41164 (13.15)        36057 (15.16)            -12.1%     +14.9%
  fanout 6             39277 (13.58)        32301 (17.04)            -17.6%     +25.2%
  fanout 8             37652 (14.31)        27656 (20.08)            -26.4%     +40.5%
  fanout 8 VERBATIM    67773 (7.21)         66800 (7.26)              -1.1%      +1.2%
  fanout 8 nostem idx  66973 (7.12)         66680 (7.15)              -0.1%      +0.9%
  fanout 8 explicit OR 23100 (24.62)        22559 (25.15)             -2.3%      +2.3%
  stem base form       38443 (14.23)        27573 (20.01)            -27.6%     +40.7%
  stem long form       38303 (14.14)        27555 (20.27)            -27.5%     +43.2%
  2 stems (AND)        18401 (31.58)        12322 (47.74)            -33.9%     +51.0%
  2 stems (OR)         13094 (44.96)        9728 (59.78)             -24.6%     +32.9%
  stem+num+tag         27978 (20.32)        21409 (26.42)            -22.7%     +30.4%
  stem miss            77303 (6.00)         77277 (6.10)              -0.4%      +1.6%
  ```
  Perf test results against the current main (with base scoring). Fanout-x means x words in the index are stemmed to the same root, so searching one term would resulting in going through all of them and calculate the stem root score. As fanout number increases, the performance difference also increases.

**Comment by boda26** [comment]
Benchmark shows CountDistinctKeys was the major perf regression. I pushed a new commit, incrementing distinct docs at ingestion, so search can get them in O(1) time. The new benchmark results confirmed that perf regression is gone, and is aligned with the nostem version.
```
scenario             baseline rps(p50)    candidate rps(p50)      rps Δ%    p50 Δ%
  fanout 1 unstemmed   38523 (13.74)        38754 (13.54)             +2.5%      -1.0%
  fanout 1 stemmed     33013 (16.58)        31938 (16.98)             -2.2%      +2.7%
  fanout 2             34774 (15.35)        33797 (15.91)             -2.7%      +3.2%
  fanout 4             35049 (15.26)        35186 (15.62)             +0.3%      +2.5%
  fanout 6             35147 (15.65)        33960 (15.83)             -1.9%      +1.3%
  fanout 8             33204 (16.48)        32610 (16.40)             +0.3%      -0.7%
  fanout 8 VERBATIM    59565 (8.02)         59952 (8.02)              +1.9%      -0.4%
  fanout 8 nostem idx  61439 (8.06)         60476 (8.03)              -0.8%      -0.4%
  fanout 8 explicit OR 19483 (28.94)        20271 (28.72)             +3.5%      -0.8%
  fanout 8 OR VERBATIM 19042 (28.83)        19411 (28.64)             +2.3%      -0.4%
  stem base form       29506 (18.29)        29098 (18.96)             -0.5%      +3.7%
  stem long form       33170 (16.30)        32787 (16.45)             -2.1%      +0.1%
  2 stems (AND)        15345 (37.50)        15430 (37.12)             +0.6%      -1.6%
  2 stems (OR)         10382 (55.42)        10604 (55.87)             +3.2%      -0.8%
  stem+num+tag         22995 (25.53)        22232 (25.79)             -3.4%      +1.5%
  stem miss            67595 (6.95)         66421 (7.19)              -1.9%      +2.9%
  ```

Frank-Gu-81:
Hi @boda26 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @chinguyen21 — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

src/query/search.cc
```diff
@@ -859,29 +888,34 @@ std::optional<float> ScoreNode(const Predicate *predicate,
...
+      bool matched = false;
       uint32_t doc_len = 0;
-      for (const auto &postings : leaf.postings) {
-        if (auto entry = postings->LookupKey(key)) {
-          tf += entry->tf;
-          doc_len = entry->doc_len;
+      for (const TermGroup &group : leaf.groups) {
+        uint32_t tf = 0;
+        for (const auto &postings : group.postings) {
+          if (auto entry = postings->LookupKey(key)) {
```
zackcam, line 907: A comment from ai:
>the extra-step scoring path is field-blind. All TEXT fields share one posting tree, so a key's presence in a term's posting list does not mean the term occurred in a requested field. LookupKey(key) here (and TermGroup at line 654) carries no field mask, so @body:running @rating:[0 100] scores a doc that holds running only in title, and a stemmed @title:run matches an inflection in a NOSTEM field. The in-iterator path gates correctly via InsertValidKeyIterator → ContainsFields(field_mask) in term.cc, using the query mask for the original word and stem_field_mask for variants, so the two paths disagree on the same query.


I think this actually is fixed in or can have the same incorporated fix as the suffix, prefix and fuzzy pr you also have up?

boda26: I already have a fix in the prefix scoring PR. It would be easier to merge it first, and fix the stemming part should be simple

BCathcart:
> Benchmark shows CountDistinctKeys was the major perf regression. I pushed a new commit, incrementing distinct docs at ingestion, so search can get them in O(1) time. The new benchmark results confirmed that perf regression is gone, and is aligned with the nostem version.

Awesome work, I was anticipating an optimization phase around stemming after the initial results but looks like you've already completely eliminated it!

src/indexes/tag.h
```diff
@@ -105,6 +105,16 @@ class Tag : public IndexBase {
...
 
+  // Returns whether `key` carries tag value `value`. `value` is normalized
+  // (lowercased unless case-sensitive) before lookup, so callers pass the raw
+  // query value. Unlike GetValue, this avoids parsing/allocating the document's
+  // tag set per call: it looks the value up in the rax and tests the key
+  // against that value's posting bag. Lock-free like GetValue, relying on the
+  // read-side invariant that the index is not mutated while the time-sliced
+  // mutex is held in read mode.
+  bool ContainsKey(absl::string_view value, const InternedStringPtr &key) const
+      ABSL_NO_THREAD_SAFETY_ANALYSIS;
```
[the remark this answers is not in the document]
chinguyen21: Can you fix this? Don't see it use anywhere?

src/indexes/text/text_index.cc
```diff
@@ -426,7 +464,7 @@ std::string TextIndexSchema::GetAllStemVariants(
     absl::string_view search_term,
     absl::InlinedVector<absl::string_view, kStemVariantsInlineCapacity>
         &words_to_search,
-    uint64_t stem_enabled_mask, bool lock_needed) {
+    uint64_t stem_enabled_mask, bool lock_needed, uint32_t *out_distinct_docs) {
   // Stem the search term
   std::string stemmed(search_term);
   lexer_.StemWordInPlace(stemmed, lexer_.GetStemmer());
```
chinguyen21, line 470: Should we need `min_stem_size_` here similar to what you added in the ingestion above?

chinguyen21: Also add a test would be good

testing/text_index_schema_test.cc
```diff
@@ -31,8 +34,70 @@ class TextIndexSchemaTest : public vmsdk::ValkeyTest {
...
+  CommitStemmed(*schema, StringInternStore::Intern("doc3"), "swimming");
+
+  // doc1 holds two inflections of "run" but is one document; summing the two
+  // posting lists instead would give 3.
+  EXPECT_EQ(StemDistinctDocs(*schema, "running"), 2);
+  EXPECT_EQ(StemDistinctDocs(*schema, "swimming"), 1);
+}
+
+// Deletion must undo exactly the one increment the document made.
+TEST_F(TextIndexSchemaTest, StemDistinctDocsDecrementsOnDelete) {
```
chinguyen21, line 80: This is good test for the single decrement for a two-inflection doc. Could you add: two docs under one root, overwrite one with a different inflection (count stays 2), then with an unrelated word (drops to 1)?

src/indexes/tag.h
```diff
@@ -105,6 +105,16 @@ class Tag : public IndexBase {
...
       ABSL_NO_THREAD_SAFETY_ANALYSIS;
 
+  // Returns whether `key` carries tag value `value`. `value` is normalized
+  // (lowercased unless case-sensitive) before lookup, so callers pass the raw
+  // query value. Unlike GetValue, this avoids parsing/allocating the document's
+  // tag set per call: it looks the value up in the rax and tests the key
+  // against that value's posting bag. Lock-free like GetValue, relying on the
+  // read-side invariant that the index is not mutated while the time-sliced
+  // mutex is held in read mode.
+  bool ContainsKey(absl::string_view value, const InternedStringPtr &key) const
```
chinguyen21, line 115: Should delete this

chinguyen21 (commented):
LGTM, just need some minimal fixes to get it merged

src/query/search.cc
boda26: fixed

src/indexes/text/text_index.cc
boda26: No I think it's not needed here. min_stem_size applied at ingestions is enough to limit the size, and when GetAllStemVariants are called later, it's already with in the min size.

testing/text_index_schema_test.cc
boda26: Added two steps in this test

chinguyen21 (approved):
LGTM, the remaining is to check DCO rules

src/indexes/text/term.cc
```diff
@@ -21,7 +21,9 @@ TermIterator::TermIterator(
     const FieldMaskPredicate query_field_mask, const bool require_positions,
     const FieldMaskPredicate stem_field_mask, bool has_original,
     float leaf_weight, uint32_t num_doc_contain_term,
-    const TextIndexSchema* text_index_schema, const scoring::Scorer* scorer,
+    uint32_t stem_num_doc_contain_term, uint32_t root_num_doc_contain_term,
+    bool has_root, const TextIndexSchema* text_index_schema,
+    const scoring::Scorer* scorer,
```
BCathcart, line 26: nit: I'd suggest grouping all the items specifically used for scoring into a struct defined in term.h to pass in here. I'd prefer a bit clearer separation now that the param list is getting so large.

In general I'd recommend taking a look at the nits on the term expansion PR and apply any you think are reasonable to this PR.

src/indexes/text/text_index.cc
```diff
@@ -364,6 +369,20 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
         existing = text_index_->GetPrefix().FindPostingsTarget(word_str);
       }
```
BCathcart, line 370: This is old code but I don't think there is a reason to look up the postings in the main tree. It's the same underlying object as the per-key tree has which we're walking. No need for another traversal.

src/indexes/text/text_index.cc
```diff
@@ -364,6 +369,20 @@ void TextIndexSchema::DeleteKeyData(const InternedStringPtr &key) {
...
 
+      // Only stem-enabled fields were counted, and the field mask is only
+      // readable before the key leaves the postings below.
+      if ((stem_text_field_mask_ != 0u) && existing) {
```
BCathcart, line 374: Maybe make existing an assert since it should always be existing

BCathcart (approved):

Frank-Gu-81:
## Note for maintainers: Do not merge this PR yet. #1400 should be merged before this.

src/indexes/text/term.cc
boda26: Refactored to a TermScoringParams struct.

PR #1355, opened 2026-09-03
Run integration tests in parallel (reduces runtime from ~12 minutes to ~95 seconds)

yairgott:
### Summary

This PR enables parallel test execution across both integration test suites:
1. **Pytest Integration Test Suite** (`integration/`)
2. **Legacy Abseil Integration Test Suite** (`testing/integration/`)

By default, `--parallel` / `-j` now automatically detects and utilizes all available CPU cores (`auto`), **reducing the overall integration test runtime from ~12 minutes down to ~95 seconds** (e.g. on 48 cores):
- **Legacy Abseil suite**: reduced from **~113s to 21.6s**
- **Pytest integration suite**: reduced from **~10+ minutes to ~75s**
- **Full test suite total**: reduced from **~12 minutes to ~95 seconds**

---

#### 1. Parallel Execution Orchestration
- **Top-Level `build.sh`**: Added `--parallel[=N]` / `-j[=N]` CLI options (defaulting to `auto`). Forwards the parallelism configuration to both test runners.
- **Legacy Abseil Suite (`testing/integration/run.sh`)**: Added `pytest-xdist>=3.5.0` integration with `--dist=load`. Automatically detects and provisions `pytest-xdist` in the test venv.
- **Pytest Suite (`integration/run.sh`)**: Runs pytest with `-n <workers> --dist=loadscope` by default.

#### 2. Collision-Free Port & Directory Partitioning
- **Band Partitioning (`integration/conftest.py`)**: Introduced `SafePortTracker` using non-overlapping port bands across workers:
  - Base client ports: `[10000, 19999]` (stride of 100 ports per worker, up to 100 workers)
  - Cluster bus ports: `[20000, 29999]` (`+10000` offset)
  - Coordinator gRPC / TLS: `[30294, 40294]` (`+20294` / `+20295` offset)
  - User-scoped lock directories (`/tmp/valkey_port_locks_<uid>`) to prevent permission conflicts in shared environments.
- **Legacy Port Tracker (`testing/integration/utils.py`)**: Partitioned cluster ports per worker (`7000 + (worker_idx % 100) * 50`) and isolated worker directories for `TEST_TMPDIR` and `TEST_UNDECLARED_OUTPUTS_DIR`.

#### 3. Robust Server Lifecycle & Leak Prevention
- **Pytest Teardown Support**: Added `teardown_method` to `ValkeySearchTestCaseCommon` in `integration/valkey_search_test_case.py` (since framework base classes do not inherit from `unittest.TestCase`, Pytest ignores `tearDown()`).
- **Guaranteed Cleanup on Setup Failure**: Wrapped `setup_test` fixtures in `try ... yield ... finally:` in both `ValkeySearchTestCaseBase` and `ValkeySearchClusterTestCase`, ensuring nodes started prior to any setup timeout or error are always cleanly shut down.
- **Best-Effort Cleanup with Error Aggregation**: In `ReplicationGroup.cleanup` and `teardown_method`, every server is given an opportunity to exit before any recorded teardown errors are re-raised, preventing both process leaks and silent failure masking.
- **Safety-Net Signal Traps**: Added `EXIT INT TERM` traps in both test runners invoking `pkill -9 valkey-server` upon script completion, error, or manual cancellation (`Ctrl+C`).

#### 4. Stability & Timing Under High CPU Contention
- In `integration/test_postfilter.py`, converted immediate assertions on `ft._debug PAUSEPOINT test` to `waiters.wait_for_true` to prevent thread scheduling races under 48-worker loads.

---

### Verification

- **Legacy Abseil Tests**: 47/47 passed concurrently across 48 workers in **21.63s** (down from ~113s sequentially).
- **Pytest Tests**: 393/393 passed across 48 workers in **~75s**.
- **Process Cleanup**: Verified with `ps -ef | grep valkey-server` on both host and inside the devcontainer: **zero** leaked or zombie processes.
- **Formatting**: Checked and passed via `.devcontainer/run_in_docker.sh ./ci/clang-format-changes.sh`.

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -1,0 +1,180 @@
+"""
+Pytest configuration and collision-free port tracking for Valkey Search integration tests.
+Ensures concurrent workers allocated via pytest-xdist never collide on client ports,
+cluster bus ports, or search coordinator ports.
+"""
+
+import fcntl
+import logging
+import os
+import socket
+import tempfile
+from pathlib import Path
+import pytest
+
+# If running under pytest-xdist, isolate the LOGS_DIR per worker process
+worker_id = os.environ.get("PYTEST_XDIST_WORKER")
+if worker_id and "LOGS_DIR" in os.environ:
+    base_logs_dir = os.environ["LOGS_DIR"]
+    if not base_logs_dir.endswith(f"/{worker_id}"):
+        worker_logs_dir = os.path.join(base_logs_dir, worker_id)
+        os.environ["LOGS_DIR"] = worker_logs_dir
+        os.makedirs(worker_logs_dir, exist_ok=True)
+
+
+class SafePortTracker(object):
+    """Provides collision-free ports for valkey-server across concurrent pytest workers.
+
+    Port ranges are partitioned by xdist worker ID into disjoint blocks.
+    Within each worker's block, ports are allocated monotonically with wraparound
+    to prevent reusing sockets in TIME_WAIT.
+    Persistent file locks (flock) without unlinking protect against cross-process
+    and external collisions without inode recreation races.
+    """
+
+    CLUSTER_BUS_PORT_OFFSET = 10000
+    SEARCH_COORDINATOR_PORT_OFFSET = 20294
+
+    # Worker base ports: [10000, 19999] (stride = 100 ports per worker, up to 100 workers)
+    # Cluster bus ports: [20000, 29999]
+    # Coordinator ports: [30294, 40294]
+    # None of the three bands overlap with each other.
+    BASE_PORT = 10000
+    PORTS_PER_WORKER = 100
+    MAX_WORKERS = 100
+
+    LOCKS_DIR = os.path.join(
+        tempfile.gettempdir(),
+        f"valkey_port_locks_{os.getuid() if hasattr(os, 'getuid') else 0}",
+    )
+
+    # Worker-local monotonic offset for round-robin port cycling
+    _worker_port_offset = 0
+
+    def __init__(self, node_id=None):
+        self.node_id = node_id
+        os.makedirs(self.LOCKS_DIR, exist_ok=True)
+        self.locked_fds = {}  # port -> fd
+
+        w_id = os.environ.get("PYTEST_XDIST_WORKER", "master")
+        if w_id.startswith("gw"):
+            try:
+                self.worker_idx = int(w_id[2:])
+            except ValueError:
+                self.worker_idx = 0
+        else:
+            self.worker_idx = 0
+
+        if self.worker_idx >= self.MAX_WORKERS:
+            raise RuntimeError(
+                f"Worker index {self.worker_idx} exceeds maximum supported workers ({self.MAX_WORKERS})"
+            )
+
+        self.worker_start_port = (
+            self.BASE_PORT + self.worker_idx * self.PORTS_PER_WORKER
+        )
+        self.worker_end_port = self.worker_start_port + self.PORTS_PER_WORKER
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        for port, fd in list(self.locked_fds.items()):
+            self._unlock_fd(fd)
+        self.locked_fds.clear()
+
+    @classmethod
+    def _unlock_fd(cls, fd):
+        try:
+            fcntl.flock(fd, fcntl.LOCK_UN)
+        except OSError:
+            pass
+        try:
+            os.close(fd)
+        except OSError:
+            pass
+
+    def _try_lock_port(self, port):
+        lock_path = os.path.join(self.LOCKS_DIR, f"port_{port}.lock")
+        fd = None
+        try:
+            fd = os.open(lock_path, os.O_CREAT | os.O_RDWR, 0o666)
+            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except (OSError, BlockingIOError):
+            if fd is not None:
+                try:
+                    os.close(fd)
+                except OSError:
+                    pass
+            return None
+
+        # Verify the port can actually be bound
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            try:
+                sock.bind(("0.0.0.0", port))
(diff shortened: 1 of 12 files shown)


integration/run.sh
```diff
@@ -2,12 +2,17 @@
...
 ROOT_DIR=$(readlink -f $(readlink -f $(dirname $0))/..)
+export ROOT_DIR
 
 . ${ROOT_DIR}/scripts/common.rc
-setup_valkey_server
 
 LOG_NOTICE "Root directory is: ${ROOT_DIR}"
 
+function cleanup() {
+  pkill -9 valkey-server 2>/dev/null || true
```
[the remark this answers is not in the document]
yairgott: pkill -9 in the entry/exit trap is an intentional catch-all safety net to sweep up orphaned or leaked daemon processes left behind if a worker crashes or receives SIGINT/SIGTERM.

This pattern was pre-existing across the codebase and changing process tree management is outside the scope of parallel test enablement.

integration/conftest.py
```diff
@@ -0,0 +1,182 @@
...
+    Port ranges are partitioned by xdist worker ID into disjoint blocks.
+    Within each worker's block, ports are allocated monotonically with wraparound
+    to prevent reusing sockets in TIME_WAIT.
+    Persistent file locks (flock) without unlinking protect against cross-process
+    and external collisions without inode recreation races.
+    """
+
+    CLUSTER_BUS_PORT_OFFSET = 10000
+    SEARCH_COORDINATOR_PORT_OFFSET = 20294
+    SEARCH_COORDINATOR_TLS_PORT_OFFSET = 20295
```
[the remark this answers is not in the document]
yairgott: The premise that candidate ports lock `candidate_port + SEARCH_COORDINATOR_TLS_PORT_OFFSET` on each allocation is incorrect.

In `integration/conftest.py`, locking the TLS offset is explicitly guarded by:
```python
if candidate_port == 6378:
    ports_to_lock.append(
        candidate_port + self.SEARCH_COORDINATOR_TLS_PORT_OFFSET
    )

File: scripts/common.rc [hunk 3]
```diff
@@ -141,14 +141,16 @@ function install_test_framework() {
 
   pushd ${test_framework_path} >/dev/null
   valkey_module_installed=$(pip_check_if_module_installed valkey)
-  if [[ "${valkey_module_installed}" == "no" ]]; then
+  xdist_module_installed=$(pip_check_if_module_installed pytest-xdist)
+  if [[ "${valkey_module_installed}" == "no" ]] || [[ "${xdist_module_installed}" == "no" ]]; then
```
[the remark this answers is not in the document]
yairgott: `scripts/common.rc` is a shared environment setup helper rather than a package/version manager.
Python integration test dependencies and exact pinned versions (such as `numpy==2.2.2`) are declared in their respective `requirements.txt` files and installed into dedicated, isolated virtual environments managed by each test runner (`.build-<config>/venv` for `testing/integration` and `.build-<config>/integration/env` for OSS integration). Re-implementing multi-package version validation in shell within `common.rc` would duplicate `pip`'s dependency resolver and add unnecessary maintenance overhead.

testing/integration/run.sh
```diff
@@ -162,7 +189,7 @@ fi
 function cleanup() {
     local exit_code=$1
     printf "Cleanup before exit..."
-    pkill valkey-server || true
+    pkill -9 valkey-server || true
     deactivate >/dev/null 2>&1 || true
     cd ${ROOT_DIR}
     printf "${GREEN}done${RESET}\n"
```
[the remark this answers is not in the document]
yairgott: As noted in the discussion on `integration/run.sh` (https://github.com/valkey-io/valkey-search/pull/1355#discussion_r3927054327), exit trap cleanup acts as an intentional catch-all safety net to sweep up orphaned daemon processes if a worker crashes or receives SIGINT/SIGTERM.
To address the concern about impacting unrelated `valkey-server` instances running on the host when executing outside the devcontainer, we have scoped process cleanup by the invocation's unique test temporary directory:
```bash
pkill -9 -f "valkey-server.*${TEST_TMPDIR}"

File: integration/run.sh [hunk 5]
```diff
@@ -16,6 +23,7 @@ Usage: run.sh [options...]
     --asan                  When passed, the integration will load the module under .build-release-asan/ | .build-debug-asan/
     --tsan                  When passed, the integration will load the module under .build-release-tsan/ | .build-debug-tsan/
     --capture               Disable pytest output capture (shows print statements in real-time).
+    --parallel[=N] | -j[=N] Run tests in parallel with N workers (default: all CPU cores).
```
zackcam, line 26: https://pytest-xdist.readthedocs.io/en/stable/distribution.html
>Use -n logical to use the number of logical CPU cores rather than physical ones. This currently requires the [psutil](https://pypi.org/project/psutil/) package to be installed; if it is not or if it fails to determine the number of logical CPUs, fall back to -n auto behavior.

Nice that it doesn't fail should we make it so we need psutil as a requirement or note this somewhere?

integration/conftest.py
```diff
@@ -0,0 +1,185 @@
...
+            SafePortTracker._worker_port_offset = (
+                SafePortTracker._worker_port_offset + 1
+            ) % self.PORTS_PER_WORKER
+
+            ports_to_lock = [
+                candidate_port,
+                candidate_port + self.CLUSTER_BUS_PORT_OFFSET,
+                candidate_port + self.SEARCH_COORDINATOR_PORT_OFFSET,
+            ]
+            if candidate_port == 6378:
```
zackcam: Can this be hit still? I see:

``` 
 # Worker base ports: [10000, 19999] (stride = 100 ports per worker, up to 100 workers)
 # Cluster bus ports: [20000, 29999]
 # Coordinator ports: [30294, 40294]
```

which doesn't include this port

yairgott: Good catch! You're completely right, it can never be hit. I've removed SEARCH_COORDINATOR_TLS_PORT_OFFSET.

integration/run.sh
yairgott: Great point.  I've updated the runner scripts so that logical resolves through num_proc() with the same 32-worker safety cap as auto. That way it consistently uses logical cores across all environments without needing psutil.

integration/conftest.py
allenss-amazon: Don't we already have a port reservation/allocator in the base pytest module? Why can't that be used?

zackcam (approved):

PR #1358, opened 2026-09-04
Score on the main thread without the time-slice mutex, and implement FT.AGGREGATE ADDSCORES

zackcam:
### Commit 1: Get background corpus information from background thread
Search() now pre-builds the recompute scorer at the end of its background-thread run, under the reader lock it already holds. The snapshot of corpus state (total docs, per-term IDF, average document length) therefore matches the state the carried neighbor scores were computed against, so a recomputed score is on the same scale as the scores it ranks against. A LockPolicy enum on the constructor distinguishes callers that already hold the reader lock from those that need it acquired.

### Commit 2: Score on the main thread without the time-slice mutex
Score() no longer takes the reader lock. Every per-key read now takes the same fine-grained lock its writer holds: text term frequency and document length go through the key's own text index under the word's bucket mutex, tag membership takes the tag index mutex, and the document score takes the mutated-records mutex. Routing text lookups through the key's own tree also fixes a pre-existing bug where a Postings object pinned at construction could go stale and lead to an outdated score.

A no_content FT.AGGREGATE previously fetched every field of every key on the main thread and then discarded all of it. That fetch is now skipped, the same way FT.SEARCH NOCONTENT skips it. This closes a safety gap (the fetch drove a recompute on a path that never ran the key contention check) and removes pointless work. Behavior change: a no_content aggregate no longer re-validates keys and prunes keys that were deleted, expired, or mutated out of the filter between scoring and reply, matching FT.SEARCH NOCONTENT semantics.

### Commit 3: Implement FT.AGGREGATE ADDSCORES
ADDSCORES parsed into a flag that nothing read, and a non-vector score had no route into an aggregation record because the score column and score write were gated on vector queries. Non-vector queries now default the score name to __score, matching Redis. With ADDSCORES, the score column is registered and each record carries the document's relevance score, so @__score works in LOAD, SORTBY, APPLY, and GROUPBY. The score is the carried search-time score; with no LOAD the query stays no_content and the fetch stays skipped.

Testing
Unit: all suites pass, including new tests pinning the staleness fix (fails without the per-key tree routing), the lock removal (Score() callable under a held reader lock, which deadlocked before), the scorer pre-build gate, and the ADDSCORES flag.

TSAN: clean across the query test suite. The word-bucket lock race test was verified to fail with the exact predicted race when the lock is removed.

Integration: new cluster tests cover the FT.SEARCH recompute on the local responder, the no_content aggregate fetch skip, and the with-LOAD aggregate recompute. New ADDSCORES tests cover the score field, SORTBY ordering by relevance, and the no-LOAD case.


zackcam:
> **Reviewers for this PR**
> 
> * **First Pass Reviewer: @Aksha1812** — Please do your best to do a detailed review on the PR and get a response on your feedback. Once the first pass is done, notify the maintainer assigned to this PR to follow up on the final review and getting the PR merged. You can reach out to the people owning the relevant code paths for more help on the review.
> * **Maintainer Reviewer: @allenss-amazon** — Once the first review is done, please follow up with a final review and help to merge the change in.
> 
> Assigned automatically to the least-assigned members of the reviewer pools in `.github/reviewer-pools.json`. Use `/reviewer` or `/remove-reviewer` to adjust.

Allen you can ignore this review if you want for Brennan as he has been doing the scoring reviews and has high context on this change

src/commands/ft_aggregate.cc
```diff
@@ -255,11 +261,23 @@ absl::StatusOr<std::pair<size_t, size_t>> ProcessNeighborsForProcessing(
...
 
-  query::ProcessNeighborsForReply(
-      ctx, parameters.index_schema->GetAttributeDataType(), neighbors,
-      parameters, vector_identifier);
+  // no_content means LOAD requested no attributes, so return_attributes is
+  // empty and GetContent would fetch every field of every key only for
+  // CreateRecordsFromNeighbors to discard it. Skip it, as FT.SEARCH NOCONTENT
+  // does in HandleEarlyReplyScenarios — and with it the stale-match, expiry and
+  // slot-ownership pruning that only the fetch performs.
+  if (!parameters.no_content) {
```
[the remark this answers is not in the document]
zackcam: This from what i can tell is the same behaviour that matches what FT.SEARCH NOCONTENT already does. SO do not think we need to address this

Frank-Gu-81:
Hi @zackcam 👋 — flagging this as a **P1 launch blocker** for **valkey-search 1.3 RC1**. We're cutting the release branch the **morning of Sept 14** (RC1 lands Sept 15), so all P1s need to be merged before then.

First-pass reviewer: @Aksha1812 — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

src/query/search.cc
```diff
@@ -1492,6 +1546,22 @@ absl::Status Search(SearchParameters &parameters, SearchMode search_mode) {
...
+    // and the main-thread content fetch, and the snapshot taken here (corpus
+    // stats, per-term IDF) matches the corpus the surviving candidates were
+    // just scored with. Skipped when no reply-side recompute can happen: no
+    // predicate (match-all), no results, or a reply that fetches no content.
+    const Predicate *root_predicate =
+        parameters.filter_parse_results.root_predicate.get();
+    if (root_predicate != nullptr &&
+        !parameters.search_result.neighbors.empty() &&
+        parameters.WillFetchContentOnMainThread()) {
+      parameters.recompute_scorer = std::make_unique<SingleDocumentScorer>(
```
boda26, line 1560: The recompute scorer pre-build is unconditional now, but recompute only happens when docs are mutated mid-flight which is a rare situation. SingleDocumentScorer's constructor runs a full ResolveLeaves, which is costly and not used if mutation did not happen. You can consider handling ScoreTextQuery's ResolveLeaves to the scorer, or build recompute scorer conditionally here.

zackcam: Building conditionally is not possible here because whether a document mutates mid-flight is only knowable later on the main thread, and building there is the design this PR removes. I think the duplicated work here is minimal as it is small queries while making changes would potentially undo the work that this pr was meant for

boda26 (approved):

integration/test_aggregate_addscores.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+        for key, mapping in DOCS.items():
+            client.hset(key, mapping=mapping)
+        waiters.wait_for_true(
+            lambda: client.execute_command("FT.AGGREGATE", IDX, "hello",
+                                           "LOAD", "1", "@n")[0] == 2
+        )
+        return client
+
+    def test_addscores_is_accepted(self):
```
BCathcart, line 54: Implicitly tested in the other tests

integration/test_aggregate_addscores.py
BCathcart, line 1: I asked Claude to check compatibility and it came back with the following. I haven't verified them but assuming there's something real here it further motivates a thorough compatibility testing suite for scoring-related behaviour.

Four divergences:

**Non-default scorers are rejected.** `kScorerByStr` registers only `BM25STD` (scorer.h:42), so `SCORER TFIDF`, `TFIDF.DOCNORM`, `DISMAX`, `DOCSCORE`, `HAMMING`, `BM25STD.NORM`, `BM25STD.TANH` all fail here. The default itself matches — Redis has defaulted to BM25STD since the `BM25`→`BM25STD` rename in Redis Open Source 8.4. Whether our BM25STD values match Redis's for the same corpus is untested; Redis's multiplies by the `SCORE_FIELD` doc score and applies a slop penalty. Pre-existing scoring gap that ADDSCORES now makes visible to clients.

**`@__score` resolves even without ADDSCORES.** The score column is registered unconditionally in `ParseCommand` (ft_aggregate.cc:182-189), so `SORTBY 2 @__score DESC` without ADDSCORES is silently accepted and sorts on an unset field. Redis errors at pipeline build with ``Property `__score` not loaded nor in schema``. Lenient, not breaking, but a behavioral difference.

**Vector queries share the one score column.** `IsVectorQuery()` wins (ft_aggregate.cc:304-313) and the column is named by the KNN `AS` alias, holding the vector distance. So ADDSCORES on a KNN/hybrid query gives you the distance under the KNN alias, not a separate full-text `__score` — in Redis those are two distinct keys (`distFieldName` vs `scoreKey`), and nothing there rejects the combination.

**No-LOAD rows skip revalidation.** With no LOAD the query stays `no_content`, so the main-thread fetch — and with it stale-match/expiry/slot pruning — is skipped and the carried `Neighbor.score` is emitted as-is (ft_aggregate.cc:315-324). Deliberate, and documented in the test as parity with `FT.SEARCH NOCONTENT WITHSCORES`, but it means ADDSCORES with no LOAD can return scores for keys a loading path would have dropped. I didn't trace Redis's expired-doc path on an equivalent query, so the cross-engine half is unconfirmed.

BCathcart: I created an issue for scoring compatibility: #1387

BCathcart (commented):
I was digging into the scoring recompute integration tests and while working to make them more robust started hitting behaviour I didn’t understand. Claude tells me that #1217 changed things that broke assumptions made in this PR. I’ll dive into these and raise a PR to your branch @zackcam hopefully today or tomorrow.

integration/test_aggregate_addscores.py
zackcam: >Non-default scorers are rejected. kScorerByStr registers only BM25STD (scorer.h:42), so SCORER TFIDF, TFIDF.DOCNORM, DISMAX, DOCSCORE, HAMMING, BM25STD.NORM, BM25STD.TANH all fail here. The default itself matches — Redis has defaulted to BM25STD since the BM25→BM25STD rename in Redis Open Source 8.4. Whether our BM25STD values match Redis's for the same corpus is untested; Redis's multiplies by the SCORE_FIELD doc score and applies a slop penalty. Pre-existing scoring gap that ADDSCORES now makes visible to clients.

This one of rejecting non BM25STD is what we expect for right now so this one particularly probably is fine to be ignored

BCathcart:
The rough high level idea driving this PR was reducing locking by passing information we need from the background to the main thread. I theorized that we could re-use the scorer object created on the background thread and avoid re-fetching some schema-level data and reduce contention. The first commit started it and I then had Claude build on top of this idea to fully remove the time-slice mutex acquisitions from the main thread.

I was manually cleaning these things up in https://github.com/zackcam/valkey-search/pull/3 when I had to re-think if this is the right approach. First off, the following is all that's needed for BM25 at the schema-level:
- `N` — total documents in the index
- `dt` — documents containing the term (per-term, but an index-wide count)
- `avgdl` — average document length across the corpus

`N` and `avgdl` should be trivial to fetch under a small lock. `dt` along with key-level `tf` and `dl` can be fetched from the Postings object for the term. Unfortunately the postings object cached by ResolveLeaves() within the scoring object on the background thread can't be trusted on the main thread. It could have been deleted and recreated or it may not have even existed and now does because the key update added the first instance of the term. To re-fetch the postings we would need to have the term, which is simply enough to extract from the predicate itself for exact term matching, but for term expansion and stemming we're basically just back to square one with having to do a ResolveLeaves-like search to fetch.

I think we need to stop trying to make the scorer construction approach designed for the background work for the main thread as is. VerifyFilter doesn't touch the main TextIndex and relies on the immutable properties of the per-key TextIndex as well as takes a broad RWLock for accessing the stem tree. I think the answer may be to do the scoring through the per-key trees as well. We could cache postings objects during the VerifyFilter predicate evaluations to avoid doing additional tree traversals during scoring. I was already thinking of improving the use of the ResolveLeaves cache between pre-filter evaluation and scoring in the background, so this would be a similar pattern. The VerifyFilter predicate tree walk may not actually resolve all leaves if there is an OR though, but we could cache what was resolved to speed up ScoreNode as it walks the predicate to avoid duplicate per-key tree traversals to fetch postings objects.

zackcam:
Closing in lieu of https://github.com/valkey-io/valkey-search/pull/1400 brennan has pulled the work from this needed here

PR #1360, opened 2026-09-07
New memory allocation machinery

allenss-amazon:
# Run C++ static initializers after ValkeyModule_Alloc is established

## The problem

C++ static initializers in a loadable module are invoked by `dlopen` itself. Thus static
initializers that allocate memory must use a memory allocator that is provided by the
linking process itself. This is incompatible with the desire for the search module to
perform all of it's allocation through `ValkeyModule_Alloc`/`Free` which are only available
after `dlopen` has completed and the `ValkeyModule_OnLoad` module function has called 
`ValkeyModule_Init` to setup the function pointers for the ValkeyModule API.

Previously, a special mechanism of overrides provided compile-time redirection to a
specialized allocator. That allocator operated in two modes: one before
`ValkeyModule_Alloc` was set up and one after. In the "before" mode, allocations were
forwarded to libc's `malloc`/`free` and tracked in a hash map. In the "after" mode
allocations were no longer tracked, but every `free` had to consult the hash map to
determine which allocator's free routine needed to be invoked.

While this solved the problem, at some performance cost, it also required source-level
changes in every block of code that performed memory allocation. That becomes
increasingly difficult as more libraries are added, and it is undesirable for system
libraries such as libc and libstdc++.

## What this PR does

This PR removes that machinery and replaces it with a new one.

The linker is directed to rename the section holding the list of static initializers,
so that `dlopen` does not invoke them. Valkey core then calls `ValkeyModule_OnLoad`,
which calls `ValkeyModule_Init` to establish the addresses of all `ValkeyModule_*`
functions, and vmsdk immediately runs the static initializers to complete module
initialization.

Renaming `new`/`delete`/`malloc` is no longer needed. Instead, `malloc`, `free`,
`calloc` and the rest are defined as part of the module's own code; they perform
module-scope bookkeeping and delegate to `ValkeyModule_Alloc` and friends. The version
script marks these symbols local, which is what makes every reference inside the module
bind to them rather than to libc's.

There is no longer a safety net detecting premature allocations, and this is deliberate. The old allocator had a
fallback mode for the window before `ValkeyModule_Alloc` was available; the new one does
not. Nothing in the module allocates in that window, because the static initializers are
deferred until after `ValkeyModule_Init` has run. If something ever does, the call
faults immediately at the offending call site and Valkey's crash handler prints a
backtrace that resolves to the exact line — which localizes the problem better than the
old bookkeeping did.

For unit tests, the object file containing the allocator is not linked in; in sanitizer
builds its contents are compiled out. In both cases the linker connects the code to the
standard malloc machinery. The net result is that module-wide bookkeeping for memory
allocation does not work in these test environments. Only an integration test using the
non-sanitizer build can obtain accurate module-wide memory allocation statistics.

## A scenario that defeats both approaches

There is a scenario that the previous solution failed to handle, and that a naive
version of the new one would fail to handle as well. It is not detected by any tool and
causes memory corruption during production operation.

The failure happens when a library allocates memory without going through the override
machinery and then transfers ownership of that memory to the caller, expecting the
caller to free it. Under the old machinery this failed because the allocation was not in
the hash map, so the free routine forwarded it to `ValkeyModule_Free`, corrupting
jemalloc's internal data structures. We have already seen exactly this problem with `std::ostringstream` and `std::getline`.

This PR addresses the problem in three ways.

1. Dynamically linked libraries are deprecated. libstdc++ is now statically linked, and
   all future user-level libraries are expected to be statically linked as well.

2. The shared libraries that remain are a known-good set. Each has been examined for the
   ownership-transfer pattern, and none of them hands the module memory that the module
   is expected to release. A library that allocates and frees entirely on its own side —
   one whose matching free routine lives in the same library — poses no integrity problem, and that
   covers all of them. The set is pinned by the post-link check, so a library that has
   not been examined cannot become a dependency without the build failing. Note, it's still the case
that the memory allocated by these libraries will not show up in the Valkey memory usage statistics.
As long as these allocations are small it shouldn't be a problem.

3. libc is on that list, with three exceptions. `strdup`, `realpath` and `getcwd` return
   memory the caller is expected to release, and libc offers no matching free routine to
   release it with. All three are replaced by local implementations that allocate from
   the module allocator, so they can be used normally.

## Enforcement

None of this is visible in the source: it depends on link options and symbol binding, so
a reviewer cannot see it and a future change can undo it silently.
`ci/check_module_allocators.sh` runs after the link and enforces six invariants:

1. The static initializers are deferred — the module's `DT_INIT_ARRAY` holds only
   crtbegin's `frame_dummy`, and the renamed section is non-empty.
2. The module defines each allocator entry point itself, imports none of them from libc,
   and gives each one local binding.
3. The module has no undefined `GLIBCXX` symbols, so no C++ heap object can cross a DSO
   boundary.
4. Nothing the module exports collides with a symbol libstdc++.so.6 also defines.
5. No unhandled libc function that returns caller-owned memory is referenced.
6. The set of shared libraries the module depends on is unchanged.

The last of these is about the same hazard as the third. Every DSO the module links
against is another allocator boundary, and a new dependency is one nobody has reviewed.
The current non-libc dependencies are safe: libsystemd is used only for four socket
predicates that return `int`; the OpenSSL objects the module obtains are each paired with
the matching OpenSSL free routine, and the raw buffers (`ASN1_STRING_to_UTF8`, the
`i2d_*` family) are released by `OPENSSL_free` inside libcrypto, so both halves happen on
the far side of the boundary. OpenSSL is deliberately left dynamic: linking it statically
would mean rebuilding the module for every OpenSSL CVE instead of picking up a
distribution update.

Two of these failures are silent. If the allocators end up with global rather than local
binding, they are preempted by libc's and the module simply stops using
`ValkeyModule_Alloc` — nothing crashes, and no memory is accounted for. If new code calls
a libc function that hands back memory, the result is heap corruption in production and
nothing earlier. Each check has been verified to fail against a build that violates it.


vmsdk/src/memory_allocation_c_api.cc
```diff
@@ -0,0 +1,240 @@
...
+    if (cwd == nullptr) {
+      errno = ENOMEM;
+      return nullptr;
+    }
+    if (syscall(SYS_getcwd, cwd, capacity) >= 0) {
+      return cwd;
+    }
+    const int saved_errno = errno;
+    free(cwd);
+    if (saved_errno != ERANGE || size != 0) {
```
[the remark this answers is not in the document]
allenss-amazon: Mimics the existing function behavior in libc

Aksha1812:
> /priority 1360 P2

@allenss-amazon this only works in the triage board itself https://github.com/valkey-io/valkey-search/issues/1346 . it would be difficult for the dashboard to read all comments from all PRs to pick this information , if we want such information to be updated in the PR itself , we can use labels assigning priority and have the dashboard read through labels of the PR. Ig this is misleading "General commands — act on a PR (the priority tables below):" , but "the priority tables below"  in the dashboards context it means updating the per PR section in the reviewer dashboard on the issue itself . Let me know if i need to be more explicit in the dashboard instructions to make that clear

Frank-Gu-81:
Hi @allenss-amazon 👋 — this is tracked as a **P2** for **valkey-search 1.3**. P2s aren't RC1 blockers, but we'd love to land them for **GA**.

First-pass reviewer: @Frank-Gu-81 — if your first-pass review is already done, please ignore this message; otherwise, please prioritize getting this PR reviewed.

vmsdk/src/memory_allocation_c_api.cc
```diff
@@ -0,0 +1,240 @@
...
+  return new_ptr;
+}
+
+void* aligned_alloc(size_t alignment, size_t size) noexcept {
+  void* ptr = ValkeyModule_Alloc(AlignSize(size, alignment));
+  if (ABSL_PREDICT_TRUE(ptr != nullptr)) {
+    vmsdk::ReportAllocMemorySize(ValkeyModule_MallocUsableSize(ptr));
+  }
+  return ptr;
+}
```
Frank-Gu-81, line 133: Does this guarantee the return address is aligned to the `alignment`?

allenss-amazon: It doesn't. And Valkey doesn't export a function that implements this. Will open an issue to resolve this.

Frank-Gu-81: Sounds good. Thanks!

ci/check_module_allocators.sh
```diff
@@ -0,0 +1,277 @@
...
+# the local entries the version script produces are filtered out.
+#
+LIBSTDCXX=$(gcc -print-file-name=libstdc++.so.6 2>/dev/null || true)
+if [ -n "${LIBSTDCXX}" ] && [ -f "${LIBSTDCXX}" ]; then
+    CLASHES=$(comm -12 \
+        <(nm -D --defined-only "${MODULE_SO}" |
+              awk '$2 ~ /^[TDWVBRi]$/ {sub(/@@?.*$/, "", $3); print $3}' |
+              sort -u) \
+        <(nm -D --defined-only "${LIBSTDCXX}" |
+              awk '{sub(/@@?.*$/, "", $3); print $3}' | sort -u))
```
Frank-Gu-81: Check 4 appears to filter out GNU-unique symbols before comparing exports against libstdc++.so.6. Since the comment says the dangerous case is `STB_GNU_UNIQUE` locale facet IDs, this looks like the checker can miss the exact collision it is supposed to catch. Should we include that symbol class in the module-side nm filter or is this already covered?

Frank-Gu-81 (approved):
I do not have other outstanding comments other than the one in `ci/check_module_allocators.sh`. The rest of the file/logic looks good to me.

ci/check_module_allocators.sh
allenss-amazon: good catch. Fixing it now.

allenss-amazon: Fixed in dd1e0a3 — and it was worse than "can miss".

`nm` renders `STB_GNU_UNIQUE` as a lowercase `u`, which is indistinguishable at a glance from a local symbol. The `/^[TDWVBRi]$/` filter was there to drop the LOCAL entries the version script produces, and it was silently dropping every unique symbol along with them — i.e. the entire class the check is named for. libstdc++.so.6 exports 106 of them.

Measured on a build with `-Wl,--exclude-libs,ALL` removed:

| | collisions reported |
|---|---|
| before | 4526 (106 missed, incl. `std::num_put<char>::id`) |
| after | 4632 |

So it did still fail that build — but only because thousands of unrelated symbols collided too. Had a change exported just the facet ids, the check would have passed and the module would have crashed at load as soon as another C++ module (valkey-json, in the integration tests) was present. Exactly the case the check exists for.

The fix reads the binding from `readelf`'s Bind column instead, where the class is spelled `UNIQUE` and can't be mistaken for `LOCAL`, and accepts GLOBAL/WEAK/UNIQUE. Both sides of the comparison now go through a single `exported_syms` helper so they can't drift apart, and check 2 uses it too since it asks the same question about the allocator entry points. Verified both directions: the facet ids now appear on a bad build, and a build with `--exclude-libs,ALL` still passes.

Frank-Gu-81: New changes looks good to me! Thanks.

Frank-Gu-81 (approved):

Issue #1353, opened 2026-09-02
[BUG] FT.SEARCH Compatibility Issues


BCathcart:
Claude seems to have developed pretty good knowledge about RedisSearch APIs and was able to find a handful of valid divergences in the scoring PR #1287. I set it on the task to find existing `FT.SEARCH` incompatibilities and the results are below. I'll also publish a draft PR attached to this issue with a failing compatibility test for each case. This was done a couple weeks ago so some should already be fixed and there could be new ones since.

## FT.SEARCH: 8 divergences from RediSearch, none covered by tests

Comparing byte-identical commands against RediSearch 2.10.20 (`search` 21020, from
`redis/redis-stack-server`) and a source build of `main`, all pinned `DIALECT 2`:

| # | Divergence | RediSearch | valkey-search | Test |
|---|---|---|---|---|
| 1 | `SORTBY` ignored under `NOCONTENT` — `ASC` == `DESC` | `[3,'n:2','n:3','n:1']` | `[3,'n:1','n:2','n:3']` | `test_sortby_ignored_under_nocontent` |
| 2 | `WITHSORTKEYS` + `NOCONTENT` drops the sort-key elements, so reply arity is wrong | `len=7` | `len=4` | `test_withsortkeys_dropped_under_nocontent` |
| 3 | String `SORTBY` sorts by raw bytes instead of collating case-insensitively | `apple, Banana, Cherry, date` | `Banana, Cherry, apple, date` | `test_string_sortby_is_case_sensitive`, `test_tag_sortby_is_case_sensitive` |
| 4 | Sort-key prefix is always `#`; strings should use `$` | `$apple` | `#apple` | `test_string_sortkey_prefix` |
| 5 | Absent sort key emits the bare string `#` instead of nil | `(nil)` | `#` | `test_sortkey_without_sortby_is_not_nil`, `test_sortkey_of_missing_field_is_not_nil` |
| 6 | Numeric sort keys and `RETURN` values echo the raw hash bytes | `#2.5`, `#1000` | `#2.500`, `#1e3` | `test_numeric_sortkey_not_normalized` |
| 7 | `RETURN 0` keeps `NOCONTENT` set, so a later `RETURN <field>` is ignored | `[1,'o:1',['title','hello world']]` | `[1,'o:1']` | `test_return_zero_overridden_by_later_return` |
[...]

Issue #1351, opened 2026-09-02
[BUG] Memory leak during test_dropindex_with_background_queries


BCathcart:
Caught in this ASAN test run: https://github.com/valkey-io/valkey-search/actions/runs/33572270173/job/100068579792?pr=1323

```
../../../integration/test_fulltext_inflight_blocking.py::TestFullTextInFlightBlockingCMD::test_dropindex_with_background_queries PASSED [ 41%]
=================================================================
==21851==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 64 byte(s) in 2 object(s) allocated from:
    #0 0x7f98c69a5548 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:95
    #1 0x7f98ba2958af in valkey_search::indexes::text::InvasivePtr<valkey_search::indexes::text::Postings> valkey_search::indexes::text::InvasivePtr<valkey_search::indexes::text::Postings>::Make<>() /src/indexes/text/invasive_ptr.h:63
    #2 0x7f98ba2958af in AddKeyToPostings /src/indexes/text/text_index.cc:65
    #3 0x7f98ba2958af in valkey_search::indexes::text::TextIndexSchema::CommitKeyData(valkey_search::InternedStringPtr const&) /src/indexes/text/text_index.cc:286
    #4 0x7f98ba1ff406 in valkey_search::IndexSchema::SyncProcessMutation(ValkeyModuleCtx*, absl::lts_20240722::flat_hash_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, valkey_search::IndexSchema::DocumentMutation::AttributeData, absl::lts_20240722::container_internal::StringHash, absl::lts_20240722::container_internal::StringEq, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, valkey_search::IndexSchema::DocumentMutation::AttributeData> > >&, valkey_search::InternedStringPtr const&) /src/index_schema.cc:733
    #5 0x7f98ba20067a in valkey_search::IndexSchema::ProcessSingleMutationAsync(ValkeyModuleCtx*, bool, valkey_search::InternedStringPtr const&, vmsdk::StopWatch*) /src/index_schema.cc:1030
    #6 0x7f98ba2015d2 in operator() /src/index_schema.cc:903
[...]

PR #1364, opened 2026-09-08
fixing redis search divergence #4: sort key prefix always use #; strings should use $

JasonLi314:
fixing #1353 item 4: Sort-key prefix is always #; strings should use $

Notice this fixed behavior is gated so only emulated version 1.3.0+ will see this change.


integration/test_non_vector.py
```diff
@@ -1041,6 +1041,47 @@ def test_tag_and_negate_at_scale(self):
         create_bulk_data_standalone(client)
         validate_tag_and_negate_queries(client)
 
+    def test_withsortkeys_prefix_by_field_type(self):
+        """
+            Regression test for issue #1353, divergence item 4.
```
neerajr0: Nit, add issue link here for future reference: https://github.com/valkey-io/valkey-search/issues/1353.

src/commands/ft_search.cc
```diff
@@ -203,6 +203,17 @@ void SerializeNonVectorNeighbors(ValkeyModuleCtx *ctx,
...
+
+  const bool is_numeric = [&]() -> bool {
+    if (!command.with_sort_keys || !command.sortby_parameter.has_value()) {
+      return false;
+    }
+    auto idx = command.index_schema->GetIndex(command.sortby_parameter->field);
+    return idx.ok() &&
+           idx.value()->GetIndexerType() == indexes::IndexerType::kNumeric;
+  }();
+  const std::string prefix_str = is_numeric ? "#" : "$";
```
neerajr0: AI is saying the same change is needed in `SerializeNeighbors` https://github.com/JasonLi314/valkey-search/blob/4447aa075b1a22ea217e3f356fe2d406650b6dc8/src/commands/ft_search.cc#L134, can you confirm?

src/commands/ft_search.cc
neerajr0, line 1: Can you also confirm no other field types use a prefix besides `#`?

integration/test_non_vector.py
```diff
@@ -1041,6 +1041,47 @@ def test_tag_and_negate_at_scale(self):
...
+        result = client.execute_command(
+            "FT.SEARCH", "sortkey_prefix_idx", "@m:{all}", "SORTBY", "n", "ASC",
+            "WITHSORTKEYS", "RETURN", "1", "n", "DIALECT", "2")
+        assert result == [
+            3,
+            b"skp:2", b"#1", [b"n", b"1"],
+            b"skp:3", b"#2", [b"n", b"2"],
+            b"skp:1", b"#3", [b"n", b"3"],
+        ]
+
```
neerajr0, line 1121: Consider also verifying against a type that is neither string nor numeric to confirm the prefix is still `#`.

integration/test_non_vector.py
JasonLi314: other code seems to use issue/pr numbers only not full link. I will keep the convention.

integration/test_non_vector.py
JasonLi314: good catch. added test. also one correction: we should only use `#` for numbers and all other uses `$`.

src/commands/ft_search.cc
JasonLi314: fixed

src/commands/ft_search.cc
JasonLi314: fixed

BCathcart:
A couple small things:
- Since there is cluster startup and teardown overhead, let's generally prefer testing multiple searches against the same index within the same test invocation instead of across parameterizations. I'd like to get to a world where our integration tests are fast enough it doesn't matter, but we're pretty far away from that right now.
- As a project we should have a discussion around our strategy for compatibility tests. Really anything that verifies API behaviour can, and maybe should, be a compatibility test. We've mostly followed the practice of writing regular integrations tests for new features and then scrambling to write some compatibility tests after the fact. I'll bring this up in the OSS meeting tomorrow.

integration/test_non_vector.py
```diff
@@ -1087,6 +1042,97 @@ def test_tag_and_negate_at_scale(self):
...
+        # order-insensitive and prefix-only to stay independent of
+        # tie-ordering (divergence 8) and of the SORTBY-under-NOCONTENT fix
+        # (issue #1215).
+        result = client.execute_command(
+            "FT.SEARCH", self.SORTKEY_PREFIX_IDX, "@m:{all}", "SORTBY", "vec",
+            "ASC", "WITHSORTKEYS", "RETURN", "1", "z", "DIALECT", "2")
+        assert result[0] == 3
+        sort_keys = result[2::3]
+        assert len(sort_keys) == 3
+        assert all(key.startswith(b"$") for key in sort_keys)
```
BCathcart: I'm curious does this return the raw bytes of the vector after "$" as the sort key value?

BCathcart (approved):

BCathcart:
@JasonLi314 can you also please check when SORTBY functionality was introduced to figure out which previous versions (1.0, 1.1, 1.2) this needs to be backported to?

BCathcart:
After looking at #1366 I think we need to decide as a whole on how to treat API changes to restore compatibility with Redis search. I was considering this as a bug fix that should be backported but I want to get alignment on that before doing so.

Secondly, you can see all the testing in that PR is done under the compatibility testing framework. I do think that makes more sense here as well. Can you take a look at migrating your tests into the compatibility framework? If there are any challenges in doing so please raise them. Developers should be incentivized to validate our APIs using it and if it's annoying to do so let's improve it.

integration/test_non_vector.py
JasonLi314: yes, and i've changed test to assert whole byte vector

neerajr0:
Thanks for addressing above. AI is pointing out the behavior for `WITHSORTKEYS` without `SORTBY`. Can you confirm if this needs a fix?

> 1. Missing sort key should be nil, not a prefixed empty string. GetSortKeyValue returns "" when there's no sortby_parameter or the doc lacks the field. WITHSORTKEYS without SORTBY is accepted (nothing in PostParseQueryString binds them), so valkey-search now replies the one-byte string "$". The change silently flipped that case from "#" to "$" — both wrong, and untested either way. Same for a doc that doesn't have the SORTBY field.
 
> 2. This directly answers BCathcart's vector question — and the test locks in the wrong answer. The new ("vec", b"$") case asserts $ + 8 raw bytes. Asserting $<raw bytes> as expected codifies a valkey-search-only behavior as compatible. I'd flag it as needing a live RediSearch check before it lands, which is exactly the argument for BCathcart's compat-framework request.

JasonLi314:
1. integration-tests-asan failure seems to be unrelated to this change. Issue tracked in #1351

2. > @JasonLi314 can you also please check when SORTBY functionality was introduced to figure out which previous versions (1.0, 1.1, 1.2) this needs to be backported to?
  
    SORTBY was first added in `1.2.0-rc1` so only 1.2 needs backporting.

3. I think this one should be a gated fix as it does change what client sees and affects how their parser and application logic works.
4. I will port integ test to comp test framework right now.

JasonLi314:
@neerajr0
1. This is divergence 5 listed in #1353 which will be fixed separately
2. I ran redis search and confirmed its behavior, current integration test assertions are correct, but yes I'm migrating to comp test.

neerajr0 (approved):

BCathcart:
Ok I understand our compatibility philosophy better now. I agree it should be gated to the next release (1.3.0) which means it only becomes the default behaviour on the next major version. Because of that I'm dropping the priority but I'm sure we can still land this for 1.3.0.

JasonLi314:
I have made this change gated so it's only effective for emulated version 1.3.0+.

I have moved test itself to compatibility test, though I've retained an integ test for the emulated behavior.

integration/compatibility/generate_sortkey.py
```diff
@@ -0,0 +1,49 @@
...
+@pytest.mark.parametrize("key_type", ["hash"])
+class TestSortKeyPrefixCompatibility(BaseCompatibilityTest):
+    ANSWER_FILE_NAME = "sortkey-answers.pickle.gz"
+
+    # All field types valkey-search accepts in FT.CREATE (GEO is rejected).
+    SORTABLE_FIELDS = ("z", "t", "n", "f", "vec")
+
+    def test_withsortkeys_prefix_by_field_type(self, key_type):
+        self.setup_data(SORTKEY_PREFIX_DATA_SET, key_type)
+        time.sleep(0.5)  # tiny data set; give indexing a moment
```
BCathcart, line 24: Indexing is done synchronously so sleeps aren't needed in any of these tests

integration/compatibility/generate_sortkey.py
```diff
@@ -0,0 +1,49 @@
...
+        for field in self.SORTABLE_FIELDS:
+            self.check("FT.SEARCH", f"{key_type}_idx1", "@m:{all}",
+                       "SORTBY", field, "ASC", "WITHSORTKEYS",
+                       "RETURN", "1", field, "DIALECT", "2")
```
BCathcart, line 28: Just an observation not specific to this PR: moving the tests to the compatibility framework actually gives less visibility to the reader about what we're asserting. You can't really read the tests to get an understanding of the code. This might also be why I've seen developers prefer to write regular integration tests instead of compatibility tests. It's a little less intuitive. It could be worth having the expected results made visible somewhere. This could be auto generated

BCathcart (commented):
Did a quick final pass. LGTM.

Just an fyi for the future, it's better to push new commits than amend and force push the existing one. Force pushing prevents the reviewer from being able to see the diff since the last review

integration/compatibility/generate_sortkey.py
JasonLi314: Yes this actually caused some of the divergence issue undetected because the exact assertions are not clear. From my understanding `WITHSORTKEYS` compatibility tests are basically skipped silently.

integration/compatibility/generate_sortkey.py
JasonLi314: will fix in next pr

PR #1068, opened 2026-05-27
Fixup compatibility testing to latest Redis server

allenss-amazon:
Included PR #1063 

Change default compatibility Redis from redis-stack-server to redis:latest

1. Fix divide-by-zero incompatibilty
2. Fix parsing precedence 
3. Fix text quote parsing.

All fixes are gated by the emulate-release compatibility machinery, tagged to 1.2.1 release.

PR #1366, opened 2026-09-09
Move compatibility reference to redis:latest and fix what it exposed (supersedes #1068, #1239)

allenss-amazon:
Supersedes #1068 and #1239. Both moved the compatibility reference server from
`redis-stack-server` to `redis:latest` and fixed what that exposed, but they were
developed independently and shared no commits, so four engine fixes existed twice
in byte-identical form under the same compatibility labels at different release
gates. Neither could absorb the other: merging them auto-merged two changes
*without* conflict markers into broken code, a duplicate
`case indexes::IndexerType::kText:` in one switch and the expression generator
registered twice in `GENERATORS`.

This branch is cut fresh from `main` and populated by hand, taking each
overlapping item from whichever branch did it better, then rebased onto #1271.

## Why redis:latest

Redis 8 bundles Redisearch natively, so the suite compares against current
mainline behavior rather than the older redis-stack build. The expression sweep
(`generate_expr.py`) had never been registered in `GENERATORS`, so it was neither
regenerated nor replayed; it is now, and it passes.

## Compatibility fixes

All gated on `search.emulate-release` at 1.3.0 and documented in
`COMPATIBILITY.md`.

| Gate | Fix |
|---|---|
| `ft_aggregate_divide_by_zero` | Division by zero follows IEEE 754: `+inf`, `-inf`, `NaN` for `0/0`, rather than collapsing to a flat `NaN` |
| `ft_aggregate_pow_precedence` | `^` gets its own precedence level above `*` and `/`, and is right-associative |
| `ft_aggregate_logical_precedence` | `&&` binds tighter than `\|\|`, instead of sharing one left-associative level |
| `ft_search_apostrophe_phrase` | An unescaped `'` is a phrase delimiter equivalent to `"`, paired left to right |
| `empty_string_not_numeric` | `""` is not a number, so `abs("")` is `nan` and `(0)==("")` is false |
| `reduce_sum_avg_empty_nan` | `SUM` and `AVG` over a group that folded nothing answer `nan`, not `0` |

Two fixes are deliberately ungated. A TEXT-indexed field is now accepted as an
`APPLY` operand, which turns a hard error into a result, so nothing could have
depended on the old behavior. And `timefmt` no longer spins on a format that
produces no output: `strftime` returns 0 both when the buffer is too small and
when the format yields nothing, and a vector blob passed as the format truncates
at its first NUL to exactly that. Both ends are fixed, the reachable trigger and
the unbounded loop below it. That one is remotely reachable and worth a look.

## Retuning #1271 to the new reference

Three rules from #1271 were measured against redis-stack and answer differently
on Redis 8. Each surfaced only once the pickles were regenerated against the new
reference.

| | redis:latest | redis-stack |
|---|---|---|
| `MIN` / `MAX` over nothing | `inf` / `-inf` | `0` |
| `SUM` / `AVG` over nothing | `nan` | `0` |
| `MIN` over a TAG field or an array | `inf` | `0` |
| `APPLY (@absent)&&(1)` | record kept, `x = 0` | record dropped |

`reduce_minmax_numeric` keeps its structure and its gate; only the value it lands
on changes, and a non-number is now skipped rather than read as `0`.
`expr_missing_propagates` is narrowed so `&&` is exempt, because a missing operand
there is falsy on Redis 8. Neither pre-1.3.0 path is touched.

Redis 8 also *truncates* the result stream after a row whose `APPLY` referenced an
absent field, losing later documents that do have it, while `total_results` still
reports the full match count. That is not reproduced here, because matching it
means deliberately returning incomplete results. It is recorded in
`known_differences.md` along with everything else the two reference engines
disagree about.

## Tests

All four answer pickles are regenerated against `redis:latest`, #1271's array
sweep included. The sources hash covers every `.py` in the directory, so they
regenerate together.

- Unit tests: 23 suites, all pass
- Compatibility replay: 5 of 5 answer files, zero data mismatches
- Full integration suite: 496 passed, 8 skipped

`known_differences.md` records what a passing suite does not cover: the result
stream truncation above, a GROUPBY key absent on some documents, `FIRST_VALUE`
over an empty group, and the float formatting the harness absorbs rather than
matching, since `std::to_chars` shortest round-trip is a deliberate choice.

https://claude.ai/code/session_017Zo94xWTehNvzHVkVLPdFV


integration/compatibility/known_differences.md
```diff
@@ -0,0 +1,132 @@
...
+* **`integration/run.sh` zaps every valkey-server on the host.** Its `zap`
+  helper matches on process name and `kill -9`s all of them, so two integration
+  runs on one machine kill each other's servers. Failures look like
+  `Connection closed by server` or `Error 111 connecting`, and clear on a re-run.
+
+* **The `kText` fix is ungated.** Allowing a TEXT field as an `APPLY`/`FILTER`
+  operand turns a hard error into a result, so no client can have depended on
+  the old behavior and no `search.emulate-release` gate was added. That is
+  inconsistent with how the other fixes here are handled, and may warrant a gate
+  plus a COMPATIBILITY.md row.
```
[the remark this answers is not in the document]
allenss-amazon: This behavior change doesn't require compatibility gate.

BCathcart:
> All gated on search.emulate-release at 1.3.0 and documented in COMPATIBILITY.md.

We're supposed to maintain compatibility across minor versions, so if we consider this a breaking change then shouldn't we gate until the next major version? If we consider it a bug that applications shouldn't be relying on, then no need to gate at all.

This also gets to a question around how we treat API incompatibility with Redis. @JasonLi314 is solving incompatibilities found in #1353 as well. We claim API compatibility with Redis so I think we have a justification to make these API changes in patch/minor versions. If doing so would break existing applications then I'd rather hold off, but I don't know if hitting these cases would break existing redis clients that applications rely on and not fixing them is simply leaving the functionality unusable.

integration/compatibility/array-input-answers.pickle.gz
Frank-Gu-81, line 1: [FAILED ../../../integration/compatibility_test.py::TestAnswersCMD::test_answers[array-input-answers.pickle.gz]](https://github.com/valkey-io/valkey-search/actions/runs/34654894906/job/103445031770?pr=1366)

Looks like it failed on the following 4 cases:
```
 1. JSON with (@items)==(@items2)
 2. JSON with (@items)!=(@items2)
 3. HASH with (@items)==(@items2)
 4. HASH with (@items)!=(@items2)
```
That came from `generate_array.py`:
```
 RL: items=[1,9], items2=[1,2], result=1
 VK: items=[1,9], items2=[1,2], result=0
```

So array comparison exposes some incompatibility here where Redis compares non-empty arrays using only their first elements, while `Value::Compare` compares all elements lexicographically. This makes valid `FT.AGGREGATE APPLY` expressions return opposite values for cases such as `[1,9] == [1,2]`, and currently fails the HASH and JSON compatibility runs.

src/expr/expr.cc
```diff
@@ -546,22 +554,32 @@ struct Compiler {
...
+    static const std::vector<DyadicOp> kFixedLorOps{{"||", &FuncLor}};
+    static const std::vector<DyadicOp> kLegacyLogicalOps{{"||", &FuncLor},
+                                                         {"&&", &FuncLand}};
+    auto fixed = [&] { return DoDyadic(ctx, &Compiler::LandOp, kFixedLorOps); };
+    auto legacy = [&] {
+      return DoDyadic(ctx, &Compiler::CmpOp, kLegacyLogicalOps);
     };
-    return DoDyadic(ctx, &Compiler::LandOp, ops);
+    return VALKEY_SEARCH_COMPATIBILITY_FIX(
+        1, 3, 0, "ft_aggregate_logical_precedence", fixed, legacy);
```
Frank-Gu-81, line 579: The compatibility gates increment their `INFO` counters whenever the parser or evaluator reaches them, even if the affected behavior is not exercised. For example, `APPLY @price + 1` increments both precedence counters despite containing no `^`, `&&`, or `||`; `@price / 2` increments the divide-by-zero counter for every record; and an ordinary text query increments the apostrophe-phrase counter.

I think it could confuse operators on determining whether upgrading is safe or not. A few places I see:
1. `expr.cc:607`
2. `value.cc:559`
3. `filter_parser.cc:845`

integration/compatibility/known_differences.md
```diff
@@ -0,0 +1,132 @@
...
+values that differ byte-for-byte through `compare_number_eq`, which uses
+`math.isclose` and treats `nan` and `-nan` as equal.
+
+## 2. Where the two reference engines disagree
+
+These are not valkey-search defects. They are places where `redis:latest` and
+`redis/redis-stack-server` answer differently, so the choice of reference
+decides what "compatible" means. Recorded because several rules in
+`COMPATIBILITY.md` were originally measured against redis-stack and adjusted
+when the reference moved.
```
BCathcart, line 127: Did the redis search major version effectively change when switching to redis:latest? I know they do versioning a bit different now but I'm wondering if this is showing that Dialect 2 is a constantly moving target or it could have been potentially been stable until their major version change. It would be nicer to claim a major version of valkey search is API compatible with a major version of redis search rather than valkey search is compatible with whichever way Dialect 2 was blowing at the moment

BCathcart (approved):

src/expr/expr.cc
allenss-amazon: True and difficult to determine with the current parser structure.

integration/compatibility/array-input-answers.pickle.gz
allenss-amazon: Yes, I've removed the array equality tests as IMO, Redis is broken. This gets documented in the divergence file.

integration/compatibility_test.py
```diff
@@ -329,8 +339,22 @@ def compare_row(l, r, key_type):
...
+                continue
+            # Values differ byte-for-byte. If both are numeric, fall back to the
+            # tolerant numeric compare -- it treats nan/-nan as equal and uses
+            # math.isclose, absorbing the two engines' differing float precision
+            # and negative-zero formatting on any server-computed numeric field
+            # (APPLY results, GROUPBY reducers). Non-numeric values (concat/
+            # lower/substr/timefmt string results, tags, keys) stay an exact
+            # match.
+            if _is_numeric(lv) and _is_numeric(rv) and compare_number_eq(lv, rv):
+                continue
```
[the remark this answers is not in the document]
allenss-amazon: fixed by remove the ambiguous cases.

BCathcart:
> Did the redis search major version effectively change when switching to redis:latest? I know they do versioning a bit different now but I'm wondering if this is showing that Dialect 2 is a constantly moving target or it could have been potentially been stable until their major version change. It would be nicer to claim a major version of valkey search is API compatible with a major version of redis search rather than valkey search is compatible with whichever way Dialect 2 was blowing at the moment

^ Answering my resolved question, the docs explicitly call out Redis 8 including breaking changes for search.

The Redis docs also quote "Any breaking changes to the syntax are introduced in a new dialect, while RediSearch continues to support older dialects", but Claude found that this doesn't hold:

> Redis 8.0 broke it twice: it shipped parsing changes to APPLY expressions and stricter argument validation without scoping them to a new dialect

I do think it's worth mapping a Valkey Search version compatibility to a major version of Redis from now on. The first version of Valkey Search targeting compatibility with Redis 8 is 1.3.0. The previous versions were targeting RedisSearch 2.10.20 before the versioning system change. If we find Redis search has breaking changes within a major version, then we'll have to be a bit more specific but it's in everyone's best interest if that doesn't happen.

My proposed COMPATIBIILITY.md section: #1386.

PR #1373, opened 2026-09-09
FT.AGGREGATE: SORTBY ignores a following LIMIT, breaking paging past the first page

allenss-amazon:
`FT.AGGREGATE` `SORTBY` truncated its input to its own `max_`, which defaults to
10, before any later stage ran. A following `LIMIT` therefore only ever saw ten
records.

```
FT.AGGREGATE idx * LOAD 2 @__key @n1 SORTBY 2 @n1 ASC LIMIT 0 1000   -> 10 rows
FT.AGGREGATE idx * LOAD 2 @__key @n1 SORTBY 2 @n1 ASC LIMIT 500 100  ->  0 rows
```

The second is the damaging one. Paging through sorted aggregate results returns
correct data for the first page and then silently nothing, because the offset
falls past the end of the ten records that survived the sort. It is an empty
reply, not an error, so a client cannot tell.

## The rule

The bound is a property of the whole pipeline, not of the `SORTBY` clause.
Measured over 10000 documents on **both** `redis:latest` and
`redis/redis-stack-server`, which agree on every case:

| Query | rows |
|---|---|
| `SORTBY`, no `MAX`, no `LIMIT` | 10 (the default) |
| `SORTBY MAX 50`, no `LIMIT` | 50 (`MAX` sets the bound) |
| `SORTBY MAX 0`, no `LIMIT` | 10 (`MAX 0` means unset) |
| `SORTBY` + `LIMIT 0 1000` | 1000 (a `LIMIT` raises it) |
| `SORTBY MAX 5` + `LIMIT 0 100` | 100 (a `LIMIT` outranks `MAX`) |
| `SORTBY` + `LIMIT 500 100` | 100 (the offset counts too) |
| `LIMIT 0 1000` then `SORTBY` | 1000 (already bounded upstream) |

## The fix

`ResolveSortByBounds()` runs once parsing is complete and walks the stages. A
`LIMIT` after the `SORTBY` raises the bound to its offset plus count, saturating
rather than wrapping so a huge offset just means retain everything. A `LIMIT`
before it means the stream is already bounded and nothing should be truncated
again. Otherwise an explicit `MAX` stands, and `MAX 0` falls back to the default
of 10.

It cannot be done while parsing a single clause, which is why the default was
wrong in the first place: `SORTBY` is parsed before the `LIMIT` that governs it
is known.

Gated at 1.3.0 as `compatibility-sortby_max_follows_limit` and documented in
`COMPATIBILITY.md`, like the other observable `FT.AGGREGATE` changes. With the
gate off the parsed `MAX` is used unchanged, so nothing outside the
compatibility replay changes behavior.

## Why the suite missed it

It already paired `SORTBY` with `LIMIT` in six queries, but every one asks for at
most 5 rows at an offset of at most 2. Those fit inside the ten-record window
whatever its size, so none of them could ever observe the truncation. The one
larger value in the suite, `LIMIT 0 40`, is never paired with a `SORTBY`.

The new generator case uses a count past the bound and an offset past it, on the
15-document `sortable numbers` dataset, plus the `MAX` interactions and a `LIMIT`
ahead of the `SORTBY`. It records 168 answers.

## Tests

- Unit tests: 23 suites, all pass
- Compatibility replay: all three answer files, zero data mismatches
- Full integration suite: 495 passed, 8 skipped, 1 unrelated flake

The flake is `test_fulltext.py::TestFullText::test_stemming`. It passes on its
own and the whole `TestFullText` class passes on its own (42 passed), so it only
fails inside the full-suite run. This PR touches only `FT.AGGREGATE` parsing, and
with the gate at its default it is behaviorally identical to `main`, so it cannot
reach `FT.SEARCH` stemming.

https://claude.ai/code/session_017Zo94xWTehNvzHVkVLPdFV


src/commands/ft_aggregate_parser.cc
```diff
@@ -423,6 +426,63 @@ AggregateParameters::MakeReference(const absl::string_view name, bool create) {
...
+    const size_t parsed_max = sortby->max_;
+    // MAX 0 is how Redis spells "no MAX", so it falls back to the default.
+    size_t resolved = parsed_max == 0 ? SortBy::kDefaultMax : parsed_max;
+
+    // Only an immediate neighbour bounds this SORTBY. A LIMIT further along
+    // the pipeline is separated by a stage that changes the record count, so
+    // its window is not a bound on what the sort has to keep.
+    const Limit *adjacent = i + 1 < stages.size()
+                                ? dynamic_cast<Limit *>(stages[i + 1].get())
+                                : nullptr;
```
Frank-Gu-81, line 468: I think this way it would only recognizes a `LIMIT` directly adjacent to `SORTBY`? So, something below would be unrecongnized?

```
SORTBY, APPLY, LIMIT
SORTBY, FILTER, LIMIT
```

integration/compatibility/generate.py
```diff
@@ -518,6 +518,35 @@ def test_aggregate_limit(self, key_type, dialect, vector_data_type):
...
+        self.check(dialect, f"{base} limit 20 5")
+        # An explicit MAX smaller than the LIMIT: the LIMIT wins.
+        self.check(dialect, f"{base} max 3 limit 0 15")
+        # MAX 0 means no MAX, so the default bound applies.
+        self.check(dialect, f"{base} max 0")
+        # MAX alone sets the bound when no LIMIT follows.
+        self.check(dialect, f"{base} max 12")
+        # A LIMIT ahead of the SORTBY has already bounded the stream.
+        self.check(dialect,
+            f"ft.aggregate {key_type}_idx1 * load 3 @__key @n1 @n2 limit 0 14 sortby 2 @n1 asc")
```
Frank-Gu-81, line 579: Would it be beneficial to add another case where limit and sortby is not adjacent to each other? Anyways, not a blocker.

src/commands/ft_aggregate_parser.cc
allenss-amazon: If there's no intervening stage that INCREASES the number of records, then this is a viable optimization. But the GROUPBY case can increase the number of records. 

This is a courtesy optimization anyways. Nothing prevents a user from specifying the MAX clause which has the same effect.

src/commands/ft_aggregate.cc
```diff
@@ -194,6 +194,11 @@ absl::Status AggregateParameters::ParseCommand(vmsdk::ArgsIterator &itr) {
     return absl::InvalidArgumentError("Only Dialects 2, 3 and 4 are supported");
   }
 
+  // SORTBY's retention bound depends on the LIMIT stages around it, so it can
+  // only be settled once the whole pipeline is known. Must precede
+  // GetSerializationRange(), which reads the resolved stages.
+  ResolveSortByBounds(*this);
```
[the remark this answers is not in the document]
allenss-amazon: False increments of that counter are not a concern.

Frank-Gu-81 (approved):

src/commands/ft_aggregate_parser.cc
Frank-Gu-81: Makes sense. Thanks for the clarification.

integration/compatibility/generate.py
allenss-amazon: done.

Issue #31, opened 2025-02-03
Restrict Index names to allow future per-slot index functionality.


allenss-amazon:
I propose that in the 1.0 release that index names be restricted in their format. In particular index names that look like they have a slot tag should be rejected as invalid names.

A future feature would allow "per-slot" indexes. This would easily be expressed by adding a slot-tag to the index name and validating that each prefix string also had the same slot-tag. 

Per-slot indexes have several advantages.

1. The query performance of a cluster can scale horizontally when using per-slot indexes (assuming there are enough of them).
2. Clients can easily route directly to the correct destination shard simply by treating an index name as a key.
3. They are much easier to slot migrate.

Implementation is relatively simple, and can easily be done in a future release. By reserving the the slot-tag notations for the 1.0 release we ensure that a future implementation of this feature won't conflict with any currently named indexes.

Issue #30, opened 2025-02-02
Revise Startup memory allocation machinery


allenss-amazon:
C++ static initializers that allocate memory aren't really handled by the Redis loadable module machinery. The problem is that the static initializers are executed BEFORE the call to the `...Module_OnLoad` which provides the address of `...Module_Allocate`. Nor does the Valkey core export it's own internal `malloc` function. Thus during the execution of the static initializers some mechanism must be provided to perform memory allocations.

The current codebase has a fairly complicated mechanism that uses a private version of malloc (private-malloc) and maintains a separate hash set to track allocations done using the private-malloc so that subsequent free operations can be directed to the correct allocator. This incurs additional overhead for every memory free operation as the tracking machinery is globally locked to perform a hash table lookup and occasionally removal.

Two alternative solutions to this overhead have been suggested:

1. Have the Valkey core export `zmalloc`, `zfree`, etc. so that the `dlopen` machinery can link those symbols, allowing the static initializers to use those functions directly. 

2. Another solution that doesn't require core changes is to use the linker's `wrap` facility to swap the compiler provided `_init` function for a our own no-op function. Now `dlopen` will invoke our wrapped, empty, initialization function. Later when the core invokes `...Module_OnLoad` the code can invoke the renamed `_init` after the `...Module_Alloc` function pointer is known.
3.

allenss-amazon:
Fixed in 1.3

PR #1376, opened 2026-09-10
Sync board priority with PR P1/P2/P3 labels (both directions)

Aksha1812:
## What

Keeps the triage board's priority and the PR's `P1`/`P2`/`P3` labels consistent, in **both directions**. Set priority with `/priority <#> P2` and the label appears on the PR; add/change the label on the PR and the board follows. 

## How

`reconcilePriority()` does a per-PR three-way merge against the value the two were last synced to (`state.prioritySynced[n]`):

- **label changed** since last sync → adopt the label into the board
- **board changed** since last sync → write the board value onto the PR's labels
- **both changed** (a genuine conflict) → **label wins** — the documented source of truth, since labels are visible in the PR list and in GitHub search

Notes:

- `P4` / unset maps to "no label" (the repo has no `P4` label), so a P4 board entry is represented by the absence of a priority label and is never clobbered by label edits.
- Label writes happen **after** the issue body is persisted, so a failed label API call can't desync — `prioritySynced` already records the intended value and the next run recomputes the same op.
- Multiple priority labels on one PR resolve to the strongest (lowest number).
- A cross-repo read (`TARGET_REPO`, e.g. the fork demo) mirrors **label → board only** and never writes back to the other repo.

Bumps the workflow permission `pull-requests: read` → `write` so the Action can edit labels.

## Testing

Offline harness exercising `reconcilePriority` across 34 assertions: agreement, bootstrap (label-present vs board-only), one-sided changes, the both-changed conflict, label removal, P4 preservation, multi-label, the cross-repo mirror path, the failed/partial-write retry-until-converged path, and exact-label-set convergence (a stray extra label keeps retrying instead of reading as done) — all pass.

### Live end-to-end (on a fork)

Ran the workflow against a fork's own issues and PRs — board issue [Aksha1812/valkey-search#32](https://github.com/Aksha1812/valkey-search/issues/32) set as `DASHBOARD_ISSUE`, `P1`/`P2`/`P3` labels created on the fork — exercising both directions and the full convergence lifecycle:

| Step | Setup | Result |
|---|---|---|
| **label → board** | [PR #30](https://github.com/Aksha1812/valkey-search/pull/30) had a `P1` label, board unset | Board adopted it (`priority=1`), no label write needed |
| **board → label** | board priority `P2` for [PR #31](https://github.com/Aksha1812/valkey-search/pull/31), no priority label | Run logged `Synced priority labels on 1 PR(s)`; PR #31 gained the `P2` label |
| **convergence** | next run, after the write | Pending push cleared, synced value recorded (`P2`) |
| **idempotent** | run again with nothing changed | No label writes, no duplicate labels — the sync goes quiet |

The labels are still visible on those PRs ([#30 → `P1`](https://github.com/Aksha1812/valkey-search/pull/30), [#31 → `P2`](https://github.com/Aksha1812/valkey-search/pull/31)). Confirms both sync directions work against real GitHub PRs/issues and that the pending-target retry holds the board's intent until the labels actually converge, then stops.


.github/scripts/reviewer-dashboard.js
```diff
@@ -356,6 +361,93 @@ function priorityOf(state, n) {
...
+      if (Bl === P) {
+        if (Ln === P) {                    // labels reached the target → done
+          delete state.priorityPending[n];
+          state.prioritySynced[n] = Bl;
+        } else {                           // write failed / partial → retry from current labels
+          pushOp(n, Bl, pr.priorityLabels);
+        }
+        continue;
+      }
+      delete state.priorityPending[n];     // board moved off the target → push superseded
```
Frank-Gu-81: Claude found this potential issue and it looks valid to me. Worth taking a look. Other changes looks good to me. 

Could this lose a newer board update? For example: P1 is synced, /priority # P2 successfully writes the P2 label but remains pending, then /priority # P3 arrives before the next run. We delete pending here, compare board=P3 and label=P2 against synced=P1, and “label wins,” reverting the new command to P2. I reproduced this with reconcilePriority().

Can we persist successful label writes as synchronized immediately, or otherwise recognize that the observed P2 label is the completed pending write before reconciling P3? Please add a regression test for P1 → P2 → P3 across consecutive runs.

Aksha1812: done https://github.com/valkey-io/valkey-search/pull/1376/commits/68666eb7e9963de43411647e50936b63d07f4d79

Frank-Gu-81 (approved):

.github/scripts/reviewer-dashboard.js
Frank-Gu-81: Thanks. LGTM.

PR #1379, opened 2026-09-11
Enforce command arity at registration via SetCommandInfo

chinguyen21:
Move the arity checks for FT.DROPINDEX, FT.INFO and FT._LIST out of the command handlers and into command registration, so the engine rejects wrong-arity calls before dispatch instead of each handler re-checking argc.

Add an optional `arity` field to vmsdk CommandOptions (positive N = exactly N args including the command name, -N = at least N, 0 = no enforcement) and have RegisterCommands publish it through ValkeyModule_SetCommandInfo. Set the arity for FT.DROPINDEX (2), FT.INFO (-2) and FT._LIST (1), and drop the now redundant in-handler argc checks and their unit cases.

Add an integration test asserting the engine rejects wrong-arity calls for these commands.

### DIFF
--- a/integration/test_command_arity.py
+++ b/integration/test_command_arity.py
@@ -1,0 +1,26 @@
+"""Command arity is enforced by the engine (via SetCommandInfo at
+registration), so a wrong-arity call is rejected before the handler runs."""
+
+import pytest
+import valkey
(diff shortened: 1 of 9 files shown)


BCathcart:
One thing to check is if there are any changes in the error messages customers see

PR #1383, opened 2026-09-11
Install gcc14 static libstdc++ for benchmark builds due to #1360

Frank-Gu-81:
## Summary

Fix the benchmark runner dependency setup for builds that link `libsearch` with `-static-libstdc++`.

#1360 made static libstdc++ a build requirement for non-sanitized module builds. The benchmark workflow uses `/usr/local/bin/g++` -> GCC 14 on AL2023, but some runners had `gcc14` and `gcc14-c++` installed without `gcc14-libstdc++-static`, which caused the link step to fail with:

```
/usr/bin/ld: cannot find -lstdc++
```

  ## Changes

  - install `gcc14-libstdc++-static` in `ci/install_deps.sh`

  ## Why

  This makes benchmark hosts provision the static libstdc++ archive required by the current module link flags, instead of relying on manual host state.

### DIFF
--- a/ci/install_deps.sh
+++ b/ci/install_deps.sh
@@ -67,7 +67,7 @@ function ensure_gcc_amazon_linux() {
         LOG_INFO "✓ GCC ${current_version} already available at /usr/local/bin/gcc (>= ${GCC_MIN_VERSION})"
     else
         LOG_INFO "GCC >= ${GCC_MIN_VERSION} not found (current: ${current_version}) — installing GCC 14 from AL2023 repos..."
-        sudo dnf install -y gcc14 gcc14-c++
+        sudo dnf install -y gcc14 gcc14-c++ gcc14-libstdc++-static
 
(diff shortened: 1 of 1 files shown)


Frank-Gu-81:
Test run: https://github.com/Frank-Gu-81/valkey-search/actions/runs/34647769879

neerajr0:
Thanks for the fix. Is `libstdc++` needed for the actual benchmark workflow? If it isn't, then the workflow is unnecessarily failing for a dependency it doesn't need. Is there any way to be more resilient to this?

Frank-Gu-81:
@neerajr0 Yes, indirectly. The benchmark itself doesn't use `libstdc++.c`, but it needs to build `libsearch.so` first, which now requires it.

neerajr0 (approved):

allenss-amazon (approved):

PR #1389, opened 2026-09-14
Bug fix: only last RETURN in SEARCH should be effective

JasonLi314:
fixing #1353 item 7 when multiple `RETURN` in `FT.SEARCH` gets their command combined instead of taking last one as effective. This issue is extra apparent when `RETURN 0` doesn't take affect.


integration/compatibility/__init__.py
```diff
@@ -7,30 +7,39 @@
...
+# per-command subdirectory (e.g. search/); all paths are relative to this
+# directory and ANSWER_FILE_NAME must match the "answers" entry.
 GENERATORS = [
     {"generator": "generate.py",         "answers": "aggregate-answers.pickle.gz",   "cluster": True},
     {"generator": "generate_text.py",    "answers": "text-search-answers.pickle.gz", "cluster": False},
     {"generator": "generate_array.py",   "answers": "array-input-answers.pickle.gz",  "cluster": False},
     {"generator": "generate_expr.py",    "answers": "expr-answers.pickle.gz",         "cluster": False},
     {"generator": "generate_sortkey.py", "answers": "sortkey-answers.pickle.gz",      "cluster": False},
     {"generator": "generate_filter.py",  "answers": "filter-answers.pickle.gz",       "cluster": False},
+    {"generator": "search/generate_return.py", "answers": "search/return-answers.pickle.gz", "cluster": False},
```
mohanrajendran: I don't think we need a new `search/` namespace for this generator yet.  We can continue keeping this as a flat list for now.

integration/compatibility/generate_sortkey.py
```diff
@@ -21,7 +21,6 @@ class TestSortKeyPrefixCompatibility(BaseCompatibilityTest):
 
     def test_withsortkeys_prefix_by_field_type(self, key_type):
         self.setup_data(SORTKEY_PREFIX_DATA_SET, key_type)
-        time.sleep(0.5)  # tiny data set; give indexing a moment
```
mohanrajendran, line 24: Any reason to remove this? Indexing process is asynchronous and we should wait for it to be completed with sufficient time to not introduce flakiness.

integration/test_non_vector.py
```diff
@@ -1119,6 +1119,56 @@ def test_sortkey_prefix_gate(self):
...
+    """
+
+    def test_return_clause_gate(self):
+        client: Valkey = self.server.get_new_client()
+        assert client.execute_command(
+            "FT.CREATE", "rcg_idx", "ON", "HASH", "PREFIX", "1", "rcg:",
+            "SCHEMA", "m", "TAG", "p", "NUMERIC", "title", "TEXT") == b"OK"
+        assert client.execute_command(
+            "HSET", "rcg:1", "m", "all", "p", "10",
+            "title", "hello world") == 3
```
mohanrajendran, line 1138: Please add the indexing barrier after the HSET:

```suggestion
            "HSET", "rcg:1", "m", "all", "p", "10",
            "title", "hello world") == 3
        IndexingTestHelper.wait_for_indexing_complete_on_node(client, "rcg_idx")
```

integration/test_non_vector.py
```diff
@@ -1119,6 +1119,56 @@ def test_sortkey_prefix_gate(self):
...
+class TestReturnClauseGate(ValkeySearchTestCaseDebugMode):
+    """
+        Repeated RETURN clauses are last-one-wins (issue #1353 item 7),
+        gated on search.emulate-release: pre-1.3.0 a `RETURN 0` latched
+        no-content for the whole command and repeated clauses accumulated
+        fields. The NOCONTENT keyword is sticky on both sides of the gate.
+        debug-mode is required to set emulate-release at the module version.
+    """
+
+    def test_return_clause_gate(self):
```
mohanrajendran: `--run-integration-tests=divergence` is passed to pytest as `-k divergence`, so this test is currently filtered out. Renaming the test makes the new gate coverage run under that command; the compatibility-fixture replay still needs separate selection/wiring.

```suggestion
    def test_divergence_return_clause_gate(self):
```

integration/compatibility/__init__.py
JasonLi314: the file structure is becoming difficult to read. every single tiny new scenarioes like sortkey or filter have their own file. It's going to blow up eventually so better to fix it earlier than later.

integration/compatibility/generate_sortkey.py
JasonLi314: good catch though i need to point out this is limited to redis version >= 8.8 only with `search-workers > 0`. 

For valkey all write response are blocked until index is written so from a single client's perspective it's synchronous [per official documentation](https://valkey.io/topics/search/#:~:text=The%20client%20which%20executed%20the%20mutation%20command%20is%20blocked%20until%20all%20indexes%20are%20updated.). 

Overall we only have this issue for redis oracle runs which include these generate.py files. I will retain this time.sleep in generate_*.py files and indexing barrier for valkey integ tests purely for conventional purpose, but add a todo comment to remove all valkey indexing barrier.

integration/test_non_vector.py
JasonLi314: will fix

integration/test_non_vector.py
JasonLi314: will add barrier purely for conventional sake with todo comment to remove all of these barrier in the future as commented above.

integration/compatibility/__init__.py
mohanrajendran: Agreed, lets fix it in a follow-up PR.

JasonLi314: ok i will move new test out of the folder to maintain current convention. will do the restructuring cleanly in a separate PR.

mohanrajendran (approved):

BCathcart (approved):

PR #1394, opened 2026-09-15
Add WITHCURSOR to FT.AGGREGATE and FT.SEARCH, and the FT.CURSOR command

### LINKED ISSUE #900: [FEATURE] Support FT.CURSOR

**Additional information**

allenss-amazon:
Adds cursors to valkey-search: `FT.AGGREGATE`, `FT.HYBRID` and `FT.SEARCH` accept `WITHCURSOR`, which returns the first rows of the result and saves the rest server-side, and the new `FT.CURSOR READ` / `FT.CURSOR DEL` command pages through or releases them. The syntax and the `FT.AGGREGATE` / `FT.CURSOR` replies match Redis. The design is described in [`docs/design-notes/cursor.md`](docs/design-notes/cursor.md).

## Summary

**Syntax.** `WITHCURSOR [COUNT <count>] [MAXIDLE <ms>]` is accepted anywhere after the query string of `FT.AGGREGATE` (before, between or after pipeline stages) and `FT.SEARCH`; if it is repeated the last one wins. `COUNT` defaults to 1000 and `MAXIDLE` to 300000. Both must be at least 1 and at most a configurable maximum (`search.cursor-max-count`, default 100000; `search.cursor-max-idle-ms`, default unlimited); out of range values are an error.

**Replies.**
- `FT.AGGREGATE ... WITHCURSOR` returns `[[count, row...], cursor_id]`, the normal aggregate reply plus a continue cursor.
- `FT.HYBRID ... WITHCURSOR` returns the same `[[count, row...], cursor_id]` as `FT.AGGREGATE`: FT.HYBRID feeds its fused results through the FT.AGGREGATE pipeline, so it shares the cursor machinery. `WITHCURSOR` is parsed by the aggregate suffix parser FT.HYBRID already uses, and ends a `SEARCH`, `VSIM` or `COMBINE` clause the way `LOAD` does.
- `FT.SEARCH ... WITHCURSOR` (a Valkey extension) returns `[total_matches, [row...], cursor_id]`, where each row is an array holding the elements of one row of a non-cursor reply. The cursor pages through the rows the query would otherwise return (the `LIMIT` window, or the first k KNN results).
- `FT.CURSOR READ <index> <cursor_id> [COUNT <count>]` returns `[[count, row...], cursor_id]` for both kinds of cursor. `count = min(rows remaining, COUNT)`; Without `COUNT` the cursor's current read size is used, which starts as the `COUNT` of the `WITHCURSOR` clause that created it; a `COUNT` on a read replaces that read size for later reads. Both match Redis, measured over 3000 documents. `COUNT` is bounded by `search.cursor-max-count`.
- A cursor id of 0 means all rows have been returned; no cursor object exists (or it has just been destroyed). `FT.CURSOR DEL <index> <cursor_id>` returns OK; it applies the same index and database checks as `READ`, and a cursor whose index has been dropped can still be deleted.

**Cursor object.** When rows remain after the first reply, the command's `QueryCommand` (parameters, search results and, for aggregate, the output `RecordSet`) is handed to a `Cursor` object. The cursor is therefore a snapshot of the query output: later changes to the data are not visible, unlike Redis, where a cursor that is not fully materialized evaluates rows at read time (see `known_differences.md`). The cursor does not keep its index alive. Cursor replies reuse the command's own reply code, which for `FT.SEARCH` is now a set of `SearchCommand` members that work per row.

**Global cursor table.** `CursorTable` is indexed three ways: by id (hash map, O(1)), by destruction timestamp (`std::multimap`, O(log N), whose iterators stay valid across other inserts and erases), and by (database, index name). The third index lets index removal discard exactly that index's cursors from the SchemaManager's single removal choke point (`FT.DROPINDEX`, `FLUSHDB`/`FLUSHALL`, replica full sync, RDB load, coordinator metadata drops); a later read then replies `Cursor not found`. The per-database entry is held by pointer, so `SWAPDB` moves an index's cursors along with it in O(1). A new slot in the cron callback destroys cursors whose idle time has expired; the table entry is removed in the cron task and the cursor object is destroyed on a utility thread. Cursors are also cleared at shutdown. The App INFO field `search_num_cursors` reports the number of cursors, and `FT._DEBUG SHOW_CURSORS` lists each cursor's id and milliseconds until expiration.

**Cursor ids.** A cursor id is an unsigned 64 bit integer: the upper half is a global 31 bit counter, incremented per cursor and wrapping at 2^31 (keeping ids positive as RESP integers); the lower half is the CRC-32 of the server's `run_id`, computed once at module load (a new `vmsdk::Crc32`, zlib compatible). An id of 0, or one already in the table, is skipped.

**FT.CURSOR validation.** Reading or deleting a cursor requires the same key-prefix permissions as querying its index, so a user who could not run the originating query cannot read or release the cursor either. The named index must exist in the connection's current database but, as in Redis, need not be the index the cursor was created on. The cursor must belong to the connection's current database (valkey-search indexes are per database; Redis only supports indexes in db 0). If the cursor's own index has been dropped, and possibly recreated, `READ` destroys the cursor and replies `The index was dropped while the cursor was idle`, matching Redis; `DEL` of such a cursor succeeds. `READ` resets the cursor's idle timer. Error text matches Redis (`Cursor not found, id: <id>`, `Cursor does not exist`).

**OOM.** `FT.AGGREGATE` and `FT.SEARCH` are already `deny-oom`, so a query requesting a cursor is rejected while OOM. `FT.CURSOR` is not, so existing cursors can still be read and released.

**Timeouts and partial results.** The partial results setting has no effect on a query that requests a cursor: a timed-out query hands back the rows it gathered rather than an error, and the `FT.AGGREGATE` pipeline still runs to completion over them. An open cursor also no longer counts as an in-flight query (`search_async_queries_in_flight`), since its query is over and only its output is still held.

**Cluster mode.** The query fans out as usual; the cursor lives on the node that ran it, and `FT.CURSOR` must be sent to that node.

## Testing
- Unit tests: `testing/ft_cursor_test.cc` (FT.CURSOR syntax, READ to exhaustion, DEL, other index name, wrong db, dropped/recreated index, expiration of 0, 1 and 2 cursors in one call, touch, id generation incl. skipping 0, SHOW_CURSORS), WITHCURSOR parser cases for both commands, and FT.SEARCH cursor reply tests. `vmsdk::Crc32` is checked against zlib values.
- Integration tests: `integration/test_cursor.py`, including a power-set content test that pages a cursor and compares the rows against the same query without `WITHCURSOR`, over HASH/JSON x vector/non-vector x FT.AGGREGATE/FT.SEARCH/FT.SEARCH NOCONTENT x inside/outside MULTI x normal/timed-out termination (48 standalone cases, 12 cluster cases), plus ACL rejection, the READ default COUNT, eager discard, SWAPDB, and in-flight accounting (aggregate and search paging, row shapes vs. non-cursor replies, LIMIT window, placement of WITHCURSOR, DEL, other index / db, dropped index, MAXIDLE expiry, config limits, MULTI, OOM, and a cluster test).
- Full unit suite and full integration suite pass (626 passed, 8 skipped), including the compatibility replay, which confirms non-cursor replies are unchanged.

https://claude.ai/code/session_01KVMH2WSMYEWkWMySK2zeH6


src/cursor.cc
```diff
@@ -0,0 +1,218 @@
...
+absl::StatusOr<int64_t> ParseCursorCount(vmsdk::ArgsIterator &itr) {
+  return ParseBoundedValue(itr, "COUNT",
+                           options::GetCursorMaxCount().GetValue());
+}
+
+absl::StatusOr<CursorOptions> ParseCursorOptions(vmsdk::ArgsIterator &itr) {
+  CursorOptions cursor_options;
+  const int64_t max_idle_ms = options::GetCursorMaxIdleMs().GetValue();
+  cursor_options.max_idle =
+      absl::Milliseconds(std::min(kDefaultCursorMaxIdleMs, max_idle_ms));
```
chinguyen21, line 52: **`WITHCURSOR` without a `COUNT` returns more rows than `search.cursor-max-count` allows, and creates no cursor at all.**

This line clamps the *default* `MAXIDLE` to the configured maximum. The default `COUNT` gets no
equivalent: `CursorOptions::count` keeps its in-class initializer of 1000 unless an explicit
`COUNT` clause runs `ParseCursorCount`, which is the only place the configured maximum is
enforced. So the cap binds exactly when the client names a value, and not when it omits one.

*Repro:* 5 documents in the index:

```
CONFIG SET search.cursor-max-count 3
FT.AGGREGATE idx * WITHCURSOR COUNT 3   -> [[3, row, row, row], 4987770739]
FT.AGGREGATE idx * WITHCURSOR           -> [[5, row x5], 0]
FT.SEARCH    idx * WITHCURSOR           -> [5, [row x5], 0]
```

With an explicit `COUNT` at the cap the mechanism works: three rows, cursor stays open. With no
`COUNT`, both commands return every row in a single reply and never create a cursor, so the
`3` that was just configured has no effect. The larger the result set, the larger the reply that
the cap was set to prevent.

src/valkey_search_options.cc
```diff
@@ -74,6 +74,26 @@ static auto query_string_bytes =
...
+                          1,                      // min
+                          INT64_MAX)              // max
+        .Build();
+
+/// Register the "--cursor-max-idle-ms" flag. The largest MAXIDLE accepted by
+/// WITHCURSOR.
+constexpr absl::string_view kCursorMaxIdleMsConfig{"cursor-max-idle-ms"};
+static auto cursor_max_idle_ms =
+    config::NumberBuilder(kCursorMaxIdleMsConfig,  // name
+                          INT64_MAX,               // default
```
chinguyen21: **Suggestion: a read-only client can hold an unbounded number of query snapshots for an unbounded time.**

`MAXIDLE` defaults to 300000 ms, but its maximum here is `INT64_MAX`, so a client may name an
arbitrarily distant expiration. There is also no configuration limiting how many cursors exist at
once. Each cursor holds a full snapshot of its query's output — for `FT.AGGREGATE` the
`RecordSet`, for `FT.SEARCH` the neighbor list — so `WITHCURSOR COUNT 1 MAXIDLE <large>` in a loop
accumulates whole result sets that nothing reclaims.

This is bounded, and that is worth stating precisely: module allocations route through
`ValkeyModule_Alloc`, so cursor memory counts toward `maxmemory`, and `FT.AGGREGATE` / `FT.SEARCH`
carry `denyoom`, so new cursor-creating queries are rejected once the limit is reached. The
reachable end state is therefore not unbounded growth — it is a server parked at `maxmemory` with
writes failing, reached by a client holding only read permissions, and persisting until an
operator runs `FT._DEBUG SHOW_CURSORS` and deletes them by hand. `FT.CURSOR READ` and `DEL` keep
working in that state, which is good, but nothing sheds the load on its own.

*Suggestion:* a cap on live cursors (globally or per index), a finite default for
`search.cursor-max-idle-ms`, or both. Redis bounds this with a per-index cursor limit and a finite
idle default, and `known_differences.md` would be the natural place to record whatever is chosen
if it differs.

Not a merge blocker — reaching it needs a deliberately hostile client, and the idle sweeper plus
`denyoom` keep it from being silent. Raising it because the defaults are the part that is hard to
change later.

src/commands/ft.create.json
```diff
@@ -356,6 +356,7 @@
         ]
       }
     ],
+    "arity": -5,
```
chinguyen21: *Nit:* this `arity` addition looks unrelated to cursors.

`-5` is the right value — `FT.CREATE idx SCHEMA f TAG` is the shortest valid form — and the field
being absent before was presumably a real gap. It just is not mentioned in the PR description, and
a reader diffing this file for cursor changes will stop on it.

Worth either a line in the description or its own commit, so that whoever bisects a future
`FT.CREATE` arity change finds a reason rather than a cursor PR.

src/commands/ft_cursor.cc
```diff
@@ -0,0 +1,116 @@
...
+
+  // As in Redis, the named index must exist but need not be the one the
+  // cursor was created on.
+  const int db_num = ValkeyModule_GetSelectedDb(ctx);
+  VMSDK_RETURN_IF_ERROR(
+      SchemaManager::Instance().GetIndexSchema(db_num, index_name).status());
+
+  auto &table = CursorTable::Instance();
+  Cursor *cursor = table.Lookup(id);
+  if (cursor == nullptr || table.GetDbNum(id) != db_num) {
```
chinguyen21: *Nit:* this check tells a caller whether a cursor id exists before the ACL check runs.

A user without the cursor index's key prefixes gets `Cursor not found, id: <id>` for an id that
does not exist and the permission error for one that does, so the reply distinguishes the two.

Part of that ordering is inherent — the ACL check below applies to the cursor's *own* index, which
is only known after the lookup. But the index the caller *named* is already known above, and
`FT.SEARCH` / `FT.AGGREGATE` would have prefix-checked it. Applying the same check to
`index_name` before the lookup would close the difference without changing any of the logic that
follows.

Genuinely minor: ids are a counter plus a per-run constant rather than a secret, and this path is
already stricter than Redis, which applies no prefix check to `FT.CURSOR` at all. Noting it
because the check that exists reads as though it is meant to be authoritative.

chinguyen21 (approved):
LGTM, just one non-blocking comment

PR #1397, opened 2026-09-15
Bug fix: added deterministic sorting for sortby

### LINKED ISSUE #1353: [BUG] FT.SEARCH Compatibility Issues

JasonLi314:
Fixes #1353 item 8.

FT.SEARCH SORTBY ties (documents with equal sort values) have no tie-break, so their relative order depends on incidental retrieval order rather than the data. Two consequences: repeating the same query can return ties in a different order, and because the full-result path uses std::stable_sort while the LIMIT path uses std::partial_sort, a LIMITed reply is not a prefix of the unlimited one — paginating over ties yields duplicate or missing rows.

This adds a direction-following tie-break on the document key, making the comparator a total order so the result is a pure function of the data and query, and a LIMITed reply is a true prefix of the unlimited one.

Not a Redis-compatibility fix: RediSearch breaks ties differently. This is a determinism improvement, so it is ungated.


src/commands/ft_search.cc
```diff
@@ -314,26 +327,35 @@ void ApplySorting(std::vector<indexes::Neighbor> &neighbors,
...
-    auto it_b = b.attribute_contents->find(sortby.field);
+    auto get_sortkey_val = [&](const indexes::Neighbor &neighbor)
+        -> std::optional<absl::string_view> {
+      if (!neighbor.attribute_contents) {
+        return std::nullopt;
+      }
+      auto target_field_it = neighbor.attribute_contents->find(sortby.field);
+      if (target_field_it == neighbor.attribute_contents->end()) {
+        return std::nullopt;
+      }
```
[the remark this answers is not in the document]
JasonLi314: good catch. fixed in commit below.

src/commands/ft_search.cc
```diff
@@ -306,6 +307,18 @@ void ApplySorting(std::vector<indexes::Neighbor> &neighbors,
   bool is_numeric =
       index_result.ok() &&
       index_result.value()->GetIndexerType() == indexes::IndexerType::kNumeric;
+  // Tied neighbors order by key, following the sort direction, so the result
```
Aksha1812, line 310: nit: lets trim AI generated comments.

src/query/search.cc
```diff
@@ -512,6 +512,16 @@ absl::StatusOr<std::vector<indexes::Neighbor>> MaybeAddIndexedContent(
...
   }
+  // A SORTBY on a stored field is compared against attribute_contents, and
+  // only the main-thread fetch adds that field beyond RETURN (GetContent).
+  // Index-served content carries just the RETURN attributes, so decline.
+  const bool sort_by_vec_score =
+      parameters.sortby_parameter && parameters.score_as &&
+      parameters.sortby_parameter->field ==
+          vmsdk::ToStringView(parameters.score_as.get());
+  if (parameters.sortby_parameter && !sort_by_vec_score) {
+    return results;
```
Aksha1812, line 523: Fix looks correct . but this might cause performance to degrade for sort by queries and would force a main-thread GetContent per neighbor where it used to be fully index-served. Wondering if adding the sortby fields to the attributes list is the more efficient choice here.

src/commands/ft_search.cc
JasonLi314: done

src/query/search.cc
JasonLi314: good catch. done

src/query/search.cc
```diff
@@ -513,11 +513,44 @@ absl::StatusOr<std::vector<indexes::Neighbor>> MaybeAddIndexedContent(
...
+  const bool sort_by_vec_score =
+      parameters.sortby_parameter.has_value() && parameters.score_as &&
+      parameters.sortby_parameter->field ==
+          vmsdk::ToStringView(parameters.score_as.get());
+  if (parameters.sortby_parameter.has_value() && !sort_by_vec_score) {
+    const std::string &sortby_field = parameters.sortby_parameter->field;
+    const bool sortby_served_by_return = std::any_of(
+        parameters.return_attributes.begin(),
+        parameters.return_attributes.end(), [&](const ReturnAttribute &attr) {
+          return attr.identifier.get() &&
```
Aksha1812, line 536: `attr.identifier` is the resolved hash field name but `sortby_field` is the schema alias, these only match when the schema doesn't use `AS`, which is why the tests pass.

Try `SCHEMA a AS b NUMERIC b AS c NUMERIC` with `SORTBY b RETURN 1 c`: the check matches on the wrong field, the sortby entry gets skipped, and the comparator ends up sorting by `c`'s values. Wrong order, no error. Heads up that `attr.attribute_alias` isn't the fix either, that breaks `price_raw AS p` with `SORTBY p RETURN 1 p`. Easiest is probably to drop the dedup and always push the sortby entry, since `emplace` makes the duplicate a no-op. you can verify by adding a test or running some queries.

.vscode/settings.json
```diff
@@ -18,5 +18,6 @@
     ],
     "clangd.inactiveRegions.useBackgroundHighlight": true,
     "clangd.enableCodeCompletion": true,
-    "remote.SSH.enableAgentForwarding": false
-}
+    "remote.SSH.enableAgentForwarding": false,
+    "editor.tabSize": 2
```
Aksha1812: do we need this ?  or this got through from your local development setup

src/query/search.cc
```diff
@@ -513,11 +513,44 @@ absl::StatusOr<std::vector<indexes::Neighbor>> MaybeAddIndexedContent(
...
+
+  // collecting attributes required for both RETURN and SORTBY
+  // SORTBY is required for sorting later
   struct AttributeInfo {
-    const ReturnAttribute *attribute;
+    enum class Type { ReturnAttribute, SortByParameter };
+    Type type;
+    union {
+      const ReturnAttribute *return_attribute;
+      const SortByParameter *sortby_param;
```
Aksha1812: wondering if directly storing the attribute _contents key here would be better rather than pulling them later  https://github.com/valkey-io/valkey-search/blob/f3a74c14d95928d3b842678ec316706ac8c79417/src/query/search.cc#L646-L670 as done in the PR , if we are not using return_attribute and sortby_param individually  . that would make the implementation more cleaner

.vscode/settings.json
JasonLi314: yes, all cpp files have tab size 2 yet there are no configs for it. I've added scope so only cpp files are affected but let me know if I'm missing anything here.

src/query/search.cc
JasonLi314: fixed

src/query/search.cc
JasonLi314: excellent catch though this issue is much broader and already exists before current PR fix. It affects both knn and non-vector sorting and diverges with redis behavior. I put details in issue here #1440

src/query/search.cc
```diff
@@ -513,11 +513,38 @@ absl::StatusOr<std::vector<indexes::Neighbor>> MaybeAddIndexedContent(
...
+          return attr.identifier.get() &&
+                 vmsdk::ToStringView(attr.identifier.get()) == sortby_field;
+        });
+    if (!sortby_served_by_return) {
+      auto index = parameters.index_schema->GetIndex(sortby_field);
+      if (!index.ok()) {
+        return results;
+      }
+      attributes.push_back(AttributeInfo{.identifier = sortby_field,
+                                         .index = index.value().get()});
```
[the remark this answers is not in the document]
JasonLi314: this issue is already documented in #1440 and is out of scope for this PR. This broken behavior already exists in mainline, not introduced by this PR.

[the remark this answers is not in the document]
JasonLi314: The skip prevents exactly this. With a AS p, p AS q and SORTBY p RETURN 1 q, RETURN q's resolved identifier is the hash field p, which equals the SORTBY token, so sortby_served_by_return is true and the SORTBY entry is never inserted. The bag holds only the RETURN entry (q's own value under key p), and the reply serves q's value. The double-insert you describe can occur only if the check is removed, which is why the check exists. Verified empirically on this branch with the structurally identical schema (a AS b NUMERIC b AS c NUMERIC, SORTBY b RETURN 1 c): RETURN values are correct on both the KNN and non-vector paths (matching Redis); only the sort order diverges, which is the pre-existing #1440.

JasonLi314:
Opened 3 new issues that were discovered in this PR but does not fit in PR's scope:
#1440
#1441 
#1442

Issue #1387, opened 2026-09-13
[TEST] Scoring Compatibility


BCathcart:
Currently there are no scoring-specific tests in the compatibility framework.

1. The compatibility framework has been updated to test against redis:latest which is 8.10. The main initial chunk of scoring behaviour introduced in #1287 was tested manually against 8.6. I've proposed a version-compatibility convention in #1386 which ties a Valkey Search version to a Redis major version. Ideally testing against 8.6 vs 8.10 makes no difference since there shouldn't be breaking changes. If we do find there are breaking changes (I think @boda26 already may have seen some), then we may need to tighten the mapping to more specific Redis versions. In any case I think it's a necessary exercise to test scoring against 8.10.
      - Note: It's a little awkward that the default for existing features in 1.3 is compatibility with RedisSearch 2 and for new features it will be Redis 8 (i.e. we're split across Redis major versions). Until users actually report this as a pain point though I don't think it would be worth implementing RedisSearch 2 compatible scoring behaviour and gating breaking Redis 8 scoring changes until the next Valkey Search major release or explicit opt-in.
2. Claude called out some FT.AGGREGATE ADDSCORES compatibility issues: https://github.com/valkey-io/valkey-search/pull/1358/changes#r4001024716.

BCathcart:
Another general comment I have on compatibility testing I'll pull here for visibility (https://github.com/valkey-io/valkey-search/pull/1364#discussion_r3992513444): 
> Just an observation not specific to this PR: moving the tests to the compatibility framework actually gives less visibility to the reader about what we're asserting. You can't really read the tests to get an understanding of the code. This might also be why I've seen developers prefer to write regular integration tests instead of compatibility tests. It's a little less intuitive. It could be worth having the expected results made visible somewhere. This could be auto generated

I've seen a pattern of developers preferring to write regular integration tests and manually verify compatibility in ad-hoc ways. We should improve the compatibility framework so that's it's the intuitive default approach.

PR #1400, opened 2026-09-17
Various small scoring changes

BCathcart:
Pulling out some changes from existing PRs that attempted a new locking strategy. I think we need to pivot now (https://github.com/valkey-io/valkey-search/pull/1358#issuecomment-5705630126) but these changes are still relevant.

  - Skips unnecessary main-thread revalidation/recompute for no-content FT.AGGREGATE queries. This optimization is independent of scoring.
 - Implements FT.AGGREGATE ADDSCORES, exposing each document's relevance score as __score in the aggregation pipeline
  - Adds main thread scoring-recompute integration tests
  - Some small renaming and refactoring

Note there are some suspected compatibility issues for ADDSCORES which are tracked in #1387.


zackcam (approved):

src/query/search.cc
```diff
@@ -439,9 +439,20 @@ void EvaluatePrefilteredKeys(
   if (needs_dedup) {
     result_keys.reserve(max_keys);
   }
+  // Same gate as the non-vector drain loop: a solved query's fetchers are exact
+  // rather than candidate generators, so the per-key re-check only repeats work
+  // the fetcher already did -- for text that means a prefix or suffix radix
+  // walk of up to max-term-expansions words, or a stem variant lookup, per
+  // candidate. Only AND with numeric/tag (which keeps just the smallest child's
+  // fetchers) and negation (which scans the universal set) leave work for the
+  // evaluator.
```
BCathcart:
```suggestion
  // Skip per-key predicate re-evaluation when the query is fully solved by the
  // entries fetchers and only yields valid keys. The non-vector path only
  // reaches here for unsolved queries but this check benefits the hybrid
  // pre-filter path. Note we don't score during this drain on purpose because
  // the vast majority of keys are expected to be filtered out or miss the
  // final KNN top-k.
```

BCathcart:
Conditions for the greptile concern:
>  - They have a document/schema field literally named __score — a name that already looks reserved and that Valkey claims under ADDSCORES
>  - They run FT.AGGREGATE (not FT.SEARCH — its LOAD path isn't involved) on a non-vector query
 > - They reference that field by name via LOAD, LOAD *, or a pipeline stage
  >- And they don't use ADDSCORES — with ADDSCORES the shadowing is the documented Redis behavior anyway

Seems like quite the edge case. I added it as part of the Scoring Compatibility follow-up task though.

BCathcart:
Follow-up: Better unify the non-vector and hybrid code paths. Now `EvaluatePrefilteredKeys()` solves both de-duplication and skipping evaluation when the entries fetcher solves the query. `DoSearchNonVector()` also checks `IsUnsolvedQuery()` and if it is solved, performs de-duplication. Essentially we have a path for if it's non-vector and solved and a shared path (`EvaluatePrefilteredKeys()`) if it's vector and solved or hybrid pre-filter path. We probably don't need a totally separate path for if it's non-vector and solved.

src/query/search.cc
BCathcart, line 1384:
```suggestion
  // TODO: Unify code path to rely on de-dup and unsolved checking
  // in EvaluatePrefilteredKeys
  if (!requires_prefilter_evaluation) {
```

integration/test_scoring_cluster.py
```diff
@@ -271,11 +289,178 @@ def test_hybrid_text_vector(self):
...
+        res = self._run_with_mutation_in_flight(
+            coordinator, key,
+            "FT.SEARCH", IDX_NAME, "@t:{red}", "LIMIT", "0", "100",
+            "WITHSCORES",
+        )
+        assert self._counter(coordinator, "search_predicate_revalidation") == 1
+        pairs = _pairs(res)
+        assert sorted(k for k, _ in pairs) == sorted(DOCS)
+        scores1 = [s for _, s in pairs]
+        assert all(s > 0 for s in scores1)
```
BCathcart, line 411:
```suggestion
        scores1 = [s for _, s in pairs]
        assert all(s > 0 for s in scores1)
        scores1 != scores0
        sorted(scores1) != sorted(scores0)
```

integration/test_scoring_cluster.py
```diff
@@ -271,11 +289,178 @@ def test_hybrid_text_vector(self):
...
+            "FT.AGGREGATE", IDX_NAME, "@t:{red}", "ADDSCORES",
+            "LOAD", "1", "__key",
+        )
+        assert self._counter(coordinator, "search_predicate_revalidation") == 1
+        pairs = _agg_pairs(res)
+        assert sorted(k for k, _ in pairs) == sorted(DOCS)
+        scores2 = [s for _, s in pairs]
+        assert scores2 != scores1
+        # NOTE: FT.AGGREGATE results aren't implicitly sorted by score
+        assert sorted(scores2, reverse=True) == scores0
```
BCathcart:
```suggestion
        # NOTE: FT.AGGREGATE results aren't implicitly sorted by score
        sorted_scores2 = sorted(scores2, reverse=True) 
        sorted_scores2 != scores1
        sorted_scores2 == scores0
```

integration/test_scoring_cluster.py
```diff
@@ -271,11 +289,182 @@ def test_hybrid_text_vector(self):
...
+            "FT.SEARCH", IDX_NAME, "@t:{red}", "LIMIT", "0", "100",
+            "WITHSCORES",
+        )
+        assert self._counter(coordinator, "search_predicate_revalidation") == 1
+        pairs = _pairs(res)
+        assert sorted(k for k, _ in pairs) == sorted(DOCS)
+        scores1 = [s for _, s in pairs]
+        assert all(s > 0 for s in scores1)
+        assert scores1 != scores0
+        assert sorted(scores1) != sorted(scores0)
```
BCathcart:
```suggestion
        # TODO: Uncomment after https://github.com/valkey-io/valkey-search/issues/1439 is fixed
        # assert scores1 != scores0
        # assert sorted(scores1) != sorted(scores0)
```

integration/test_scoring_cluster.py
```diff
@@ -271,11 +289,183 @@ def test_hybrid_text_vector(self):
...
+            "LOAD", "1", "__key",
+        )
+        assert self._counter(coordinator, "search_predicate_revalidation") == 1
+        pairs = _agg_pairs(res)
+        assert sorted(k for k, _ in pairs) == sorted(DOCS)
+        scores2 = [s for _, s in pairs]
+        assert scores2 != scores1
+        # NOTE: FT.AGGREGATE results aren't implicitly sorted by score
+        sorted_scores2 = sorted(scores2, reverse=True)
+        assert sorted_scores2 != scores1
```
BCathcart:
```suggestion
        # TODO: Uncomment after https://github.com/valkey-io/valkey-search/issues/1439 is fixed
        # assert sorted_scores2 != scores1
```

src/query/search.h
```diff
@@ -277,8 +277,8 @@ struct SearchParameters {
   // where nothing (e.g. SORTBY) requires loading and reordering the full result
   // set first. When true, the query can complete on the background thread and
   // skip content loading.
-  bool NoProcessingRequired() const {
-    return no_content && !RequiresCompleteResults();
```
KarthikSubbarao: We don't have a dependency on `RequiresCompleteResults` anymore? Why?

BCathcart: I remember deciding processing on main thread and requiring complete results were separate dimensions and making it look like processing depended on whether or not we required results was misleading

KarthikSubbarao (approved):

PR #1421, opened 2026-09-19
Add FP64 vector distance support

### LINKED ISSUE #1401: [FEATURE] Support Float64 in vector search
RedisSearch supports Float64 in vector search; I see that this has been documented previously in https://github.com/valkey-io/valkey-search/issues/317

1) Can we know how trivial or non-trivial it would be to add Float64 support
2) If lack of Float64 support is a deliberate design decision and deviation from RedisSearch, explain why
3) Please provide guidance for RedisSearch users who wish to transition to Valkey Search and are using Float64

As for the "valid use case" mentioned in #317 - I'm sure that there was a valid use case for RedisSearch itself, which is why it's supported.

bandalgomsu:
Add FLOAT64 vector support to provide higher-precision vector indexing and search. This also improves compatibility for Redis Search users migrating to Valkey Search with existing FLOAT64 vector schemas.

While FLOAT32 is sufficient for most embedding workloads, some use cases require higher input precision, a wider numeric range, and double-precision distance calculations.

Closes : #1401


src/indexes/vector_base.h
```diff
@@ -62,17 +62,17 @@ class VectorRecord {
...
-  inline float GetReciprocalMagnitude() const { return reciprocal_magnitude_; }
+  inline double GetReciprocalMagnitude() const { return reciprocal_magnitude_; }
 
  private:
   // Constructor is private, called via placement new in Construct.
-  VectorRecord(absl::string_view vector, float reciprocal_magnitude);
+  VectorRecord(absl::string_view vector, double reciprocal_magnitude);
 
-  const float reciprocal_magnitude_;
+  const double reciprocal_magnitude_;
```
yairgott, line 75: Changing reciprocal_magnitude_ to double adds 4 bytes per VectorRecord. While negligible for high-dim FLOAT32, users often choose FLOAT16/BFLOAT16 specifically for RAM efficiency where every byte matters.

We could avoid this regression by templatizing VectorRecord with `using MagnitudeT = std::conditional_t<std::is_same_v<T, double>, double, float>; ` or having a dedicated `VectorRecordFp64`.

WDYT?

third_party/hnswlib/space_l2_fp64.h
```diff
@@ -0,0 +1,53 @@
...
+  // difference can overflow even though each stored element is valid FP64.
+  double scale = 0.0;
+  for (size_t i = 0; i < qty; ++i) {
+    scale = std::max(scale, std::abs(pVect1[i]));
+    scale = std::max(scale, std::abs(pVect2[i]));
+  }
+  if (scale == 0.0) return 0.0;
+
+  double scaled_sum_sq = 0.0;
+  for (size_t i = 0; i < qty; ++i) {
```
yairgott, line 27: Doing `pVect1[i] / scale - pVect2[i] / scale` executes two 64-bit floating point divisions per dimension. Since `fstdistfunc_` is on the critical hot path during indexing and search, this adds noticeable latency and hinders vectorization. Could we precompute `const double reciprocal_scale = 1.0 / scale;` outside the loop and multiply inside instead? Additionally, for typical vectors that don't approach overflow limits, a fast single-pass accumulation loop would avoid the two-pass scaling penalty entirely.

src/commands/ft_search.cc
```diff
@@ -90,7 +90,7 @@ void ReplyScore(ValkeyModuleCtx *ctx, ValkeyModuleString &score_as,
 
 // Reply with just the score value as a top-level element (Redis WITHSCORES
 // format: score appears between document ID and attributes array).
-void ReplyScoreTopLevel(ValkeyModuleCtx *ctx, float score) {
+void ReplyScoreTopLevel(ValkeyModuleCtx *ctx, double score) {
   auto score_value = absl::StrFormat("%.12g", score);
```
yairgott, line 94: Here and in any other instances of using `"%.12g"`: Since `FLOAT64` is chosen specifically for double precision (~15–17 decimal digits), `%.12g` truncates up to 5 digits of precision in search replies. Using `%.17g` (or `std::numeric_limits<double>::max_digits10`) would preserve full double precision.

src/indexes/vector_base.cc
```diff
@@ -185,7 +329,8 @@ absl::StatusOr<RecordResult> VectorBase::AddRecord(const InternedStringPtr &key,
     return RecordResult::kInvalidData;
   }
   auto vector_record = data.ConsumeVector();
-  float magnitude = 1.0f / vector_record->GetReciprocalMagnitude();
+  float magnitude =
```
yairgott, line 332: Maintaining the magnitude as `float` for `TrackKey` can overflow to `+inf` for FP64 vectors or underflow to `0.0f`. Updating `TrackKey` and `TrackedKeyMetadata.magnitude` to `double` would handle the full FP64 range.

src/indexes/vector_base.cc
```diff
@@ -233,7 +378,8 @@ absl::StatusOr<RecordResult> VectorBase::ModifyRecord(
...
```
yairgott, line 381: Same comment as for line 332.

yairgott:
@bandalgomsu, Thanks for driving this! The Redis Search migration compatibility, the backward-compatible coordinator protocol in `coordinator.proto`, and the test coverage look really solid. 

I've added a few comments which are related to performance, precision, and memory. Can you please have a look?

Thanks!

third_party/hnswlib/space_l2_fp64.h
bandalgomsu: I agree that performance is important here, especially since this is on the hot path.
I noticed that the existing FP16 and FP32 L2 implementations use direct single-pass accumulation without special handling for extreme values that may overflow.
Would it make sense for FP64 to follow the same behavior as well, rather than adding overflow handling specifically for FP64?

src/indexes/vector_base.h
bandalgomsu: The fixed-size allocator rounds allocations up to a multiple of 8 bytes, so for common FLOAT16/BFLOAT16 dimensions like 128/256/512/768/1024/1536, this change does not increase the allocated size. Some other dimensions do incur an 8-byte increase. Do you think it's worth handling those cases where the vector size isn't a multiple of 8 bytes?

[allocator.cc](https://github.com/valkey-io/valkey-search/blob/main/src/utils/allocator.cc#L111)

For example (FLOAT16):
- dim=128 (256 bytes): float 4+256=260 → 264, double 8+256=264 → 264 (no change)
- dim=130 (260 bytes): float 4+260=264 → 264, double 8+260=268 → 272 (+8)

